PR #26, opened 2021-04-22
Rename the phylum binary

louislang:
Current binary produced is `phylum-cli`. This renames the binary to simply `phylum`.


eeclfrei (approved):

peterjmorgan:
🙌

Issue #20, opened 2021-04-21
Improve `auth` subcommand


louislang:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
A reorganization of the `auth` subcommand. Running `ph auth` should return the high level help information for this command, along with the list of subcommands of `auth` and their descriptions.

This command should have several of its own subcommands:

- [x] `ph auth register` - Should go into an interactive mode and ask the user for their `email`, `password`, `confirm password` and `name`. We currently request first and last name, this should be updated to simply `name`. After successful registration, we should create a new API key named `cli` that will be used for authentication moving forward.
- [x] `ph auth login` - Log a user in with existing credentials. Also in interactive mode. After login, we should get or create the API key `cli` and use that for authentication moving forward.
- [x] `ph auth keys` - Provide management facilities for API keys. This command should similarly have its own subcommands: `create`, `list` and `remove`.
- [x] `ph auth status` - Return an indicator of current authentication status.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #28, opened 2021-04-23
Add new subcommand for `auth`

louislang:
Update the authentication mechanism to conform to requested changes in https://github.com/phylum-dev/cli/issues/20.

Adds the `auth` subcommand which has it's own subcommands:

* `ph auth keys` - Manage user keys
* `ph auth register` - Interactively register a new user 
* `ph auth login` - Interactively login a user

### DIFF
--- a/lib/src/bin/.conf/cli.yaml
+++ b/lib/src/bin/.conf/cli.yaml
@@ -17,31 +17,54 @@ args:
         help: Set the timeout (in seconds) for requests to the Phylum api
         takes_value: true
 subcommands:
+    - auth:
+        about: Manage authentication, registration, and API keys
+        subcommands:
+            - register:
+                about: Register a new account
+            - login:
+                about: Login to an existing account
+            - keys:
+                about: Manage API keys
+                after_help: "EXAMPLES
+
+
+                             # List existing active API keys
+
+
+                             \tphylum auth list
+
+
+                             # Create new API key
+
+
+                             \tphylum auth keys create
+
+
+                             # Remove (deactivate) an API key
+
+
+                             \tphylum auth keys remove <api-key>
+
+
+                             "
+                subcommands:
+                    - create:
+                        about: Create a new API key
+                    - list:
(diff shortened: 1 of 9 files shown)


lib/src/bin/.conf/cli.yaml
```diff
@@ -17,31 +17,54 @@ args:
...
+
+
+                             "
+                subcommands:
+                    - create:
+                        about: Create a new API key
+                    - list:
+                        about: List current API keys
+                    - remove:
+                        about: Deactive an API key
```
eeclfrei, line 57: `s/Deactive/Deactivate/`

lib/src/bin/phylum-cli.rs
```diff
@@ -126,7 +129,7 @@ fn handle_status(api: &mut PhylumApi, req_type: &PackageType, matches: clap::Arg
             // get everything
             let resp = api.get_status();
             log::debug!("==> {:?}", resp);
-            print_response(&resp, pretty_print);
+            //print_response(&resp, pretty_print);
```
eeclfrei, line 132: Should this be commented out?

lib/src/bin/phylum-cli.rs
```diff
@@ -225,11 +228,207 @@ fn handle_submission(api: &mut PhylumApi, config: Config, matches: clap::ArgMatc
     exit_status
 }
 
+/// Register a user. Drops the user into an interactive mode to get the user's
+/// details.
+fn handle_auth_register(api: &mut PhylumApi, config: &mut Config, config_path: &str) {
+    let name: String = Input::with_theme(&ColorfulTheme::default())
+        .with_prompt("Your name")
+        .interact_text()
```
eeclfrei, line 236: It looks like `interact_*` can fail for various i/o issues.  Instead of `unwrap`ing we should probably bubble up the error via `handle_auth` and then exit process with an error code if this is the case.

eeclfrei (approved):

PR #161, opened 2022-01-14
Update sha2 crate due to RUSTSEC-2021-0100

ejortega:
https://rustsec.org/advisories/RUSTSEC-2021-0100.html

Says 0.9.8 isn't affected but 0.9.8 has been yanked, so bumped up the version.


eeclfrei:
Cool, I noticed this in the build a while ago and was on my list to update...

eeclfrei (approved):

PR #165, opened 2022-01-21
Update install script to support m1/arm

eeclfrei:


lib/install.sh
```diff
@@ -52,9 +66,15 @@ fi
 
 # Copy the specific platform binary.
 platform=$(get_platform)
-arch="x86_64"
+arch=$(get_arch)
 bin_name="phylum-${platform}-${arch}"
-check_copy "${bin_name}" "${HOME}/.phylum/phylum" 
+
+if [[ "$arch" == "aarch64" ]]; then
+    cat "${bin_name}" > "${HOME}/.phylum/phylum"
```
louislang, line 73: Should this be moved into `check_copy` itself? Seems like this will only get used for `aarch64`. It's _probably_ safe to use it for all architectures/platforms?

lib/install.sh
```diff
@@ -25,6 +28,17 @@ get_platform() {
     fi
 }
 
+# Get the platform name.
+get_arch() {
+    local platform_arch
+    platform_arch=$(uname -p)
```
louislang, line 34: `uname -p` on the m1 comes back as `arm`. When we go to copy the file, we look for `phylum-macos-arm`, which doesn't exist. We do have `phylum-macos-aarch64` in the zip though. Should we rename the produced file to `phylum-macos-arm`?

lib/install.sh
eeclfrei: Held off on that for now, since some of the `check_copy` calls are to directories as the destination.

lib/install.sh
eeclfrei: Updated to match against first 3 chars (i.e. `arm.*`), so it _should_ copy the right bin now.

louislang: Ok cool, I'll re-test on an m1 here in a bit!

eeclfrei: Awesome thanks!

louislang (approved):
Verified working on m1 mac with continued support for intel machines.

PR #172, opened 2022-01-28
Remediation refactor

andreaphylum:
Opening this draft PR in order to collect feedback over time, as this is going to move around a number of important parts.

- I'm going to leave old code commented out until we are sure we are not breaking features.
- I factored out `restson` as its own crate under the `lib/` directory, but I expect to fully remove it before closing this PR.
- Part of the process of this PR is going to be reorganizing the repository in a [Cargo workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html).


andreaphylum:
A few notes about progress:

- Some logic that was separated between GitHub actions definitions and shell scripts has now been factored in the `dist` xtask. Running `cargo xtask dist` now builds the project in Docker, copies out the relevant binaries, and generates the distribution files (`settings.yaml`, `install.sh`, shell completions) separately under three `target/dist/phylum-{os}-{arch}` directories, and can be run locally (depending only on the existence of the `docker` binary).
- GitHub actions can now be simplified as follows: run `cargo xtask dist`, sign files, create zipballs.
- The output of a release is three different zipballs, one for each build target; in each of those, the `install.sh` script has been simplified and is now platform-agnostic.
- Completions aren't located in the source tree anymore but are generated on the fly by the xtask.
- Dockerfiles have been moved under the `xtask` crate's source tree as they are related more to the build process than to the `cli` source.
- Simplified the Dockerfile for the Linux musl build.
- Moved `clap` definitions from the deprecated `yaml` format to the declarative API.

louislang (approved):
lgtm. Would like @DanielJoyce's input here since he's been working on this area a lot recently.

eeclfrei (approved):

DanielJoyce:
Hmm, thought I reveiwed this a while ago.

PR #186, opened 2022-03-11
Fix release workflow

### LINKED ISSUE #185: Release workflow does not produce artifacts
# Overview
The [Release New workflow](https://github.com/phylum-dev/cli/actions/workflows/release.yml) does not produce release artifacts. It appears to run successfully, but the `release` step fails to find the expected artifacts:

```
Run softprops/action-gh-release@v0.1.7
🤔 Pattern 'phylum-linux-x86_64.zip' does not match any files.
🤔 Pattern 'phylum-macos-x86_64.zip' does not match any files.
🤔 Pattern 'phylum-macos-aarch64.zip' does not match any files.
🤔 Pattern 'phylum-linux-x86_64.minisig' does not match any files.
🤔 Pattern 'phylum-macos-x86_64.minisig' does not match any files.
🤔 Pattern 'phylum-macos-aarch64.minisig' does not match any files.
👩‍🏭 Creating new GitHub release for tag v1.3.0-rc1...
🤔 phylum-linux-x86_64.zip,phylum-macos-x86_64.zip,phylum-macos-aarch64.zip,phylum-linux-x86_64.minisig,phylum-macos-x86_64.minisig,phylum-macos-aarch64.minisig not include valid file.
🎉 Release ready at https://github.com/phylum-dev/cli/releases/tag/v1.3.0-rc1
```

# How To Reproduce
1. Create a release candidate tag for the `cli` repo and push it. This should be one that has specifically `rc` in the tag (e.g., `v1.3.0-rc2`) so as not to create a "true" release.
2. Wait for the `Release New` workflow under the `Actions` tab to pick up the tag push and wait for it to finish.
3. Check the artifact output to verify that the expected files are not there.

# Expected Behavior
The collection of files required for releasing a new version of the CLI are included in the output of running the `Release` workflow (currently called `Release New`).

# Additional Context
There is an additional request, while working this ticket, to enable the ability to download release artifacts without actually *releasing* the CLI. That is, it should be possible to download the artifacts from a build to test them without needing to create a new release tag. The retention period for these artifacts was suggested to be 7 days.

maxrake:
# Overview
The CLI release workflow was not producing any artifacts. The last big refactor changed the way the artifacts are produced and named. The workflow and documentation has been updated to match. Further, the `release.sh` script was updated to make it easier to push a release. Linting and formatting was done throughout.

# Checklist
- [ ] ~Have you created sufficient tests?~

# Issue
Closes #185

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,8 @@ on:
     branches:
       - master
       - development
+    tags:
+      - '*'
 
 jobs:
   Build-Release-Artifacts:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,18 +73,11 @@ jobs:
         uses: actions/download-artifact@v3
         with:
           name: release-archives
-      # TODO: see if it is possible to replace this action with an inline expression
-      - name: Determine prerelease status
-        uses: haya14busa/action-cond@v1
-        id: preRelease
-        with:
-          cond: ${{ contains(github.ref, 'rc') }}
-          if_true: 'true'  # string value
-          if_false: 'false'  # string value
-      - name: Release
+      - name: Create GitHub release
         uses: softprops/action-gh-release@v0.1.14
         with:
-          prerelease: ${{ steps.preRelease.outputs.value }}
+          # This check is already filtered down to only 'refs/tags/' and shouldn't "overmatch"
+          prerelease: ${{ contains(github.ref, 'rc') }}
           fail_on_unmatched_files: true
           token: ${{ secrets.GITHUB_TOKEN }}
           files: phylum-*.zip
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,28 +3,31 @@
 name: Release New
 
 on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
   push:
     tags:
       - '*'
-
-  # Allows you to run this workflow manually from the Actions tab
-  workflow_dispatch:
 
 jobs:
   release:
     runs-on: ubuntu-latest
     steps:
-      - run: |
+      - name: Install host dependencies
+        run: |
           sudo add-apt-repository -y ppa:dysfunctionalprogramming/minisign
           sudo apt update
           sudo apt install -yq build-essential zip minisign
-      - uses: actions/checkout@master
-      - uses: actions-rs/toolchain@v1
+      - name: Checkout the repo
+        uses: actions/checkout@master
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-      - name: Build Linux
+      - name: Build Distributions
         run: cargo xtask dist
-      - name: Sign files
+      # TODO: determine if it makes more sense to sign the entire zipfile or just the binaries w/i the zipfiles
+      - name: Sign binaries
         run: |
           echo -e $MINISIGN_KEY > minisign.key
           echo $MINISIGN_PASSWORD | minisign -Sm target/dist/phylum-linux-x86_64/phylum -s minisign.key -t 'Phylum - Future of software supply chain security'
@@ -36,9 +39,9 @@ jobs:
       - name: Build zipfiles
         working-directory: ./target/dist
         run: |
-          zip -r phylum-linux-x86_64 phylum-linux-x86_64
-          zip -r phylum-macos-x86_64 phylum-macos-x86_64
-          zip -r phylum-macos-aarch64 phylum-macos-aarch64
+          zip -r phylum-linux-x86_64.zip phylum-linux-x86_64
+          zip -r phylum-macos-x86_64.zip phylum-macos-x86_64
+          zip -r phylum-macos-aarch64.zip phylum-macos-aarch64
       - name: Determine prerelease status
         uses: haya14busa/action-cond@v1
         id: preRelease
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -1,61 +1,84 @@
-# Modified Release github action. Replace the current with this when ready.
+# This is the workflow for building release artifacts.
+# Optionally, creating a release with those artifacts is possible by pushing a tag.
 ---
-name: Release New
+name: Release
```
maxrake, line 4: Making this change (putting it back, really) causes the Actions tab to show multiple "Release" workflows:

Screen Shot 2022-03-11 at 1 01 03 PM

The correct one can be determined by hovering over the link to view the file name (should be `release.yml`). I just disabled the unused workflows...but I'm wondering if there is a way to remove them permanently. Are they needed for something...or some amount of time/retention?

.github/workflows/release.yml
```diff
@@ -1,61 +1,84 @@
...
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-      - name: Build Linux
+
+      # TODO: determine if it is possible to break out the build into multiple targets, to enable parallel job execution
+      - name: Build Distributions
         run: cargo xtask dist
```
maxrake, line 38: The way it is now...the builds take over 22 minutes to complete...most of it on this step. If we split the builds out, it is possible to parallelize them with multiple jobs...bringing the time down to more like 8-10 minutes. I can do this here or in a separate issue...depending on a larger conversation about the best way to package and release the artifacts for user consumption. I've already talked with @furi0us333 about this and I believe he plans to have a larger discussion with @louislang

README.md
```diff
@@ -15,32 +17,59 @@ The command line interface (CLI) allows users to submit their project package de
...
-```
-5. [Create a new Phylum project](https://docs.phylum.io/docs/projects#creating-a-new-project) in your project directory
-```
-phylum projects create <project-name>
-```
-6. [Submit your package lock file](https://docs.phylum.io/docs/analyzing-dependencies)
-```
-phylum analyze <package-lock-file.ext>
-```
+## Quickstart for Linux or macOS
```
maxrake, line 20: The [User docs](https://docs.phylum.io/v1.3/docs/getting-started) were updated to match. Well...a suggested edit was submitted. It would be great if *all* the docs were co-located with the code, so that updates can be made, reviewed, pushed, and published synchronously.

release.sh
```diff
@@ -1,26 +1,21 @@
...
-read version
+read -r version
 
 printf "changelog: "
-read changelog
+read -r changelog
 
 sed -E -i "1 s#^#* $version - $changelog\n#" CHANGELOG
-sed -E -i "s/^version = \"([^\"]*)\"/version = \"$version\"/" lib/Cargo.toml
-sed -E -i "s/^version: \"([^\"]*)\"/version: \"$version\"/" lib/src/bin/.conf/cli.yaml
```
furi0us333, line 11: I know this file no longer exists, but I would like either @louislang or @andreaphylum to confirm that there aren't other places we need to update the version.

furi0us333 (changes requested):
We are going to use our own runners even though GitHub does not does not count minutes on public repositories.

.github/workflows/test.yml
```diff
@@ -2,40 +2,48 @@
...
     branches:
       - master
       - development
 
-  # Allows you to run this workflow manually from the Actions tab
-  workflow_dispatch:
-
 jobs:
   test:
     runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
maxrake: @furi0us333, can you confirm the `self-hosted` runners are set up for this `cli` repo? Making the change and pushing the commit kicked off a `Test` workflow, but the job has not been picked up by a runner yet and it's been an hour.

furi0us333 (changes requested):
Upon additional investigation, it looks like I made a poor choice, you will want to revert these changes back. Sorry about that.

https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security

image

.github/workflows/release.yml
louislang: I know @andreaphylum implemented this xtask, so I'll defer to him on it. But parallelizing builds is always a plus imo.

README.md
louislang: Chatted with @furi0us333 about this yesterday. Automating the docs would be wonderful.

release.sh
louislang: Versions were previously being set in the `cli.yaml`. This now gets set by an env var (`CARGO_PKG_VERSION`) at compile time. See: https://github.com/phylum-dev/cli/blob/1dd193bd8ee0dd85fb6027e6ab48d51c8b62efc8/cli/src/app.rs#L17

So we probably need to make sure that env var is set properly.

maxrake: The `CARGO_PKG_VERSION` envvar is set by cargo and exposed when the crate is compiled. It is populated from the value in the project's manifest file...`cli/Cargo.toml` in this case. So...ensuring the manifest is updated to a new version when releasing, like this script does, should be enough.
Ref: ["Environment variables Cargo sets for crates" section in "The Cargo Book"](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates)

louislang: 👍 Figured it was something like that, just wanted to follow the thread there since we ditched the `cli.yaml` a bit ago!

.github/workflows/release.yml
andreaphylum: The xtask is really just [Rust code](https://github.com/phylum-dev/cli/blob/development/xtask/src/main.rs#L60) executed sequentially. We could definitely break it down in more subcommands, ideally one per parallelizable branch. Other than that, if we have a way of precompiling the Docker images and pulling them from somewhere instead of compiling them as a build step, I assume we could further reduce the CPU time. Unfortunately I'm not too well versed with GitHub actions, so not sure if that is feasible, but it would be a pretty big win in that case.

.github/workflows/release.yml
furi0us333: I did not explicitly see language on deleting a Workflow in the GH docs. My guess is they are meant to remain for historical purposes. If we really want to remove them, we can try deleting all of the Workflow Runs in each Workflow and see if that does it, but that might not be necessary as they are now disabled anyways.

README.md
furi0us333: I am working on this and hope to have a solution soon.

furi0us333 (approved):
LGTM

.github/workflows/release.yml
maxrake: Created #188 as a new issue to track this task. Resolving this conversation...

PR #202, opened 2022-03-21
Rename history.md to history-cli.md

furi0us333:
# Overview

Changing page slug as readme.com didn't like `history` for some reason.


DanielJoyce (approved):

maxrake (approved):
Not sure why it broke before...but this is worth a shot.

janasheehan (approved):

PR #203, opened 2022-03-21
Update docs.yml

furi0us333:
# Overview

The categories are different between versions, so we will not be able to update a development version of the docs at this time. Removing development strategy from the action to keep it from failing on push to `development`.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -3,7 +3,6 @@
 name: Sync `docs` directory to ReadMe
 
 on:
-  workflow_dispatch:
   push:
     branches:
       - master
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,7 +7,6 @@ on:
   push:
     branches:
       - master
-      - development
   
 jobs:
   Update-Documentation:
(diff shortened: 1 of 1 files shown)


janasheehan (approved):

.github/workflows/docs.yml
```diff
@@ -18,13 +17,6 @@ jobs:
         uses: actions/checkout@v3
        
       - name: Update v1-latest using rdme
-        if: ${{ github.ref_name == 'master' }}
```
maxrake, line 21: This line is likely still warranted. Without it, b/c the `workflow_dispatch` is set, it would seem to be possible for anyone to manually update the docs.

maxrake (changes requested):
Please confirm or deny the suspicion about allowing anyone to update the docs with the manually triggered workflow.

.github/workflows/docs.yml
furi0us333: @maxrake Is the conditional preferred to removing the `workflow_dispatch` trigger?

maxrake: Thinking more about it and talking with @kylewillmon, I believe the correct move is to remove the `workflow_dispatch` trigger.

maxrake (approved):
LGTM.

PR #204, opened 2022-03-22
Cherry pick doc automation

furi0us333:
# Overview

Cherry picking doc automation work to get it running in `master`


docs/building.md
```diff
@@ -0,0 +1,24 @@
...
+2. Build the project
+
+   ```sh
+   cargo build
+   ```
+
+3. You can use the executable directly as `./target/debug/phylum` or install it like so:
+
+   ```sh
+   cargo install --path cli
```
kylewillmon, line 23: I don't believe this is valid for `master`

kylewillmon (commented):
Merging anything other than the `development` branch into `master` would require rebasing the `development` branch to keep a linear history.

If we want to cherry-pick just to cut a new tag, that would be fine, but that should never be merged into `master`.

kylewillmon:
Using the new tag approach in this case would require that we modify the workflow to run on a tag.

But I think it makes sense to wait for `development` unless there is some benefit I am missing of doing this merge now.

PR #205, opened 2022-03-23
Parallel builds

### LINKED ISSUE #188: Parallelize the build
# Overview
The way it is now...the builds take ~17-22 minutes to complete...most of it on the `Build distributions` step of the Release workflow. If the builds are split out, it is possible to parallelize them with multiple jobs...bringing the time down to more like 8-10 minutes.

> The xtask is really just [Rust code](https://github.com/phylum-dev/cli/blob/development/xtask/src/main.rs#L60) executed sequentially. We could definitely break it down in more subcommands, ideally one per parallelizable branch. Other than that, if we have a way of precompiling the Docker images and pulling them from somewhere instead of compiling them as a build step, I assume we could further reduce the CPU time. Unfortunately I'm not too well versed with GitHub actions, so not sure if that is feasible, but it would be a pretty big win in that case.

_Originally posted by @andreaphylum in https://github.com/phylum-dev/cli/pull/186#discussion_r825925055_

## Acceptance Criteria

- [ ] Each build target can be compiled independently from the other build targets
- [ ] Build target names are standardized so as to ensure uniqueness and traceability from the rust build target to the download artifact
- [ ] Optional: if possible, make use of existing/pre-compiled Docker images

## Associated User Story
--

kylewillmon:
# Overview

Build the CLI binaries in parallel

## Checklist

## Issue

closes #188


.github/workflows/release.yml
```diff
@@ -27,38 +94,34 @@ jobs:
...
-          echo $MINISIGN_PASSWORD | minisign -Sm target/dist/phylum-linux-x86_64/phylum -s minisign.key -t 'Phylum - the future of software supply chain security'
-          echo $MINISIGN_PASSWORD | minisign -Sm target/dist/phylum-macos-x86_64/phylum -s minisign.key -t 'Phylum - the future of software supply chain security'
-          echo $MINISIGN_PASSWORD | minisign -Sm target/dist/phylum-macos-aarch64/phylum -s minisign.key -t 'Phylum - the future of software supply chain security'
+          for archive in phylum-*/;
+          do
+            archive=$(echo ${archive} | sed -e "s/\/$//")
+            echo $MINISIGN_PASSWORD | minisign -Sm ${archive}/phylum -s minisign.key -t 'Phylum - the future of software supply chain security'
+            cp -R shell-completions ${archive}/completions
+            cp cli/cli/src/install.sh ${archive}/install.sh
+            cp cli/cli/src/settings.yaml ${archive}/settings.yaml
```
maxrake, line 112:
```suggestion
            cp cli/cli/src/settings.yaml ${archive}/settings.yaml
            chmod 600 ${archive}/settings.yaml
```
This would be a good spot to restrict the file permissions of the `settings.yaml` file. It contains an `offline_access` token that should only be visible to the user.

maxrake (dismissed):
With these changes, it appears the following Dockerfiles are no longer used:

* `xtask/src/dockerfiles/Dockerfile.macos`
* `xtask/src/dockerfiles/Dockerfile.linux`

Is there a purpose for keeping these files? Is it [still?] possible to build the various targets locally or is the expectation that the CI infrastructure will handle all of that?

The changes in this PR are very much welcome and bring the build time down from ~18-22 minutes to ~9-10 minutes...with the ability to add additional build targets w/o adding much (if any) additional time.

.github/workflows/release.yml
kylewillmon: Hmmm... I wasn't aware that was a problem. The CLI also now creates the `settings.yaml` file if it doesn't already exist. I think this is worth an issue of its own.

kylewillmon: Create #208 to look into this.

kylewillmon:
> Is there a purpose for keeping these files?

Good point. I don't think so.

> Is it [still?] possible to build the various targets locally or is the expectation that the CI infrastructure will handle all of that?

CI will handle that now. Cross-compiling with `cargo` is not terribly difficult, but the setup varies based on host OS and architecture and I see no reason that we need to take on that responsibility.

maxrake (approved):
LGTM

Issue #187, opened 2022-03-15
`phylum update` subcommand is broken


furi0us333:
# Overview
Due to changes in the way release assets are packaged in the current pre-release, the updater fails to find the file it is looking for.

# How To Reproduce
Steps to reproduce this behavior:
1. Run `phylum update -p`
2. See error

# Expected Behavior
The updater successfully downloads and installs the correct version.

# Additional Context
![image](https://user-images.githubusercontent.com/34108612/158298755-b15cb51b-8da4-48d5-ab2f-7c4283be312d.png)

The new package strategy is nice as you no longer have to download all builds, so this likely requires some fixes in the updater code (https://github.com/phylum-dev/cli/blob/development/cli/src/update.rs). We will probably need to generate minisigs off the platform zips and check those to verify the package.

Current `master` Assets
![image](https://user-images.githubusercontent.com/34108612/158299472-bd91d660-eb33-4fc3-803d-0c75771ab23b.png)

Current `development` Assets
![image](https://user-images.githubusercontent.com/34108612/158299536-6d6708eb-a780-41b4-9698-9ee329d58362.png)

louislang:
I don't have super strong opinions. My one qualm against this change is basically `<arch>-<company>-<system>` which _perhaps_ is less clear? `macos` is more prominent than  `darwin`, for example.

maxrake:
Understood. The idea behind this suggestion was to pave the way for making the `phylum` install be more like the [rustup.sh](https://rustup.rs/) install script...where the script is run on target and determines the correct binary to acquire. Consumers of the phylum CLI tool wouldn't need to know about all the various target triples and which one they need b/c the install script would do that for them (Windows being the exception). Looking at that script, it appears that the `sys` part of the target triple is determined with a `uname -s` call.

Since the CLI is built with Rust, the thought was that the Rust targets...and maybe even a good chunk of the rustup script...could be used as is. Even still, mapping from `darwin` to `macos` is not a big deal...so long as no future naming conflicts emerge.

kylewillmon:
If we want `phylum update` to work from `1.2.0` to `1.3.0`, we must include the assets that the already-released `1.2.0` binary expects in the `1.3.0` release.

Or, to put it differently: The easiest way to resolve this issue is to change the assets back to the layout used for the last release.

The layout change was made in #172, but I don't see an explanation there for why the layout was changed. Is anyone opposed to going back to the old layout? (cc: @andreaphylum)

[...]

PR #209, opened 2022-03-24
Sign zip archives instead of binaries

kylewillmon:
It's probably beneficial to sign the zip archives instead of just signing the binary. Here are a few reasons why:

* Signing the archives allows verification before extracting files
* Signing the archives confirms the integrity of `install.sh` and other files as well as the binary itself

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,22 +100,17 @@ jobs:
       - name: Download release artifacts
         uses: actions/download-artifact@v3
 
-      - name: Sign binaries and prep archives
+      - name: Prep archives
         run: |
-          echo -e $MINISIGN_KEY > minisign.key
           for archive in phylum-*/;
           do
             archive=$(echo ${archive} | sed -e "s/\/$//")
-            echo $MINISIGN_PASSWORD | minisign -Sm ${archive}/phylum -s minisign.key -t 'Phylum - the future of software supply chain security'
             cp -R shell-completions ${archive}/completions
             cp cli/cli/src/install.sh ${archive}/install.sh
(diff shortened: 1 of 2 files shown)


maxrake:
This idea came up in the #development channel on 10 MAR 2022:

> **Charles Coggins** For the CLI artifacts:
Does it make more sense to sign just the phylum binary…which gets included in a zipfile of other files…or sign the entire zip archive? Currently, only the phylum binary is signed.
> **louis** Binary needs to be signed. I'm pretty sure we're pulling down the binary and checking its signature, not the entire zip archive for our auto update capability.
> **louis** and just for ref, here's where we check the signature for the file: https://github.com/phylum-dev/cli/blob/development/cli/src/update.rs#L249

Not saying this is a bad idea...just that it might have some additional ramifications...

kylewillmon:
Right. But that code is already broken, per #187

maxrake (changes requested):
The changes in this review do not function as intended. A `v1.3.0-rc5` tag was created and pushed to trigger the `Release` workflow and the `Release` job within it. That triggered [this action](https://github.com/phylum-dev/cli/actions/runs/2055237333), which shows a failure: `Process completed with exit code 127.` This appears to be due to the `minisign` application not being installed/available within the runner.

Further, there is concern that the release artifact will not include the `*.minisign` files. Finally, it might be nice to _also_ have the ability to verify the artifacts every time the `Release` workflow runs (e.g., for PRs and pushes to `development` and `main` branches) and not just when a release is created from a tag. Moving the signing step back into the `Build-Release-Artifacts` job might address several of these concerns.

maxrake (approved):
Verified artifacts...LGTM.

PR #218, opened 2022-03-30
Move code out of `src/bin/phylum`

kylewillmon:
Move code out of `src/bin` so that tooling (i.e., rustdoc, the rust VS Code extension, etc.) works better.

There is no reason that all of that code should be in `src/bin` rather than in the crate itself. Eventually, we may even consider moving much of the code that is in `phylum.rs` into the crate as well and having that file just be a simple call to a function in the library.


kylewillmon:
Things are relatively quiet in this repo right now, and I'm hoping to make this change while it's easy

maxrake (approved):
LGTM

eeclfrei (approved):

PR #219, opened 2022-03-30
settings.yaml file permissions

### LINKED ISSUE #208: Retrict access to settings.yaml
We should restrict the file permissions of the `settings.yaml` file. It contains an `offline_access` token that should only be visible to the user.

_Originally pointed out by @maxrake in https://github.com/phylum-dev/cli/pull/205#discussion_r834473640_

kylewillmon:
# Overview

Set the file permissions for `settings.yaml` when it is created by the CLI. Additionally, this PR removes the `settings.yaml` file from the release.

## Checklist

## Issue

Closes #208


eeclfrei:
Sorry I'm a bit out of the loop on this one - so we're just creating the `settings.yaml` on first run now?  Just want to verify that we're setting reasonable defaults for the `auth_info` and `connection` uri's.

kylewillmon:
> Sorry I'm a bit out of the loop on this one - so we're just creating the `settings.yaml` on first run now? Just want to verify that we're setting reasonable defaults for the `auth_info` and `connection` uri's.

The change was made in #196 to use default values if the config file isn't found. You can see those defaults [here in the code](https://github.com/phylum-dev/cli/blob/9439550cb2e30444d80ec1041830097c2b8b3536/cli/src/config.rs#L47).

eeclfrei:
So it was - missed that somehowl.

eeclfrei (dismissed):

maxrake (changes requested):
Everything _looks_ good in the code. However, when doing (still manual) acceptance testing there are failures:

```sh
❯ ph version
[1]    72919 killed     phylum version
```

This was encountered by removing all local installed phylum artifacts, downloading the artifacts from the build associated with the PR, and installing clean. All of the commands have this same behavior.

One other thing that was noticed...but unrelated to this review...is that the top-level project `README.md` is not synced up with the more current `docs/quickstart.md`. Perhaps there is a way to reference one file from the other so that only one source need be updated.

maxrake (approved):
It works now! Very nice usage of the `install` command.
LGTM.

Issue #216, opened 2022-03-25
Make install script portable


maxrake:
# Overview

Change the CLI `install.sh` script to `/bin/sh` for portability. This will pave the way for future re-factors of the script and also allow for more target system support.

_Originally suggested by @kylewillmon in https://github.com/phylum-dev/cli/issues/214#issuecomment-1079072318_

# Acceptance Criteria
- [ ] The shebang line in the `install.sh` script is `/bin/sh`
- [ ] QA passes with `shellcheck` on strict mode (i.e., `-o all`)
- [ ] Functionality does not change

## Associated User Story
--

PR #225, opened 2022-04-05
Use zip archives for `phylum update`

kylewillmon:
# Overview

Implement `phylum update` by downloading the appropriate zip file and running `install.sh`

## Checklist

### DIFF
--- a/cli/src/update/unix.rs
+++ b/cli/src/update/unix.rs
@@ -1,6 +1,8 @@
-use std::io;
+use std::io::{self, Cursor};
 use std::process::Command;
+use std::str;
 
+use anyhow::Context;
 use log::debug;
 use minisign_verify::{PublicKey, Signature};
 use reqwest::Client;
@@ -20,9 +22,9 @@ const GITHUB_URI: &str = "https://api.github.com";
 pub async fn needs_update(current_version: &str, prerelease: bool) -> bool {
     let updater = ApplicationUpdater::default();
     match updater.get_latest_version(prerelease).await {
-        Some(latest) => updater.needs_update(current_version, &latest),
-        None => {
-            log::debug!("Failed to get the latest version for update check");
+        Ok(latest) => updater.needs_update(current_version, &latest),
+        Err(e) => {
+            log::debug!("Failed to get the latest version for update check: {:?}", e);
             false
         }
     }
@@ -34,7 +36,7 @@ pub async fn do_update(prerelease: bool) -> anyhow::Result<String> {
     let ver = updater
         .get_latest_version(prerelease)
         .await
-        .ok_or_else(|| anyhow::anyhow!("Failed to get version metadata"))?;
+        .context("Failed to get the latest version")?;
     updater
         .do_update(ver)
         .await
@@ -118,35 +120,25 @@ impl ApplicationUpdater {
     }
 
     /// Check for an update by querying the Github releases page.
-    async fn get_latest_version(&self, prerelease: bool) -> Option<GithubRelease> {
-        async fn inner(github_uri: &str, prerelease: bool) -> anyhow::Result<GithubRelease> {
-            if prerelease {
-                let url = format!("{}/repos/phylum-dev/cli/releases", github_uri);
-                let r = http_get(&url).await?;
-                let releases = r.json::<Vec<GithubRelease>>().await?;
-                // Use the first one in the list, which should be the most recent
-                releases
-                    .first()
-                    .cloned()
-                    .ok_or_else(|| anyhow::anyhow!("no releases found"))
-            } else {
-                let url = format!("{}/repos/phylum-dev/cli/releases/latest", github_uri);
-                let r = http_get(&url).await?;
-                Ok(r.json::<GithubRelease>().await?)
-            }
-        }
-
-        let ver = match inner(&self.github_uri, prerelease).await {
-            Ok(ver) => ver,
-            Err(e) => {
-                log::warn!("Failed getting latest version: {:?}", e);
-                return None;
-            }
+    async fn get_latest_version(&self, prerelease: bool) -> anyhow::Result<GithubRelease> {
+        let ver = if prerelease {
+            let url = format!("{}/repos/phylum-dev/cli/releases", self.github_uri);
+            let r = http_get(&url).await?;
+            let releases = r.json::<Vec<GithubRelease>>().await?;
+            // Use the first one in the list, which should be the most recent
+            releases
+                .first()
+                .cloned()
+                .ok_or_else(|| anyhow::anyhow!("no releases found"))?
+        } else {
+            let url = format!("{}/repos/phylum-dev/cli/releases/latest", self.github_uri);
+            let r = http_get(&url).await?;
+            r.json::<GithubRelease>().await?
         };
 
         log::debug!("Found latest version: {:?}", ver);
 
-        Some(ver)
+        Ok(ver)
     }
 
     /// Compare the current version as reported by Clap with the version currently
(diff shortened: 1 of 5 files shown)


maxrake (dismissed):
I reviewed all the files. I don't know Rust well enough to claim that those files are good...but they _appear_ to be good.

Trying the build artifacts resulted in no big issues found. I couldn’t try updating to the latest non-prerelease b/c that is still 1.2.0, which has the old artifact name/scheme...and the failure experienced there is expected. Updating to a prerelease version worked.

It would be nice to have others who are more familiar with Rust complete the review...but I'm providing my approval now based on the fact that the changes are confirmed to do what they say, based on manual verification testing.

cli/src/update/unix.rs
```diff
@@ -59,14 +57,50 @@ impl Default for ApplicationUpdater {
...
+/// Generic function for fetching data via HTTP GET.
+async fn http_get(url: &str) -> anyhow::Result<reqwest::Response> {
+    let client = Client::builder().user_agent("phylum-cli").build()?;
+    let response = client.get(url).send().await?;
+    Ok(response)
+}
+
+/// Download the specified Github asset. Returns a bytes object containing the contents of the asset.
+async fn download_github_asset(latest: &GithubReleaseAsset) -> anyhow::Result<bytes::Bytes> {
+    let r = http_get(latest.browser_download_url.as_str()).await?;
```
cd-work, line 69:
```suggestion
    let r = http_get(&latest.browser_download_url).await?;
```

cli/src/update/unix.rs
```diff
@@ -59,14 +57,50 @@ impl Default for ApplicationUpdater {
...
+    Ok(r.bytes().await?)
+}
+
+const SUPPORTED_PLATFORMS: &[&str] = &[
+    "x86_64-unknown-linux-musl",
+    "x86_64-apple-darwin",
+    "aarch64-apple-darwin",
+];
+
+/// Determine the current platform. Error if unsupported
```
cd-work, line 79:
```suggestion
/// Determine the current platform. Error if unsupported.
```

I'm not familiar with all the code conventions yet, but seems like trailing full stop is the norm.

cli/src/update/unix.rs
```diff
@@ -59,14 +57,50 @@ impl Default for ApplicationUpdater {
...
+        "apple-darwin"
+    } else {
+        "unsupported"
+    };
+
+    let platform = format!("{}-{}", arch, os);
+    if SUPPORTED_PLATFORMS.contains(&platform.as_str()) {
+        Ok(platform)
+    } else {
+        Err(anyhow::anyhow!("unsupported platform: {}", &platform))
```
cd-work, line 103:
```suggestion
        Err(anyhow::anyhow!("unsupported platform: {}", platform))
```

cli/src/update/unix.rs
```diff
@@ -83,116 +117,36 @@ impl ApplicationUpdater {
...
-            Ok(client) => {
-                let response = client.get(url).send().await;
-
-                match response {
-                    Ok(response) => f(response).await,
-                    Err(_) => None,
-                }
+    /// Check for an update by querying the Github releases page.
+    async fn get_latest_version(&self, prerelease: bool) -> Option<GithubRelease> {
+        async fn inner(github_uri: &str, prerelease: bool) -> anyhow::Result<GithubRelease> {
```
cd-work, line 122: I'd assume an inner function was chosen here solely for `async`? Async blocks should work better without having to put functions inside of functions:

```rust
        let release = async {
            if prerelease {
                let url = format!("{}/repos/phylum-dev/cli/releases", self.github_uri);
                let r = http_get(&url).await?;
                let releases = r.json::<Vec<GithubRelease>>().await?;
                // Use the first one in the list, which should be the most recent
                releases
                    .first()
                    .cloned()
                    .ok_or_else(|| anyhow::anyhow!("no releases found"))
            } else {
                let url = format!("{}/repos/phylum-dev/cli/releases/latest", self.github_uri);
                let r = http_get(&url).await?;
                Ok(r.json::<GithubRelease>().await?)
            }
        };

        match release.await {
            Ok(ver) => {
                log::debug!("Found latest version: {:?}", ver);
                Some(ver)
            },
            Err(e) => {
                log::warn!("Failed getting latest version: {:?}", e);
                None
            }
        }
```

The match block changes here are mostly a style suggestion, I think it improves consistency between the two different match arms.

cli/src/update/unix.rs
```diff
@@ -230,119 +184,57 @@ impl ApplicationUpdater {
...
         // Get the URL for each asset from the Github JSON response in `latest`.
         debug!("Finding the github assets in the Github JSON response");
-        let bin_asset_url = self.find_github_asset(&latest, bin_asset_name)?;
-        let bash_asset_url = self.find_github_asset(&latest, shell_asset_name)?;
-        let minisign_name = format!("{}.minisig", bin_asset_name);
-        let sig_asset_url = self.find_github_asset(&latest, minisign_name.as_str())?;
+        let zip_asset =
+            self.find_github_asset(&latest, format!("{}.zip", archive_name).as_str())?;
+        let sig_asset =
+            self.find_github_asset(&latest, format!("{}.zip.minisig", archive_name).as_str())?;
```
cd-work, line 197:
```suggestion
        let zip_asset = self.find_github_asset(&latest, &format!("{}.zip", archive_name))?;
        let sig_asset =
            self.find_github_asset(&latest, &format!("{}.zip.minisig", archive_name))?;
```

cli/src/update/unix.rs
```diff
@@ -230,119 +184,57 @@ impl ApplicationUpdater {
...
-            Ok(x) => x,
-            Err(_) => {
-                return Err(std::io::Error::new(
-                    io::ErrorKind::Other,
-                    "Could not create the path for the installation",
-                ))
-            }
-        };
+        debug!("Verifying the package signature");
+        if !self.has_valid_signature(zip.as_ref(), std::str::from_utf8(sig.as_ref())?) {
```
cd-work, line 204:
```suggestion
        if !self.has_valid_signature(&zip, str::from_utf8(sig.as_ref())?) {
```

I'd suggest importing `str` here, just to keep things concise. Also means in the future nobody has to worry about whether it's worth it to start importing it now, or to also use the fully qualified path so they don't have to touch unrelated code.

cli/src/update/unix.rs
```diff
@@ -230,119 +184,57 @@ impl ApplicationUpdater {
...
+            anyhow::bail!("The update binary failed signature validation");
+        }
 
-        if !self.has_valid_signature(final_bin.as_str(), sig.as_str()) {
-            fs::remove_file(&installed_bin_path)?;
-            fs::remove_file(&installed_bash_path)?;
+        debug!("Extracting package to temporary directory");
+        let temp_dir = tempfile::tempdir()?;
+        ZipArchive::new(std::io::Cursor::new(zip))?.extract(temp_dir.path())?;
```
cd-work, line 210: I would import `Cursor` here.

cli/src/update/unix.rs
```diff
@@ -230,119 +184,57 @@ impl ApplicationUpdater {
...
 
-            return Err(std::io::Error::new(
-                io::ErrorKind::Other,
-                "Possible attack attempt! Binary changed after initial signature verification and was removed.",
-            ));
-        }
+        debug!("Running the installer");
+        let working_dir = temp_dir.path().join(archive_name);
+        let status = Command::new(working_dir.join("install.sh"))
```
cd-work, line 214: Going the safe route here makes sense, but just `"./install.sh"` should work since we're setting the working directory to `working_dir` anyway.

cd-work (changes requested):
I haven't run it myself, but these are some Rust style issues I've found.

Most of them are based on style guidelines from other projects, so please let me know if they conflict with any style guidelines in place already.

cli/src/update/unix.rs
kylewillmon: That's true, but I like the safety of the full path for now.

After #216 is completed, it would probably be good to change this to `Command::new("/bin/sh").arg("install.sh")...` to avoid relying on the executable bit being set. As of now, this could fail if `/tmp` is mounted `noexec`

cd-work (approved):

PR #227, opened 2022-04-06
Add email to `phylum auth status`

### LINKED ISSUE #174: Show current user in `auth status` response
# Overview
When a user runs `phylum auth status` we currently inform them that they are using a long-lived token. We should provide additional context and inform the user _who_ they are currently authenticated as. For example:

```
$ phylum auth status
You are currently authenticated as foo@bar.com
```

# Acceptance Criteria
- [ ] When a user runs `phylum auth status` they get a message indicating who they are authenticated as
- [ ] If the user is not authenticated, `auth status` prints a message alerting them to this fact and asking them to authenticate with `phylum auth login`

## Associated User Story
https://www.notion.so/phylum/Current-User-Context-0cde5251e45f4aa7a1764875ef51560e

cd-work:
Previously the `auth status` command would only check that the token is
present in the configuration file, however it would neither confirm its
validity nor print information about the user of that token.

By requesting the user's information from the authentication API, it's
possible to both provide more insight about the user currently logged in
and verify that the token is actually valid instead of just present.

Currently only the user's E-Mail is displayed, since this is the
information used for logging in.

Closes #174.

---

# Overview

Provide information on what this PR does and any context necessary to understand its implementation.

## Checklist

## Issue

What issue(s) does this PR close? Use the `closes #174` here.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -29,7 +29,7 @@ maplit = "1.0.2"
 minisign-verify = "0.1.8"
 nom = "6.1.2"
 open = "2.0.1"
-phylum_types = { path = "/home/chris/programming/rust/phylum-types" }
+phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "master" }
 prettytable-rs = "0.8.0"
 rand = "0.8.4"
 reqwest = { version = "0.11.3", features = ["blocking", "json", "rustls-tls"], default-features = false }
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -29,7 +29,7 @@ maplit = "1.0.2"
 minisign-verify = "0.1.8"
 nom = "6.1.2"
 open = "2.0.1"
-phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "master" }
+phylum_types = { path = "/home/chris/programming/rust/phylum-types" }
 prettytable-rs = "0.8.0"
 rand = "0.8.4"
 reqwest = { version = "0.11.3", features = ["blocking", "json", "rustls-tls"], default-features = false }
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -19,6 +19,7 @@ mod job;
 mod project;
 mod user_settings;
 
+use crate::auth::fetch_oidc_server_settings;
 use crate::auth::handle_auth_flow;
 use crate::auth::handle_refresh_tokens;
 use crate::auth::AuthAction;
@@ -177,6 +178,12 @@ impl PhylumApi {
             .get::<AuthStatusResponse>(job::get_auth_status(&self.api_uri))
             .await?
             .authenticated)
+    }
+
+    /// Get information about the authenticated user
+    pub async fn user_info(&self, auth_info: &AuthInfo) -> Result<UserInfo> {
+        let oidc_settings = fetch_oidc_server_settings(auth_info).await?;
+        self.get(oidc_settings.userinfo_endpoint.into()).await
     }
 
     /// Create a new project
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -22,7 +22,7 @@ mod user_settings;
 use crate::auth::fetch_oidc_server_settings;
 use crate::auth::handle_auth_flow;
 use crate::auth::handle_refresh_tokens;
-use crate::auth::AuthAction;
+use crate::auth::{AuthAction, UserInfo};
 use crate::config::AuthInfo;
 use crate::types::AuthStatusResponse;
 use crate::types::PingResponse;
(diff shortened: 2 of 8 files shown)


cli/Cargo.toml
```diff
@@ -29,7 +29,7 @@ maplit = "1.0.2"
 minisign-verify = "0.1.8"
 nom = "6.1.2"
 open = "2.0.1"
-phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "master" }
+phylum_types = { path = "/home/chris/programming/rust/phylum-types" }
```
cd-work, line 32: I've added the `UserInfo` to the `phylum-types` locally, since it seems like that's the correct location for HTTP request types. I'll send a PR for that if the approach taken in this PR seems reasonable.

cli/src/bin/phylum.rs
```diff
@@ -153,17 +153,24 @@ async fn handle_commands() -> CommandResult {
...
+    if let Some(matches) = matches.subcommand_matches("auth") {
+        return handle_auth(
+            config,
+            &config_path,
+            matches,
+            app_helper,
+            timeout,
+            ignore_certs,
+        )
+        .await;
```
cd-work, line 171: The `handle_auth` method itself already returns a `CommandResult`, so I think it makes a lot more sense to just propagate that upwards here.

cli/src/commands/auth.rs
```diff
@@ -67,7 +91,7 @@ pub async fn handle_auth(
             }
         }
     } else if matches.subcommand_matches("status").is_some() {
-        handle_auth_status(&config);
+        handle_auth_status(config, timeout, ignore_certs).await?;
```
cd-work, line 94: Currently only error values from `handle_auth_status` are propagated, which should be fine since it's otherwise always `Void` anyway. But generally I think it would make sense to have each of these branches return their function's value, instead of having a `CommandValue::Void.into()` returned at the end.

I've decided against refactoring this to keep this PR minimal.

cli/src/print.rs
```diff
@@ -11,26 +11,26 @@ use crate::summarize::Summarize;
...
 }
 
 #[macro_export]
 macro_rules! print_user_failure {
-    ($($tts:tt)*) => {
+    ($($tts:tt)*) => {{
         eprint!("❗ ");
         eprintln!($($tts)*);
-    }
+    }}
```
cd-work, line 33: These macros omit the double brace syntax required for using them in things like match arms without trailing semicolon. It's usually good practice to allow that to make macros feel more like native Rust syntax. So I've added them for everything here.

cli/src/commands/auth.rs
```diff
@@ -27,12 +27,34 @@ async fn handle_auth_login(mut config: Config, config_path: &Path) -> CommandRes
...
+) -> CommandResult {
+    // Create a client with our auth token attached.
+    let api = PhylumApi::new(
+        &mut config.auth_info,
+        &config.connection.uri,
+        timeout,
+        ignore_certs,
+    )
+    .await
+    .map_err(|err| anyhow!("Error creating client").context(err))?;
```
kylewillmon, line 43: The error and context are backwards here. `err` is the error and `"Error creating client"` should be the context.

```suggestion
    .map_err(|err| err.context("Error creating client").into())?;
```

maxrake (commented):
I viewed the code...but Rust is still not something I feel confident approving. I tried the artifacts from this PR and they indeed do what was asked in the corresponding ticket:
```
❯ ph auth status
✅ Currently authenticated as 'charles@phylum.io' with long lived refresh token
```
LGTM...but only leaving a comment instead of an approval...to let the Rust SMEs weigh in.

cli/src/api/mod.rs
```diff
@@ -179,6 +180,12 @@ impl PhylumApi {
             .authenticated)
     }
 
+    /// Get information about the authenticated user
+    pub async fn user_info(&self, auth_info: &AuthInfo) -> Result<UserInfo> {
```
kylewillmon, line 184: The `auth_info` here will always be the same `auth_info` that was passed in to `PhylumApi::new(...)`. I'm not sure there's an easy way to refactor and fix that, but it feels odd to have to pass in the same object twice.

cli/src/auth/oidc.rs
```diff
@@ -244,3 +245,15 @@ pub async fn handle_refresh_tokens(
     let oidc_settings = fetch_oidc_server_settings(auth_info).await?;
     refresh_tokens(&oidc_settings, refresh_token).await
 }
+
+/// Represents the userdata stored for an authentication token.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct UserInfo {
```
kylewillmon, line 251: `serde` will fail the deserialization if any of these fields are not present. But we're only actually using 1 field at the moment. Should we be more permissive here?

cli/src/commands/auth.rs
kylewillmon: I agree. That refactor is needed, but that can be a separate PR.

cli/src/print.rs
kylewillmon: Good catch.

cli/src/commands/auth.rs
```diff
@@ -27,12 +27,34 @@ async fn handle_auth_login(mut config: Config, config_path: &Path) -> CommandRes
...
-        print_user_success!("Currently authenticated with long lived refresh token");
-    } else {
-        print_user_warning!("User is not currently authenticated");
+pub async fn handle_auth_status(
+    mut config: Config,
+    timeout: Option<u64>,
+    ignore_certs: bool,
+) -> CommandResult {
+    // Create a client with our auth token attached.
+    let api = PhylumApi::new(
```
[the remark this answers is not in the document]
cd-work: Hm that's a good point. I was wondering if I should make use of API at all or side-step it and instead try to manually send a request. But I didn't want to duplicate any code.

Would you suggest sending the Request without the help of `PhylumApi` or instead maybe just check first if there is a token and early-return in case one is not present yet.

Alternatively a parameter could be added to `PhylumApi` to perform auth, but I don't really like that myself.

cli/src/commands/auth.rs
cd-work: I actually just copied that from `phylum.rs`, so then it's probably wrong there too.

cli/src/auth/oidc.rs
cd-work: That seems reasonable, I guess it doesn't hurt to just wrap it in `Option` to show these still might be there but we don`t **need** them? That way when someone else makes changes in the future they know which fields should be present and we don't have to remove them completely.

kylewillmon (approved):

PR #228, opened 2022-04-07
Add test and documentation for setting the Phylum API token through the PHYLUM_API_KEY environment variable.

### LINKED ISSUE #157: Support passing in offline_access token via env var
Right now it has to be set in settings yaml

We should support passing it in via env var.

mdx97:
# Overview

@kylewillmon [mentioned](https://github.com/phylum-dev/cli/issues/157#issuecomment-1071889106) in #157 that there has already been code added to support passing the Phylum API token through environment variable, but that some tests and documentation should be added. This PR does just that!

## Checklist

## Issue

closes #157 (?)


cli/src/config.rs
```diff
@@ -222,4 +223,19 @@ mod tests {
...
+    fn test_pass_api_key_through_env() {
+        write_test_config();
+        let temp_dir = temp_dir();
+        let test_config_file = temp_dir.as_path().join("test_config");
+        env::set_var("PHYLUM_API_KEY", "ENV VARIABLE TOKEN");
+
+        let config: Config = read_configuration(&test_config_file).unwrap();
+
+        assert_eq!(
+            String::from(&config.auth_info.offline_access.unwrap()),
```
mdx97, line 237: This was the most immediately obvious way of doing string comparison on the `RefreshToken` type. If there is a better / more idiomatic way of doing this let me know!

cd-work: I don't think there's a better way to do this without touching `phylum-types`. Ideally that would have `PartialEq` derived, but unfortunately it does not.

Though you can skip the `String::from` for the token:

```rust
        assert_eq!(
            String::from(&config.auth_info.offline_access.unwrap()),
            token
        );
```

cli/src/config.rs
```diff
@@ -222,4 +223,19 @@ mod tests {
         let config: Config = parse_config(&test_config_file).unwrap();
         assert_eq!(config.request_type, PackageType::Npm);
     }
+
+    #[test]
+    fn test_pass_api_key_through_env() {
+        write_test_config();
+        let temp_dir = temp_dir();
+        let test_config_file = temp_dir.as_path().join("test_config");
+        env::set_var("PHYLUM_API_KEY", "ENV VARIABLE TOKEN");
```
cd-work, line 232: I'd suggest extracting the `"ENV VARIABLE TOKEN"` string into a variable so it can be reused for comparison later. That should make it easier to change in the future in case a random token string isn't valid anymore.

```suggestion
        let token = "ENV VARIABLE TOKEN";
        env::set_var("PHYLUM_API_KEY",  token);

        write_test_config();
        let temp_dir = temp_dir();
        let test_config_file = temp_dir.as_path().join("test_config");
```

cd-work (changes requested):

DanielJoyce:
I don't have anything else to add beyond the other review. In the future I'd like us to move to using Figment which is the config crate used by Rocket, and supports merging env from different sources, so we don't have to roll our own support.

cli/src/config.rs
cd-work: The latest version of `phylum-types` now does implement `PartialEq`, so it should be possible to clean this up even further.

cd-work (approved):

cd-work (approved):

PR #217, opened 2022-03-28
Create phylum auth token command

mdx97:
# Overview

This PR implements the `phylum auth token` command to print out the user's current long-lived token that is stored in their `settings.yaml` file.

I will also be creating a companion PR in the `phylum-types` repository to implement `Display` for `RefreshToken`.
## Checklist

- [x] Does this PR have an associated issue?
- [x] Have you ensured that you have met the expected acceptance criteria?
- [ ] Have you created sufficient tests?
- [x] Have you updated all affected documentation?

## Issue
closes #175

PR #234, opened 2022-04-07
Follow development branch of phylum-types

kylewillmon:
Tracking the `master` branch of `phylum-types` creates an over-complicated workflow. We are already using `Cargo.lock` to lock the dependency to a specific revision until `cargo update` is run.

It is much more useful for us to develop against the `development` branch so that we can quickly find and fix any problems that might arise. This PR makes that switch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,7 +38,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --manifest-path Cargo.toml --all-features --all --tests --examples -- -D clippy::all -D warnings
+          args: --locked --all-features -- -D warnings
 
   test-matrix:
     strategy:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1311,7 +1311,7 @@ dependencies = [
 [[package]]
 name = "phylum_types"
 version = "0.1.0"
-source = "git+https://github.com/phylum-dev/phylum-types?branch=master#582271f37bbceac2304b9839af5bc16449bf0062"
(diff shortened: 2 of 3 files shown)


kylewillmon:
To list a few recent examples of work that has been hindered by our reliance on `master`:

* #217 has been blocked by changes that have hit the `development` branch of `phylum-types`, but are not yet in `master`
* In #227, the `UserInfo` type was added to this crate instead of `phylum-types` because it is seen as difficult to get code into `master`
* [These comments](https://github.com/phylum-dev/cli/pull/228#discussion_r844586921) on #228 discuss working around issues with `phylum-types` that would be simple to resolve if it just meant a PR to `development`

The alternative is that every code change in `phylum-types` requires 2 PRs (one to get to `development` and then another to get to `master`). And, indeed, this has been seen:

* phylum-dev/phylum-types#12 followed a day later by phylum-dev/phylum-types#13
* phylum-dev/phylum-types#14 followed a day later by phylum-dev/phylum-types#15
* phylum-dev/phylum-types#16 followed 10 days later by phylum-dev/phylum-types#18

It's hard to see how this double-PR system helps anything. So I'd like to get rid of it.

eeclfrei (approved):
I agree it's not buying us much and this will improved velocity.

maxrake (approved):
I agree that the development/master staging/production split is not the best approach for all situations...this being one of them. Maybe if all code was in a mono-repo...but that has other concerns. With all the various repos and interactions between them, it can be tricky to maintain synchronicity across the stack for each environment.

PR #235, opened 2022-04-08
Migrate install script to POSIX sh

### LINKED ISSUE #216: Make install script portable

cd-work:
This migrates the phylum CLI install script to posix /bin/sh, which will
allow for execution on more platforms.

Closes #216.

### DIFF
--- a/cli/src/install.sh
+++ b/cli/src/install.sh
@@ -1,31 +1,30 @@
-#!/bin/bash
+#!/bin/sh
 
 RED='\033[0;31m'
 GREEN='\033[0;32m'
 NC='\033[0m'
 
 # Don't continue after failure:
-set -euo pipefail
+set -eu
 
 error() {
-    echo -e "${RED}    ERROR${NC} ${1}"
+    printf "%b    ERROR%b %s\n" "${RED}" "${NC}" "${1}"
 }
 
 success() {
-    echo -e "${GREEN}    OK${NC} ${1}"
+    printf "%b    OK%b %s\n" "${GREEN}" "${NC}" "${1}"
 }
 
 banner() {
-    echo -e "\n    ${GREEN}phylum-cli${NC} installer\n"
+    printf "\n    %bphylum-cli%b installer\n\n" "${GREEN}" "${NC}"
 }
 
 # Get the platform name.
 get_platform() {
-    local platform_str
     platform_str=$(uname)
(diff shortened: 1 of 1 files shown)


cli/src/install.sh
```diff
@@ -126,13 +125,12 @@ copy_files() {
...
+cd "$(dirname "$0")"
 banner
 create_directory
 copy_files
 patch_bashrc
 patch_zshrc
-popd >/dev/null
 
 success "Successfully installed phylum."
 rc_file=$(set -e; get_rc_file)
```
kylewillmon, line 136: We can remove `set -e` in our command substitutions now. That was only a problem for `/bin/bash` (as pointed out by [SC2311](https://github.com/koalaman/shellcheck/wiki/SC2311))

kylewillmon (approved):
LGTM. You could go a bit further and remove `set -e` in command substitutions. They're unnecessary now, but they're also not doing much harm, so I approve either way.

cli/src/install.sh
```diff
@@ -1,31 +1,30 @@
...
+#!/bin/sh
 
 RED='\033[0;31m'
 GREEN='\033[0;32m'
 NC='\033[0m'
 
-# Don't continue after failure:
-set -euo pipefail
+# Fail when trying to expand unset variables
+set -u
```
kylewillmon, line 8: Sorry if my comment wasn't clear. We still need the "set -e" here. Just not in the command substituon "$(...)" elsewhere in the flie.


```suggestion
set -eu
```

cli/src/install.sh
kylewillmon, line 136:
```suggestion
rc_file=$(get_rc_file)
```

kylewillmon (changes requested):

kylewillmon (approved):

kylewillmon (approved):

PR #236, opened 2022-04-08
Add uninstall script

### LINKED ISSUE #206: Provide support for uninstall
# Overview
As a user,
I would like to be able to uninstall the phylum CLI tool,
in order to bring my system back to the state it was in before the phylum CLI tool was installed.

# Acceptance Criteria
- [ ] A command is available to automatically uninstall or, at a minimum, documentation provided to manually uninstall
- [ ] No traces of the phylum CLI tool remain after uninstall
- [ ] All user config files (e.g., profile and rc files) are updated to remove any entries added by the phylum CLI tool install
- [ ] Documentation is updated to describe the process

## Associated User Story
*EDIT:* [CLI Uninstall Script](https://www.notion.so/phylum/CLI-Uninstall-Script-c52c9fc6f486419ab220b3ca1da15afb)

cd-work:
This adds a new script to the published archives which allows for easy
uninstallation of the phylum CLI.

This implementation intentionally deviates from the behavior suggested
in #206, since removal of configuration files is usually not expected by
the user. So only files generated by the installer will be removed.

To ensure it's still possible to clean up all phylum CLI files including
configurations, the `--purge` option is available.

Closes #206.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -104,6 +104,7 @@ jobs:
             archive=$(echo ${archive} | sed -e "s/\/$//")
             cp -R shell-completions ${archive}/completions
             cp cli/cli/src/install.sh ${archive}/install.sh
+            cp cli/cli/src/phylum_uninstall.sh ${archive}/phylum_uninstall.sh
             chmod a+x ${archive}/phylum
             zip -r ${archive}.zip ${archive}
           done
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -103,8 +103,8 @@ jobs:
           do
             archive=$(echo ${archive} | sed -e "s/\/$//")
             cp -R shell-completions ${archive}/completions
-            cp cli/cli/src/install.sh ${archive}/install.sh
-            cp cli/cli/src/phylum_uninstall.sh ${archive}/phylum_uninstall.sh
+            cp cli/scripts/install.sh ${archive}/install.sh
+            cp cli/scripts/phylum-uninstall.sh ${archive}/phylum-uninstall.sh
             chmod a+x ${archive}/phylum
             zip -r ${archive}.zip ${archive}
           done
--- a/README.md
+++ b/README.md
@@ -69,6 +69,20 @@ The command line interface (CLI) allows users to submit their project package de
    phylum analyze <package-lock-file.ext>
    ```
 
+## Uninstalling the Phylum CLI
+
+The Phylum CLI can be uninstalled using the `phylum_uninstall.sh` script, which
+should be in your `$PATH` and at `~/.phylum/phylum_uninstall.sh` if you followed
+the installation instructions:
+
+    ```
+    # Uninstall all files created by the installer
+    phylum_uninstall.sh
+
+    # Remove everything, including configuration files
+    phylum_uninstall.sh --purge
+    ```
+
 ---
 
 ## Questions/Issues
--- a/scripts/phylum-uninstall.sh
+++ b/scripts/phylum-uninstall.sh
@@ -37,12 +37,12 @@ remove_from_path_and_alias() {
 
     if grep -q ".phylum:\$PATH" "${rc_path}"; then
         # shellcheck disable=SC2016 # Expanding variables here would be a bad idea
-        sed -i '/^export PATH="$HOME\/.phylum:$PATH"$/d' "${rc_path}"
+        sed -i'' '/^export PATH="$HOME\/.phylum:$PATH"$/d' "${rc_path}"
         success "Removed ${HOME}/.phylum/ from ${shell}'s \$PATH."
     fi
 
     if grep -q 'alias ph=' "${rc_path}"; then
-        sed -i "/^alias ph='phylum'$/d" "${rc_path}"
+        sed -i'' "/^alias ph='phylum'$/d" "${rc_path}"
         success "Removed ph alias for phylum from ${shell}."
     fi
 }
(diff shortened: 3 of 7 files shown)


cd-work:
Since all of `~/.phylum` is added to the path, this will automatically also include the uninstall script, which should be fine. But as a result I thought it would be unwise just to call it `uninstall.sh`. 

I've decided to place the uninstall script next to the `install.sh`, but I personally find it somewhat weird to find these in the Rust `src` directory. I actually had to `find` through the repo to locate it because that was the last place I'd expect it in. So if there's a desire to move the install script to a different location, this might be a good time?

Currently the uninstallation process is only documented in the README.md. However the quickstart guide seems to be almost identical to the readme so it might make sense to include it there too, even though removal of the CLI is not technically something required to "start"?

kylewillmon:
> I've decided to place the uninstall script next to the `install.sh`, but I personally find it somewhat weird to find these in the Rust `src` directory. I actually had to `find` through the repo to locate it because that was the last place I'd expect it in. So if there's a desire to move the install script to a different location, this might be a good time?

I fully support moving `install.sh` to a more reasonable location. Either the root or a new top-level `scripts` directory makes the most sense to me.
 
> Currently the uninstallation process is only documented in the README.md. However the quickstart guide seems to be almost identical to the readme so it might make sense to include it there too, even though removal of the CLI is not technically something required to "start"?

We definitely need to document uninstall somewhere in the `docs` folder since those are published to [`docs.phylum.io`](https://docs.phylum.io/docs). Maybe `docs/quickstart.md` (which should be kept in sync with the top-level `README.md`) or maybe in a new `docs/uninstall.md` (which would probably mean removing it from the top-level `README.md`)

peterjmorgan:
> Since all of `~/.phylum` is added to the path, this will automatically also include the uninstall script, which should be fine. But as a result I thought it would be unwise just to call it `uninstall.sh`.

That makes sense, but perhaps uninstall-phylum.sh might be better? (just thinking about tab completion)

cd-work:
> That makes sense, but perhaps uninstall-phylum.sh might be better? (just thinking about tab completion)

Yeah that makes sense, I think dash is usually more common for command names.

> I fully support moving install.sh to a more reasonable location. Either the root or a new top-level scripts directory makes the most sense to me.

I personally like script subdirectories, just to keep the root clean. So I'll move them there for now.

> We definitely need to document uninstall somewhere in the docs folder since those are published to docs.phylum.io. Maybe docs/quickstart.md (which should be kept in sync with the top-level README.md) or maybe in a new docs/uninstall.md (which would probably mean removing it from the top-level README.md)

I think just moving it to quickstart is probably the best choice. Just as a little sidenote at the end so users can read it during installation and hopefully remember for the future.

maxrake (changes requested):
Posting this now for action...but will continue to review...

.github/workflows/release.yml
```diff
@@ -103,7 +103,8 @@ jobs:
           do
             archive=$(echo ${archive} | sed -e "s/\/$//")
             cp -R shell-completions ${archive}/completions
-            cp cli/cli/src/install.sh ${archive}/install.sh
+            cp cli/scripts/install.sh ${archive}/install.sh
+            cp cli/scripts/phylum-uninstall.sh ${archive}/phylum-uninstall.sh
```
maxrake, line 107:
```suggestion
            cp cli/scripts/uninstall-phylum.sh ${archive}/uninstall-phylum.sh
```
Suggestion is for tab completion...b/c typing `ph<tab>` or `phylum<tab>` should show the CLI binary options and not other executables.

There is a case that could be made...for another issue...to provide the uninstall feature within the CLI binary itself instead of a standalone script. I've seen other tools do this...and then they only add a single line/entry to the *rc conf files.

README.md
```diff
@@ -69,6 +69,20 @@ The command line interface (CLI) allows users to submit their project package de
    phylum analyze <package-lock-file.ext>
    ```
 
+## Uninstalling the Phylum CLI
+
+The Phylum CLI can be uninstalled using the `phylum_uninstall.sh` script, which
+should be in your `$PATH` and at `~/.phylum/phylum_uninstall.sh` if you followed
```
maxrake, line 75: Reminder to ensure this section has file names that match whatever the ultimate name is: `uninstall-phylum.sh` is the current suggestion.

Another reminder is that an odd quirk of the way the docs are set up in this repo...and published to a README.com site at <https://docs.phylum.io/docs/welcome> ...is that the `README.md` file should be manually kept in sync with the `docs/quickstart.md` file. Looks like that is the case now, but just repeating here instead of also adding a comment to that file.

scripts/phylum-uninstall.sh
```diff
@@ -0,0 +1,117 @@
...
+
+    remove_from_path_and_alias "${rc_path}" "zsh"
+}
+
+cleanup_bashrc() {
+    rc_path="${HOME}/.bashrc"
+
+    if grep -q '.phylum/completions/phylum.bash' "${rc_path}"; then
+        # shellcheck disable=SC2016 # Expanding variables here would be a bad idea
+        sed -i '/^source $HOME\/.phylum\/completions\/phylum.bash$/d' "${rc_path}"
```
maxrake, line 70: There is an error on this line...but not familiar enough with `sed` to say what:
```sh
❯ uninstall-phylum.sh --purge

    phylum-cli uninstaller

    OK Removed all files.
sed: 1: "/Users/coggins/.bashrc": command c expects \ followed by text
```
BTW...the name `uninstall-phylum.sh` here is just b/c I renamed the file in the `install.sh` script while trying all this out.

maxrake (changes requested):
There is an error in the uninstall script that halts processing. There may be more to fix after proceeding past that point. Submitting review now for changes and will continue testing...either on my own now or as changes are made.

.github/workflows/release.yml
cd-work: Hm I think worrying about completions at all is probably an indication that it's not a great idea to have this in the `$PATH` to begin with?

> There is a case that could be made...for another issue...to provide the uninstall feature within the CLI binary itself instead of a standalone script. I've seen other tools do this...and then they only add a single line/entry to the *rc conf files.

I've considered this but I don't think that's the best solution considering users already interact with an `inshall.sh`, so it's kinda natural that the archive would also come with an `uninstall.sh` rather than putting that inside `phylum` directly.

kylewillmon: `install.sh` is going to go through some changes and users may not directly interact with it for very long. The end goal is something like:

```
curl http://get.phylum.io | sh -
```

See [this user story](https://www.notion.so/phylum/Install-Script-Improvement-880c22380d4041b388b7937a603ad6aa) for some of the details.

Plus, I think `phylum uninstall` matches well with the current `phylum update` command.

cd-work:
Closed in favor of #239.

PR #238, opened 2022-04-08
Add Python poetry.lock support

### LINKED ISSUE #237: Add poetry support
# Overview

As a Phylum user, I want to submit a poetry.lock file generated using the Poetry tool for analysis, so that I do not have to export a requirements.txt file.

# Acceptance Criteria
- [ ] CLI tool correctly identifies poetry.lock file format
- [ ] CLI tool correctly extracts all relevant package/version pairs

## Associated User Story
https://www.notion.so/phylum/Poetry-Support-3bd330c59d4b419e832842d02ae890ee

cd-work:
This adds support for analyzing the Python `poetry.lock` package lock
format.

Since this format includes a metadata field with a version indicator, a
warning is emitted when a breaking change that isn't supported by
`phylum` was detected.

While it would be possible to abort completely, it's not unlikely that a
format change does not touch the information required by `phylum` at
all. So we attempt to continue anyway and just let the user know that
some packages might not have been analyzed.

Closes #237.

---

## Checklist

### DIFF
--- a/cli/src/commands/lock_files.rs
+++ b/cli/src/commands/lock_files.rs
@@ -147,7 +147,7 @@ mod tests {
             ("tests/fixtures/requirements.txt", PackageType::PyPi),
             ("tests/fixtures/Pipfile", PackageType::PyPi),
             ("tests/fixtures/Pipfile.lock", PackageType::PyPi),
-            ("tests/fixtures/Poetry.lock", PackageType::PyPi),
+            ("tests/fixtures/poetry.lock", PackageType::PyPi),
         ];
 
         for (file, expected_type) in &test_cases {
--- a/cli/src/commands/lock_files.rs
+++ b/cli/src/commands/lock_files.rs
@@ -40,6 +40,12 @@ pub fn try_get_packages(path: &Path) -> Option<(Vec<PackageDescriptor>, PackageT
         return packages.ok().map(|pkgs| (pkgs, PackageType::PyPi));
     }
 
+    let packages = Poetry::new(path).ok()?.parse();
+    if packages.is_ok() && !packages.as_ref().unwrap().is_empty() {
+        log::debug!("Submitting file as type poetry lock");
+        return packages.ok().map(|pkgs| (pkgs, PackageType::PyPi));
+    }
+
     let packages = Pom::new(path).ok()?.parse();
     if packages.is_ok() && !packages.as_ref().unwrap().is_empty() {
         log::debug!("Submitting file as type pom xml");
@@ -98,6 +104,10 @@ pub fn get_packages_from_lockfile(path: &str) -> Option<(Vec<PackageDescriptor>,
             let parser = PipFile::new(path).ok()?;
             parser.parse().ok().map(|pkgs| (pkgs, PackageType::PyPi))
         }
+        "poetry.lock" => {
+            let parser = Poetry::new(path).ok()?;
+            parser.parse().ok().map(|pkgs| (pkgs, PackageType::PyPi))
+        }
         "effective-pom.xml" => {
             let parser = Pom::new(path).ok()?;
             parser.parse().ok().map(|pkgs| (pkgs, PackageType::Maven))
@@ -137,6 +147,7 @@ mod tests {
             ("tests/fixtures/requirements.txt", PackageType::PyPi),
             ("tests/fixtures/Pipfile", PackageType::PyPi),
             ("tests/fixtures/Pipfile.lock", PackageType::PyPi),
+            ("tests/fixtures/Poetry.lock", PackageType::PyPi),
         ];
 
         for (file, expected_type) in &test_cases {
--- a/cli/src/lockfiles/python.rs
+++ b/cli/src/lockfiles/python.rs
@@ -111,7 +111,7 @@ impl Parseable for Poetry {
             );
         }
 
-        Ok(lock.packages.drain(..).map(Package::into).collect())
+        Ok(lock.packages.drain(..).map(PackageDescriptor::from).collect())
     }
 }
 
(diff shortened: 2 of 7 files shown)


cli/src/lockfiles/python.rs
```diff
@@ -88,6 +90,60 @@ impl Parseable for PipFile {
...
+            package_type: PackageType::PyPi,
+        }
+    }
+}
+
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "kebab-case")]
+struct PoetryMetadata {
+    lock_version: String,
+}
```
cd-work, line 145: I wasn't *completely* sure where to put this. It's *kinda* like a parser, but the `parsers.rs` module is a bit more hand-rolled than this.

I'd personally make the `parsers.rs` file into a directory, extract all the modules into separate files, then put this in `parsers/poetry.rs`. But maybe there's also some better place for it?

cli/tests/fixtures/poetry.lock
```diff
@@ -0,0 +1,1123 @@
+[[package]]
```
cd-work, line 1: Not sure if there's any recommendation on what kind of input files are usually used for tests. I just copied the `poetry.lock` from `poetry` itself. However this is kinda big and the density of interesting edgecases is probably not particularly high.

cd-work (commented):
This was actually surprisingly easy, so I'm not sure if I missed something but it seems to work fine for the lockfiles I've tested. Let me know if there's something I forgot to handle.

cli/tests/fixtures/poetry.lock
maxrake: Not sure either...but this file should work for now. It can be changed later, as we learn more about interesting edge cases.

cli/src/lockfiles/python.rs
```diff
@@ -169,4 +225,25 @@ mod tests {
...
+                assert_eq!(pkg.version, "0.10.2");
+                assert_eq!(pkg.package_type, PackageType::PyPi);
+            } else if pkg.name == "certifi" {
+                assert_eq!(pkg.version, "2021.10.8");
+                assert_eq!(pkg.package_type, PackageType::PyPi);
+            } else if pkg.name == "html5lib" {
+                assert_eq!(pkg.version, "1.1");
+                assert_eq!(pkg.package_type, PackageType::PyPi);
+            }
+        }
```
kylewillmon, line 247: The `if` inside the loop scares me, because it allows completely skipping the asserts. Imagine if the parser accidentally failed to set `name`: Then we'd iterate through all the packages and never check anything because `pkg.name` is never set to `"toml"`, `"certifi"`, or `"html5lib"`.

Using something like `.find(...).expect()` to ensure 

```suggestion
        let toml = pkgs
            .iter()
            .find(|p| p.name == "toml")
            .expect("Package 'toml' not found");
        assert_eq!(toml.version, "0.10.2");
        assert_eq!(toml.package_type, PackageType::PyPi);

        let certifi = pkgs
            .iter()
            .find(|p| p.name == "certifi")
            .expect("Package 'certifi' not found");
        assert_eq!(certifi.version, "2021.10.8");
        assert_eq!(certifi.package_type, PackageType::PyPi);

        let html5lib = pkgs
            .iter()
            .find(|p| p.name == "html5lib")
            .expect("Package 'html5lib' not found");
        assert_eq!(html5lib.version, "1.1");
        assert_eq!(html5lib.package_type, PackageType::PyPi);
```

It looks like you just copied the style of tests from `parse_pipfile()` and `lock_parse_pipfile()` above. They have the same problem...

You could extract that pattern into an `assert_package(...)` function if you want to fix all of them. And then use something like this in all of those tests:
```
assert_package(pkgs, "toml", "0.10.3", PackageType::PyPi);
```

cli/src/commands/lock_files.rs
```diff
@@ -137,6 +147,7 @@ mod tests {
             ("tests/fixtures/requirements.txt", PackageType::PyPi),
             ("tests/fixtures/Pipfile", PackageType::PyPi),
             ("tests/fixtures/Pipfile.lock", PackageType::PyPi),
+            ("tests/fixtures/Poetry.lock", PackageType::PyPi),
```
maxrake, line 150:
```suggestion
            ("tests/fixtures/poetry.lock", PackageType::PyPi),
```
...to match the name in source and also the way it gets generated by `poetry` itself.

maxrake (commented):
Reviewed the parts I understand...calling for support from @kylewillmon for the Rust parts.

cli/src/lockfiles/python.rs
```diff
@@ -88,6 +90,60 @@ impl Parseable for PipFile {
...
+        // Warn if the version of this lockfile might not be supported.
+        if !lock.metadata.lock_version.starts_with("1.") {
+            log::warn!(
+                "Expected poetry lockfile version ^1.0.0, found {}. \
+                Attempting to continue, but results might be inaccurate.",
+                lock.metadata.lock_version
+            );
+        }
+
+        Ok(lock.packages.drain(..).map(Package::into).collect())
```
kylewillmon, line 114: I think this would be more readable using `PackageDescriptor::from`:

```suggestion
        Ok(lock.packages.drain(..).map(PackageDescriptor::from).collect())
```

I'm guessing the line ended up like this because the natural way to write it is `.map(|p| p.into())`, but clippy doesn't like that. This is one place where I think clippy is wrong....

kylewillmon (dismissed):

cli/src/lockfiles/python.rs
kylewillmon: It's not much code, so I think it's fine to leave it here.

But if you want to make `parsers.rs` into a directory, I'm okay with that, too.

cli/src/lockfiles/python.rs
cd-work: As you've mentioned, I just copied the existing tests. But I agree with you that it's too easy to pass these incorrectly. I'll add a function and fix the other tests too.

cli/src/lockfiles/python.rs
cd-work: The line ended up that way because it's shorter. I had it as `PackageDescription::from` originally.

Though I do agree that it's more descriptive to show the destination rather than the source type. So it's probably easier for readers to go for the longer solution.

cli/src/lockfiles/python.rs
cd-work: Maybe keep it like this for now and save the complete refactor for a later date.

kylewillmon (approved):

kylewillmon (approved):

PR #240, opened 2022-04-08
Merge to `master` for `v2.0.0` tag

eeclfrei:


furi0us333 (dismissed):

maxrake (approved):
All of this code has already been reviewed previously...
Even still...as a sanity check, a [v2.0.0-rc1 tag](https://github.com/phylum-dev/cli/releases/tag/v2.0.0-rc1) was created on the latest commit that is part of this review. Then, the `TestGHA` test framework was run against that tag and it was discovered that there were some issues with the `install-phylum-latest-action`. Those were identified and fixed in a group setting...with @louislang and @eeclfrei viewing while @maxrake made the changes. With the updates, the CLI and the `phylum-analyze-pr-action` are working as expected now. Therefore, this review is getting a LGTM.

PR #243, opened 2022-04-11
Use git_version for version numbers

### LINKED ISSUE #124: `version` does not display release candidate versions
**Describe the bug**
The `version` command does not display release candidate versions.

**To Reproduce**
1. Install a pre-release version of the phylum CLI
2. Run `phylum version`
3. Observe that only the major.minor.patch semver is shown

**Expected behavior**
Display the actual version of the software, including any pre-release semvers

**Screenshots**
image

kylewillmon:
# Overview

Use [`git_version`](https://docs.rs/git-version/latest/git_version/) to provide the version number for the CLI.

If the CLI is built from a tagged commit, it will display the tag. Otherwise, it displays the most recent tag, followed by the number of commits difference and the hash of the current commit. If changes have been made to the working directory, a `-modified` suffix is added. Here is an example from this branch:

```
❯ ./target/debug/phylum version
✅ phylum (Version v2.0.0-4-gd43325e)
```

## Checklist

## Issue

Fixes #124

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,7 +56,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout the repo
-        uses: actions/checkout@master
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -753,6 +753,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "git-version"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f6b0decc02f4636b9ccad390dcbe77b722a77efedfa393caf8379a51d5c61899"
+dependencies = [
+ "git-version-macro",
+ "proc-macro-hack",
+]
+
+[[package]]
+name = "git-version-macro"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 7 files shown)


maxrake:
A pre-release version was created from this branch...as [v2.0.1-rc1](https://github.com/phylum-dev/cli/releases/tag/v2.0.1-rc1) and the artifacts from that version were installed to test this bug fix:

```sh
❯ ph version
✅ phylum (Version d43325e)
```

This output does not appear to meet the expected behavior from #124:
> Display the actual version of the software, including any pre-release semvers

The expected version output for this test was `Version 2.0.1-rc1`.

maxrake:
The current working assumption is that the `actions/checkout` action in the workflow is not fetching enough history for the `git_version` crate to find the correct (any?) tag. It might be as simple as using a [fetch-depth of 0](https://github.com/actions/checkout#Fetch-all-history-for-all-tags-and-branches).
Here is a [GitHub issue for that action](https://github.com/actions/checkout/issues/217#:~:text=This%20happens%20because%20by%20default,will%20not%20fetch%20the%20tags.) where the issue and possible solutions were first discussed.

maxrake (changes requested):
Viewed the Rust files...seems okay...

Tested the artifacts and found the version displayed when building in CI is only showing the hash. Suggestions for possible causes and fixes are provided in other comments.

kylewillmon:
I've added `fetch-depth: 0` so that we get correct versions in CI builds.

Additionally, I've updated it to pass the `--tags` flag to `git describe` so that lightweight tags are considered as well as annotated tags.

Finally, I am intentionally not passing `--always` so that if someone actually does a shallow checkout, the will instead get the `cargo:` version like so:
```
❯ ./target/debug/phylum version
✅ phylum (Version cargo:2.0.1)
```

maxrake (approved):
Works as expected now.
LGTM

PR #244, opened 2022-04-11
ci: Run doc workflow on tag instead of master

kylewillmon:
Publish the docs when a new release is tagged rather than when a merge to `master` occurs. This is required now that we no longer have a `master` branch.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,13 +4,14 @@ name: Sync `docs` directory to ReadMe
 
 on:
   push:
-    branches:
-      - master
+    tags:
+      - '*'
   
 jobs:
   Update-Documentation:
     name: Update the documentation
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


maxrake (dismissed):
It should work as written, but a suggestion was made to be more specific about _when_ the docs workflow will trigger.

maxrake:
Now we just need to rename `development` to `main`...

maxrake (approved):
LGTM

furi0us333:
> Question for @furi0us333: is the `v1-latest` version specified to the `rdme` command meant to be tied to anything? That is, should it be `v2-latest` now to match the CLI release version or is `v1` referring to a generic documentation version that may never change?

Good question. I am sure an argument can be made either way, but my gut says it is not worth trying to keep up with CLI versioning and instead let it represent the docs version. That is, if we overhaul the docs, we can bump the version.

PR #247, opened 2022-04-12
Add yarn v2 lockfile support

### LINKED ISSUE #246: Yarn v2 lockfile support
# Overview

The new yarn v2 version updates the lockfile to a yaml-based format. Support for this new format should be added.

# Acceptance Criteria

- [ ] CLI tool correctly identifies yarn.lock version 2 file format
- [ ] CLI tool correctly extracts all relevant package/version pairs from yarn.lock version 2 file
- [ ] CLI tool continues to correctly identify yarn.lock version 1 file format
- [ ] CLI tool continues to correctly extract all relevant package/version pairs from yarn.lock version 1 file

## Associated User Story

https://www.notion.so/phylum/Yarn-Version-2-Support-7ddae3c74b934242a115a4ed394149ff

cd-work:
This adds support for the new yaml-based yarn lockfile introduced with
yarn version 2.

Similar to the new package-lock.json, the yarn lockfile now annotates
the different packages with where they were resolved from. Currently
only packages that resolve to npm are submitted for analysis.

To detect which version of the yarn lock is present, the entire file is
first attempted to parse with `serde_yaml` and if that fails, the v1
parser is used as fallback instead.

Closes #246.


maxrake (approved):
The code seems fine. I'm still not a Rust reviewer and wouldn't mind if you _added_ (but keep me on) someone who is for any PRs that include Rust code changes.

One suggestion is to add an entry to the `docs/analyzing-dependencies.md` file...for `yarn.lock` as a supported lock file type. It might even be nice to note there somehow that both the current and legacy version are supported.

Also, I am a bit curious how the `phylum-analyze-pr-action` included support for both versions of `yarn.lock` files if the CLI didn't. Maybe it was a pre-emptive step and, with the changes in this PR, the GHA will actually work now. Adding a test case for `yarn.lock` v2 file formats in the `TestGHA` repo might help to prove that point. A [issue 25 in that repo](https://github.com/phylum-dev/TestGHA/issues/25) was created to track it.

cd-work:
@kylewillmon can you give the Rust portion a review?

cli/src/lockfiles/javascript.rs
```diff
@@ -52,8 +52,38 @@ impl Parseable for YarnLock {
...
+                    .get(&"resolution".into())
+                    .and_then(|package| package.as_str())
+                    .and_then(|package| package.rsplit_once("@npm:"))
+                    .map(|(name, _version)| name);
+
+                let version = package
+                    .get(&"version".into())
+                    .and_then(|version| version.as_str());
+
+                name.zip(version)
```
kylewillmon, line 79: This is a **very** permissive that will just skip any package that it can't parse.

It's probably fine, but it makes me nervous that there might be some edge case that isn't in our test file that would cause it to skip a package and nobody would notice.

kylewillmon (approved):

cli/src/lockfiles/javascript.rs
cd-work: That is accurate, I don't have an issue with throwing hard errors on this either, it's probably more appropriate.

Since the `serde_yaml` parser validates v1/2 already, there's no necessity to be lax here, parsing **should** work for every package. So at the very least we should throw a warning?

kylewillmon: I think a hard error is appropriate here. We want to know if it ever breaks.

There is a top-level `__metadata` key that will need to be ignored. I believe npm disallows package names that start with `_`, so it should be safe to throw out those keys.

cli/src/lockfiles/javascript.rs
```diff
@@ -20,7 +21,7 @@ impl Parseable for PackageLock {
 
     /// Parses `package-lock.json` files into a vec of packages
     fn parse(&self) -> ParseResult {
-        let parsed: Value = serde_json::from_str(&self.0)?;
+        let parsed: JsonValue = serde_json::from_str(&self.0)?;
 
         let into_descriptor = |(name, v): (String, &Value)| {
```
maxrake, line 26:
```suggestion
        let into_descriptor = |(name, v): (String, &JsonValue)| {
```
Just guessing at this...

maxrake (changes requested):
Other than the failure to build...

One suggestion is to add an entry to the `docs/analyzing-dependencies.md` file...for `yarn.lock` as a supported lock file type. It might even be nice to note there somehow that both the current and legacy version are supported.

Another suggestion is to update the test in `cli/src/commands/lock_files.rs`: 

```rust
...
    fn it_can_identify_lock_file_types() {
        let test_cases = [
            ("tests/fixtures/Gemfile.lock", PackageType::RubyGems),
            ("tests/fixtures/yarn.lock", PackageType::Npm),
            ("tests/fixtures/yarn.lock", PackageType::Npm),
            ("tests/fixtures/package-lock.json", PackageType::Npm),
...
```

Note that it currently has a repeat test fixture with `tests/fixtures/yarn.lock` used twice. One of them can be replaced with the newly created `yarn-v1.lock` file from this review.

cli/src/lockfiles/javascript.rs
cd-work: Already better at Rust than I am! :smile:

maxrake (commented):
Completing for all files and requested changes, except `cli/src/lockfiles/javascript.rs` ...which still has a lot of material I don't fully understand.

maxrake (approved):

cli/src/lockfiles/javascript.rs
maxrake: LOL...that will likely never be true...but I am learning a lot from you.

kylewillmon (approved):

PR #248, opened 2022-04-12
Replace Language/Type with Ecosystem

### LINKED ISSUE #129: Consolidate `Type` and `Language` fields to Ecosystem
**Describe the issue**
It seems that it would be most accurate if we deprecate `Type` and `Language` for the project analysis in favor of something like Ecosystem.

**Describe why the issue is needed**
At the project level, the analysis may contain multiple languages.

**Describe the proposed solution**
Eliminate `Language` and change `Type` to Ecosystem.

**Estimated Subtasks**
TBD

**Additional Information**
This will likely require us to at least change the Python `Type` entry as it states Python and not PyPi.

image

cd-work:
Since a package uploaded to NPM might not necessarily be JavaScript, the
existing `phylum analyze` description could be misleading.

To improve on this, the `Language` and `Type` fields have been
consolidated into an `Ecosystem` field which just prints the package
repository rather than a guess at the language used in the package.

Closes #129.

### DIFF
--- a/cli/src/summarize.rs
+++ b/cli/src/summarize.rs
@@ -155,25 +155,16 @@ where
             resp.job_id.to_string(),
         ),
         (
-            "Type",
-            ecosystem.render(),
-            "Language",
-            ecosystem.language().to_string(),
-        ),
-        (
             "User ID",
             resp.user_email.to_string(),
             "View in Phylum UI",
             format!("https://app.phylum.io/projects/{}", resp.project),
         ),
     ];
-    let summary = details.iter().fold("".to_string(), |acc, x| {
-        vec![
-            acc,
-            format!("{:>16}: {:<36} {:>24}: {:<36}", x.0, x.1, x.2, x.3),
-        ]
-        .join("\n")
+    let mut summary = details.iter().fold("".to_string(), |acc, x| {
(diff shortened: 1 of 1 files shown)


cli/src/summarize.rs
```diff
@@ -154,26 +154,17 @@ where
...
-    let summary = details.iter().fold("".to_string(), |acc, x| {
-        vec![
-            acc,
-            format!("{:>16}: {:<36} {:>24}: {:<36}", x.0, x.1, x.2, x.3),
-        ]
-        .join("\n")
+    let mut summary = details.iter().fold("".to_string(), |acc, x| {
+        format!("{}\n{:>16}: {:<36} {:>24}: {:<36}", acc, x.0, x.1, x.2, x.3)
     });
+    summary = format!("{}\n       Ecosystem: {}", summary, ecosystem.render());
```
cd-work, line 167: I've decided to put this at the bottom since it's both the easiest implementation and I think it looks a bit strange to have the right side of the grid empty with other stuff below it.

If the formatting should be kept more consistent with the previous behavior and the gap in the middle is *desired*, please let me know.

furi0us333 (changes requested):
@cd-work I don't think you meant to add the `.phylum_project` file to your commit.

furi0us333 (dismissed):
LGTM

furi0us333 (approved):
LGTM

PR #251, opened 2022-04-12
Follow XDG directories spec

### LINKED ISSUE #224: Adhere to XDG Base Directory standard
# Overview
We currently place all of our files in the `~/.phylum` directory. Many other modern tools for Linux and macos have migrated to the [XDG Base Directory Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) and are storing files in `~/.local/bin`, `~/.local/share`, `~/.local/state`, and `~/.config`.

Following this standard would provide benefits for us. For example, we may find that `~/.local/bin` is already in the `PATH` on many distributions and we can avoid modifying shell rc files on many targets. Here is the relevant note from the spec:

> User-specific executable files may be stored in $HOME/.local/bin. Distributions should ensure this directory shows up in the UNIX $PATH environment variable, at an appropriate place.

cd-work:
Previously phylum's installer would put all files into `~/.phylum` and
retrieve the configuration file from there. This leads to a rather
cluttered home directory which is why the XDG directories spec is
usually preferred.

The CLI will now check the XDG config directory first
(`${XDG_CONFIG_HOME:-${HOME}/.config}/phylum/settings.yaml`), then fall
back to the old config path if that isn't present.

The installer will put the completions in the XDG data directory
(`${XDG_DATA_HOME:-${HOME}/.local/share}/phylum`) and the binary in
`${HOME}/.local/bin/`.

Since the `${HOME}/.local/bin` path is sometimes included in the path
already, without being present in the shell config itself, we check the
current `$PATH` first to try and detect if we need to add it or not.

Closes #224.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -167,15 +167,16 @@ pub fn get_home_settings_path() -> Result<PathBuf> {
         .map(PathBuf::from)
         .unwrap_or_else(|| home_path.join(".config"));
 
-    let config = config_dir.join("phylum").join("settings.yaml");
-    let fallback_config = home_path.join(".phylum").join("settings.yaml");
+    let config_path = config_dir.join("phylum").join("settings.yaml");
+    let old_config_path = home_path.join(".phylum").join("settings.yaml");
 
-    // Pick ~/.phylum/settings.yaml only when it exists and the XDG directory does not.
-    if !config.exists() && fallback_config.exists() {
-        Ok(fallback_config)
-    } else {
-        Ok(config)
+    // Migrate the config from the old location.
+    if !config_path.exists() && old_config_path.exists() {
+        fs::create_dir_all(config_path.parent().unwrap())?;
+        fs::rename(old_config_path, &config_path).unwrap();
     }
+
+    Ok(config_path)
 }
 
 #[cfg(test)]
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1,7 +1,8 @@
 use std::env;
-use std::fs;
 use std::io;
+use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
+use std::fs::{self, Permissions};
 
 use anyhow::{anyhow, Result};
 use chrono::{DateTime, Local};
@@ -172,6 +173,7 @@ pub fn get_home_settings_path() -> Result<PathBuf> {
 
     // Migrate the config from the old location.
     if !config_path.exists() && old_config_path.exists() {
+        fs::set_permissions(&old_config_path, Permissions::from_mode(0o600))?;
         fs::create_dir_all(config_path.parent().unwrap())?;
         fs::rename(old_config_path, &config_path).unwrap();
     }
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1,8 +1,9 @@
 use std::env;
-use std::io;
-use std::os::unix::fs::PermissionsExt;
-use std::path::{Path, PathBuf};
 use std::fs::{self, Permissions};
+use std::io;
+#[cfg(unix)]
+use std::os::unix::fs::{OpenOptionsExt, PermissionsExt};
+use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, Result};
 use chrono::{DateTime, Local};
@@ -74,7 +75,6 @@ fn create_private_file<P: AsRef<Path>>(path: P) -> io::Result<fs::File> {
     opts.write(true).create(true).truncate(true);
     #[cfg(unix)]
     {
-        use std::os::unix::fs::OpenOptionsExt;
         opts.mode(0o600);
     }
 
@@ -173,7 +173,10 @@ pub fn get_home_settings_path() -> Result<PathBuf> {
 
     // Migrate the config from the old location.
     if !config_path.exists() && old_config_path.exists() {
-        fs::set_permissions(&old_config_path, Permissions::from_mode(0o600))?;
+        #[cfg(unix)]
+        {
+            fs::set_permissions(&old_config_path, Permissions::from_mode(0o600))?;
+        }
         fs::create_dir_all(config_path.parent().unwrap())?;
         fs::rename(old_config_path, &config_path).unwrap();
     }
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1,5 +1,7 @@
 use std::env;
-use std::fs::{self, Permissions};
+use std::fs;
+#[cfg(unix)]
+use std::fs::Permissions;
 use std::io;
 #[cfg(unix)]
 use std::os::unix::fs::{OpenOptionsExt, PermissionsExt};
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -160,7 +160,22 @@ pub fn get_home_settings_path() -> Result<PathBuf> {
     let home_path =
         home::home_dir().ok_or_else(|| anyhow!("Couldn't find the user's home directory"))?;
 
-    Ok(home_path.join(".phylum").join("settings.yaml"))
+    // Resolve XDG config directory.
+    let config_dir = env::var("XDG_CONFIG_DIR")
+        .ok()
+        .filter(|s| !s.is_empty())
+        .map(PathBuf::from)
+        .unwrap_or_else(|| home_path.join(".config"));
+
+    let config = config_dir.join("phylum").join("settings.yaml");
+    let fallback_config = home_path.join(".phylum").join("settings.yaml");
+
+    // Pick ~/.phylum/settings.yaml only when it exists and the XDG directory does not.
+    if !config.exists() && fallback_config.exists() {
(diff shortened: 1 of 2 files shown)


cli/src/install.sh
```diff
@@ -90,44 +95,42 @@ patch_bashrc() {
...
         # Don't be suspicious. Don't be suspicious
-        xattr -d com.apple.quarantine "${HOME}/.phylum/phylum"
+        xattr -d com.apple.quarantine "${bin_dir}/${bin_name}"
     fi
 
+    # TODO: Why are we modifying a file unrelated to the installation process?
     # Ensure correct permissions on settings.yaml (if it exists).
     if [ -f "${HOME}/.phylum/settings.yaml" ]; then
         chmod 600 "${HOME}/.phylum/settings.yaml"
     fi
```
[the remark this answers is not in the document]
kylewillmon: This is for the upgrade use case. In the 1.2.0 and older releases, the installer *did* create the `settings.yaml` file. But for a long time, it did nothing with the permissions.

Today, the CLI itself creates the `settings.yaml` file and sets the permissions upon creation.

This `chmod` attempts to bridge the gap between users with a fresh install that get a `settings.yaml` file created by the CLI with `0600` permissions and users upgrading from an old release where `install.sh` created the file (likely with `0644` permissions).

cli/src/install.sh
```diff
@@ -53,12 +57,15 @@ add_to_path_and_alias() {
...
+    then
+        echo "export PATH=\"${bin_dir}:\$PATH\"" >> "${rc_path}"
+        success "Updated ${shell}'s \$PATH to include ${bin_dir}."
     fi
 
+    # TODO: Is it necessary to set an alias? Phylum doesn't exactly take long to
+    # type and it might be better to leave this up to the user?
     if ! grep -q 'alias ph=' "${rc_path}"; then
         echo "alias ph='phylum'" >> "${rc_path}"
         success "Created ph alias for phylum in ${shell}."
```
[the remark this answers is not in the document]
kylewillmon: I agree that we shouldn't be adding it in the installer. But this probably deserves its own issue so that we can have some discussion about it.

cli/src/config.rs
```diff
@@ -160,7 +160,22 @@ pub fn get_home_settings_path() -> Result<PathBuf> {
...
+
+    let config = config_dir.join("phylum").join("settings.yaml");
+    let fallback_config = home_path.join(".phylum").join("settings.yaml");
+
+    // Pick ~/.phylum/settings.yaml only when it exists and the XDG directory does not.
+    if !config.exists() && fallback_config.exists() {
+        Ok(fallback_config)
+    } else {
+        Ok(config)
+    }
```
kylewillmon, line 178: I think it would be better to just migrate the file
```suggestion
    // Migrate the config from the old location
    if !config.exists() && fallback_config.exists() {
        fs::create_dir_all(config.parent().unwrap())?;
        fs::rename(fallback_config, config)?;
    }
    Ok(config)
```

cli/src/install.sh
```diff
@@ -53,12 +57,15 @@ add_to_path_and_alias() {
     shell=${2}
 
     # shellcheck disable=SC2016 # we don't want to expand $PATH here
-    if ! grep -q '.phylum:\$PATH' "${rc_path}"; then
```
kylewillmon, line 56: This script is also responsible for upgrade. And as a part of that upgrade, we need to remove the `export PATH="$HOME/.phylum:$PATH"` from the rc files that the previous `install.sh` may have put there.

The same goes for the lines added for tab completion below. We need to remove the old lines, if they exist.

kylewillmon (changes requested):

cli/src/install.sh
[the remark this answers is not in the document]
cd-work: Yeah I agree, I just wanted to bring it up here since I was contemplating it in the other (un)installer-related PRs already.

louislang:
> I think it would likely make sense to put this PR on ice until we've resolved how we want to make things like completions/path modification optional.

I think this makes sense. Should we mark this as a draft so it doesn't accidentally get merged by someone?

cd-work:
> I think this makes sense. Should we mark this as a draft so it doesn't accidentally get merged by someone?

I don't actually think it's possible to convert existing PRs to drafts? Unless you're suggesting to close and re-open as draft.

cd-work:
Decided to move things into a separate file in this PR, I think that's the easiest choice to avoid having to move things around multiple times.

samtay:
Worth mentioning that technically [Apple specifies](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW6) using `$HOME/Library/Application Support` for config/cache dirs like this. For example if you use the [directories crate](https://crates.io/crates/directories), you can see in their [example](https://github.com/dirs-dev/directories-rs/tree/0a890f0c80acb96db736906b9309bac98fb844db#example) that this is what they use for a project dir on Mac.

Personally I think this is unexpected for a CLI tool, so I'd vote to keep it under `~/.config/`.

However it does raise the question, what is our story for Windows here? Looking at other apps (e.g. [neovim](https://github.com/neovim/neovim/issues/9352), [vs code](https://code.visualstudio.com/docs/getstarted/settings#_settings-file-locations)), it seems many use `%APPDATA%`, but I can't tell if there's a particular standard. Not a Windows user myself.

cd-work:
> Personally I think this is unexpected for a CLI tool, so I'd vote to keep it under ~/.config/.

Yes I agree, I've considered the `dirs` crate and decided to go this route instead. I would say for data dirs it might be a bit less clear but for config stuff `~/.config` should definitely be the right place.

> However it does raise the question, what is our story for Windows here? Looking at other apps (e.g. neovim, vs code), it seems many use %APPDATA%, but I can't tell if there's a particular standard. Not a Windows user myself.

It seems like right now Windows support only exists through Linux shims anyway. But I would recommend `%APPDATA%` should that ever change.

cli/src/install.sh
```diff
@@ -48,87 +53,97 @@ get_rc_file() {
...
+        xattr -c "${bin_dir}/${bin_name}"
     fi
 
-    # Ensure correct permissions on settings.yaml (if it exists).
-    if [ -f "${HOME}/.phylum/settings.yaml" ]; then
-        chmod 600 "${HOME}/.phylum/settings.yaml"
+    # Correct inaccurate settings.yaml permissions set by previous installers.
+    if [ -f "${config_dir}/settings.yaml" ]; then
+        chmod 600 "${config_dir}/settings.yaml"
     fi
```
kylewillmon, line 117: The previous installer that we're fixing would have put this file in the old location, so this doesn't actually accomplish what it says.

So I think we've got three options:

1. Use the old location for this section of the install script
2. Correct the permissions in the migration code in `config.rs` instead of here
3. Declare that we're not worried about this anymore and delete these lines

I'm okay with any of the 3 options.

kylewillmon (approved):

cli/src/install.sh
cd-work: > The previous installer that we're fixing would have put this file in the old location, so this doesn't actually accomplish what it says.

It does' but it assumes that the user ran `phylum update`. The configuration file is migrated by `phylum` sooner or later, but that doesn't do any permission changes.

We could also try to fix all permission changes in the move (in Rust) though, then it should be fixed once and forever and we can remove it from the installer?

kylewillmon: > It does' but it assumes that the user ran `phylum update`. The configuration file is migrated by `phylum` sooner or later, but that doesn't do any permission changes.

Good point... So if we want to fix this in `install.sh`, we would have to check both the old location (in case of a manual update where a user downloads the new zip and runs `install.sh`) and the old location (in case of `phylum update`)... That's annoying
 
> We could also try to fix all permission changes in the move (in Rust) though, then it should be fixed once and forever and we can remove it from the installer?

Yep. That would do it.

cli/src/config.rs
```diff
@@ -1,7 +1,8 @@
 use std::env;
-use std::fs;
 use std::io;
+use std::os::unix::fs::PermissionsExt;
```
kylewillmon, line 3: We need a guard around this to make sure we can build on Windows.

kylewillmon: We need a guard around this to make sure we can build on Windows.

cli/src/config.rs
```diff
@@ -1,6 +1,8 @@
 use std::env;
-use std::fs;
+use std::fs::{self, Permissions};
```
kylewillmon, line 2: Somewhat nitpicky, but this will warn on Windows as an unused import.

```suggestion
use std::fs;
#[cfg(unix)]
use std::fs::Permissions;
```

kylewillmon (approved):

cli/src/config.rs
cd-work: Great point.

kylewillmon (approved):

PR #253, opened 2022-04-12
Sign release tags and be descriptive

kylewillmon:
This PR aims to improve `release.sh` so that our releases use signed tags and more descriptive commit messages.

It also gives the user more information to work with by listing a log of commits since the latest release.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,4 +1,10 @@
-#!/bin/bash
+#!/bin/sh
+
+LATEST=$(git describe --tags --abbrev=0)
+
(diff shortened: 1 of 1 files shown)


eeclfrei (approved):

maxrake (approved):
The latest _tag_ is shown...which could be a pre-release instead of a release...

```sh
❯ git describe --tags --abbrev=1
v0.0.1-rc0-git-version-test

❯ git log --oneline HEAD "^v0.0.1-rc0-git-version-test"
--NO-OUTPUT-RETURNED--
```

...but still an improvement...especially for the signed and annotated tag.

PR #257, opened 2022-04-13
Format "Last updated" field with ISO 8601

### LINKED ISSUE #256: Output of "Last updated" in `phylum package` should be human readable
# Overview

Currently when executing `phylum package`, the analyzation result will return the `Last updated` field as a unix timestamp in milliseconds. I'm personally not good enough at math to calculate the exact date from that in my head and I'm sure many others share this trouble.

So I think it would be cool to convert this time automatically and print it in a human-readable format. I'd suggest ISO 8601 just because dates are hard and that way it will at least be clear.

# Acceptance Criteria
- [ ] "Last updated" output is in ISO 8601 format

cd-work:
Instead of printing the "Last updated" field in `phylum package` as a
millisecond unix timestamp, it is now printed in ISO 8601, making it
human-readable.

Closes #256.

### DIFF
--- a/cli/src/render.rs
+++ b/cli/src/render.rs
@@ -1,10 +1,12 @@
-use crate::types::PingResponse;
-use crate::utils::table_format;
 use ansi_term::Color::{Blue, Green, Red, White, Yellow};
+use chrono::{Local, TimeZone};
 use phylum_types::types::job::*;
 use phylum_types::types::package::*;
 use phylum_types::types::project::*;
 use prettytable::*;
+
+use crate::types::PingResponse;
+use crate::utils::table_format;
 
 pub trait Renderable {
     fn render(&self) -> String;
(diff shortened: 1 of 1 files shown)


cli/src/render.rs
```diff
@@ -268,9 +270,11 @@ impl Renderable for PackageType {
 
 impl Renderable for PackageStatusExtended {
     fn render(&self) -> String {
+        let time = Local.timestamp((self.basic_status.last_updated / 1000) as i64, 0);
```
kylewillmon, line 273: From the docs:

> Panics on the out-of-range number of seconds and/or invalid nanosecond, for a non-panicking version see [timestamp_opt](https://docs.rs/chrono/latest/chrono/offset/trait.TimeZone.html#method.timestamp_opt).

I'm not sure what values are "out-of-range" in that context, but it's probably better for us to print `"UNKNOWN"` or something if that ever happens rather than to panic.

kylewillmon (approved):

cli/src/render.rs
cd-work: Seems highly unlikely unless something goes seriously wrong in the backend, but I guess it doesn't hurt.

kylewillmon (approved):

kylewillmon (approved):
Yet another example of why we need a merge bot (or merge queues)...

Doing the same review twice just to satisfy "Require branches to be up to date before merging" is annoying...

PR #259, opened 2022-04-13
phylum-init.sh download script

kylewillmon:
Installing `phylum` in the future will be as simple as:

```
curl https://get.phylum.io | sh -
```

This PR creates the script that will be hosted at the URL.


peterjmorgan:
Can we add in the verification of the signature on the download file to this?

kylewillmon:
> Can we add in the verification of the signature on the downloaded installer to this?

Yep. I had just realized that I missed that

kylewillmon:
One note: Adding the signature check also means that the user must have `minisign` installed in order to perform the install. Unless there is an objection, I will add `--no-verify` option to allow installing even without that dependency.

cd-work:
> One note: Adding the signature check also means that the user must have minisign installed in order to perform the install. Unless there is an objection, I will add --no-verify option to allow installing even without that dependency.

I doubt most people have minisig installed, which effectively would mean that most downloads don't even check for integrity at that point? That does not seem ideal for something retrieved over the internet.

Though for me the script would fail anyway since I also don't have `unzip` installed.

kylewillmon:
> Though for me the script would fail anyway since I also don't have `unzip` installed.

This has also been a concern for me while writing this script... I am adding a Concerns section to the [associated user story](https://www.notion.so/phylum/Install-Script-Improvement-880c22380d4041b388b7937a603ad6aa) so that we list these somewhere. We should brainstorm ideas and find a path to get away from these issues.

cd-work:
> This has also been a concern for me while writing this script... I am adding a Concerns section to the associated user story so that we list these somewhere. We should brainstorm ideas and find a path to get away from these issues.

Tar seems like a pretty simple suggestion to solve that one. I'm not sure why you'd distribute zips when tar and gzip should be installed on pretty much any unix system by default?

scripts/phylum-init.sh
```diff
@@ -0,0 +1,150 @@
...
+        exit 1
+    fi
+}
+
+# All of the required commands (outside of POSIX)
+require_command mktemp
+require_command curl
+require_command unzip
+
+# Parse command line arguments
```
cd-work, line 76: You have command line arguments here, but no way they're ever communicated to the user. Even if you don't want to add a dedicated `--help` command, I think we should at least log all the options whenever an unrecognized parameter is detected.

cd-work (changes requested):

scripts/phylum-init.sh
kylewillmon: It's simple enough to add `--help` if we're going to print usage for unrecognized parameters. Done.

scripts/phylum-init.sh
```diff
@@ -68,6 +68,21 @@ require_command() {
...
+phylum-init.sh [options]
+
+Fetch and install the phylum CLI.
+
+Options
+    -t, --target       Specify the target triple to install
+    -y, --yes          Do not prompt for comfirmation during install
+    --no-verify        Disable signature verification (NOT RECOMMENDED)
+    -h, --help         Show this help message
+
```
cd-work, line 82: This is extremely nitpicky, but generally I'd avoid unnecessary newlines at the end of a command. Would be perfectly fine with this, but there's no need to take up more space than necessary.

The standard is generally *not* to include a newline. At least that's what any command I could come up with does.

cd-work (dismissed):
Looks good, other than the previously mentioned issues wrt unzip/minisig, which you can't do much about in this PR, everything seems to work.

maxrake:
How will the script get to https://get.phylum.io (or maybe https://sh.phylum.io and https://win.phylum.io to be like `rustup`)? Is it expected to be manually transferred there each time there are updates? Or, will that site simply be a redirect to the file in the default branch of the repository? Something else?

kylewillmon:
> How will the script get to https://get.phylum.io (or maybe https://sh.phylum.io and https://win.phylum.io to be like `rustup`)? Is it expected to be manually transferred there each time there are updates? Or, will that site simply be a redirect to the file in the default branch of the repository? Something else?

¯\_(ツ)_/¯

This is still experimental for now. The first iteration will probably be manually pushed. Once we've decided that we like it, we can establish a procedure for updating it (e.g., publishing it from a Github Actions) and explain how to use it in the documentation and such.

maxrake:
All information level findings...but this is what `shellcheck` is reporting for the file:

image

scripts/phylum-init.sh
kylewillmon: Simple enough and that makes sense

maxrake (approved):
The code looks good. Trying the script out worked. There might be room for adding the `curl --progress-meter` option...or similar...to help users with slower connections know that _something_ is happening and how much longer it will take. That can be a separate issue, though. This is great stuff and will pave the way for even more.

LGTM

scripts/phylum-init.sh
maxrake: it was changed...resolving...

PR #260, opened 2022-04-14
Set appropriate exit codes on failure

### LINKED ISSUE #229: Set appropriate exit codes without program malfunction
# Overview

Currently the only way to get a non-zero exit code is when an error is generated. To the user this looks like there was a malfunction in the application itself.

However, as a user I'd want commands like `phylum auth status` to set an appropriate exit code when I'm not logged in, while still just printing a concise and human-friendly message that informs me that I'm not logged in, rather than indicating a malfunction of the program itself.

Not only will this play nicer with UI elements like error-indicators in the `$PS1`, but it also makes it possible to use commands in shell scripts without having to parse text.

While `phylum auth status` is the only command where I know it would make sense, there might be others that could benefit from this.

# Acceptance Criteria
- [ ] `phylum auth status` sets non-zero exit code when the user is not logged in
- [ ] The text output of `phylum auth status` is not changed

## Associated User Story
--

cd-work:
Several paths of the Phylum CLI would print warnings/errors to the user,
but still set and exit code of `0`. This makes it difficult for for
automated tools to verify success and can confuse users.

As a solution, the `CommandValue` enum has been changed, replacing the
`CommandValue::Void` variant with `CommandValue::Code`. This code is
then used on exit instead of the fixed `0` which was used for
`CommandValue::Void`.

This also introduces a few simplifications to the error/CommandValue
handling, removing unnecessary code and inlining unnecessary
abstractions. This hopefully improves transparency of the process and
makes the usage consistent.

Some functions relying on `CommandValue::Void` as a replacement to `()`
without ever propagating it have been switched to `anyhow::Result<()>`.
Generally there's little reason to use `CommandValue` unless the goal
is for it to end up directly as the command's exit code.

Some functions also had to be adjusted to ensure the code is properly
propagated. This should capture all instances where the result of a
function returning `CommandResult` was ignored.

Closes #229.

### DIFF
--- a/cli/src/bin/phylum.rs
+++ b/cli/src/bin/phylum.rs
@@ -13,22 +13,13 @@ use phylum_cli::commands::auth::*;
 use phylum_cli::commands::jobs::*;
 use phylum_cli::commands::packages::*;
 use phylum_cli::commands::projects::handle_projects;
-use phylum_cli::commands::{CommandResult, CommandValue};
+use phylum_cli::commands::{CommandResult, CommandValue, ExitCode};
 use phylum_cli::config::*;
 use phylum_cli::print::*;
 use phylum_cli::update;
 use phylum_cli::{print_user_failure, print_user_success, print_user_warning};
 use phylum_types::types::common::JobId;
 use phylum_types::types::job::Action;
-
-/// Exit with status code 0 and optionally print a message to the user.
-pub fn exit_ok(message: Option<impl AsRef<str>>) -> ! {
-    if let Some(message) = message {
-        info!("{}", message.as_ref());
-        print_user_success!("{}", message.as_ref());
-    }
-    process::exit(0)
-}
 
 /// Print a warning message to the user before exiting with exit code 0.
 pub fn exit_warn(message: impl AsRef<str>) -> ! {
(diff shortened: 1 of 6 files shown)


cli/src/bin/phylum.rs
```diff
@@ -146,7 +130,9 @@ async fn handle_commands() -> CommandResult {
         let res = update::do_update(matches.is_present("prerelease")).await;
         spinner.stop();
         println!();
-        return res.map(CommandValue::String);
+        let message = res?;
+        print_user_success!("{}", message);
```
kylewillmon, line 134: You are expertly dodging land mines here.... I definitely would have written this as:

```
print_user_success!("{}", res?);
```

But now I realize that thanks to our definition of `print_user_success!`, that would print the ✅ even on failure... 🤦

cli/src/commands/auth.rs
```diff
@@ -89,32 +93,30 @@ pub async fn handle_auth(
...
-            Err(error) => {
-                return Err(anyhow!(
-                    "User login failed: {}",
-                    error.root_cause().to_string()
-                ));
-            }
+            Err(error) => Err(anyhow!(
+                "User login failed: {}",
+                error.root_cause().to_string()
+            )),
```
kylewillmon, line 112: This seems like another place that `ExitCode::AuthenticationFailure` would be appropriate

kylewillmon (approved):

cli/src/bin/phylum.rs
cd-work: Yeah, I initially considered doing this, but then thought that's just a bug waiting to happen regardless of the current `print_user_success` implementation. Better to be explicit and avoid any potential for misinterpretation.

cli/src/commands/auth.rs
cd-work: Yeah, though I didn't want to impact the CLI's text output. Currently this does set an error code and I think it's better just to keep that error code and the text output consistent rather than switching for consistency with a mostly unrelated command.

kylewillmon (approved):

kylewillmon (approved):

PR #262, opened 2022-04-14
Truncate excessive project names

### LINKED ISSUE #127: Long project names can mess up formatting
**Describe the bug**
If a project is created with a long name it can mess up the formatting on a few of the CLI views.

**To Reproduce**
1. Create a project with a very long name
2. Run `phylum projects list` and observe formatting
Note: this also impacts several other views

**Expected behavior**
CLI handles long project names.  Perhaps truncating and adding ellipsis?  Or limiting the accepted project name size and throwing an error if it is exceeded?

**Screenshots**
`projects list` view
image

`history` view
image

`analyze` view
image

cd-work:
Since a lot of output in the Phylum CLI relies on tables, it's important
that text does not exceed the maximum cell length. The most likely
culprit for this is the `project`, which can be freely defined by the
user.

All instances of the project name output are now truncated to the
respective cell length and a truncation is indicated using the ellipsis
symbol.

To ensure this works for project names in any language, the length of
the cells is calculated using the `unicode-width` crate.

Closes #127.

### DIFF
--- a/cli/src/print.rs
+++ b/cli/src/print.rs
@@ -1,9 +1,10 @@
-use std::io;
-use std::io::Write;
+use std::borrow::Cow;
+use std::io::{self, Write};
 
 use ansi_term::Color::{Blue, Cyan};
 use clap::Command;
 use serde::Serialize;
+use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
 
 use crate::api::PhylumApiError;
 use crate::filter::Filter;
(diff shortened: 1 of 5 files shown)


cli/src/print.rs
```diff
@@ -90,3 +91,21 @@ pub fn print_sc_help(app: &mut Command, subcommand: &str) {
     }
     println!();
 }
+
+/// Limit a string to a specific length, using an ellipsis to indicate truncation.
+pub fn truncate(text: &str, max_length: usize) -> Cow<str> {
```
cd-work, line 96: I've decided to put this in the `print` module since it's directly related to printing things on the terminal. While I could have put it in `utils`, I'm generally not a big fan of a generic `utils` module so I did not want to further contribute to its existence.

kylewillmon: I agree. And I'd support a PR to move `utils::table_format` to `print::table_format` and delete `utils.rs`

cli/src/print.rs
```diff
@@ -1,9 +1,10 @@
-use std::io;
-use std::io::Write;
+use std::borrow::Cow;
+use std::io::{self, Write};
 
 use ansi_term::Color::{Blue, Cyan};
 use clap::Command;
```
janasheehan, line 5: i cannot believe these labels....std? clap? who comes up with those!

kylewillmon (approved):

kylewillmon (approved):

PR #265, opened 2022-04-14
Remove utils module

cd-work:
Generally a `utils` module is often somewhat of a code smell, since its
generic name robs its members of more descriptive namespaces.

To avoid this, the only function currently present in the `utils` module
has been moved to the more appropriate `print` module and the module
itself was deleted.


cd-work:
Removing the `utils` module as suggested by @kylewillmon in #262.

kylewillmon (dismissed):

maxrake (dismissed):
Even I could follow these changes since it doesn't take much Rust knowledge.

LGTM.

kylewillmon (approved):

PR #271, opened 2022-04-18
fix: Parse package extras in Python requirements.txt files

### LINKED ISSUE #267: Python dependencies with square brackets are ignored
# Overview
Python dependencies such as `celery[redis]==1.2.3` are ignored in a `requirements.txt` files.

# Additional Context
I've added the `square-reqs` branch with and updated test file to show the issue

kylewillmon:
# Overview

This PR adjusts our parsing for `requirements.txt` files so that we ignore package extras (i.e., stuff in square brackets after the package name).

I've also added a new `ws` utility to the `parsers` module that makes eating whitepspace a bit easier. This utility was taken basically verbatim from `nom::recipes`, but I use `space0` instead of `multispace0` so that it doesn't eat newlines.

## Checklist

## Issue

Fixes #267

### DIFF
--- a/cli/src/lockfiles/parsers/pypi.rs
+++ b/cli/src/lockfiles/parsers/pypi.rs
@@ -83,7 +83,8 @@ fn package(input: &str) -> Option<PackageDescriptor> {
                 v.trim_start_matches("-e").to_string(),
             ),
             None => {
-                let (version, name) = filter_pip_name(s).ok()?;
+                let (version, pkg) = filter_pip_name(s).ok()?;
+                let (_, name) = filter_package_name(pkg).ok()?;
                 (
                     name.to_string(),
                     version.trim_start_matches('@').to_string(),
@@ -135,6 +136,15 @@ mod test {
             Some(PackageDescriptor {
                 name: "requests".into(),
                 version: "2.27.1".into(),
+                package_type: PackageType::PyPi,
+            })
+        );
+
+        assert_eq!(
+            package("git-for-pip-example[PDF] @ git+https://github.com/matiascodesal/git-for-pip-example.git@v1.0.0"),
+            Some(PackageDescriptor {
+                name: "git-for-pip-example".into(),
+                version: "git+https://github.com/matiascodesal/git-for-pip-example.git@v1.0.0".into(),
                 package_type: PackageType::PyPi,
             })
         );
--- a/cli/src/lockfiles/parsers/pypi.rs
+++ b/cli/src/lockfiles/parsers/pypi.rs
@@ -3,7 +3,7 @@ use nom::{
     bytes::complete::{tag, take_until},
     character::complete::{alphanumeric1, char, not_line_ending},
     combinator::{opt, recognize, rest, verify},
-    multi::{many0, many1},
+    multi::{many0, many1, separated_list0},
     sequence::{delimited, pair, terminated},
 };
 use phylum_types::types::package::{PackageDescriptor, PackageType};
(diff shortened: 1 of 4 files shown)


cli/src/lockfiles/parsers/pypi.rs
```diff
@@ -94,3 +114,29 @@ fn package(input: &str) -> Option<PackageDescriptor> {
         package_type: PackageType::PyPi,
     })
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn package_with_extras() {
```
maxrake, line 123: An additional [failing] test could be added here to point out an issue:

```rust
        assert_eq!(
            package("git-for-pip-example[PDF] @ git+https://github.com/matiascodesal/git-for-pip-example.git@v1.0.0"),
            Some(PackageDescriptor {
                name: "git-for-pip-example".into(),
                version: "git+https://github.com/matiascodesal/git-for-pip-example.git@v1.0.0".into(),
                package_type: PackageType::PyPi,
            })
        );
```

This test case comes from attempting to replicate an example from [pip install documentation for extras](https://pip.pypa.io/en/stable/cli/pip_install/#examples), [PEP 508 for extras](https://peps.python.org/pep-0508/#extras), and the current (working/valid?) example in `cli/tests/fixtures/complex-requirements.txt`:

```
...
git-for-pip-example @ git+https://github.com/matiascodesal/git-for-pip-example.git@v1.0.0 
[EOF]
```

The test case fails, with the following message:

```sh
running 1 test
thread 'lockfiles::parsers::pypi::test::package_with_extras' panicked at 'assertion failed: `(left == right)`
  left: `Some(PackageDescriptor { name: "git-for-pip-example[pdf]", version: "git+https://github.com/matiascodesal/git-for-pip-example.git@v1.0.0", package_type: PyPi })`,
 right: `Some(PackageDescriptor { name: "git-for-pip-example", version: "git+https://github.com/matiascodesal/git-for-pip-example.git@v1.0.0", package_type: PyPi })`', cli/src/lockfiles/parsers/pypi.rs:142:9
...
```

maxrake (changes requested):
While some of the new Rust code is not completely understood, the tests were run both before and after the code changes and appear to work as intended. However, when an additional test case was added...for a package specified from a VCS source...the test failed to identify the package name correctly. More detail is in other comments.

Marking as "request changes" but it may be the case that the discovered issued is more associated with incorrectly handling "url_req" specifications from the [PEP 508 Grammar](https://peps.python.org/pep-0508/#grammar)...in which case a different issue may be opened to track it.

cli/src/lockfiles/parsers/pypi.rs
```diff
@@ -94,3 +114,29 @@ fn package(input: &str) -> Option<PackageDescriptor> {
...
+        assert_eq!(
+            package("celery [ redis ] == 5.0.5"),
```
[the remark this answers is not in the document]
kylewillmon: As you mention, `complex-requirements.txt` is a weird test case. I'd go further and say that it needs to be removed in the near future.

So instead, I have added a `Django` example to `requirement.txt`. That file contains only pass test cases, so it seemed like a more logical place to add another pass test case.

I used `Django` instead of your suggestions so that we expand our coverage to as many real world examples as possible.

cli/src/lockfiles/parsers/pypi.rs
kylewillmon: I've added this test case and modified the code to fix it, but this is definitely something we should revisit later. I don't believe that we truly support URL specified dependencies. And if/when we do add support for them, we should revisit how we represent them in the `PackageDescriptor` object. The URL in the `version` field feels unnatural

maxrake (approved):
Changes from the last review look good. More work is needed...in the future and on a different issue...to improve handing of URI based requirements.

LGTM

PR #272, opened 2022-04-19
Indicate version type in update notification

### LINKED ISSUE #125: Update Available message language
**Describe the issue**
The Update Available message does not take into consideration pre-release versions.

image

**Describe the proposed solution**
We could possibly change the language to side-step this.  Proposed language:

You are not running the latest stable version of the Phylum CLI.
`phylum update`
The above command will update to the latest stable release.

I am not sure if we should also change the title to "Warning" or something like some other Linux CLI tools do?
image

**Estimated Subtasks**
TBD

cd-work:
This patch clarifies the message about a new Phylum CLI version by
explicitly pointing out that a new *stable* version is available when
the user is currently running from an existing stable version.

Closes #125.

### DIFF
--- a/cli/src/bin/phylum.rs
+++ b/cli/src/bin/phylum.rs
@@ -17,19 +17,20 @@ use phylum_cli::commands::{CommandResult, CommandValue, ExitCode};
 use phylum_cli::config::*;
 use phylum_cli::print::*;
 use phylum_cli::update;
+use phylum_cli::CURRENT_VERSION;
 use phylum_cli::{print_user_failure, print_user_success, print_user_warning};
 use phylum_types::types::common::JobId;
 use phylum_types::types::job::Action;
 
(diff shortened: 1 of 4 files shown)


cli/src/bin/phylum.rs
```diff
@@ -99,8 +100,9 @@ async fn handle_commands() -> CommandResult {
         }
     }
 
-    if check_for_updates && update::needs_update(false).await {
-        print_update_message();
+    let prerelease = CURRENT_VERSION.contains('-');
```
kylewillmon, line 103: As far as I know, we have never actually updated the cargo version for an RC release, so `prerelease` will always be `false`.

cd-work:
Closing due to https://github.com/phylum-dev/cli/issues/125#issuecomment-1102824097.

PR #273, opened 2022-04-19
Filter non-PyPi dependencies from poetry lockfile

### LINKED ISSUE #269: Python Poetry parser does not account for package source
# Overview
The current implementation of the Python `poetry.lock` file parser does not account for `package.source` sub-tables. That is the mechanism the [`poetry-core` library uses for specifying dependency types](https://github.com/python-poetry/poetry-core/blob/master/src/poetry/core/packages/types.py), when the type is not a "basic" dependency that can be fully specified in the `package` table entry alone. Such types include: `DirectoryDependency`, `FileDependency`, `URLDependency`, and `VCSDependency`. Therefore, when the `poetry.lock` file contains a dependency with a `package.source` sub-table, it gets treated as a "basic" `Dependency` type instead. This can result in analyzing the wrong dependency, when the same package/version pair exists in PyPI.

# How To Reproduce
1. View the [phylum-dev/vulnerabilities-admin private repo `poetry.lock` file](https://github.com/phylum-dev/vulnerabilities-admin/blob/development/poetry.lock), specifically the [entry for the `common` package](https://github.com/phylum-dev/vulnerabilities-admin/blob/development/poetry.lock#L265-L278)
2. Notice that the `common` package there is specifying a VCSDependency of a _private_ git repo, [phylum-dev/common](https://github.com/phylum-dev/common), which indeed has a Python package defined in the [setup.py](https://github.com/phylum-dev/common/blob/development/setup.py) file there of `common v0.1.0`
3. Perform a `phylum analyze` on the `poetry.lock` file
4. View the results in the Web UI to see that the package was attempted to be downloaded...but for the [`common==0.1` package that exists in the _public_ PyPI](https://pypi.org/project/common/0.1/#files)
5. This is a form of [dependency confusion](https://dhiyaneshgeek.github.io/web/security/2021/09/04/dependency-confusion/), but within the CLI parser; the package installer (Poetry) will still know how to parse and install the correct package

# Expected Behavior
The packages in the `poetry.lock` file that have a `package.source` sub-table...which specifies something other than a basic dependency available in the PyPI public package registry...will be filtered out by CLI parser and not included in the list of packages to analyze.

# Additional Context
* The [poetry `Locker` class](https://github.com/python-poetry/poetry/blob/fb13b3a676f476177f7937ffa480ee5cff9a90a5/src/poetry/packages/locker.py#L45-L49) is really the only "documentation" that exists for the `poetry.lock` file format
  * This link is for the current version of that lockfile, v1.1
* The Poetry project repo has some test files and fixtures that might help
  * [`python-poetry/poetry/tests/packages/test_locker.py`](https://github.com/python-poetry/poetry/blob/da9680b0faee435a9d4fb52ac4b11b7a18c48069/tests/packages/test_locker.py)
  * [`python-poetry/poetry/tests/fixtures`](https://github.com/python-poetry/poetry/tree/da9680b0faee435a9d4fb52ac4b11b7a18c48069/tests/fixtures)
* The build backend for Poetry is broken out into it's own repo, [`poetry-core`](https://github.com/python-poetry/poetry-core)
  * The package types are defined there: [`python-poetry/poetry-core/src/poetry/core/packages`](https://github.com/python-poetry/poetry-core/tree/master/src/poetry/core/packages) and [`python-poetry/poetry-core/src/poetry/core/packages/types.py`](https://github.com/python-poetry/poetry-core/blob/master/src/poetry/core/packages/types.py) as a good starting point
* The CLI parser for Python `requirements.txt` files [appears to have logic to further filter out dependencies](https://github.com/phylum-dev/cli/blob/8a6ecf5344301115fa1a8b69f2d3b1f1b72be461/cli/src/lockfiles/python.rs#L27) by using the [`parsers::pypi::parse` module](https://github.com/phylum-dev/cli/blob/8a6ecf5344301115fa1a8b69f2d3b1f1b72be461/cli/src/lockfiles/parsers.rs#L157)
  * Perhaps something similar can be done for `poetry.lock` files

cd-work:
This patch filters the poetry lockfile to ensure only packages which are
present in PyPi are submitted, to avoid potential dependency confusion.

To ensure the poetry lockfile parser's robustness, a new test lockfile
has been created which makes use of every possible dependency type other
than `legacy`.

Closes #269.

### DIFF
--- a/cli/src/lockfiles/python.rs
+++ b/cli/src/lockfiles/python.rs
@@ -114,6 +114,7 @@ impl Parseable for Poetry {
         Ok(lock
             .packages
             .drain(..)
+            .filter(|package| package.source.is_none())
             .map(PackageDescriptor::from)
             .collect())
     }
@@ -130,6 +131,7 @@ struct PoetryLock {
 struct Package {
     name: String,
     version: String,
+    source: Option<Value>,
 }
 
 impl From<Package> for PackageDescriptor {
@@ -253,17 +255,17 @@ mod tests {
         let parser = Poetry::new(Path::new("tests/fixtures/poetry.lock")).unwrap();
 
         let pkgs = parser.parse().unwrap();
-        assert_eq!(pkgs.len(), 62);
+        assert_eq!(pkgs.len(), 43);
 
         let expected_pkgs = [
             PackageDescriptor {
-                name: "toml".into(),
-                version: "0.10.2".into(),
+                name: "cachecontrol".into(),
+                version: "0.12.10".into(),
                 package_type: PackageType::PyPi,
             },
             PackageDescriptor {
-                name: "certifi".into(),
-                version: "2021.10.8".into(),
+                name: "flask".into(),
+                version: "2.1.1".into(),
                 package_type: PackageType::PyPi,
             },
             PackageDescriptor {
@@ -275,6 +277,19 @@ mod tests {
 
         for expected_pkg in expected_pkgs {
             assert!(pkgs.contains(&expected_pkg));
+        }
+    }
+
+    /// Ensure sources other than PyPi are ignored.
+    #[test]
+    fn poetry_ignore_other_sources() {
+        let parser = Poetry::new(Path::new("tests/fixtures/poetry.lock")).unwrap();
+
+        let pkgs = parser.parse().unwrap();
+
+        let invalid_package_names = ["toml", "directory-test", "requests", "poetry"];
+        for pkg in pkgs {
+            assert!(!invalid_package_names.contains(&pkg.name.as_str()));
         }
     }
 }
--- a/cli/src/lockfiles/python.rs
+++ b/cli/src/lockfiles/python.rs
@@ -114,7 +114,12 @@ impl Parseable for Poetry {
         Ok(lock
             .packages
             .drain(..)
-            .filter(|package| package.source.is_none())
+            .filter(|package| {
+                package
+                    .source
+                    .as_ref()
+                    .map_or(true, |source| source.source_type == "git")
+            })
             .map(PackageDescriptor::from)
             .collect())
     }
@@ -131,17 +136,33 @@ struct PoetryLock {
 struct Package {
     name: String,
     version: String,
-    source: Option<Value>,
+    source: Option<PackageSource>,
 }
 
 impl From<Package> for PackageDescriptor {
     fn from(package: Package) -> Self {
+        let version = package
+            .source
+            .and_then(|source| {
+                let reference = source.resolved_reference.as_ref();
+                reference.map(|reference| format!("{}#{}", source.url, reference))
+            })
+            .unwrap_or(package.version);
+
         Self {
             name: package.name,
-            version: package.version,
             package_type: PackageType::PyPi,
+            version,
         }
     }
+}
+
+#[derive(Deserialize, Debug)]
+struct PackageSource {
+    #[serde(rename = "type")]
+    source_type: String,
+    url: String,
+    resolved_reference: Option<String>,
 }
 
 #[derive(Deserialize, Debug)]
@@ -255,7 +276,7 @@ mod tests {
         let parser = Poetry::new(Path::new("tests/fixtures/poetry.lock")).unwrap();
 
         let pkgs = parser.parse().unwrap();
-        assert_eq!(pkgs.len(), 43);
+        assert_eq!(pkgs.len(), 44);
 
         let expected_pkgs = [
             PackageDescriptor {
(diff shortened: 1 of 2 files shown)


maxrake:
> I think we might want to capture maybe VCS dependencies at a minimum. The [request manager](https://github.com/phylum-dev/request-manager/blob/8f4d1bb04dc20066d2f9bdf5be939a03295857d8/src/models.py#L480) will allow a URL as the version.

If VCS support is added, will the CLI or the request manager (or both?) be responsible for knowing if the VCS is public/accessible or private/inaccessible? The [phylum-dev/vulnerabilities-admin private repo `poetry.lock` file](https://github.com/phylum-dev/vulnerabilities-admin/blob/development/poetry.lock), specifically the [entry for the `common` package](https://github.com/phylum-dev/vulnerabilities-admin/blob/development/poetry.lock#L265-L278) is an example of a private GitHub repository that I would guess request manager is unable to access.

ejortega:
> > I think we might want to capture maybe VCS dependencies at a minimum. The [request manager](https://github.com/phylum-dev/request-manager/blob/8f4d1bb04dc20066d2f9bdf5be939a03295857d8/src/models.py#L480) will allow a URL as the version.
> 
> If VCS support is added, will the CLI or the request manager (or both?) be responsible for knowing if the VCS is public/accessible or private/inaccessible? The [phylum-dev/vulnerabilities-admin private repo `poetry.lock` file](https://github.com/phylum-dev/vulnerabilities-admin/blob/development/poetry.lock), specifically the [entry for the `common` package](https://github.com/phylum-dev/vulnerabilities-admin/blob/development/poetry.lock#L265-L278) is an example of a private GitHub repository that I would guess request manager is unable to access.

This could be older info and maybe has changed. @louislang or @eeclfrei asked me to modify some of the scoring issues at one point due to VCS entries included in NPM lockfiles. I believe the handling of private/public would be done by another service and not the CLI.

maxrake (approved):
The changes were confirmed to work with the `vulnerabilities-admin` poetry.lock file that initiated this issue. So...LGTM.

As for the VCS support...either adding that here or in a new issue/PR is acceptable.

ejortega:
@cd-work [python requirements parser ](https://github.com/phylum-dev/cli/blob/18a0a03752b7fc24cd4961ba2314e200859fc35f/cli/src/lockfiles/parsers/pypi.rs#L37).

cli/src/lockfiles/python.rs
```diff
@@ -130,18 +136,35 @@ struct PoetryLock {
...
+    source: Option<PackageSource>,
 }
 
 impl From<Package> for PackageDescriptor {
     fn from(package: Package) -> Self {
+        let version = package
+            .source
+            .and_then(|source| {
+                let reference = source.resolved_reference.as_ref();
+                reference.map(|reference| format!("{}#{}", source.url, reference))
```
maxrake, line 148: There is still some uncertainty...but the current thought is that the format for the version on this line is not quite what the `request-manager` expects. Feeding the [`poetry.lock` file from `vulnerabilities-admin`](https://github.com/phylum-dev/vulnerabilities-admin/blob/poetry_month/poetry.lock) results in the following error...which appears to be due to failing to match any of the version regex's in request-manager...specifically [this one](https://github.com/phylum-dev/request-manager/blob/bb1d4bfecabb63a8aba79ed86df91af31dd1fb82/src/models.py#L480).

```sh
❯ ./phylum analyze -l poetry_type_test_4 -F poetry.lock
[2022-04-19T17:16:34Z ERROR phylum] Execution failed: Improperly formatted creation request
❗ Error: Execution failed caused by: Improperly formatted creation request
```

My regex foo is limited, but perhaps @ejortega can assist.

maxrake (changes requested):
Changes appear good...and updates to the unit tests do show them to pass...but a functional test, using the production system, shows a failure. See the comment for more detail.

cli/src/lockfiles/python.rs
cd-work: The regex definitely doesn't match `https` URLs. I'm not sure any format that is supported in `poetry.lock` is also supported by this regex.

ejortega: I'm testing out some changes with request-manager and api

maxrake:
A bug ticket was created in the `request-manager` repo to track the issue blocking this PR:
* [Model doesn't handle certain VCS urls](https://github.com/phylum-dev/request-manager/issues/80)

cli/src/lockfiles/python.rs
cd-work: The URLs work correctly now.

maxrake (approved):
Confirmed the changes (on staging environment) do work as intended. There is a different issue I noticed while doing that testing...but it is not related to this PR or even the CLI. So...LGTM.

cli/src/lockfiles/python.rs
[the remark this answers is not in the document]
maxrake: looks good on staging now.

PR #279, opened 2022-04-19
Allow phylum-init.sh to work with curl

### LINKED ISSUE #278: phylum-init.sh trys and fails to get user confirmation when stdin is curl
# Overview
Initial testing of the script at `sh.phylum.io` shows that it cannot get confirmation from the user and automatically aborts the install.

```
❯ curl -o - --silent https://sh.phylum.io/ | sh -s          
Release archive URL: https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-apple-darwin.zip
Continue install? [y/N] Aborting install
```

kylewillmon:
This PR modifies `phylum-init.sh` to read from `/dev/tty` when prompting for confirmation. It also adds a check for stdout to make sure that the user will see our prompt.

Fixes #278


kylewillmon:
The updated script isn't on `sh.phylum.io` yet, but it can be tested directly from Github:

```
curl -o - --silent https://raw.githubusercontent.com/phylum-dev/cli/tty-pipes/scripts/phylum-init.sh | sh -
```

maxrake (approved):
Tried it with the following invocations:
* `curl -o - --silent https://raw.githubusercontent.com/phylum-dev/cli/tty-pipes/scripts/phylum-init.sh | sh -`
* `curl -o - --silent https://raw.githubusercontent.com/phylum-dev/cli/tty-pipes/scripts/phylum-init.sh | sh -s -- --yes`

Both worked. LGTM

cd-work (commented):
This will not work on systems where /dev/tty does not exist, while also making it impossible for those systems to download to a file and execute it from there.

Does it maybe make sense to only do this when it's absolutely necessary?

See https://github.com/rust-lang/rustup/pull/2951 for more details.

kylewillmon:
> Does it maybe make sense to only do this when it's absolutely necessary?

Sure. Checking `[ -t 0 ]` is simple enough

cd-work (approved):

PR #282, opened 2022-04-20
Rename projects subcommand to project

### LINKED ISSUE #281: `phylum projects` should be `phylum project`
Most CLI tools use the singular version of words for their tool commands. Unfortunately, the `phylum` tools has a `projects` subcommand rather than a `project` subcommand. The plural version is longer to type and unnatural in commands that only work on a single project (i.e., `phylum projects create`, `phylum projects link`, etc.).

We should update the tool to use `project` as the subcommand, keeping `projects` around as an alias to `project`, so that current users and scripts are not broken by the change.

kylewillmon:
This is the minimal change to rename `phylum projects` to `phylum project` and update the appropriate documentation.

I did not worry about renaming modules and functions (i.e., `projects::handle_projects()` -> `project::handle_project()`), because those are not outward facing and that would be a much bigger change. But I am happy to do that if others feel that it is needed.

Fixes #281


docs/projects.md
```diff
@@ -9,19 +9,19 @@ Projects allow you to track dependencies for a given software repository (e.g. a
 To create a new project and associate the working directory with it, simply run:
```
cd-work, line 9: I'd say the filename and title of this documentation is still accurate, but I'd be curious if you've considered changing it? Since it's talking about projects in general, I would say that "project" is weird.

cd-work (changes requested):
> I did not worry about renaming modules and functions (i.e., projects::handle_projects() -> project::handle_project()), because those are not outward facing and that would be a much bigger change. But I am happy to do that if others feel that it is needed.

I would definitely suggest doing that especially for things directly referring to the CLI argument like `handle_projects`.

For other modules it gets a bit more complicated, though I think with project it shouldn't be an issue. I wanted to bring up `cli/src/api/project.rs` as one thing that should probably stay `projects.rs`, since the API endpoint is actually `/projects/`, but it seems like that's the one thing that's actually already `project` even though maybe it shouldn't be.

Also you forgot this:
```
./cli/src/commands/jobs.rs:153:                "Failed to find a valid project configuration. Did you run `phylum projects create <project-name>`?"
```

docs/projects.md
kylewillmon: I did consider it, but decided not to. As you point out, this page is talking about the general topic of "Projects" and not the specific subcommand `project`.

The work that @furi0us333 is doing will restructure our docs around the subcommands and at that point it will rightfully be `project` (singular).

kylewillmon:
> I would definitely suggest doing that especially for things directly referring to the CLI argument like `handle_projects`.

Will do.

> For other modules it gets a bit more complicated, though I think with project it shouldn't be an issue. I wanted to bring up `cli/src/api/project.rs` as one thing that should probably stay `projects.rs`, since the API endpoint is actually `/projects/`, but it seems like that's the one thing that's actually already `project` even though maybe it shouldn't be.

Good point, but I'm not sure what I want to do about it... The API endpoint probably should have been `/project/`, but that's much harder to change without breaking things.
 
> Also you forgot this:

Yep. Thank you.

cli/src/commands/project.rs
```diff
@@ -31,7 +31,7 @@ pub async fn get_project_list(api: &mut PhylumApi, pretty_print: bool) {
 /// Handle the project subcommand. Provides facilities for creating a new project,
```
kylewillmon, line 31: I didn't even need to change this one.

cd-work (approved):

PR #289, opened 2022-04-22
Create directory with mode 700 by default

### LINKED ISSUE #285: Config file directory permissions
The `phylum` tool [creates the directory](https://github.com/phylum-dev/cli/blob/f136965c0588cd5592dd355e4e857c809bba8280/cli/src/config.rs#L97) when necessary. But it does not restrict the permissions at all.

We should restrict the permissions of this directory if we are the ones to create it. The [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) that we are moving towards suggests this:

>  If, when attempting to write a file, the destination directory is non-existent an attempt should be made to create it with permission 0700. If the destination directory exists already the permissions should not be changed. The application should be prepared to handle the case where the file could not be written, either because the directory was non-existent and could not be created, or for any other reason. In such case it may choose to present an error message to the user.

cd-work:
Following the XDG spec, new directories will now be created with the
mode 700 set instead of 755.

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

> If, when attempting to write a file, the destination directory is
> non-existent an attempt should be made to create it with permission
> 0700. If the destination directory exists already the permissions
> should not be changed. The application should be prepared to handle
> the case where the file could not be written, either because the
> directory was non-existent and could not be created, or for any other
> reason. In such case it may choose to present an error message to the
> user.

Closes #285.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -179,7 +179,9 @@ pub fn get_home_settings_path() -> Result<PathBuf> {
         {
             fs::set_permissions(&old_config_path, Permissions::from_mode(0o600))?;
         }
-        fs::create_dir_all(config_path.parent().unwrap())?;
+        let config_dir = config_path.parent().unwrap();
+        fs::create_dir_all(&config_dir)?;
+        fs::set_permissions(&config_dir, Permissions::from_mode(0o700))?;
         fs::rename(old_config_path, &config_path).unwrap();
     }
 
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1,10 +1,11 @@
 use std::env;
 use std::fs;
 #[cfg(unix)]
-use std::fs::Permissions;
-use std::io;
+use std::fs::{DirBuilder, Permissions};
+use std::io::{self, Write};
 #[cfg(unix)]
-use std::os::unix::fs::{OpenOptionsExt, PermissionsExt};
+use std::os::unix::fs::{DirBuilderExt, OpenOptionsExt, PermissionsExt};
+#[cfg(unix)]
 use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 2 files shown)


kylewillmon (changes requested):
This solves the issue for the bin and data directories (i.e., `~/.local/bin/` and `~/.local/share/phylum`, respectively), but not for the config directory (`~/.config/phylum`) which is created in `config.rs` (either [here](https://github.com/phylum-dev/cli/blob/432e4e7829c90b20a011996ae4b582b570ab783d/cli/src/config.rs#L100) or [here](https://github.com/phylum-dev/cli/blob/432e4e7829c90b20a011996ae4b582b570ab783d/cli/src/config.rs#L182) depending on the scenario).

cd-work:
> This solves the issue for the bin and data directories (i.e., ~/.local/bin/ and ~/.local/share/phylum, respectively), but not for the config directory (~/.config/phylum) which is created in config.rs (either here or here depending on the scenario).

I've set it for the case where the config directory is created by the `phylum` CLI. I don't think we should bother with it when the user manually created that directory.

kylewillmon:
> I've set it for the case where the config directory is created by the phylum CLI. I don't think we should bother with it when the user manually created that directory.

I agree that we shouldn't change the permissions if the user manually created the directory. The other case I linked is the `save_config` function, which is what creates the config directory in the fresh install case.

maxrake (commented):
Missed in the first review...but two nits for `cli/src/install.sh`:

* This comment does not match the code (should be `zshrc`):

  ```sh
      # Remove old entries from bashrc.
      sed -i'' "/^fpath+=(\"\$HOME\/.phylum\/completions\")$/d" "${HOME}/.zshrc"
      sed -i'' "/^export PATH=\"\$HOME\/.phylum:\$PATH\"$/d" "${HOME}/.zshrc"
      sed -i'' "/^alias ph='phylum'$/d" "${HOME}/.zshrc"
  ```

* The `config_dir` variable on line 10 is unused:
  * `config_dir="${XDG_CONFIG_HOME:-${HOME}/.config}/phylum"`
  * Maybe this was left on purpose, to point out where the config is located since it's not otherwise obvious in this script

maxrake (dismissed):
My nits were fixed...thanks!

I'm sure you are already aware...but marking as needs changes since it doesn't build on Windows.

kylewillmon (approved):

kylewillmon (approved):

PR #290, opened 2022-04-22
Add `--project` option to `phylum history`

### LINKED ISSUE #277: Remove `project` subcommand from `ph history`
# Overview
The `phylum history` command includes a `project` subcommand which feels out of place. We should remove the functionality from this location and users should leverage `phylum project` or `phylum project list` to acquire the same output.

# Expected Behavior
The `phylum history` command does not include a `project` subcommand.

# Additional Context
image

cd-work:
This removes the `phylum history project` subcommand from the
documentation and prints a warninng when running it without the `--json`
flag.

Users are instructed to instead use the `phylum project` command.

Closes #277.

### DIFF
--- a/cli/src/commands/jobs.rs
+++ b/cli/src/commands/jobs.rs
@@ -139,7 +139,10 @@ pub async fn handle_history(api: &mut PhylumApi, matches: &clap::ArgMatches) -> 
                 Blue.paint("phylum analyze <lock_file>")
             );
         } else {
-            println!("Projects and most recent runs\n",);
+            if pretty_print {
+                println!("Projects and most recent runs\n",);
+            }
+
             print_response(&resp, pretty_print, None);
         }
     }
--- a/cli/src/commands/jobs.rs
+++ b/cli/src/commands/jobs.rs
@@ -99,22 +99,40 @@ pub async fn handle_history(api: &mut PhylumApi, matches: &clap::ArgMatches) -> 
 
         if let Some(project_name) = project_name {
             if project_job_id.is_none() {
+                print_user_warning!(
+                    "`phylum history project <PROJECT>` is deprecated, \
+                    use `phylum history --project <PROJECT>` instead"
+                );
+
                 let resp = api.get_project_details(project_name).await;
                 print_response(&resp, pretty_print, None);
             } else {
+                print_user_warning!(
+                    "`phylum history project <PROJECT> <JOB_ID>` is deprecated, \
+                    use `phylum history <JOB_ID>` instead"
+                );
+
                 // TODO The original code had unwrap in it above. This needs to
                 // be refactored in general for better flow
                 let job_id = resolve_job_id(project_job_id.expect("No job id found"))?;
                 action = get_job_status(api, &job_id, verbose, pretty_print, display_filter).await
             }
         } else {
+            print_user_warning!(
+                "`phylum history project` is deprecated, use `phylum project` instead"
+            );
+
             get_project_list(api, pretty_print).await;
         }
     } else if matches.is_present("JOB_ID") {
         let job_id = resolve_job_id(matches.value_of("JOB_ID").expect("No job id found"))?;
         action = get_job_status(api, &job_id, verbose, pretty_print, display_filter).await;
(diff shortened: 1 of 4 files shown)


cli/src/commands/jobs.rs
```diff
@@ -94,6 +94,12 @@ pub async fn handle_history(api: &mut PhylumApi, matches: &clap::ArgMatches) ->
         .and_then(|v| Filter::from_str(v).ok());
 
     if let Some(matches) = matches.subcommand_matches("project") {
+        if pretty_print {
+            print_user_warning!(
+                "`phylum history project` is deprecated, use `phylum project` instead"
+            );
+        }
+
```
kylewillmon, line 102: This should be moved inside the `if project_job_id.is_none()` section below since that is the case that duplicates the `phylum project` command

cli/src/commands/jobs.rs
```diff
@@ -94,6 +94,12 @@ pub async fn handle_history(api: &mut PhylumApi, matches: &clap::ArgMatches) ->
         .and_then(|v| Filter::from_str(v).ok());
 
     if let Some(matches) = matches.subcommand_matches("project") {
+        if pretty_print {
```
kylewillmon, line 97: I'm guessing this is to avoid causing problems in a piped scenario. But `print_user_warning!()` writes to stderr not stdout, so it's not actually an issue.

cd-work: Hm that's a fair point, I just didn't want to change any output that might be likely to be computer-read.

cli/src/commands/jobs.rs
```diff
@@ -114,16 +120,22 @@ pub async fn handle_history(api: &mut PhylumApi, matches: &clap::ArgMatches) ->
...
+        } else {
+            let resp = api.get_status().await;
+
+            if let Err(Some(StatusCode::NOT_FOUND)) = resp.as_ref().map_err(|e| e.status()) {
+                print_user_warning!(
+                    "No results found. Submit a lockfile for processing:\n\n\t{}\n",
+                    Blue.paint("phylum analyze <lock_file>")
+                );
+            } else {
+                println!("Projects and most recent runs\n",);
```
kylewillmon, line 135: Not really related to this issue, but this should be wrapped in `if pretty_print`
```suggestion
                if pretty_print {
                    println!("Projects and most recent runs\n",);
                }
```

kylewillmon (dismissed):

kylewillmon:
> @kylewillmon Do you think there's any value in keeping that old jobs table format, or are you fine with getting rid of it completely?

By "getting rid of it completely", you mean changing `phylum history project <name>` to return the same format as `phylum history --project <name>`? I think that would be acceptable.
 
> I think the thresholds are likely best retrieved with `phylum project thresholds <name>` or something like that, but one could also take the entirety of `phylum history project <name>` and put it under `phylum project <name>`?

I agree with `phylum project thresholds <name>`. But I also think we can (and perhaps should) just go without a command for getting project thresholds. Current customers just use the UI for this.

cd-work:
> By "getting rid of it completely", you mean changing phylum history project <name> to return the same format as phylum history --project <name>? I think that would be acceptable.

I basically mean what is done in this PR right now. Deprecating `phylum history project <name>` for now and should it ever get removed completely in the future, the format would be gone. Since that's the only place using that specific format. Currently there's no other way to print single-line project job status or thresholds afaik.

> I agree with phylum project thresholds <name>. But I also think we can (and perhaps should) just go without a command for getting project thresholds. Current customers just use the UI for this.

Seems fine to me, then at least we don't need to bloat up this PR even more.

kylewillmon:
> I basically mean what is done in this PR right now. Deprecating `phylum history project <name>` for now and should it ever get removed completely in the future, the format would be gone. Since that's the only place using that specific format. Currently there's no other way to print single-line project job status or thresholds afaik.

Got it. Yes, I'm good with that.

kylewillmon (approved):

kylewillmon (approved):

kylewillmon (approved):

PR #300, opened 2022-04-25
Use value hints to improve tab completion

### LINKED ISSUE #299: Tab completion does not complete filename in `phylum analyze`
Tab completion for `phylum analyze` does not complete filenames. So in a directory with a `requirements.txt` file, if you type:

```
phylum analyze req<TAB>
```

It does not tab complete. The result is that I am required to correctly spell requirements.... And my spelling isn't great.

I have only tested this in `zsh`, so it may not be a problem for `bash`

kylewillmon:
Add `value_hint` calls to filename arguments so that `clap_complete` generates better tab completion files and `zsh` allows tab completion of file arguments.

Fixes #299

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -1,4 +1,4 @@
-use clap::{arg, Command};
+use clap::{arg, Command, ValueHint};
 use git_version::git_version;
 
 const VERSION: &str = git_version!(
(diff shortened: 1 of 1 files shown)


maxrake (approved):
The changes were confirmed with a local build and test. It also appears that each of the affected commands were updated...so...

LGTM

PR #302, opened 2022-04-26
Hide the --prerelease arg in `phylum update`

### LINKED ISSUE #301: Hide `--prerelease` option in `ph update` command
# Overview
We do not want to encourage the use of prerelease builds, so this functionality should not be advertised. Rather than remove the option, it was decided to just suppress it from the help/docs.

https://docs.rs/clap/3.1.12/clap/struct.Arg.html#method.hide

kylewillmon:
This PR hides the `--prerelease` argument of `phylum update`.

I did not remove the help doc for argument because it is still available with tab completion (in zsh and fish):

```
❯ phylum update -<TAB>     
--help        -h  -- Print help information                                                                                                                                                                                                                                                              
--prerelease  -p  -- Update to the latest prerelease (vs. stable, default: false)                                                                                                                                                                                                                        
```

Fixes #301

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -33,7 +33,7 @@ pub fn app<'a>() -> clap::Command<'a> {
                 .about("Check for a new release of the Phylum CLI tool and update if one exists")
                 .arg(arg!(
                     -p --prerelease "Update to the latest prerelease (vs. stable, default: false)"
-                ))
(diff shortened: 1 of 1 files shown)


furi0us333:
FWIW, I left the option off of the version 2 docs I am working on.

cd-work (approved):

PR #311, opened 2022-04-27
Bump dependencies

kylewillmon:
This should take some of the workload off of poor ol' Dependabot.

With the exception of `nom`, all of these updates are SemVer compatible updates. Dependabot still has the job of alerting us of breaking changes.

### DIFF
--- a/cli/src/commands/uninstall.rs
+++ b/cli/src/commands/uninstall.rs
@@ -136,5 +136,5 @@ impl Shell {
 
 /// Get the user's home directory.
 fn home_dir() -> Result<PathBuf> {
-    home::home_dir().ok_or(anyhow!("Unable to find home directory"))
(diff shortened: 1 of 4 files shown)


cli/src/commands/uninstall.rs
```diff
@@ -136,5 +136,5 @@ impl Shell {
 
 /// Get the user's home directory.
 fn home_dir() -> Result<PathBuf> {
-    home::home_dir().ok_or(anyhow!("Unable to find home directory"))
+    home::home_dir().ok_or_else(|| anyhow!("Unable to find home directory"))
```
maxrake, line 139: Sounds like a threat to me!

maxrake (approved):
Assuming the status checks come back clean, this all LGTM.

PR #312, opened 2022-04-27
Bump dependencies with breaking changes

kylewillmon:
This PR bumps dependencies including many dependencies where the update includes a breaking change. I read through the Changelogs for these dependencies and updated our code accordingly


samtay (approved):
Ran a basic smoke test through a few commands, set thresholds, etc. and didn't observe any issues.

PR #314, opened 2022-04-27
fix(install): Check for files before running sed on them

### LINKED ISSUE #313: Installer fails when shell RC files do not exist
The installer fails if it is run on a host where either `.zshrc` or `.bashrc` do not exist.

kylewillmon:
It was noticed that the installer fails if either `.zshrc` or `.bashrc` does not exist. This PR fixes that.

Fixes #313


samtay (approved):
Tested locally, LGTM!

maxrake (approved):
Approving...but will need to coordinate/adjust/merge with the work done in #317 to fix #316

Issue #215, opened 2022-03-25
Enforce shellcheck in CI


maxrake:
# Overview

Recent updates to the CLI install script included adhering to standards as provided by the `shellcheck` QA tool. However, these standards can drift back out of compliance as future work occurs in the project. To combat this, the project should:

* add `shellcheck` to CI
  * Use version 0.8.0 or newer in order to make use of the `-o all` option
* fail builds when not adhering to the strict standards
* consider adding pre-commit options for developers
* add documentation or pointers on how to add `shellcheck` scanning to IDE(s)

_Originally suggested by @kylewillmon in https://github.com/phylum-dev/cli/issues/214#issuecomment-1079072318_

# Acceptance Criteria
- [ ] All shell scripts in the repo are scanned with `shellcheck` in CI
- [ ] The CI checks fail when there are QA findings
- [ ] The shell checks occur automatically for PRs and pushing to the `main' branch
- [ ] The shell checks can be triggered manually

## Associated User Story
--

maxrake:
The current [installed software for the GitHub hosted `ubuntu-latest` runner](https://github.com/actions/virtual-environments/blob/ubuntu20/20220227.1/images/linux/Ubuntu2004-Readme.md) makes use of an older version of `shellcheck` (v0.7.0-2build2). Starting with v0.8.0, the `-o all` option/feature was introduced...which is like a strict mode for scanning. This option should be used. Some options for getting this version include:
* use a [shellcheck GitHub action](https://github.com/marketplace/actions/shellcheck) with a specific version specified (or even just allow for the latest release)
* manually download and use a specific version, as outlined [here](https://github.com/koalaman/shellcheck/issues/704)

PR #315, opened 2022-04-27
Set permissions with umask instead of mkdir -m

kylewillmon:
Using `mkdir -m` has two issues. First, it always sets the mode rather
than only setting it when it creates the directory. Second, when
combined with `-p`, it only sets that mode for the last directory in the
chain. This method of using umask fixes both problems.

### DIFF
--- a/cli/src/install.sh
+++ b/cli/src/install.sh
@@ -101,7 +101,7 @@ copy_files() {
     bin_name="phylum"
 
     # Ensure binary directory exists.
-    mkdir -pm 700 "${bin_dir}"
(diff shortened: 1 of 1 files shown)


kylewillmon:
I have to give credit to `shellcheck` for this one. I noticed that it was showing [SC2174](https://github.com/koalaman/shellcheck/wiki/SC2174) for these lines. The fact that we didn't notice this sooner is a point for #215

samtay (approved):
Hmm I went to try this out and I'm getting some errors, but they're unrelated to this PR. Changes here LGTM, I'll open up some issues for the other things I'm observing.

PR #317, opened 2022-04-28
Fix sed on macOS

### LINKED ISSUE #316: Installer fails on macOS
# Overview
Installer is broken on macOS, due to assuming gnu-sed.

# How To Reproduce
1. Run installer on mac (without gnu-sed)
2. Observe:
```
❯ ./install.sh 

    phylum-cli installer

sed: 1: "/Users/samtay/.bashrc": bad flag in substitute command: 's'
```

# Expected Behavior
Work out of the box or document that `gsed` should be installed as `sed`.

# Additional Context
Might need a little wrapper that detects `uname` or an ugly `sed -i.bak ; rm file.bak` thing.

samtay:
# Overview

Makes the `./install.sh` work on macOS

The `-i.bak` and `rm $file.bak` is a little awkward, but I think this is preferable to something like
```bash
if [[ $(uname) = "Darwin" ]]; then
  sed -i '' 's/red/green/g' sed.txt
else
  sed -i'' 's/red/green/g' sed.txt
fi
```
because it also handles the case of MacOS with gnu-sed installed (which I think is quite common, at least I am in this camp 🙂 ).

## Checklist

## Issue
Closes #316.

### DIFF
--- a/cli/src/install.sh
+++ b/cli/src/install.sh
@@ -23,6 +23,11 @@ banner() {
     printf "\n    %bphylum-cli%b installer\n\n" "${GREEN}" "${NC}"
 }
 
+cross_sed() {
+    sed -i.bak "$1" "$2"
+    rm "${2}.bak"
+}
+
 # Get the platform name.
 get_platform() {
     platform_str=$(uname)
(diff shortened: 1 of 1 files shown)


cli/src/install.sh
```diff
@@ -23,6 +23,11 @@ banner() {
     printf "\n    %bphylum-cli%b installer\n\n" "${GREEN}" "${NC}"
 }
 
+cross_sed() {
+    sed -i.bak "$1" "$2"
+    rm "${2}.bak"
```
maxrake, line 28:
```suggestion
    sed -i'.bak' "$1" "$2"
    rm -f "${2}.bak"
```
The single quotes appear to be necessary...at least from the [blog post](https://singhkays.com/blog/sed-error-i-expects-followed-by-text/) used to confirm this behavior as a fix.

The `rm -f` will ensure no prompts for confirmation.

maxrake (changes requested):
Small change requested.

cli/src/install.sh
samtay: Huh, after looking at that blog post I'm going to just try `sed -i'' -e` before merging this. Good find, I had searched a bit for a simple incantation like that but came up dry.

maxrake: FWIW...I am on a M1 Mac and tried the `sed -i''` method and still had issues...but it is definitely possible I missed something. Trying whatever method is chosen on a non-macOS system is the other piece of this puzzle...don't want to break it for those users instead.

cli/src/install.sh
```diff
@@ -23,6 +23,11 @@ banner() {
     printf "\n    %bphylum-cli%b installer\n\n" "${GREEN}" "${NC}"
 }
 
+cross_sed() {
+    sed -i'.bak' "$1" "$2"
```
kylewillmon, line 27: I'm curious what happens if someone already has a `.bashrc.bak` file for some other reason. Does this command fail? Does it get overwritten?

maxrake (approved):
LGTM

cli/src/install.sh
samtay: Ah, I thought this worked at first, but it ends up leaving a `~/.zshrc-e` file.

cli/src/install.sh
samtay: It doesn't fail, but yes, it does get overwritten.

maxrake: It does not fail...and I _think_ it gets overwritten. The local tests I did for this incantation were done sequentially, before `rm -f`'ing the file. That all worked as well...but I can understand why removing each file directly would be advantageous. I can also see where you might be going with this...to check for the existence of a `*.bak` file _before_ making a new one or at least before deleting it unilaterally.

cli/src/install.sh
samtay: To test this I've been alternating between having `gsed` preferred (`PATH="/opt/homebrew/opt/gnu-sed/libexec/gnubin:$PATH"`) vs the mac default BSD preferred (`PATH="/usr/bin:$PATH"`) but yea a quick sanity check on a linux box would be good.

cli/src/install.sh
kylewillmon: Nah. Just curious. This is probably fine

samtay: I mean, saying this makes me feel like a bad person 😆 but an easy mitigation would be an odd extension, something  like `sed -i'.phylum.bak' ; rm -f ${2}.phylum.bak`.

maxrake (dismissed):

cd-work (commented):
It might be easier to just use `perl`?

The following seems to work for example:

```
perl -i -pe 's/<LINE>\n//g' <FILE>
```

kylewillmon (approved):

cd-work (approved):
Works on Linux.

PR #318, opened 2022-04-28
fix(install): Use printf instead of echo

kylewillmon:
Using escape sequences in echo relies on XSI-conformance in addition to
POSIX conformance. Using printf is more portable.

Ref: https://github.com/koalaman/shellcheck/wiki/SC2028
Ref: https://www.man7.org/linux/man-pages/man1/echo.1p.html

### DIFF
--- a/cli/src/install.sh
+++ b/cli/src/install.sh
@@ -68,7 +68,7 @@ autoload -U compinit && compinit" \
     > "${phylum_rc}"
 
     if ! grep -q "source ${phylum_rc}" "${rc_path}"; then
-        echo "\nsource ${phylum_rc}" >> "${rc_path}"
+        printf "\nsource %s\n" "${phylum_rc}" >> "${rc_path}"
     fi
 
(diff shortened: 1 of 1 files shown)


maxrake (approved):
Beat me to it...`shellcheck` on strict mode FTW.

One note for the changes here...the entries are added with new lines but the uninstall command does not appear to check for those same new lines. That leaves shell config files with extra blank lines after an uninstall. FWIW...that same behavior was present before this PR, but now there are additional new line. Not a showstopper, but maybe worth mentioning.

kylewillmon:
> One note for the changes here...the entries are added with new lines but the uninstall command does not appear to check for those same new lines. That leaves shell config files with extra blank lines after an uninstall. Not a showstopper, but worth mentioning.

It's possible that the user has put something on that blank line in the time between install and uninstall. And we wouldn't want to accidentally join two lines, so I think the behavior of the uninstaller is correct.

PR #320, opened 2022-04-28
Add `--bearer` parameter to `phylum auth token`

### LINKED ISSUE #319: Add `auth token --bearer` parameter to print Phylum API key
# Overview

The current `phylum auth token` command prints the Phylum refresh token, which is a long-lived token that isn't particularly useful on its own.

To make it easy for people to send their own API requests without having to handle the conversion from long-lived refresh token to short-lived access token, a new `--bearer` parameter should print the short-lived token instead. This could then be used directly inside of curl with `-H "Authorization: Bearer $(phylum auth token --bearer)"`.

See https://github.com/phylum-dev/cli/issues/298#issuecomment-1109994212 for additional motivation behind this.

# Acceptance Criteria

- [ ] New `--bearer` parameter prints short-lived access token

cd-work:
This patch adds the new `--bearer` parameter which will print the
short-lived API access token rather than the long-lived refresh token.

Using this new command it should be much easier to make `curl` requests
to the phylum API, without having to handle login manually.

Closes #319.

### DIFF
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -4,6 +4,7 @@ use anyhow::{anyhow, Context, Result};
 use clap::Command;
 
 use crate::api::PhylumApi;
+use crate::auth;
 use crate::commands::{CommandResult, ExitCode};
 use crate::config::{save_config, Config};
 use crate::print::print_sc_help;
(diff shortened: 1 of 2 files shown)


cli/src/commands/auth.rs
```diff
@@ -65,18 +66,24 @@ pub async fn handle_auth_status(
...
                 "User is not currently authenticated, please login with `phylum auth login`"
             );
-            Ok(ExitCode::NotAuthenticated.into())
+            return Ok(ExitCode::NotAuthenticated.into());
         }
+    };
+
+    if matches.is_present("bearer") {
+        let tokens = auth::handle_refresh_tokens(&config.auth_info, refresh_token).await?;
+        println!("{}", String::from(&tokens.access_token));
```
kylewillmon, line 82: It seems like `AccessToken` should implement `Display`, especially since `RefreshToken` already does.

kylewillmon (dismissed):

kylewillmon (approved):

PR #321, opened 2022-04-28
Simplify the PR template

kylewillmon:
This patch removes the boilerplate from our pull request template. What
is left is only the checklist, which is important to ensure that nothing
is missed before a PR is merged.

If developers are using detailed commit messages (like this one), PRs
will automatically be created with the detailed commit message followed
by the checklist, which is exactly how we want them.

Fix #321.

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,14 +1,7 @@
-# Overview
-
-Provide information on what this PR does and any context necessary to understand its implementation.
 
 ## Checklist
 
-- [ ] Does this PR have an associated issue?
+- [ ] Does this PR have an associated issue (i.e., `closes #<issueNum>` in description above)?
 - [ ] Have you ensured that you have met the expected acceptance criteria?
 - [ ] Have you created sufficient tests?
 - [ ] Have you updated all affected documentation?
-
-## Issue
-
-What issue(s) does this PR close? Use the `closes #<issueNum>` here.


maxrake (approved):
Works for me. Detailed commit messages are great...and having them automatically populate the PR is even better. Approving, but also mentioning @furi0us333 in case he has some additional SOC2 related insight/background to share.

cd-work:
Is this checklist something motivated by SOC compliance or are there other motivational factors?

I personally feel like pull request templates create a lot of overhead at the wrong place. If you're aware that your PR is missing things, it should likely be created as a draft instead of a finished PR. And the person validating compliance with the acceptance criteria should be a separate party (the reviewer), not the author of the PR?

I understand its motivation as a reminder what kind of things you might want to look at when writing a PR, but that would only be necessary the first or second time. After that I'd assume you know the basic mechanics of the repository. Considering that the majority of contributions in any project come from existing contributors, that seems inefficient.

louislang:
SOC only demands that we track changes in an associated issue tracker. So as long as there is _an issue_ it probably doesn't matter that it's tied to the PR or not. Tying to a PR just helps us down the road should we need to triage something.

kylewillmon:
> SOC only demands that we track changes in an associated issue tracker. So as long as there is _an issue_ it probably doesn't matter that it's tied to the PR or not. Tying to a PR just helps us down the road should we need to triage something.

I am curious about this. Our issue tracker is Github. And Github uses both issues and PRs to track issues. It even uses the same numbering and reference format for PRs and issues (e.g., #321). So for small changes, it seems like the PR itself should be enough to track the issue.

samtay (approved):

PR #322, opened 2022-04-28
Use phylum-types Display implementations

cd-work:


kylewillmon (approved):

kylewillmon:
Looks like you picked up a version bump of the `http` library in this as well

cd-work:
> Looks like you picked up a version bump of the http library in this as well

Is there any policy on version bumps? I just `cargo update`'ed  for the latest phylum-types and then rebased on top of the commit that already updated.

Generally I feel like completely unrelated updates when they aren't necessary (like updating in every PR), probably just causes unnecessary review overhead. But otoh updating generally shouldn't hurt considering crates should comply with semver.

samtay (dismissed):

kylewillmon:
> Is there any policy on version bumps? I just `cargo update`'ed for the latest phylum-types and then rebased on top of the commit that already updated.

Now that we have Dependabot enabled, we'll see PRs created for these updates every Monday. We haven't really set a policy for doing updates outside of that.

The way to avoid this is to be more specific when you run `cargo update`:

```
cargo update -p phylum_types
```

But in this case, it's just 1 dep and it's probably fine.

maxrake (approved):

PR #323, opened 2022-04-29
chore: refresh the release script

maxrake:
While releasing the last handful of CLI versions it became clear that
the release script used to do so could use a refresh. For one, it
automated _everything_, which felt risky and most of the time the script
would be aborted before it could tag and push the changes. Another
minor annoyance was not remembering the format of the version or
changelog entry. Additionally, the git log summary would only show from
the last tag, not taking into account that the last tag could be from a
pre-release.

These issues have been addressed, along with some minor
refactoring and formatting. It is still just a basic shell script and
there could certainly be better ways yet.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,29 +1,48 @@
 #!/bin/sh
 
-LATEST=$(git describe --tags --abbrev=0)
+# Releasing a new version of the CLI is initiated with a tag and completed with
+# the Release workflow in CI. Run this script from the `main` branch and follow
+# the prompts to initiate a new release. There are a few manual steps that are
+# provided at the end. This is to ensure a chance to review the automated work
+# and not accidentally release a new version.
 
-printf "Latest release: %s\n\nGit log:\n" "${LATEST}"
+LATEST=$(git describe --tags --abbrev=0 --exclude="*-rc*")
+printf "Latest release: %s\n\n" "${LATEST}"
+printf "Git log since latest release:\n"
 git log --oneline HEAD "^${LATEST}"
 printf "\n"
 
-printf "version: "
+printf "version (w/o a leading 'v'): "
 read -r version
+TAG=v${version}
 
-printf "changelog: "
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -1,29 +1,48 @@
...
-git tag --sign -m "${TAG} - ${changelog}" "${TAG}"
-git push "${TAG}"
+git commit -m "${commit_message}"
+
+cat << __instructions__
+
+The automation is done. Run the following commands manually,
+in sequence, to tag the release and push the changes:
+
+    git tag --sign -m "${TAG} - ${changelog}" ${TAG}
```
kylewillmon, line 45: I think this should be done automatically. As long as the push is manual, everything else can be automated

release.sh
```diff
@@ -1,29 +1,48 @@
...
 
-echo Tagging / pushing "${TAG}", press any key to proceed...
+printf "Press enter to proceed with the commit ..."
 read -r
-git push
-git tag --sign -m "${TAG} - ${changelog}" "${TAG}"
-git push "${TAG}"
+git commit -m "${commit_message}"
+
+cat << __instructions__
```
kylewillmon, line 40: Before printing these instructions, it might be nice to show some details about the changes that have been made.

For example, showing the output of one or more of these commands:

* `git log main ^origin/main` (maybe with a `--stat`)
* `git diff origin/main main`
* `git show "$TAG"` (if the tag is automated)

release.sh
```diff
@@ -1,29 +1,48 @@
 #!/bin/sh
```
samtay, line 2: Seems like a script where we should exit early and loudly on failure.
```suggestion

set -eu

```

maxrake: added in the latest commit

samtay (approved):
Lgtm!

kylewillmon (approved):

PR #329, opened 2022-04-29
Add explicit option to disable thresholds from CLI

### LINKED ISSUE #288: CLI allows "disabling" total project threshold
In the `phylum project set-thresholds` command, using a score of 0 (zero) for the "TOTAL PROJECT Threshold" claims that this is "disabling" the threshold:

```
✔ TOTAL PROJECT Threshold · 0

Disabling total project risk domain
```

However, in the UI, there is no option to disable the total project threshold. The individual risk domains all have sliders next to them, but the total project threshold does not:

total-project-threshold

So this threshold is actually enabled with a value of 0

cd-work:
Previously, when setting the project thresholds from the CLI, a value of
`0` would automatically disable the specified threshold. This does not
correctly represent the backend configuration, since some thresholds
cannot be disabled and it is possible to have a threshold enabled with a
value of `0`.

To address both of these issues, an explicit `Disabled` option has been
added to the interactive `set-thresholds` command, which allows
controlling the `active` state of the threshold independently from its
value.

To ensure that the "Total Project" threshold is not disabled, a special
case has been added which will inform the user that a value must be set
for this threshold.

Closes #288.

### DIFF
--- a/cli/src/commands/project.rs
+++ b/cli/src/commands/project.rs
@@ -5,8 +5,6 @@ use anyhow::anyhow;
 use chrono::Local;
 use uuid::Uuid;
 
-use phylum_types::types::user_settings::Threshold;
-
 use super::{CommandResult, ExitCode};
 use crate::api::PhylumApi;
 use crate::config::{get_current_project, save_config, ProjectConfig, PROJ_CONF_FILE};
@@ -123,7 +121,10 @@ pub async fn handle_project(api: &mut PhylumApi, matches: &clap::ArgMatches) -> 
         );
         println!();
 
-        println!("Specify the thresholds and actions for {}. A threshold of zero will disable the threshold.", format_args!("{}", White.paint(project_name)));
+        println!(
+            "Specify the thresholds and actions for {}. Accepted values are 0-100 or 'Disabled'.",
+            format_args!("{}", White.paint(project_name))
+        );
         println!();
 
         let project_details = match api.get_project_details(project_name).await {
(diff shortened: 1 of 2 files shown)


maxrake (dismissed):
Confirmed the expected behavior by building and testing locally. There is room for additional documentation to be added somewhere in the `docs` directory...but that may already be in progress by @furi0us333 in the `brad/docs_v2` branch.

LGTM.

cli/src/commands/project.rs
```diff
@@ -142,17 +143,21 @@ pub async fn handle_project(api: &mut PhylumApi, matches: &clap::ArgMatches) ->
...
-        ]
-        .iter()
-        {
-            let (threshold, action) = prompt_threshold(threshold_name).unwrap_or((0, "none"));
+        ] {
+            let threshold = match prompt_threshold(threshold_name) {
+                Ok(threshold) => threshold,
+                Err(_) => {
+                    print_user_failure!("Failed to read user input");
+                    continue;
```
matt-phylum, line 158: Is it right that if an error occurs it is just reported to stderr and then we go on to the next threshold without updating `user_settings`? Would it be better if the user was prompted again or if an error was propagated so the operation aborts? It's similar to what the old code did so maybe it's okay.

cd-work: Yes, reporting to stderr and going to the next without updating settings is the goal. You wouldn't want to update them if the user failed to produce a valid input I'd imagine.

The previous version actually was a bit more devious. If it failed for some reason (generally this shouldn't fail), it would disable the threshold. I've changed that because I think it's usually not a good idea to change something when an error occurred.

maxrake (approved):

matt-phylum (approved):
Should the threshold names and whether or not those thresholds can be disabled be coming from a dependency or an API call? It seems like they mean something somewhere in the backend. Otherwise LGTM.

PR #330, opened 2022-04-29
Automatically run `cargo update` in a workflow

kylewillmon:
This runs a basic `cargo update` in a Github Action and creates a PR
with any changes. Combined with limiting Dependabot to major version
updates, this should resolve concerns about too many PRs and being too
strict in our Cargo.toml files.

Ref: Discussion on #325

### DIFF
--- a/.github/workflows/cargo-update.yml
+++ b/.github/workflows/cargo-update.yml
@@ -1,0 +1,27 @@
+---
+name: Test
+
+on:
+  workflow_dispatch:
+  # Run every Monday
+  schedule:
+    - cron: '30 5 * * 1'
+
+jobs:
+  cargo-update:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@v3
+
+      - name: Cargo update
+        run: cargo update
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: "chore: Bump dependencies"
+          body: |
+            Bump dependencies in Cargo.lock for all SemVer-compatible updates.
+          branch: auto-cargo-update
--- a/.github/workflows/cargo-update.yml
+++ b/.github/workflows/cargo-update.yml
@@ -1,5 +1,5 @@
 ---
-name: Test
+name: Update Deps
 
 on:
   workflow_dispatch:
@@ -21,7 +21,7 @@ jobs:
         uses: peter-evans/create-pull-request@v4
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          commit-message: "chore: Bump dependencies"
+          commit-message: "build: Bump dependencies"
           body: |
             Bump dependencies in Cargo.lock for all SemVer-compatible updates.
           branch: auto-cargo-update
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,9 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
+    ignore:
+      - dependency-name: "*"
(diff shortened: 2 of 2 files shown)


cd-work (approved):
> Combined with disabling Dependabot for major version
updates

Just to clarify your typo: Dependabot is disabled for everything **but** major versions. So major versions will still automatically get separate PRs.

I much prefer this over the previous state of things, though one thing that surprises me is that dependabot doesn't have the option to do this on its own? I feel like it should be capable of creating **one** PR to just bump all the minor stuff without having to define a separate action for it?

kylewillmon:
> Just to clarify your typo: Dependabot is disabled for everything **but** major versions. So major versions will still automatically get separate PRs.

You are correct... I wrote it backwards. I'll amend to correct it.

> I much prefer this over the previous state of things, though one thing that surprises me is that dependabot doesn't have the option to do this on its own? I feel like it should be capable of creating **one** PR to just bump all the minor stuff without having to define a separate action for it?

The feature was requested a full three years ago in dependabot/dependabot-core#2174, but there hasn't been much progress from what I can tell.

.github/workflows/cargo-update.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v3
+
+      - name: Cargo update
+        run: cargo update
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: "chore: Bump dependencies"
```
maxrake, line 24:
```suggestion
          commit-message: "build: Bump dependencies"
```
The conventional commit types, for the default Angular format, in use by the tool(s) that are likely to be used to enforce this standard are [HERE](https://github.com/commitizen/conventional-commit-types/blob/master/index.json) and `build` seems more appropriate than `chore`:
> Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)

cd-work (dismissed):

.github/workflows/cargo-update.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@v3
+
+      - name: Cargo update
+        run: cargo update
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
```
maxrake, line 22: There is a question about whether `committer` and `author` should be specified here...to ensure verified commits. It should be possible to specify the GitHub Actions user - `github-actions@github.com` - to satisfy that requirement...but maybe not worth making that change until the need for it is proven.

`title` may also need to be explicitly specified, if the PRs end up getting titles with the default of `Changes by create-pull-request action`.

.github/workflows/cargo-update.yml
```diff
@@ -0,0 +1,27 @@
+---
+name: Test
```
maxrake, line 2:
```suggestion
name: update_deps
```
...or whatever...but there is already a `Test` workflow.

maxrake (changes requested):
The workflow name will need to change.

.github/workflows/cargo-update.yml
kylewillmon: I don't think setting `committer` and `author` is enough to get verified commits. But I'm not sure it matters anyways. Unlike Dependabot, this will require a human to do the merge. And with a "squash and merge", the commit will be rewritten and (hopefully) signed at that point.

Interesting question about the title... I should say that I pulled most of the options here directly from [this example](https://github.com/peter-evans/create-pull-request/blob/main/docs/examples.md#update-cargo-dependencies).

I'd like to see what happens without specifying those things. We can always update it later.

maxrake (approved):
The output behavior can be observed and tweaked later, if desired.

LGTM

PR #332, opened 2022-04-29
fix: Resolve project create errors

### LINKED ISSUE #331: `phylum project create` error
# Overview
There seems to be a false error in the new `v3.0.0` around project creation.

# How To Reproduce
1. Update to CLI `v3.0.0`
2. Attempt to create a project `phylum project create test`
3. See error for invalid project name, even with simple name

# Expected Behavior
Project names that abide by the stated limitations are accepted.

# Additional Context
image

kylewillmon:
This is the quickest way to fix #331. Even in staging, the API is not
currently accepting the `group_name` parameter for project creation and
this is causing an error (which is getting misrepresented as being about
the characters in the project name).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1266,7 +1266,7 @@ dependencies = [
 [[package]]
 name = "phylum_types"
 version = "0.1.0"
-source = "git+https://github.com/phylum-dev/phylum-types?branch=development#c9960ef94be75e5829d62eb3e2937e7e77dbc141"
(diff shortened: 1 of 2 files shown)


cd-work:
This is an alternative solution which works and should not require any changes in the future:
https://github.com/phylum-dev/phylum-types/pull/30

maxrake (dismissed):
It points to the correct commit.

cd-work (approved):

samtay (approved):

PR #343, opened 2022-05-04
Fix yarn lockfile parsing with patched deps

### LINKED ISSUE #342: Add support for Yarn v3
# Overview
We currently have support for Yarn v1 and v2, but not v3.

# Acceptance Criteria
- [ ] Yarn v3 lock files are correctly parsed and processed

This might be helpful: https://dev.to/arcanis/yarn-3-0-performances-esbuild-better-patches-e07

cd-work:
Yarn allows patching dependencies, which produces a local patch file
applied on top of a dependency. Before this patch, this would cause an
error in the lockfile parser.

To provide the most value possible, this patch adds support for parsing
these patched dependencies and extracts the original package, while
discarding the local patch information. This allows analyzing the
unpatched dependency for vulnerabilities.

Closes #342.

### DIFF
--- a/cli/src/lockfiles/javascript.rs
+++ b/cli/src/lockfiles/javascript.rs
@@ -91,16 +91,24 @@ impl Parseable for YarnLock {
             let resolution = package
                 .get(&"resolution".into())
                 .and_then(YamlValue::as_str)
+                .filter(|s| !s.is_empty())
                 .ok_or_else(|| "Failed to parse resolution field in yarn lock file".to_owned())?;
 
             // Ignore workspace-local dependencies like project itself ("project@workspace:.").
-            if resolution.contains("@workspace:") {
+            if resolution[1..].contains("@workspace:") {
                 continue;
             }
 
-            let (name, _version) = resolution
-                .rsplit_once("@npm:")
-                .ok_or_else(|| "Failed to parse name in yarn lock file".to_owned())?;
+            let (name, _version) = if let Some(index) = resolution[1..].find("@patch:") {
+                // Extract npm version from patched dependencies.
+                resolution[index + "@patch:".len() + 1..]
+                    .rsplit_once("@npm")
+                    .ok_or_else(|| "Failed to parse patch in yarn lock file".to_owned())?
+            } else {
+                resolution
+                    .rsplit_once("@npm")
+                    .ok_or_else(|| "Failed to parse name in yarn lock file".to_owned())?
+            };
 
             let version = package
                 .get(&"version".into())
@@ -198,7 +206,7 @@ mod tests {
 
         let pkgs = parser.parse().unwrap();
 
-        assert_eq!(pkgs.len(), 50);
+        assert_eq!(pkgs.len(), 51);
 
         let expected_pkgs = [
             PackageDescriptor {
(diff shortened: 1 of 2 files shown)


cli/src/lockfiles/javascript.rs
```diff
@@ -91,16 +91,24 @@ impl Parseable for YarnLock {
             let resolution = package
                 .get(&"resolution".into())
                 .and_then(YamlValue::as_str)
+                .filter(|s| !s.is_empty())
                 .ok_or_else(|| "Failed to parse resolution field in yarn lock file".to_owned())?;
 
             // Ignore workspace-local dependencies like project itself ("project@workspace:.").
-            if resolution.contains("@workspace:") {
+            if resolution[1..].contains("@workspace:") {
```
cd-work, line 98: Decided it would be a good idea to exclude the `@` at the beginning of package scopes here, to ensure we're only looking at the yarn data. According to some quick searches online, it seems like this is the only place where `@` is allowed in the package name.

cli/src/lockfiles/javascript.rs
```diff
@@ -91,16 +91,24 @@ impl Parseable for YarnLock {
...
-                .rsplit_once("@npm:")
-                .ok_or_else(|| "Failed to parse name in yarn lock file".to_owned())?;
+            let (name, _version) = if let Some(index) = resolution[1..].find("@patch:") {
+                // Extract npm version from patched dependencies.
+                resolution[index + 8..]
+                    .rsplit_once("@npm")
+                    .ok_or_else(|| "Failed to parse patch in yarn lock file".to_owned())?
+            } else {
+                resolution
+                    .rsplit_once("@npm")
```
cd-work, line 109: In the lockfile which caused the issue, the `:` after npm is encoded as `%3A`. I've decided to remove it from both `rsplit_once` instances since it shouldn't be necessary for matching and could potentially cause problems in both places if urlencoding is allowed here.

cli/src/lockfiles/javascript.rs
```diff
@@ -91,16 +91,24 @@ impl Parseable for YarnLock {
...
+            if resolution[1..].contains("@workspace:") {
                 continue;
             }
 
-            let (name, _version) = resolution
-                .rsplit_once("@npm:")
-                .ok_or_else(|| "Failed to parse name in yarn lock file".to_owned())?;
+            let (name, _version) = if let Some(index) = resolution[1..].find("@patch:") {
+                // Extract npm version from patched dependencies.
+                resolution[index + 8..]
```
matt-phylum, line 104: Should this be `"@patch:".len()+1`?

matt-phylum (dismissed):

cli/src/lockfiles/javascript.rs
cd-work: I can use `"@patch:".len() + 1`, I don't think splitting is necessary when you already know the size.

matt-phylum (dismissed):
re: `split_once`, I was thinking it would be cleaner not to have to type "@patch:" twice, and remembered doing something similar with split before. It'd be something like

```rust
let (name, _version) = if let Some((_, patch)) = resolution[1..].split_once("@patch:") {
    patch
        .rsplit_once("@npm")
        .ok_or_else(|| "Failed to parse patch in yarn lock file".to_owned())?
} else {
```

`split_once` translates to something like `value.find(pattern).map(|index| (&value[..index], &value[index+pattern.len()..]))` so it's not expensive. Also, for the way we're doing this code here, it remembers the `+1` for us because its view of the string always started from the second character. So I think it's a little better than my original suggestion, but either should be fine.

cd-work:
Oh right, good comment about the split once. I was actually hung up on the `+ 1` because I was thinking that would somehow prevent us from using the result of the split. But the part we omit from the beginning is part of what we don't need anyway, so we don't need it anyway.

matt-phylum (approved):

kylewillmon (approved):

PR #345, opened 2022-05-04
Fix yarn lockfile with unknown resolvers

### LINKED ISSUE #344: Add support for yarn v2 remote dependencies
# Overview

When a remote dependency like http(s)/ssh is specified in the yarn lockfile, the parser currently fails. While remote dependencies aren't analyzed, they shouldn't cause a complete failure.

# How To Reproduce

```
"express@ssh://git@github.com:expressjs/express":
  version: 4.18.1
  resolution: "express@ssh://git@github.com:expressjs/express#commit=d854c43ea177d1faeea56189249fff8c24a764bd"
  checksum: b96b66e0e7134c4c9ed2fdaf1d15397f330586a11fee909054c0705afc1bf976e234d5891d0ec294bb8d7c1872ce04291ed370b2280daee2adb207a5bbd39e67
  languageName: node
  linkType: hard
```

# Expected Behavior

The lockfile should be parsed without any errors, ignoring the remote dependency.

cd-work:
The yarn lockfile allows various different resolvers other than the
standard one which pulls from npm. Some of these include the filesystem
(@workspace), http/s (@http/s), and ssh (@ssh).

To ensure our parser is robust against all types of possible resolvers,
this patch checks for the presence of `@npm` to ensure that the
dependency can be resolved through npm. All other dependencies are
silently ignored.

This comes with the trade-off that it is easier for a failure to parse a
dependency, that can be resolved through npm, to be hidden from the
user, opening up the possibility for hidden vulnerabilities. It also
assumes that no package scope starts with `@npm`, otherwise the parser
might throw an error.

Closes #344.

### DIFF
--- a/cli/src/lockfiles/javascript.rs
+++ b/cli/src/lockfiles/javascript.rs
@@ -100,11 +100,16 @@ impl Parseable for YarnLock {
                 .get(&"resolution".into())
                 .and_then(YamlValue::as_str)
                 .filter(|s| !s.is_empty())
-                .ok_or_else(|| anyhow!("Failed to parse resolution field in yarn lock file"))?;
+                .ok_or_else(|| anyhow!("Failed to parse yarn resolution field"))?;
 
             let (name, mut resolver) = match resolution[1..].split_once('@') {
                 Some((name, resolver)) => (&resolution[..name.len() + 1], resolver.to_owned()),
-                None => return Err(anyhow!("Failed to parse yarn depenency resolution")),
+                None => {
+                    return Err(anyhow!(
+                        "Failed to parse yarn resolution field for '{}'",
+                        resolution
+                    ))
+                }
             };
 
             // Extract original resolver from patch.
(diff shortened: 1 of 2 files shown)


louislang (dismissed):

matt-phylum (approved):

kylewillmon (changes requested):
At a minimum, I think we need to print a warning for any packages that we ignore.

I've put additional concerns on the related issue (#344)

kylewillmon (dismissed):

cli/src/lockfiles/javascript.rs
```diff
@@ -102,30 +102,53 @@ impl Parseable for YarnLock {
...
+            };
+
+            // Extract original resolver from patch.
+            if let Some((_, patch)) = resolver.split_once("patch:") {
+                // Exctract resolver from `@scope/package@RESOLVER#patch`.
+                let patch = patch[1..].split_once('@');
+                let subresolver = patch.and_then(|(_, resolver)| resolver.split_once('#'));
+                resolver = match subresolver {
+                    Some((resolver, _)) => resolver.to_owned(),
+                    None => return Err(anyhow!("Failed to parse yarn patch dependency")),
```
kylewillmon, line 117: Some context here might be good.

```suggestion
                    None => return Err(anyhow!("Failed to parse yarn patch dependency for package '{}'", name)),
```

kylewillmon (approved):

PR #350, opened 2022-05-05
docs: update documentation to version 2 format

furi0us333:
Improve user documentation to better explain commands/subcommands.


docs/phylum.md
```diff
@@ -0,0 +1,37 @@
...
+* [phylum help](https://docs.phylum.io/docs/phylum_help)
+* [phylum history](https://docs.phylum.io/docs/phylum_history)
+* [phylum package](https://docs.phylum.io/docs/phylum_package)
+* [phylum ping](https://docs.phylum.io/docs/phylum_ping)
+* [phylum update](https://docs.phylum.io/docs/phylum_update)
+* [phylum version](https://docs.phylum.io/docs/phylum_version)
+
+### Options
+  `-c`, `--config <FILE>`
+  &emsp; Sets a custom configutation file
```
maxrake, line 24:
```suggestion
  &emsp; Use a specific configuration file
```
Typo...plus it's not really _setting_ a config file so much as it is _using_ a specific one. It is understood that the CLI code will need to change if this full suggestion is accepted...but at least the typo should be fixed here.

docs/phylum.md
```diff
@@ -0,0 +1,37 @@
...
+category: 6255e67693d5200013b1fa3e
+hidden: false
+---
+
+Interact with the Phylum API.
+
+### Core Commands
+* [phylum analyze](https://docs.phylum.io/docs/phylum_analyze)
+* [phylum auth](https://docs.phylum.io/docs/phylum_auth)
+* [phylum projects](https://docs.phylum.io/docs/phylum_projects)
```
maxrake, line 12:
```suggestion
* [phylum project](https://docs.phylum.io/docs/phylum_projects)
```
`phylum projects` does work...but as an alias. The current subcommand name is `project`...as can be seen in the output of a `phylum help` command.

maxrake (commented):
Super quick review here...didn't get the fine tooth comb out...but found at least a couple of things now that are worth noting.

docs/phylum.md
furi0us333: Ah yes, this happened after I had created this page. The link also needed to be changed, I have updated both.

docs/phylum.md
```diff
@@ -0,0 +1,37 @@
...
+  
+  `--no-check-certificate`
+  &emsp; Don't validate the server certificate when performing API requests
+  &emsp; DANGER: this should only be used when SSL termination breaks the Phylum certificate pinning
+  
+  `-t`, `--timeout <TIMEOUT>`
+  &emsp; Set the timeout (in seconds) for requests to the Phylum API
+  
+  `-V`, `--version`
+  &emsp; Print version information
```
cd-work, line 37: Based on the way this seems to be formatting, I think I'd suggest using a table instead?

The same applies to all the other places where flags are documented.

cd-work (changes requested):

docs/phylum.md
```diff
@@ -0,0 +1,37 @@
...
+* [phylum help](https://docs.phylum.io/docs/phylum_help)
+* [phylum history](https://docs.phylum.io/docs/phylum_history)
+* [phylum package](https://docs.phylum.io/docs/phylum_package)
+* [phylum ping](https://docs.phylum.io/docs/phylum_ping)
+* [phylum update](https://docs.phylum.io/docs/phylum_update)
+* [phylum version](https://docs.phylum.io/docs/phylum_version)
+
+### Options
+  `-c`, `--config <FILE>`
+  &emsp; Use a specific configuration file
```
matt-phylum, line 24: These options mostly match the command line help output. Should they match? Can we generate argument documentation from a common source to keep them in sync?

docs/phylum_analyze.md
```diff
@@ -0,0 +1,42 @@
+---
+title: phylum analyze
+category: 6255e67693d5200013b1fa3e
+hidden: false
+---
+
+Submit a request for analysis to the processing system
```
matt-phylum, line 7: Nit: In some files, this line ends with a period, but in other files it does not.

docs/phylum_auth_login.md
```diff
@@ -0,0 +1,10 @@
+---
+title: phylum auth login
+category: 6255e67693d5200013b1fa3e
+hidden: false
+---
+```sh
+phylum auth login
+```
+Login to an existing account.
```
matt-phylum, line 9: In some files, there is a brief summary of the command preceding its syntax, but in other files it follows the syntax.

docs/phylum_analyze.md
furi0us333: Thank you, I think I have sorted all of these now.

docs/phylum_auth_login.md
furi0us333: You can tell I was undecided. I have decided to put the summary line above and any additional context below. I believe this is all normalized now.

docs/phylum.md
furi0us333: Markdown renders differently on GitHub, I am going to go ahead and leave this as is for now and see how it renders on our documentation site. If we don't like it, I think a table would work nicely!

docs/phylum.md
furi0us333: You aren't wrong. It would be ideal to sync these things up from a common source, we should investigate how we might do that in the future.

docs/phylum_auth_status.md
```diff
@@ -0,0 +1,11 @@
+---
+title: phylum auth status
+category: 6255e67693d5200013b1fa3e
+hidden: false
+---
+
+Returns the current authentication status
```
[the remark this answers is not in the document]
furi0us333: Understood. I imagine there will be some iteration on this, so I will note this and correct in my next PR - thank you for all of the reviews!

PR #352, opened 2022-05-05
Add better error for missing lockfiles

### LINKED ISSUE #347: Unhelpful `phylum analyze` output for nonexistent lock file
# Overview

```
```

# Expected Behavior

# Additional Context

cd-work:
The existing error for missing lockfiles was somewhat opaque, making it
difficult for users to figure out how to resolve this issue.

Before:

```
```

After:

```
[2022-05-05T19:44:04Z ERROR phylum] Execution failed: Lockfile "/tmp/xxx" does not exist
❗ Error: Execution failed caused by: Lockfile "/tmp/xxx" does not exist
```

Closes #347.

### DIFF
--- a/cli/src/commands/jobs.rs
+++ b/cli/src/commands/jobs.rs
@@ -176,7 +176,7 @@ pub async fn handle_submission(
             .value_of("LOCKFILE")
             .ok_or_else(|| anyhow!("Lockfile not found"))?;
         let res = get_packages_from_lockfile(lockfile)
-            .ok_or_else(|| anyhow!("Unable to locate any valid package in package lockfile"))?;
+            .context("Unable to locate any valid package in package lockfile")?;
 
         packages = res.0;
         request_type = res.1;
--- a/cli/src/commands/lock_files.rs
+++ b/cli/src/commands/lock_files.rs
@@ -1,78 +1,82 @@
-use phylum_types::types::package::*;
 use std::path::Path;
+
+use anyhow::{anyhow, Result};
+use phylum_types::types::package::*;
 
 use crate::lockfiles::*;
 
 /// Attempt to get packages from an unknown lockfile type
-pub fn try_get_packages(path: &Path) -> Option<(Vec<PackageDescriptor>, PackageType)> {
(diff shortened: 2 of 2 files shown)


cli/src/commands/jobs.rs
```diff
@@ -175,7 +176,14 @@ pub async fn handle_submission(
         let lockfile = matches
             .value_of("LOCKFILE")
             .ok_or_else(|| anyhow!("Lockfile not found"))?;
-        let res = get_packages_from_lockfile(lockfile)
+
+        // Provide a useful error if the lockfile does not exist.
+        let lockfile_path = Path::new(lockfile);
+        if !lockfile_path.exists() {
+            return Err(anyhow!("Lockfile {:?} does not exist", lockfile));
```
matt-phylum, line 183: Anyhow has a `bail!` macro which replaces `return Err(anyhow!(`, but then it may not be as obvious the function is returning.

cli/src/commands/jobs.rs
```diff
@@ -1,5 +1,6 @@
 use std::io;
 use std::str::FromStr;
+use std::path::Path;
```
matt-phylum, line 3: `rustfmt` wants to sort these alphabetically.

```suggestion
use std::path::Path;
use std::str::FromStr;
```

matt-phylum (dismissed):

cd-work:
Initially I just wanted to go for the dead-simple "check if file exists" check, to report that and provide some extra benefit. But after thinking about it for a while, especially considering that there are many other file-related issues (like permissions), I decided that it's probably best when we report as much error information as possible.

So I decided to make this patch a little bigger, but hopefully the errors produced are now significantly more helpful, even for scenarios not specifically considered in the issue or while writing this patch.

cli/src/commands/jobs.rs
cd-work: I can see how that would be useful, but honestly `bail!` sounds to me like it would abort the program.

kylewillmon (approved):

PR #353, opened 2022-05-06
Extensions: installation management

### LINKED ISSUE #327: `phylum extension` subcommand installation management
# Overview
The user should be able to add, remove and list extensions via the appropriate `phylum extension` subcommand.

# Acceptance Criteria
- [ ] When a user runs `phylum extension add .` the extension in the current working directory should be installed.
- [ ] After a user installs a new extension, `foobar`, it should become available to the user under the `phylum` cli, e.g., running `phylum foobar` should execute the `foobar` extension.
- [ ] When a user installs a valid extension it should print a message indicating success. It should also print a quick guide on the extension to give the user some context on how the given extension works.
- [ ] When a user attempts to install an invalid extension, it should fail and inform the user as to why.
- [ ] When a user runs `phylum extension remove <extensionName>` the extension should be entirely removed from the user system.
- [ ] When a user runs `phylum extension` or `phylum extension list` a list of currently installed extensions, their versions and a short one sentence blurb on what the extension does should be shown in a table format.

## Associated User Story
https://www.notion.so/phylum/Extension-Base-CLI-Capability-bff6fcf7c0c947b79b8c49a928f59cc4

andreaphylum:
## Checklist

- Closes #327.
- Each one of the acceptance criteria in #327 is covered by one integration test. One of these tests is failing on purpose, pending decisions on how to run an extension and related features, and their implementation.
- Draft documentation in `docs/extensions.md`.

### DIFF
--- a/cli/src/bin/phylum.rs
+++ b/cli/src/bin/phylum.rs
@@ -206,9 +206,6 @@ async fn handle_commands() -> CommandResult {
         return handle_submission(&mut api, config, &matches).await;
     } else if let Some(matches) = matches.subcommand_matches("history") {
         return handle_history(&mut api, matches).await;
-    } else if let Some(matches) = matches.subcommand_matches("extensions") {
-        #[cfg(feature = "extensions")]
-        return handle_extensions(matches).await;
     } else if should_cancel {
         if let Some(matches) = matches.subcommand_matches("cancel") {
             let request_id = matches.value_of("request_id").unwrap().to_string();
@@ -217,6 +214,9 @@ async fn handle_commands() -> CommandResult {
             let resp = api.cancel(&request_id).await;
             print_response(&resp, true, None);
         }
+    } else if let Some(matches) = matches.subcommand_matches("extension") {
+        #[cfg(feature = "extensions")]
+        return handle_extensions(matches).await;
     }
 
     Ok(ExitCode::Ok.into())
--- a/cli/src/bin/phylum.rs
+++ b/cli/src/bin/phylum.rs
@@ -214,9 +214,11 @@ async fn handle_commands() -> CommandResult {
             let resp = api.cancel(&request_id).await;
             print_response(&resp, true, None);
         }
-    } else if let Some(matches) = matches.subcommand_matches("extension") {
+    } else if cfg!(feature = "extensions") {
         #[cfg(feature = "extensions")]
-        return handle_extensions(matches).await;
+        if let Some(matches) = matches.subcommand_matches("extension") {
+            return handle_extensions(matches).await;
+        }
     }
 
     Ok(ExitCode::Ok.into())
--- a/cli/src/bin/phylum.rs
+++ b/cli/src/bin/phylum.rs
@@ -10,6 +10,8 @@ use spinners::{Spinner, Spinners};
 
 use phylum_cli::api::PhylumApi;
 use phylum_cli::commands::auth::*;
+#[cfg(feature = "extensions")]
+use phylum_cli::commands::extensions::*;
 use phylum_cli::commands::jobs::*;
 use phylum_cli::commands::packages::*;
 use phylum_cli::commands::project::handle_project;
@@ -204,6 +206,9 @@ async fn handle_commands() -> CommandResult {
         return handle_submission(&mut api, config, &matches).await;
     } else if let Some(matches) = matches.subcommand_matches("history") {
         return handle_history(&mut api, matches).await;
+    } else if let Some(matches) = matches.subcommand_matches("extensions") {
+        #[cfg(feature = "extensions")]
+        return handle_extensions(matches).await;
     } else if should_cancel {
         if let Some(matches) = matches.subcommand_matches("cancel") {
             let request_id = matches.value_of("request_id").unwrap().to_string();
--- a/cli/src/commands/extensions/extension.rs
+++ b/cli/src/commands/extensions/extension.rs
@@ -24,7 +24,7 @@ impl Extension {
     }
 
     pub fn description(&self) -> Option<&str> {
-        self.manifest.description.as_ref().map(String::as_str)
+        self.manifest.description.as_deref()
     }
 
     /// Install the extension in the default path.
--- a/cli/src/commands/extensions/extension.rs
+++ b/cli/src/commands/extensions/extension.rs
@@ -29,6 +29,7 @@ impl Extension {
 
     /// Install the extension in the default path.
     pub fn install(&self) -> Result<()> {
+        println!("Installing extension {}...", self.name());
         let target_prefix = extensions_path()?.join(self.name());
 
         if target_prefix == self.path {
@@ -50,24 +51,31 @@ impl Extension {
             }
         }
 
+        println!("Extension {} installed correctly", self.name());
+
         Ok(())
     }
 
     pub fn uninstall(self) -> Result<()> {
+        println!("Uninstalling extension {}...", self.name());
         let target_prefix = extensions_path()?.join(self.name());
 
         if target_prefix != self.path {
-            return Err(anyhow!("extension is not installed, skipping"));
+            println!("{:?} {:?}", target_prefix, self.path);
+            return Err(anyhow!("extension {} is not installed, skipping", self.name()));
         }
 
         for entry in walkdir::WalkDir::new(&self.path).contents_first(true) {
             let entry = entry?.into_path();
+            println!("removing {}...", entry.to_string_lossy());
             if entry.is_dir() {
                 std::fs::remove_dir(entry)?;
             } else if entry.is_file() {
                 std::fs::remove_file(entry)?;
             }
         }
+
+        println!("Extension {} uninstalled correctly", self.name());
 
         Ok(())
     }
--- a/cli/src/commands/extensions/extension.rs
+++ b/cli/src/commands/extensions/extension.rs
@@ -1,0 +1,112 @@
+use std::{convert::TryFrom, fs::File, io::Read, path::PathBuf};
+
+use anyhow::{anyhow, Result};
+use serde::Deserialize;
+
+const MANIFEST_NAME: &str = "PhylumExt.toml";
+
+#[derive(Debug)]
+pub struct Extension {
+    path: PathBuf,
+    manifest: ExtensionManifest,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct ExtensionManifest {
+    name: String,
+    description: Option<String>,
+    entry_point: String,
+}
+
+impl Extension {
+    pub fn name(&self) -> &str {
+        &self.manifest.name
+    }
+
+    pub fn description(&self) -> Option<&str> {
+        self.manifest.description.as_ref().map(String::as_str)
+    }
+
+    /// Install the extension in the default path.
+    pub fn install(&self) -> Result<()> {
+        let target_prefix = extensions_path()?.join(self.name());
+
+        if target_prefix == self.path {
+            return Err(anyhow!("extension path and installation path are identical, skipping"));
+        }
+
+        for entry in walkdir::WalkDir::new(&self.path) {
+            let source_path = entry?.into_path();
+            let dest_path = target_prefix.join(source_path.strip_prefix(&self.path)?);
+
+            if source_path.is_dir() {
+                std::fs::create_dir_all(dest_path)?;
+            } else if source_path.is_file() {
+                if dest_path.exists() {
+                    return Err(anyhow!("{}: already exists", dest_path.to_string_lossy()));
+                } else {
+                    std::fs::copy(source_path, dest_path)?;
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Load an extension from the default path.
+    pub fn load(name: &str) -> Result<Extension, anyhow::Error> {
+        Extension::try_from(extensions_path()?.join(name))
+    }
+}
+
+// Load the extension from the specified path.
+impl TryFrom<PathBuf> for Extension {
+    type Error = anyhow::Error;
+
+    fn try_from(path: PathBuf) -> Result<Self, Self::Error> {
+        if !path.is_dir() {
+            return Err(anyhow!("{}: not a directory", path.to_string_lossy()));
+        }
+
+        let manifest_path = path.join(MANIFEST_NAME);
+        if !manifest_path.exists() {
+            return Err(anyhow!(
+                "{}: missing {}",
+                path.to_string_lossy(),
+                MANIFEST_NAME
+            ));
+        }
+
+        let mut buf = Vec::new();
+        File::open(manifest_path)?.read_to_end(&mut buf)?;
+
+        let manifest: ExtensionManifest = toml::from_slice(&buf)?;
+        let entry_point_path = path.join(&manifest.entry_point);
+
+        if !entry_point_path.exists() {
+            return Err(anyhow!(
+                "{}: entry point does not exist",
+                entry_point_path.to_string_lossy()
+            ));
+        }
+
+        if !entry_point_path.is_file() {
+            return Err(anyhow!(
+                "{}: entry point is not a file",
+                entry_point_path.to_string_lossy()
+            ));
+        }
+
+        // TODO add further validation if necessary:
+        // - Check that the name matches /^[a-z0-9-_]+$/
+        // - Check that the entry point is a supported format (.wasm?)
+        // - Check that the entry point is appropriately signed
+        Ok(Extension { path, manifest })
+    }
+}
+
+
+// Construct and return the extension path: $XDG_DATA_HOME/phylum/extensions
+pub fn extensions_path() -> Result<PathBuf, anyhow::Error> {
+    Ok(crate::config::data_dir()?.join("phylum").join("extensions"))
+}
--- a/cli/src/commands/extensions/extension.rs
+++ b/cli/src/commands/extensions/extension.rs
@@ -53,6 +53,25 @@ impl Extension {
         Ok(())
     }
 
+    pub fn uninstall(self) -> Result<()> {
+        let target_prefix = extensions_path()?.join(self.name());
+
+        if target_prefix != self.path {
+            return Err(anyhow!("extension is not installed, skipping"));
+        }
+
+        for entry in walkdir::WalkDir::new(&self.path).contents_first(true) {
+            let entry = entry?.into_path();
+            if entry.is_dir() {
+                std::fs::remove_dir(entry)?;
+            } else if entry.is_file() {
+                std::fs::remove_file(entry)?;
+            }
+        }
+
+        Ok(())
+    }
+
     /// Load an extension from the default path.
     pub fn load(name: &str) -> Result<Extension, anyhow::Error> {
         Extension::try_from(extensions_path()?.join(name))
--- a/cli/src/commands/extensions/mod.rs
+++ b/cli/src/commands/extensions/mod.rs
@@ -2,9 +2,9 @@ pub mod extension;
 
 use std::{convert::TryFrom, path::PathBuf};
 
-pub use extension::*;
-use log::info;
 use crate::commands::{CommandResult, CommandValue, ExitCode};
+pub use extension::*;
+use log::{error, info};
 
 use anyhow::Result;
 use clap::{arg, ArgMatches, Command, ValueHint};
@@ -26,8 +26,16 @@ pub fn command<'a>() -> Command<'a> {
 }
 
 /// Generate the subcommands for each extension.
-pub fn extensions_subcommands<'a>(command: Command<'a>) -> Result<Command<'a>> {
-    let extensions = list_extensions()?
+pub fn extensions_subcommands<'a>(command: Command<'a>) -> Command<'a> {
+    let extensions = match list_extensions() {
+        Ok(extensions) => extensions,
+        Err(e) => {
+            error!("Couldn't list extensions: {}", e);
+            return command;
+        }
+    };
+
+    let extensions = extensions
         .into_iter()
         .filter(|ext| {
             command
@@ -36,9 +44,9 @@ pub fn extensions_subcommands<'a>(command: Command<'a>) -> Result<Command<'a>> {
         })
         .collect::<Vec<_>>();
 
-    Ok(extensions.into_iter().fold(command, |command, ext| {
+    extensions.into_iter().fold(command, |command, ext| {
         command.subcommand(Command::new(ext.name()))
-    }))
+    })
 }
 
 /// Entry point for the `extensions` subcommand.
--- a/cli/src/commands/extensions/mod.rs
+++ b/cli/src/commands/extensions/mod.rs
@@ -1,71 +1,12 @@
-use std::{convert::TryFrom, fs::File, io::Read, path::PathBuf};
+pub mod extension;
 
+use std::{convert::TryFrom, path::PathBuf};
+
+pub use extension::*;
 use crate::commands::{CommandResult, CommandValue, ExitCode};
 
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use clap::{arg, ArgMatches, Command, ValueHint};
-use serde::Deserialize;
-
-const MANIFEST_NAME: &str = "PhylumExt.toml";
-
-#[derive(Debug)]
-pub struct Extension {
-    path: PathBuf,
-    manifest: ExtensionManifest,
-}
-
-#[derive(Deserialize, Debug)]
-pub struct ExtensionManifest {
-    name: String,
-    description: Option<String>,
-    entry_point: String,
-}
-
-// Load the extension from the specified path.
-impl TryFrom<PathBuf> for Extension {
-    type Error = anyhow::Error;
-
-    fn try_from(path: PathBuf) -> Result<Self, Self::Error> {
-        if !path.is_dir() {
-            return Err(anyhow!("{}: not a directory", path.to_string_lossy()));
-        }
-
-        let manifest_path = path.join(MANIFEST_NAME);
-        if !manifest_path.exists() {
-            return Err(anyhow!(
-                "{}: missing {}",
-                path.to_string_lossy(),
-                MANIFEST_NAME
-            ));
-        }
-
-        let mut buf = Vec::new();
-        File::open(manifest_path)?.read_to_end(&mut buf)?;
-
-        let manifest: ExtensionManifest = toml::from_slice(&buf)?;
-        let entry_point_path = path.join(&manifest.entry_point);
-
-        if !entry_point_path.exists() {
-            return Err(anyhow!(
-                "{}: entry point does not exist",
-                entry_point_path.to_string_lossy()
-            ));
-        }
-
-        if !entry_point_path.is_file() {
-            return Err(anyhow!(
-                "{}: entry point is not a file",
-                entry_point_path.to_string_lossy()
-            ));
-        }
-
-        // TODO add further validation if necessary:
-        // - Check that the name matches /^[a-z0-9-_]+$/
-        // - Check that the entry point is a supported format (.wasm?)
-        // - Check that the entry point is appropriately signed
-        Ok(Extension { path, manifest })
-    }
-}
 
 pub fn command<'a>() -> Command<'a> {
     Command::new("extension")
@@ -90,12 +31,12 @@ pub fn extensions_subcommands<'a>(command: Command<'a>) -> Result<Command<'a>> {
         .filter(|ext| {
             command
                 .get_subcommands()
-                .all(|p| p.get_name() != ext.manifest.name)
+                .all(|p| p.get_name() != ext.name())
         })
         .collect::<Vec<_>>();
 
     Ok(extensions.into_iter().fold(command, |command, ext| {
-        command.subcommand(Command::new(&ext.manifest.name))
+        command.subcommand(Command::new(ext.name()))
     }))
 }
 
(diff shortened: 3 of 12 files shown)


cli/src/commands/extensions/extension.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    pub fn description(&self) -> Option<&str> {
+        self.manifest.description.as_deref()
+    }
+
+    /// Install the extension in the default path.
+    pub fn install(&self) -> Result<()> {
+        println!("Installing extension {}...", self.name());
+        let target_prefix = extensions_path()?.join(self.name());
```
kylewillmon, line 33: This is going to need some validation checking here to make sure this doesn't escape the `extensions_path()` directory (e.g., if someone tries to set `name` to `../../../bin`)

cli/src/commands/extensions/extension.rs
```diff
@@ -0,0 +1,139 @@
...
+
+        if target_prefix == self.path {
```
kylewillmon, line 35: This should probably also check if `target_prefix` already exists (e.g., when upgrading a plugin)

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+use anyhow::Result;
+use clap::{arg, ArgMatches, Command, ValueHint};
+
+pub fn command<'a>() -> Command<'a> {
+    Command::new("extension")
+        .about("Run extensions")
+        .subcommand(
+            Command::new("add")
+                .about("Install extension")
+                .arg(arg!([PATH]).required(true).value_hint(ValueHint::FilePath)),
```
kylewillmon, line 18: The argument here should be a directory, so `ValueHint::DirPath` is more correct.

```suggestion
                .arg(arg!([PATH]).required(true).value_hint(ValueHint::DirPath)),
```

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+    };
+
+    let names = command
+        .get_subcommands()
+        .map(|n| n.get_name().to_string())
+        .collect::<HashSet<_>>();
+
+    extensions
+        .into_iter()
+        .filter(|ext| !names.contains(ext.name()))
```
kylewillmon, line 46: It might be nice to have a warning here for any extensions that get filtered out.

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+        subcmd_remove_extension(matches.value_of("NAME").unwrap()).await
+    } else {
+        // also covers the `list` case
+        subcmd_list_extensions().await
+    }
+}
+
+/// Handle the `extension add` subcommand path.
+/// Add the extension from the specified path.
+pub async fn subcmd_add_extension(path: PathBuf) -> CommandResult {
```
kylewillmon, line 66: This and the rest of the `subcmd_*` functions should probably not be `pub`

cli/src/commands/extensions/extension.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        println!("Extension {} uninstalled correctly", self.name());
+
+        Ok(())
+    }
+
+    /// Load an extension from the default path.
+    pub fn load(name: &str) -> Result<Extension, anyhow::Error> {
+        Extension::try_from(extensions_path()?.join(name))
```
kylewillmon, line 90: Just like the other use of `.join(..)`, this function also need a check to ensure that it doesn't escape `extensions_path()`.

It probably makes sense to extract the check into a `extension_path(name)` function that does `join()` and the check.

cli/src/commands/extensions/extension.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        Ok(())
+    }
+
+    pub fn uninstall(self) -> Result<()> {
+        println!("Uninstalling extension {}...", self.name());
+        let target_prefix = extensions_path()?.join(self.name());
+
+        if target_prefix != self.path {
+            println!("{:?} {:?}", target_prefix, self.path);
```
kylewillmon, line 66: This was probably accidentally left in

```suggestion
```

cli/src/bin/phylum.rs
```diff
@@ -212,6 +214,11 @@ async fn handle_commands() -> CommandResult {
             let resp = api.cancel(&request_id).await;
             print_response(&resp, true, None);
         }
+    } else if cfg!(feature = "extensions") {
+        #[cfg(feature = "extensions")]
+        if let Some(matches) = matches.subcommand_matches("extension") {
+            return handle_extensions(matches).await;
+        }
```
kylewillmon, line 221: Extensions don't use the API, so this code should be moved up to be before the creation of the `PhylumApi` object. That will allow unauthenticated users to run `extension` commands.

More importantly, that will allow the `assert_cmd` tests to work correctly even without API authentication.

cli/tests/extensions.rs
```diff
@@ -0,0 +1,298 @@
...
+
+impl AsRef<OsStr> for TmpDir {
+    fn as_ref(&self) -> &OsStr {
+        &self.0.as_os_str()
+    }
+}
+
+impl Drop for TmpDir {
+    fn drop(&mut self) {
+        // std::fs::remove_dir_all(&self.0).unwrap();
```
kylewillmon, line 292: Why is this commented out?

cli/tests/extensions.rs
```diff
@@ -0,0 +1,298 @@
...
+
+struct TmpDir(PathBuf);
+
+impl TmpDir {
+    fn new() -> Self {
+        let dir_name: String = rand::thread_rng()
+            .sample_iter(&rand::distributions::Alphanumeric)
+            .take(16)
+            .map(char::from)
+            .collect();
```
kylewillmon, line 270: It'd be nice to use something descriptive in the name (e.g., `phylum-test-...` where `...` is the 16 random alphanumeric characters)

cli/tests/extensions.rs
```diff
@@ -0,0 +1,298 @@
...
+
+impl Drop for TmpDir {
+    fn drop(&mut self) {
+        // std::fs::remove_dir_all(&self.0).unwrap();
+    }
+}
+
+fn tmp_path() -> PathBuf {
+    project_root().join("target").join("tests-tmp")
+}
```
kylewillmon, line 298: I would think a more natural temporary directory would be `std::env::temp_dir()`. Is there some reason that wouldn't work?

docs/extensions.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Extensions
+category: 61e72e3a50a88e001a92ee5d
+---
```
kylewillmon, line 4: These docs should be hidden until MVP

```suggestion
hidden: true
---
```

docs/extensions.md
```diff
@@ -0,0 +1,50 @@
...
+
+It is possible to extend the Phylum CLI with external sub-commands.
+
+To install an extension, run the following command:
+
+```sh
+phylum extension add path/to/extension
```
kylewillmon, line 11: The future evolution of this command needs to be considered.

If we are going to eventually have a marketplace, we will want `phylum extension add <ext_name>` to go get it from the marketplace and not from a local directory.

So maybe this should be `phylum extension add --local path/to/extension`?

kylewillmon (changes requested):

cli/src/commands/extensions/extension.rs
```diff
@@ -0,0 +1,139 @@
...
+pub struct Extension {
+    path: PathBuf,
+    manifest: ExtensionManifest,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct ExtensionManifest {
+    name: String,
+    description: Option<String>,
+    entry_point: String,
```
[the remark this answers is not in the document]
andreaphylum: Absolutely -- I'd consider this specific struct just as an example. If we end up going down the single WASM file road, we may not need to expose this at all.

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+// Return a list of installed extensions. Filter out invalid extensions instead of exiting early.
+fn list_extensions() -> Result<Vec<Extension>> {
+    let extension_path = extensions_path()?;
+
+    if !extension_path.exists() {
+        return Ok(Vec::new());
+    }
+
+    Ok(std::fs::read_dir(extension_path)?
+        .filter_map(|d| Extension::try_from(d.map(|d| d.path()).ok()?).ok())
```
cd-work, line 111: Is the `d` here supposed to stand for `direntry`? I'd honestly prefer `e` because I initially thought this was for `directory` which is inaccurate.

I'd also be fine with more verbosity here, even if it causes the closure to be multi-line.

cd-work, line 111: Also can you do `d.and_then(|d| Extension::try_from(d.path().ok()))` here?

cli/tests/extensions.rs
andreaphylum: That would most likely work, I went for this solution in the interest of manual inspections and holding everything under the repo (as `target` is meant to be "ephemeral" anyway). I'm open to switching it out if needs be.

cli/tests/extensions.rs
```diff
@@ -0,0 +1,298 @@
...
+    Command::cargo_bin("phylum")
+        .unwrap()
+        .env("XDG_DATA_HOME", &tmp_dir)
+        .arg("extension")
+        .arg("add")
+        .arg(fixtures_path().join("sample-extension"))
+        .assert()
+        .success();
+
+    std::env::set_var("XDG_DATA_HOME", &tmp_dir);
```
cd-work, line 28: Should import `env` here.

cli/src/commands/extensions/extension.rs
```diff
@@ -0,0 +1,143 @@
...
+        println!("Installing extension {}...", self.name());
+        let target_prefix = extensions_path()?.join(self.name());
+
+        if target_prefix == self.path {
+            return Err(anyhow!(
+                "extension path and installation path are identical, skipping"
+            ));
+        }
+
+        for entry in walkdir::WalkDir::new(&self.path) {
```
cd-work, line 41:
```suggestion
        for entry in WalkDir::new(&self.path) {
```

Should import WalkDir here.

cli/src/commands/extensions/extension.rs
```diff
@@ -0,0 +1,143 @@
...
+            let source_path = entry?.into_path();
+            let dest_path = target_prefix.join(source_path.strip_prefix(&self.path)?);
+
+            if source_path.is_dir() {
+                std::fs::create_dir_all(dest_path)?;
+            } else if source_path.is_file() {
+                if dest_path.exists() {
+                    return Err(anyhow!("{}: already exists", dest_path.to_string_lossy()));
+                } else {
+                    std::fs::copy(source_path, dest_path)?;
```
cd-work, line 51:
```suggestion
                    fs::copy(source_path, dest_path)?;
```

cli/src/commands/extensions/extension.rs
```diff
@@ -0,0 +1,143 @@
...
+                }
+            }
+        }
+
+        println!("Extension {} installed correctly", self.name());
```
cd-work, line 56: I think the wording "correctly" is a bit strange? As a user I'd usually expect "successfully", since the command should never really *incorrectly* install an extension.

cli/src/commands/extensions/extension.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        for entry in walkdir::WalkDir::new(&self.path).contents_first(true) {
+            let entry = entry?.into_path();
+            println!("removing {}...", entry.to_string_lossy());
+            if entry.is_dir() {
+                std::fs::remove_dir(entry)?;
+            } else if entry.is_file() {
+                std::fs::remove_file(entry)?;
+            }
+        }
```
cd-work, line 81: Should probably just remove the directories including all contents, rather than walking the directory and deleting everything individually?

cd-work (changes requested):

cli/src/commands/extensions/extension.rs
andreaphylum: There is `std::fs::remove_dir_all` that would serve this purpose, but which would fail were we to switch to the single-file format (granted, it would be a very minor change and easy to catch with tests). I chose this approach in the interest of being more explicit and intentional, and also having the chance of listing the removed files to the user. Furthermore, this explicitly breaks support for symlinks, which we may not want to have in the first place and whose side effects I'm not confident about. I think switching to `remove_dir_all` would be fine still if we feel like these points aren't that important (or are counterproductive).

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+        extensions.into_iter().for_each(|ext| {
+            println!("{:20}   {}", ext.name(), ext.description().unwrap_or(""));
+        });
+    }
+
+    Ok(CommandValue::Code(ExitCode::Ok))
+}
+
+// Return a list of installed extensions. Filter out invalid extensions instead of exiting early.
+fn list_extensions() -> Result<Vec<Extension>> {
```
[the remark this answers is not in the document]
andreaphylum: `list` here was intended as a verb as opposed to a noun, which would be consistent with other function names (`handle_<...>`). maybe we can use `get` or `enumerate`, or a clearer verb instead?

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+            command.subcommand(Command::new(ext.name()))
+        })
+}
+
+/// Entry point for the `extensions` subcommand.
+pub async fn handle_extensions(matches: &ArgMatches) -> CommandResult {
+    if let Some(matches) = matches.subcommand_matches("add") {
+        subcmd_add_extension(matches.value_of_t("PATH")?).await
+    } else if let Some(matches) = matches.subcommand_matches("remove") {
+        subcmd_remove_extension(matches.value_of("NAME").unwrap()).await
```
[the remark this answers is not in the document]
andreaphylum: Probably an autocomplete mistake, will fix.

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+                .about("Install extension")
+                .arg(arg!([PATH]).required(true).value_hint(ValueHint::FilePath)),
+        )
+        .subcommand(
+            Command::new("remove")
+                .about("Uninstall extension")
+                .arg(arg!([NAME]).required(true)),
+        )
+        .subcommand(Command::new("list").about("List installed extensions"))
+}
```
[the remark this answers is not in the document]
andreaphylum: I wanted to suggest the opposite -- I feel like separating the command definitions from their modules makes it harder to maintain (and test -- we may want  to add smaller unit tests that only go through `clap` at some point and I feel like their own modules would be a better place for that) and think it would be better to have a leaner `app.rs`.

cli/src/commands/extensions/extension.rs
cd-work: Those are all solid arguments, but they seem mostly focused on what *might* be needed rather than what *is* needed. I'd recommend just going for the best solution now, rather than the best solution in the future and switch to `remove_dir_all`.

andreaphylum: That makes a lot of sense; I do have a tendency to cover as much unknowns as I can, but on the other hand it's good to keep the scope tight whenever possible.

docs/extensions.md
andreaphylum: Great point. At this stage, though, I feel like we may want to create a separate issue for that and add it in only once we have a clearer direction on how to handle the marketplace.

cli/src/commands/extensions/extension.rs
andreaphylum: Added a regex filter to `Extension::try_from` for the extension name. We are now upholding the invariant that, for any valid `Extension`, its name can't result in directory traversals.

cli/src/config.rs
```diff
@@ -195,6 +194,18 @@ pub fn get_home_settings_path() -> Result<PathBuf> {
     Ok(config_path)
 }
 
+/// Resolve XDG data directory.
+pub fn data_dir() -> Result<PathBuf> {
+    let home_path =
```
matt-phylum, line 199: If for some reason we can't find the user's home directory, this code bails out even if it isn't going to use the home directory.

cli/tests/extensions.rs
matt-phylum: If you're going to put it in a system directory you might want to look at the tempfile crate. It has support for creating directories with prefixes, automatically cleaning up, and it also handles some permissions problems you can run into with temp directories, although I don't know if that's something we need to worry about when running these tests.

matt-phylum (changes requested):

cli/src/commands/extensions/extension.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        for entry in walkdir::WalkDir::new(&self.path) {
+            let source_path = entry?.into_path();
+            let dest_path = target_prefix.join(source_path.strip_prefix(&self.path)?);
+
+            if source_path.is_dir() {
+                std::fs::create_dir_all(dest_path)?;
+            } else if source_path.is_file() {
+                if dest_path.exists() {
+                    return Err(anyhow!("{}: already exists", dest_path.to_string_lossy()));
```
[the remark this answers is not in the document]
andreaphylum: I think that allowing users to install extensions is a risky and not-too-useful endeavor which would be better left to an internal package manager with a limited API exposed to the user (e.g. specify a dependency which will be solved by the CLI via a different code path, not necessarily ending up in running an installation that must then be error checked by the user).

cli/src/commands/extensions/mod.rs
```diff
@@ -100,14 +108,25 @@ pub async fn subcmd_list_extensions() -> CommandResult {
...
+    if !extensions_path.exists() {
         return Ok(Vec::new());
     }
 
-    Ok(std::fs::read_dir(extension_path)?
-        .filter_map(|d| Extension::try_from(d.map(|d| d.path()).ok()?).ok())
+    Ok(std::fs::read_dir(extensions_path)?
+        .filter_map(|dir_entry| {
+            match dir_entry
+                .map_err(|e| anyhow!("{}", e))
```
kylewillmon, line 121: There's no need to make an ad-hoc error here. `std::io::Error` will cleanly convert to `anyhow::Error`

```suggestion
                .map_err(|e| e.into())
```

cli/tests/extensions.rs
kylewillmon: Can we use an environment variable to disable this cleanup?

Something like:

```
if env::var_os("PHYLUM_TEST_SKIP_CLEANUP").is_none() {
    fs::remove_dir_all(&self.0).unwrap();
}
```

kylewillmon (dismissed):

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,132 @@
+pub mod extension;
```
kylewillmon, line 1: The combination of `pub mod extension;` here and `pub use extension::*;` below means that there are 2 valid public paths for everything in that module (i.e., `use crate::commands::extensions::extension::Extension;` or `use crate::commands::extensions::Extension;`). It's cleaner if at most one of these statements is `pub`.

```suggestion
mod extension;
```

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+        Err(e) => {
+            error!("Couldn't list extensions: {}", e);
+            return command;
+        }
+    };
+
+    let names = command
+        .get_subcommands()
+        .map(|n| n.get_name().to_string())
+        .collect::<HashSet<_>>();
```
kylewillmon, line 42: Given the size of this set, I'd be willing to bet `Vec<String>` would run faster than `HashSet<String>`. But this isn't worth performance testing, so ¯\\_(ツ)_/¯

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+/// Entry point for the `extensions` subcommand.
+pub async fn handle_extensions(matches: &ArgMatches) -> CommandResult {
+    if let Some(matches) = matches.subcommand_matches("add") {
+        handle_add_extension(matches.value_of_t("PATH")?).await
+    } else if let Some(matches) = matches.subcommand_matches("remove") {
+        handle_remove_extension(matches.value_of("NAME").unwrap()).await
+    } else {
+        // also covers the `list` case
+        handle_list_extensions().await
+    }
```
kylewillmon, line 71: Other parts of our code mention [switching to match expressions](https://github.com/phylum-dev/cli/blob/8f112df9adcfb8c06bd6465be84c190eb8fe33db/cli/src/bin/phylum.rs#L196) for subcommands. You could be the first to actually do it.

```suggestion
    match matches.subcommand() {
        Some(("add", matches)) => handle_add_extension(matches.value_of_t("PATH")?).await,
        Some(("remove", matches)) => handle_remove_extension(matches.value_of("NAME").unwrap()).await,
        Some(("list", _)) | None => handle_list_extensions().await,
        _ => unreachable!(),
    }
```

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+// Return a list of installed extensions. Filter out invalid extensions instead of exiting early.
+fn installed_extensions() -> Result<Vec<Extension>> {
+    let extensions_path = extensions_path()?;
+
+    if !extensions_path.exists() {
+        return Ok(Vec::new());
+    }
+
+    Ok(std::fs::read_dir(extensions_path)?
```
kylewillmon, line 118: Checking the error returned by `fs::read_dir` would be better than first checking if the directory exists.

```suggestion
    let dir_entry = match fs::read_dir(extensions_path) {
        Ok(d) => d,
        Err(e) => if e.kind() == ErrorKind::NotFound {
            return Ok(Vec::new());
        } else {
            return Err(e.into());
        }
    }

    Ok(dir_entry
```

cli/src/config.rs
```diff
@@ -196,6 +195,21 @@ pub fn get_home_settings_path() -> Result<PathBuf> {
     Ok(config_path)
 }
 
+/// Resolve XDG data directory.
+pub fn data_dir() -> Result<PathBuf> {
+    if let Some(data_dir) = env::var("XDG_DATA_HOME")
+        .ok()
```
kylewillmon, line 201: This is an unnecessary UTF-8 check that can be skipped with `env::var_os`
```suggestion
    if let Some(data_dir) = env::var_os("XDG_DATA_HOME")
```

docs/phylum_extension_add.md
```diff
@@ -0,0 +1,17 @@
...
+category: 6255e67693d5200013b1fa3e
+hidden: true
+---
+To install an extension, run the following command:
+
+```sh
+phylum extension add path/to/extension
+```
+
+The extension will be installed under `$XDG_DATA_HOME/phylum/extensions/<ext_name>`.
```
kylewillmon, line 12: This should also mention the default location of `$HOME/.local/share/phylum/extensions/<ext_name>` if `$XDG_DATA_HOME` isn't set.

kylewillmon (dismissed):

cli/src/commands/extensions/mod.rs
andreaphylum: Yeah, I don't think the difference may be appreciable outside of microbenchmarks, as this function is run just a handful of times anyway. I went for `HashSet` as it felt more appropriate in terms of use case understandability (i.e. I see `HashSet`, I think "check for existence").

cli/src/commands/extensions/mod.rs
andreaphylum: I think the catch-all branch may actually be reachable in the `None` form, as we want to handle the no-subcommand case. I will commit this suggestion and then check.

kylewillmon: Notice that the last pattern is `Some(("list", _)) | None`

andreaphylum: Sorry, missed that!

kylewillmon (approved):

cd-work (dismissed):
Approving just so I don't block merging. If you want me to go over the patch in detail again, please let me know.

cli/tests/extensions.rs
```diff
@@ -0,0 +1,317 @@
...
+    Command::cargo_bin("phylum")
+        .unwrap()
+        .env("XDG_DATA_HOME", &tmp_dir)
+        .arg("extension")
+        .arg("add")
+        .arg(fixtures_path().join("sample-extension"))
+        .assert()
+        .success();
+
+    env::set_var("XDG_DATA_HOME", &tmp_dir);
```
matt-phylum, line 32: Does setting this variable mean we can't run the tests in parallel? Can we accomplish what we need by running external commands?

matt-phylum (approved):

cli/src/commands/extensions/extension.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        for entry in WalkDir::new(&self.path) {
+            let source_path = entry?.into_path();
+            let dest_path = target_prefix.join(source_path.strip_prefix(&self.path)?);
+
+            if source_path.is_dir() {
+                fs::create_dir_all(&dest_path)?;
+                #[cfg(unix)]
+                fs::set_permissions(&dest_path, Permissions::from_mode(0o700))?;
```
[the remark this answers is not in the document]
andreaphylum: Good point -- I _think_ we just need the permissions set on the innermost directory, but I'm not 100% sure what the correct course of action is according to XDG spec. I don't think having `0o700` on the entire directory tree is bad, so we might as well try to apply it recursively. We do need to make sure we are not bubbling up the permission settings incorrectly, as `DirBuilder` applies the settings for all [non-existing parents](https://doc.rust-lang.org/std/fs/struct.DirBuilder.html#method.recursive).

Don't think we need to worry about Windows at all, as the XDG standard does not apply there.

cli/src/commands/extensions/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+        .subcommand(
+            Command::new("remove")
+                .about("Uninstall extension")
+                .arg(arg!([NAME]).required(true)),
+        )
+        .subcommand(Command::new("list").about("List installed extensions"))
+}
+
+/// Generate the subcommands for each extension.
+/// TODO add tests.
```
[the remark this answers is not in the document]
andreaphylum: I think so: at the moment we aren't testing cases that might arise from various configurations of the extensions directory.

cli/tests/extensions.rs
```diff
@@ -0,0 +1,317 @@
...
+////////////////////////////////////////////////////////////////////////////////
+// Acceptance criteria tests
+////////////////////////////////////////////////////////////////////////////////
+
+// When a user runs `phylum extension add .`, the extension in the current
+// working directory should be installed.
+#[test]
+fn extension_is_installed_correctly() {
+    let tmp_dir = TmpDir::new();
+    Command::cargo_bin("phylum")
```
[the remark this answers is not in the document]
andreaphylum: We may want to call different commands with different setups (see [this](https://github.com/phylum-dev/cli/pull/353/files/6dc926b83a1cdad21ac43bd24874e238badad006#diff-54f139487fae76c6099d07eec1e677e9406089191e28e14d92fee840db585275R147)); it does look like there is some duplication indeed, but I believe we would lose some flexibility by factoring that out and the current approach is a good compromise.

cli/src/commands/extensions/extension.rs
```diff
@@ -61,9 +61,14 @@ impl Extension {
             let dest_path = target_prefix.join(source_path.strip_prefix(&self.path)?);
 
             if source_path.is_dir() {
-                fs::create_dir_all(&dest_path)?;
-                #[cfg(unix)]
-                fs::set_permissions(&dest_path, Permissions::from_mode(0o700))?;
+                if cfg!(unix) {
```
matt-phylum, line 64: Is it better to use DirBuilder in both cases? I guess it has the same effect.

```rust
let mut builder = DirBuilder::new()
    .recursive(true);

#[cfg(unix)]
builder.mode(0o700);

builder.create(&dest_path)?;
```

matt-phylum (approved):

cli/tests/extensions.rs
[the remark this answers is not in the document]
matt-phylum: I was thinking we could do something like this to start with

```rust
struct TestEnvironment {
    data_home: TmpDir,
}

impl TestEnvironment {
    fn new() -> Self {
        Self {
            data_home: TmpDir::new()
        }
    }

    fn phylum_command(&self) -> Command {
        Command::cargo_bin("phylum")
            .unwrap()
            .env("XDG_DATA_HOME", &self.data_home)
    }

    fn install_test_extension(&self, name: &str) -> Assert {
        self.phylum_command()
            .args(&["extension", "add"])
            .arg(fixtures_path().join(name))
            .assert()
    }

    fn extension_home(&self) -> PathBuf {
        PathBuf::from(&self.tmp_dir)
            .join("phylum")
            .join("extensions")
    }
}
```

And then that uninstall test could like like this

```rust
let test_env = TestEnvironment::new();
test_env.install_test_extension("sample-extension").success();

let extension_path = test_env.extension_home().join("sample-extension");
assert!(WalkDir::new(&extension_path).into_iter().count() > 1);

test_env.phylum_command()
    .args(&["extensions", "remove", "sample-extension"])
    .assert()
    .success();

assert!(!extension_path.exists());
```

The fixture doesn't need to do everything or hide anything from the test.

cli/src/commands/extensions/extension.rs
andreaphylum: I thought about that too -- I didn't particularly like stylistically having to split code this way nor replicating `DirBuilder` without mode on one branch, so I went for the current solution, but I do think yours has a more consistent semantics at least.

cli/tests/extensions.rs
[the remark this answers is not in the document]
andreaphylum: I like this idea. I think it's better to capture it in an issue and get to it on a later refactoring cycle, though, as it'll require changing a decent bit of testing code.

kylewillmon:
Why is this not going to `main`? I'd really like to avoid maintaining separate branches if at all possible. And we've already got this behind a feature flag.

kylewillmon (approved):

PR #354, opened 2022-05-06
Remove redundancy from lock_files.rs

cd-work:
This patch attempts to reduce the maintenance overhead of changing the
`lock_files.rs` by providing macros and helper functions to generate the
necessary output rather than relying on repeating the same code over and
over.

### DIFF
--- a/cli/src/commands/lock_files.rs
+++ b/cli/src/commands/lock_files.rs
@@ -12,71 +12,26 @@ pub fn try_get_packages(path: &Path) -> Result<(Vec<PackageDescriptor>, PackageT
         path.to_string_lossy()
     );
 
-    fn xxx<T: Parseable>(path: &Path) -> Option<(Vec<PackageDescriptor>, PackageType)> {
-        let packages = T::new(path).ok()?.parse();
-        packages
-            .ok()
-            .filter(|pkgs| !pkgs.is_empty())
-            .map(|pkgs| (pkgs, PackageType::Npm))
+    // Try a package lock format and return the packages if there are some.
+    macro_rules! try_format {
+        ($lock:ident, $ty:literal) => {{
+            let packages = parse::<$lock>(path).ok();
+            if let Some(packages) = packages.filter(|(packages, _)| !packages.is_empty()) {
+                log::debug!("Submitting file as type {}", $ty);
+                return Ok(packages);
+            }
+        }};
     }
 
-    if let Some(packages) = xxx::<YarnLock>(path) {
-        log::debug!("Submitting file as type yarn lock");
-        return Ok(packages);
-    }
-    // let packages = YarnLock::new(path)?.parse();
-    // if let Some(packages) = packages.ok().filter(|pkgs| !pkgs.is_empty()) {
-    //     log::debug!("Submitting file as type yarn lock");
-    //     return Ok((packages, PackageType::Npm));
-    // }
-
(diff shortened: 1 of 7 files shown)


cd-work:
I'm not sure all the changes in this patch are worth it. It's honestly not as ideal as I had hoped for but it does successfully cut down on redundancy, especially for the `try_get_packages` function.

The `parse` helper function probably doesn't provide as much value as the macro, though it does still help a bit and would hopefully prevent typos or copy/paste errors especially when it comes to the package type.

cli/src/commands/lock_files.rs
```diff
@@ -12,59 +12,26 @@ pub fn try_get_packages(path: &Path) -> Result<(Vec<PackageDescriptor>, PackageT
         path.to_string_lossy()
     );
 
-    let packages = YarnLock::new(path)?.parse();
-    if let Some(packages) = packages.ok().filter(|pkgs| !pkgs.is_empty()) {
-        log::debug!("Submitting file as type yarn lock");
-        return Ok((packages, PackageType::Npm));
+    // Try a package lock format and return the packages if there are some.
+    macro_rules! try_format {
```
matt-phylum, line 16: I think this is an improvement over what we had, but I'm not sure about using a macro here. This is the best I could do without macros.

```rust
fn try_format<Format: Parseable>(path: &Path, display: &str) -> Option<(Vec<PackageDescriptor>, PackageType)> {
    let (packages, format) = parse::<Format>(path).ok()?;
    if packages.is_empty() {
        None
    } else {
        log::debug!("Submitting file as type {}", display);
        Some((packages, format))
    }
}

if let Some((packages, format)) = try_format::<YarnLock>(path, "yarn lock")
    .or_else(|| try_format::<PackageLock>(path, "package lock"))
    .or_else(|| try_format::<GemLock>(path, "gem lock"))
    .or_else(|| try_format::<PyRequirements>(path, "pip requirements.txt"))
    .or_else(|| try_format::<PipFile>(path, "pip Pipfile or Pipfile.lock"))
    .or_else(|| try_format::<Poetry>(path, "poetry lock"))
    .or_else(|| try_format::<Pom>(path, "pom xml"))
    .or_else(|| try_format::<GradleDeps>(path, "gradle dependencies"))
    .or_else(|| try_format::<CSProj>(path, "csproj"))
{
    Ok((packages, format))
} else {
    Err(anyhow!("Failed to identify lockfile type"))
}
```

matt-phylum (approved):

cli/src/commands/lock_files.rs
cd-work: I feel like the macro is significantly more readable than this. though curious if @kylewillmon has an opinion on the matter.

matt-phylum: Yeah. It'd be nice if we could do a reverse `?` to return the first one which is `Some` without all the `or_else` chaining.

cd-work: Oh yeah, that would make things significantly easier. But that would require a macro at which point we have to use one macro anyway.

kylewillmon: I also had the thought about `or_else()`, but looking at it, I agree with @cd-work that the macro is more readable.

kylewillmon (dismissed):

kylewillmon (approved):

PR #358, opened 2022-05-06
Improve error messages with HTTP failures

### LINKED ISSUE #155: Add better logging around failed response
When the cli gets back a non-200 result, it should log the issue and print out a helpful result

cd-work:
This enhances the verbosity of errors when an HTTP request failed, by
providing the full response with a response code other than 200, rather
than trying to parse the response as valid JSON.

This should shed more insights on why the request failed, especially
when the user has connectivity issues.

Closes #155.

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -1,5 +1,6 @@
 use std::time::Duration;
 
+use anyhow::anyhow;
 use phylum_types::types::auth::*;
 use phylum_types::types::common::*;
 use phylum_types::types::job::*;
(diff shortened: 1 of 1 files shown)


cli/src/api/mod.rs
```diff
@@ -63,34 +65,36 @@ enum APIResult<T> {
...
-    }
 
-    async fn delete<T: serde::de::DeserializeOwned>(&self, path: String) -> Result<T> {
-        let body = self.client.delete(path).send().await?.text().await?;
+        let response = request.send().await?;
+        let success = response.status().is_success();
+        let body = response.text().await?;
+
+        if !success {
+            return Err(anyhow!(body).into());
```
kylewillmon, line 96: It would be nice if this error also included the HTTP status code. Maybe as context?

```
    .with_context(|| format!("HTTP status error ({})", status))
```

kylewillmon (approved):

cli/src/api/mod.rs
cd-work: You can't provide a context since this is not an anyhow result. That's why I brought the question of returning anyhow results up to you in Slack.

kylewillmon (approved):

PR #362, opened 2022-05-09
Create `phylum parse` command

### LINKED ISSUE #346: `phylum parse` subcommand
# Overview
Add a subcommand that allows parsing a lock file and returning the parsed dependencies in a structured format

# Acceptance Criteria
- [ ] `phylum parse -t mvn pom.xml` would return JSON of at least the dependency name/version pairs parsed out of the pom.xml file. (this is just an example)

kylewillmon:
This patch adds a command for parsing lock files and outputting the
packages as JSON. This closes #346


cli/src/commands/parse.rs
```diff
@@ -0,0 +1,53 @@
...
+use std::io;
+use std::path::Path;
+
+use anyhow::Result;
+use phylum_types::types::package::{PackageDescriptor, PackageType};
+
+use super::lock_files;
+use crate::lockfiles::*;
+
+type ParserResult = Result<(Vec<PackageDescriptor>, PackageType)>;
```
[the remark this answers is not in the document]
kylewillmon: I've been strongly considering just moving all the code from `lock_files.rs` into this file. They're both short files and the uses strongly overlap.

furi0us333 (commented):
You will also want to add `phylum parse` under the Additional Commands section in `phylum.md`

maxrake (commented):
I confirmed the expected behavior by running local tests. Adding tests to confirm this command with the existing test fixtures would be nice. An bug unrelated to this PR was found and documented in #364 

Leaving the approval for Rust SMEs.

cli/src/commands/parse.rs
```diff
@@ -0,0 +1,53 @@
+//! `phylum parse` command for lockfile parsing
+
+use std::io;
+use std::path::Path;
+
+use anyhow::Result;
+use phylum_types::types::package::{PackageDescriptor, PackageType};
+
+use super::lock_files;
+use crate::lockfiles::*;
```
cd-work, line 10: This certainly is super confusing. You've already suggested yourself to just move everything from `lock_files.rs` into this file. I think getting rid of these "duplicate" lock file modules with slightly different names would probably avoid some confusion in the future.

cli/src/commands/parse.rs
```diff
@@ -0,0 +1,53 @@
...
+    ("npm", &lock_files::parse::<PackageLock>),
+    ("gem", &lock_files::parse::<GemLock>),
+    ("pip", &lock_files::parse::<PyRequirements>),
+    ("pipenv", &lock_files::parse::<PipFile>),
+    ("poetry", &lock_files::parse::<Poetry>),
+    ("mvn", &lock_files::parse::<Pom>),
+    ("gradle", &lock_files::parse::<GradleDeps>),
+    ("nuget", &lock_files::parse::<CSProj>),
+    ("auto", &lock_files::get_packages_from_lockfile),
+];
```
cd-work, line 25: Big fan of the dynamic dispatch of the referenced functions here. That does make things pretty clean and I tried doing something similar in the `lock_files.rs` cleanup but failed since the `Parser::new` function cannot be used easily due to different return types.

But now that the `parse` function exists already anyway, I wonder if this same method could be used to get rid of the macro in `try_get_packages`? It seems like it should be possible.

cli/src/commands/parse.rs
```diff
@@ -0,0 +1,53 @@
...
+
+pub fn handle_parse(matches: &clap::ArgMatches) -> Result<()> {
+    let lockfile_type = matches.value_of("lockfile-type").unwrap_or("auto");
+    // LOCKFILE is a required parameter, so .unwrap() should be safe.
+    let lockfile = matches.value_of("LOCKFILE").unwrap();
+
+    let parser = LOCKFILE_PARSERS
+        .iter()
+        .filter_map(|(name, parser)| {
+            if *name == lockfile_type {
```
cd-work, line 39: Could use `(*name == lockfile_type).then(|| parser)` here.

cli/src/commands/parse.rs
```diff
@@ -0,0 +1,53 @@
...
+                Some(parser)
+            } else {
+                None
+            }
+        })
+        .next()
+        .unwrap();
```
cd-work, line 46: Is this guaranteed to succeed due to clap's `lockfile_types` check?

docs/phylum_parse.md
```diff
@@ -0,0 +1,19 @@
...
+hidden: false
+---
+Parse a lockfile and output the packages as JSON
+```sh
+phylum parse [OPTIONS] <LOCKFILE>
+```
+
+### Options
+`-t`, `--lockfile-type`
+&emsp; The type of the lockfile: `yarn`, `npm`, `gem`, `pip`, `pipenv`, `poetry`, `mvn`, `gradle`, `nuget`, `auto`
```
cd-work, line 13: I think indicating default here would be nice.

cd-work (changes requested):

cli/src/commands/lock_files.rs
```diff
@@ -38,7 +38,7 @@ pub fn try_get_packages(path: &Path) -> Result<(Vec<PackageDescriptor>, PackageT
 
 /// Determine the lockfile type based on its name and parse
 /// accordingly to obtain the packages from it
-pub fn get_packages_from_lockfile(path: &str) -> Result<(Vec<PackageDescriptor>, PackageType)> {
+pub fn get_packages_from_lockfile(path: &Path) -> Result<(Vec<PackageDescriptor>, PackageType)> {
     let path = Path::new(path);
```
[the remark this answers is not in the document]
kylewillmon: Sorry. This line is a no-op now. `Path::new(...)` is returning the same `&Path` that is being passed in. I will delete this line.

The more appropriate signature here would be `<P: AsRef<Path>>(path: P)` just like [`File::open()`](https://doc.rust-lang.org/std/fs/struct.File.html#method.open) and friends. But I was trying to keep this patch short.

cli/src/commands/lock_files.rs
```diff
@@ -70,7 +70,7 @@ pub fn get_packages_from_lockfile(path: &str) -> Result<(Vec<PackageDescriptor>,
 }
 
 /// Get all packages for a specific lockfile type.
-fn parse<P: Parseable>(path: &Path) -> Result<(Vec<PackageDescriptor>, PackageType)> {
+pub(super) fn parse<P: Parseable>(path: &Path) -> Result<(Vec<PackageDescriptor>, PackageType)> {
```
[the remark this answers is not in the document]
kylewillmon: I'll change this to `pub`

I was hesitant to use `pub` here mostly because this `phylum-cli` crate is a lib crate and (in theory) may someday be published to `crates.io`. So a piece of me worries every time we add a new `pub` function that we're committing to a public API. But it's premature to worry about that. And we can switch these functions to `pub(crate)` if necessary in the future.

cli/src/commands/parse.rs
kylewillmon: I completely agree. But I may want to do that refactor into a separate PR to keep conflicts to a minimum.

cli/src/commands/parse.rs
kylewillmon: Hmmm.... So we would avoid the need for a macro (which you [previously mentioned](https://github.com/phylum-dev/cli/pull/354/files#r867137035)) by using a loop instead?

That would make sense.

kylewillmon: The other option that I have considered for cleaning all of this up is to use [the strategy pattern](https://rust-unofficial.github.io/patterns/patterns/behavioural/strategy.html) and use unit types to implement the `Parseable` trait.

cli/src/commands/parse.rs
kylewillmon: TIL [bool](https://doc.rust-lang.org/std/primitive.bool.html#method.then) has methods.... I'll make this change.

cli/src/commands/parse.rs
kylewillmon: That is correct.

```
❯ ./target/debug/phylum parse -t nil foo.txt
error: "nil" isn't a valid value for '--lockfile-type <type>'
	[possible values: yarn, npm, gem, pip, pipenv, poetry, mvn, gradle, nuget, auto]

USAGE:
    phylum parse --lockfile-type <type> <LOCKFILE>

For more information try --help
```

maxrake (commented):
This still builds and still functions as advertised. I'm ready for this feature now...in [phylum-ci #30](https://github.com/phylum-dev/phylum-ci/issues/30). A CLI release candidate or full release with this feature is needed for that work.

kylewillmon:
> This still builds and still functions as advertised. I'm ready for this feature now...in [phylum-ci #30](https://github.com/phylum-dev/phylum-ci/issues/30). A CLI release candidate or full release with this feature is needed for that work.

This is still waiting on a review. But you can feel free to tag an `-rc` on this branch if you want to go ahead with testing.

maxrake (approved):
LGTM

PR #366, opened 2022-05-10
Fix name-based lockfile detection

### LINKED ISSUE #364: Some lockfile patterns are incorrect
# Overview
Several lockfile patterns in the `get_packages_from_lockfile` function from `cli/src/commands/lock_files.rs` are incorrect. That is, they use values that are not representative of the more common/default name. Specifically:

* `Pipfile.txt` should be `Pipfile` (w/o the extension)
  * Reference: [Pipenv documentation](https://pipenv.pypa.io/en/latest/basics/#example-pipfile-pipfile-lock)
* `effective-pom.xml` should be `pom.xml`
  * A good reference could not be found
* `gradle-dependencies.txt` should be `gradle.lockfile`
  * Reference: [Gradle documentation](https://docs.gradle.org/current/userguide/dependency_locking.html#lock_state_location_and_format)

# How To Reproduce
1. Ensure the new `phylum parse` subcommand introduced in #362 is compiled/used
2. Issue a `phylum parse <LOCKFILE>` command *without* the `-t` option and with a "standard" lockfile
3. See warning message

```sh
❯ ./phylum parse ../../cli/tests/fixtures/Pipfile 
[2022-05-09T19:53:47Z WARN  phylum_cli::commands::lock_files] Attempting to obtain packages from unrecognized lockfile type: ../../cli/tests/fixtures/Pipfile
[2022-05-09T20:03:11Z WARN  phylum_cli::lockfiles::parsers::pypi] Could not determine version for package: name
[2022-05-09T20:03:11Z WARN  phylum_cli::lockfiles::parsers::pypi] Could not determine version for package: verify_ssl
...
[2022-05-09T20:03:11Z WARN  phylum_cli::lockfiles::python] Could not determine version for package: price-parser
[2022-05-09T20:03:11Z WARN  phylum_cli::lockfiles::python] Could not determine version for package: twilio
[
  {
    "name": "pypresence",
    "version": "4.0.0",
    "type": "pypi"
  },
  {
    "name": "chromedriver-py",
    "version": "91.0.4472.19",
    "type": "pypi"
  },
  {
    "name": "pycryptodome",
    "version": "3.9.8",
    "type": "pypi"
  },
  {
    "name": "requests",
    "version": "2.24.0",
    "type": "pypi"
  }
]%                                 
```

# Expected Behavior
The standard lockfile names should be recognized and passed off to the correct parser directly instead of going through the `try_get_packages` function first.

# Additional Context
For the example provided, the parsing still worked, it just passed through the `try_get_packages` function first...which _may_ not always work as expected.

cd-work:
The detection for some lockfiles was based on incorrect file names,
causing those to fall back to the generic trial-and-error-based parser.

Changed lockfiles:
  `Pipfile.txt` -> `Pipfile` (Python unlocked dependency file)
  `gradle-dependencies.txt` -> `gradle.lockfile` (Gradle lockfile)

Closes #364.


cli/src/commands/lock_files.rs
```diff
@@ -56,10 +56,10 @@ pub fn get_packages_from_lockfile(path: &str) -> Result<(Vec<PackageDescriptor>,
         "package-lock.json" => parse::<PackageLock>(path)?,
         "yarn.lock" => parse::<YarnLock>(path)?,
         "requirements.txt" => parse::<PyRequirements>(path)?,
-        "Pipfile.txt" | "Pipfile.lock" => parse::<PipFile>(path)?,
+        "Pipfile" | "Pipfile.lock" => parse::<PipFile>(path)?,
         "poetry.lock" => parse::<Poetry>(path)?,
         "effective-pom.xml" => parse::<Pom>(path)?,
-        "gradle-dependencies.txt" => parse::<GradleDeps>(path)?,
+        "gradle.lockfile" => parse::<GradleDeps>(path)?,
```
cd-work, line 62: We could in theory keep support for our old detection mechanism by just adding the new formats. However I feel like that would be difficult for people to follow when reading this in the future.

So unless there's any evidence that these formats actually exist in practice, I'd rather remove support for the old naming scheme entirely.

kylewillmon (dismissed):

cli/src/commands/lock_files.rs
```diff
@@ -86,7 +86,7 @@ mod tests {
             ("tests/fixtures/yarn.lock", PackageType::Npm),
             ("tests/fixtures/package-lock.json", PackageType::Npm),
             ("tests/fixtures/sample.csproj", PackageType::Nuget),
-            ("tests/fixtures/gradle-dependencies.txt", PackageType::Maven),
+            ("tests/fixtures/gradle.lockfile", PackageType::Maven),
```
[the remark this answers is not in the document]
cd-work: The tests did not fail, unless you mean when moving the file without changing it in the tests in which case they obviously do.

But initially I didn't move the test files over and all test still executed just fine, I'd assume due to the generic lockfile resolution.

cli/src/commands/lock_files.rs
cd-work: Thanks a lot for pointing that out. The existing file indeed seems to be somewhat odd for usage as a lock file.

Are you familiar with the reasoning behind the current Java Gradle support? If so, what would you suggest moving forward?

Seems like our best choice might be to keep it as `gradle-dependencies.txt` and add an issue about supporting the `gradle.lockfile`? It might also be worth it to expand our existing documentation about supported lockfiles with what all these are specifically used for, since that would likely also help us understand what each of these is and which one might be incorrectly named.

spellman:
@cd-work @ejortega Can y'all clarify what we want to accomplish here?
The linked issue ends with https://github.com/phylum-dev/cli/issues/364#issuecomment-1122514511, which sounds like we don't want to support `Pipfile` and which got a thumbs up. I'm reading the conversations in that issue and this PR to say
* We need to fix two incorrect file extensions -- that's this PR.
* We need to remove support for non-lock dependency files -- that's future work.

Is that correct?
If so, do we need to write an issue for that future work?

Also, what if a project has, say, a `requirements.txt` file only -- do we want to provide feedback at the CLI to encourage the developer to use a `Pipfile` and `Pipfile.lock` (or use poetry)?

ejortega:
@spellman @cd-work  This is my understanding.

1. I think we agree to remove `Pipfile.txt/Pipfile` as it's not a lockfile. The name change in this PR is fine.
2. We should properly support parsing a `gradle.lockfile`. The name change in this PR is correct. However, the current parser only works for the output of `gradle dependencies > gradle-dependencies.txt` and not an actual `gradle.lockfile`.
3. As for python projects without a lockfile (pip/poetry). I'm not positive what's being done at the moment. @furi0us333 might have more information on this. But early discussions, the process was to using a virtualenv, pip install the packages, and pip freeze.

cd-work:
> I think we agree to remove Pipfile.txt/Pipfile as it's not a lockfile. The name change in this PR is fine.

I can't say how much that is used at the moment. But it works and is already supported. As such I don't see any rush to remove it. We could certainly consider it and open an issue, but realistically that should just be deleting unused code, not removing functionality.

> We should properly support parsing a gradle.lockfile. The name change in this PR is correct. However, the current parser only works for the output of gradle dependencies > gradle-dependencies.txt and not an actual gradle.lockfile.

While this is largely true, I wouldn't say the name change is correct and I will revert that until proper gradle.lockfile support is added. There is no standard name for the current format so keeping our own (gradle-dependencies.txt) should be okay.

> As for python projects without a lockfile (pip/poetry). I'm not positive what's being done at the moment. @furi0us333 might have more information on this. But early discussions, the process was to using a virtualenv, pip install the packages, and pip freeze.

Poetry and the pip lockfile should just work. The unfrozen pipfile is just something we support in addition to that.

cd-work:
I've created https://github.com/phylum-dev/cli/issues/372 for tracking the `gradle.lockfile` issue. I think that should be the only actionable thing besides this PR that would bring some value to us at the moment.

kylewillmon (approved):

ejortega (approved):

kylewillmon (approved):

spellman:
@ejortega said:
> 1. I think we agree to remove Pipfile.txt/Pipfile as it's not a lockfile. The name change in this PR is fine.

@cd-work said:
> Poetry and the pip lockfile should just work. The unfrozen pipfile is just something we support in addition to that.

I might be being obtuse 😅 but I'm trying to understand how these statements go together. Are we saying that we _currently_ support the unfrozen pipfile but we agree that we _will_ remove support for it?

ejortega:
We should encourage the use of a lockfile when possible so we can give an accurate package score. No need to remove functionality at this time.

PR #368, opened 2022-05-11
Use the action from completed jobs only

kylewillmon:
Failing to check the completion status of the job leads to error messages like this even when the job is not yet completed.

```
❗ Error: Project failed threshold requirements, failing the build!
```

With this change, the code uses `Action::None` if the job is incomplete.

### DIFF
--- a/cli/src/commands/jobs.rs
+++ b/cli/src/commands/jobs.rs
@@ -4,6 +4,7 @@ use std::str::FromStr;
 
 use ansi_term::Color::Blue;
 use anyhow::{anyhow, Context, Result};
+use phylum_types::types::common::Status;
 use reqwest::StatusCode;
 use serde::Serialize;
 use uuid::Uuid;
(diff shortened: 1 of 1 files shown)


louislang (approved):

cd-work:
Will an incomplete job always cause a failure? I thought that the packages that have completed are still analyzed and all unchecked packages are counted as 100% safe? I feel like if there is a detected failure in a partially complete analysis, it should still fail the build?

kylewillmon:
> Will an incomplete job always cause a failure? I thought that the packages that have completed are still analyzed and all unchecked packages are counted as 100% safe? I feel like if there is a detected failure in a partially complete analysis, it should still fail the build?

I think you may be right. The current behavior seemed unexpected to me, but I was testing against placeholder data in the new API and not real data. I'll hold off on this PR until I can test against the new endpoint in phylum-dev/api#219

kylewillmon:
Confirmed in staging: Now that there is real data coming back from the API, the failure message only occurs when packages have been analyzed and the score falls below the threshold. And this makes sense even if the job isn't yet complete.

PR #382, opened 2022-05-16
Fix `phylum history` timestamp

### LINKED ISSUE #377: Jobs show incorrect date
After the latest API updates, jobs are being shown with incorrect dates:

```
❯ phylum history 40fecf21-a470-4dd8-8662-b2ddbcc11867
                                                                                                                                                      
                                                                                                                                                      
          Project: test                                                    Label: uncategorized                                                       
       Proj Score: 56                                                       Date: 1970-01-20 03:01:31 UTC                                             
         Num Deps: 11                                                     Job ID: 40fecf21-a470-4dd8-8662-b2ddbcc11867                                
          User ID:                                             View in Phylum UI: https://app.phylum.io/projects/c9385f5e-68aa-4b7d-bb96-dd1510cbba2f 
        Ecosystem: PyPI                                                                                                                               
                                                                                                                                                      
     Score       Count                                                                                                                                
     91 - 100  [   10] ████████████████████████████████████████████████████████                        Thresholds:                                    
     81 - 90   [    0]                                                                              Project Score: 100                                
     71 - 80   [    0]                                                                    Malicious Code Risk MAL: 100                                
     61 - 70   [    0]                                                                     Vulnerability Risk VLN: 100                                
     51 - 60   [    1]                                                                       Engineering Risk ENG: 100                                
     41 - 50   [    0]                                                                            Author Risk AUT: 100                                
     31 - 40   [    0]                                                                           License Risk LIC: 100                                
     21 - 30   [    0]                                                                                                                                
     11 - 20   [    0]                                                                                                                                
      0 - 10   [    0]                                                                                                                                
                                                                                                                                                      
           Status: PASS                                                                                                                               

```

cd-work:
Closes phylum-dev/api#228.

### DIFF
--- a/cli/src/summarize.rs
+++ b/cli/src/summarize.rs
@@ -134,7 +134,7 @@ where
 {
     let ecosystem = PackageType::from_str(&resp.ecosystem).unwrap_or(PackageType::Npm);
 
-    let date_time = NaiveDateTime::from_timestamp(resp.created_at / 1000, 0);
+    let date_time = NaiveDateTime::from_timestamp(resp.created_at, 0);
 
     let details = [
         (


kylewillmon (changes requested):
Gonna put a hold on this while we figure out if this issue should be moved to API

louislang:
I vote to move it to API.

cd-work:
Playing devil's advocate: It's clearly less code for us in CLI to use seconds, if it's also simpler for the API to use seconds (I'd assume so since otherwise they would send milliseconds), then why add code to API and CLI when both prefer seconds?

kylewillmon:
> Playing devil's advocate: It's clearly less code for us in CLI to use seconds, if it's also simpler for the API to use seconds (I'd assume so since otherwise they would send milliseconds), then why add code to API and CLI when both prefer seconds?

Because it's not much code and it gives us backwards compatibility.

But we will probably need to rethink our timestamp format eventually.... [Some places in the API](https://github.com/phylum-dev/phylum-types/blob/3f596a1924f640b70b1fd8aa7611aff29bc2010b/src/types/project.rs#L44) are now using `chrono::DateTime`, which serializes using RFC 3339 strings, so maybe that is the way forward?

kylewillmon:
Handled in phylum-dev/api#229

cd-work:
> But we will probably need to rethink our timestamp format eventually.... Some places in the API are now using chrono::DateTime, which serializes using RFC 3339 strings, so maybe that is the way forward?

Honestly for the CLI it shouldn't matter much. We just need to know what format it is and ideally that format should change as little as possible. The crates we use can parse pretty much anything.

PR #383, opened 2022-05-16
Fix yarn v1 parser with quoted version key

### LINKED ISSUE #380: Yarn lockfile parser fails on lockfile generated by NPM
# Overview
Apparently since NPM 7, if you have a `yarn.lock` and you run `npm install`, NPM will update your `yarn.lock` file. NPM uses slightly different syntax that is understood by Yarn but not understood by Phylum. Specifically, the thing that trips up the parser is that it puts quotes around more property keys.

# How To Reproduce
1. Set up a Yarn project.
2. Run `npm install`.
3. Run `phylum parse yarn.lock` (unreleased feature). A cryptic error will be displayed.

# Expected Behavior
`phylum parse yarn.lock` should work before and after running `npm install`.

cd-work:
This resolves an issue with the parser for the old v1 yarn format where
a lockfile generated by npm instead of yarn would cause all keys in the
lockfile to be quoted, breaking the parser.

This patch updates the parser to accept optional quotes around the
`version` key in the lockfile, since this is the only optionally quoted
field we care about.

Closes #380.


matt-phylum:
I actually have a different solution to this I've been working on. It works, but I found another issue with v1 lockfiles so I haven't opened a PR for it yet.

matt-phylum (approved):
This seems okay. The way I was solved it involved rewriting the Yarn parser. I found a couple more issues but we can fix them separately.

I think the Yarn parser rewrite is better in the long run because it's more careful and gives better errors when something is wrong, but it's definitely more change and has a greater risk of regression.

cd-work:
@matt-phylum As discussed on slack, I'd vote to merge this PR (or variation of it) and consider broader changes in a follow-up.  I think my current approach should do the job, but let me know if you have other ideas.

matt-phylum (approved):

PR #387, opened 2022-05-18
  Merge the docs workflow into release

kylewillmon:
Merge the `docs.yml` workflow into `release.yml`. Having two workflows for this was overcomplicated and the docs workflow is short enough to embed cleanly into the release workflow.


.github/workflows/docs.yml
```diff
@@ -3,8 +3,7 @@
 name: Sync `docs` directory to ReadMe
 
 on:
-  release:
-    types: [released]
+  workflow_call:
```
maxrake, line 6: This _could_ be simplified to a single line:

```
on: workflow_call
```

...but not required.

.github/workflows/release.yml
```diff
@@ -156,3 +156,9 @@ jobs:
           files: |
             phylum-*.zip
             phylum-*.zip.minisig
+
+  Deploy-Docs:
+    name: Deploy Docs
```
maxrake, line 161: "Publish" or "Update" might be slightly better than "Deploy" and more in line with the notion of documentation.

.github/workflows/release.yml
```diff
@@ -156,3 +156,9 @@ jobs:
...
+    needs: Release
+    uses: ./.github/workflows/docs.yml
+    secrets: inherit
```
maxrake, line 164: Are any secrets from this caller workflow actually needed/used in the called workflow? Is this line just pre-emptive, in case there are secrets to pass in the future?

maxrake (approved):
Just a question and some non-mandatory suggestions for change.

LGTM

.github/workflows/release.yml
kylewillmon: The docs workflow needs `secrets.README_API` and it wasn't clear from the docs if that would be set without doing it this way.

maxrake (dismissed):

PR #391, opened 2022-05-19
Bump to v3.4.0 - Group account support

kylewillmon:
This patch includes a change to split release.sh into two scripts:
bump_version.sh and tag.sh

Making this split allows us to avoid pushing directly to the default
branch, while still ensuring that the tag is created on the default
branch. With this update, we can enable stricter branch protection
rules on the repository.

Release-Version: v3.4.0
Release-Summary: Group account support

### DIFF
--- a/tag.sh
+++ b/tag.sh
@@ -1,0 +1,67 @@
+#!/bin/sh
+
+# usage: ./tag.sh [BRANCH]
+# Tag the latest release commit (i.e., a commit with a "Release-Version" trailer)
+#
+# This script should be run on the default branch after a version bump (using bump_version.sh) has
+# been merged to the default branch. It will find and tag the latest release commit. If no branch
+# argument is provided, the search will begin from HEAD.
+#
+# Note: After the commit is found, a check is performed to ensure that the commit is reachable from
+# the default branch on the remote (origin/HEAD). To skip this check, set SKIP_ORIGIN_HEAD_CHECK=1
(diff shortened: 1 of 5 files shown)


kylewillmon:
I included both script updates and the `v3.4.0` release both in this PR. But I can split those into separate PRs if desired.

louislang (approved):

cd-work (approved):

maxrake (dismissed):
Only one suggestion, that doesn't have to happen now:

Add a usage comment to `tag.sh`...for the possible input argument of a source branch and the possible override using `SKIP_ORIGIN_HEAD_CHECK`.

Ideally, there would be a `releasing.md` (or named whatever) file in the docs for explaining the release procedure. Since that file does not exist, a note in the `tag.sh` script...like exists for the `bump_version.sh` would be the next best thing.

tag.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/sh
+
+# usage: ./tag.sh [BRANCH]
+# Tag the latest release commit (i.e., a commit with a "Release-Verson" trailer)
```
maxrake, line 4:
```suggestion
# Tag the latest release commit (i.e., a commit with a "Release-Version" trailer)
```

maxrake (dismissed):
LGTM

maxrake (approved):

PR #392, opened 2022-05-19
docs: add KB article for Groups

furi0us333:
- Add KB article for Groups to better explain the use case
- Add link to Phylum UI in Quickstart/Readme

### DIFF
--- a/docs/groups.md
+++ b/docs/groups.md
@@ -1,0 +1,17 @@
+---
+title: Groups
+category: 6255e67693d5200013b1fa41
+hidden: false
+---
+
+Phylum groups allow users to view/submit projects and analysis jobs in a shared group context. This allows users to share results without needing to share account credentials.
+
+1. [Create a group](https://docs.phylum.io/docs/phylum_group_create) for your team to use
+2. [Create a project](https://docs.phylum.io/docs/phylum_project_create) using the `--group` option and your group name
+3. [Analyze](https://docs.phylum.io/docs/phylum_analyze) the desired lock file
+
+Any user that is a member of the group will be able to access the anlaysis results.
+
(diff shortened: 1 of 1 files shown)


kylewillmon:
Are we adding KB articles to the CLI docs now? Seems like that would be better placed somewhere else.

docs/phylum.md
```diff
@@ -12,6 +12,7 @@ Interact with the Phylum API
 * [phylum project](https://docs.phylum.io/docs/phylum_project)
 
 ### Additional Commands
+* [phylum group](https://docs.phylum.io/docs/phylum_group)
```
kylewillmon, line 15: Where is this file?

kylewillmon (changes requested):
I think a file is missing

furi0us333:
> Are we adding KB articles to the CLI docs now? Seems like that would be better placed somewhere else.

You will notice the `category` in the front matter is slightly different on `groups.md` which will place it in the KB section of the docs. In terms of docs as code, they just happen to live in this repo.

docs/phylum.md
furi0us333: https://github.com/phylum-dev/cli/blob/main/docs/phylum_group.md

It is new and won't be live until we release.

kylewillmon:
> they just happen to live in this repo.

Ah. I didn't notice that we already have a few of these... But we don't have all of them (i.e., [Phylum Package Score](https://docs.phylum.io/docs/phylum-package-score) and [Processing](https://docs.phylum.io/docs/processing) aren't here)...

I think these files should find a new home...

docs/phylum.md
kylewillmon: Oh... They already existed, but weren't linked. My bad.

PR #393, opened 2022-05-19
Remove unused deps in Cargo.toml

kylewillmon:
There were some dependencies in `Cargo.toml` that we weren't using (or weren't using directly). This patch removes them.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -23,7 +23,6 @@ dialoguer = "0.10.0"
 env_logger = "0.9.0"
 futures = "^0.3"
 git-version = "0.3.5"
-histogram = "0.6.9"
 home = "0.5.3"
 hyper = "0.14.16"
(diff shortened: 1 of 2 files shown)


cli/Cargo.toml
```diff
@@ -38,10 +37,9 @@ rand = "0.8.4"
 reqwest = { version = "0.11.3", features = ["blocking", "json", "rustls-tls"], default-features = false }
 routerify = { version = "3.0.0", features =["all"] }
 serde = { version = "^1.0", features = ["derive"] }
-serde_derive = "1.0"
 serde_json = "^1.0"
-serde-xml-rs = "0.5.1"
 serde_yaml = "0.8.14"
+serde-xml-rs = "0.5.1"
```
kylewillmon, line 42: Proof that machines and humans sort things differently....

PR #394, opened 2022-05-19
Remove APIResult type

kylewillmon:
This type has been flawed for a few reasons:

* Deserialization for this type only happens when an HTTP 200 status
  code is returned. But the API correctly uses 4xx and 5xx HTTP status
  codes for errors.

* The error format does not match the format used by the API.

* Using this type results in an obscure error message when
  deserialization fails:

```
data did not match any variant of untagged enum APIResult
```

## Additional context

As noted by @samtay in [this comment](https://github.com/phylum-dev/cli/pull/365#discussion_r869637268), API errors do not match this format.


janasheehan (approved):
it sort of makes sense to me... and look like it won't break anything

PR #398, opened 2022-05-20
Fix error with missing lockfile

### LINKED ISSUE #347: Unhelpful `phylum analyze` output for nonexistent lock file

cd-work:
Closes #347.

### DIFF
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -1,9 +1,9 @@
 //! `phylum parse` command for lockfile parsing
 
-use std::io;
 use std::path::Path;
+use std::{fs, io};
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use phylum_types::types::package::{PackageDescriptor, PackageType};
 
 use super::{CommandResult, ExitCode};
@@ -11,7 +11,7 @@ use crate::lockfiles::*;
 
 type ParserResult = Result<(Vec<PackageDescriptor>, PackageType)>;
 
-const LOCKFILE_PARSERS: &[(&str, &dyn Fn(&Path) -> ParserResult)] = &[
+const LOCKFILE_PARSERS: &[(&str, &dyn Fn(String) -> ParserResult)] = &[
     ("yarn", &parse::<YarnLock>),
     ("npm", &parse::<PackageLock>),
     ("gem", &parse::<GemLock>),
@@ -21,7 +21,6 @@ const LOCKFILE_PARSERS: &[(&str, &dyn Fn(&Path) -> ParserResult)] = &[
     ("mvn", &parse::<Pom>),
     ("gradle", &parse::<GradleDeps>),
     ("nuget", &parse::<CSProj>),
-    ("auto", &get_packages_from_lockfile),
 ];
 
 pub fn lockfile_types() -> Vec<&'static str> {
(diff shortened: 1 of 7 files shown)


cd-work:
I've opened this as a draft since I'm curious what others like @kylewillmon think about the approach. It's quite a big change just for improving the error with missing lockfiles.

It seems like none of the lockfile parsers actually need to accept a path, so accepting the contents of the file instead seems reasonable to me. However it doesn't actually seem to provide any tangible benefit; even for the automatic lockfile format detection we still have to clone the entire `String` since the parsers don't take a reference. That said I feel like taking a reference might be possible? Maybe having the parsers completely stateless, just taking a `&str` as parameter would be the way to go.

The best alternative I can think of would be to check if the file is readable, but there doesn't seem to be a convenient way to do this. You can check for `Path::exists`, but there's no such thing as `Path::readable`. So ignoring performance I believe the easiest way to determine if a file can be read would be to just read the first byte?

I'm also open to other suggestions.

kylewillmon:
> Maybe having the parsers completely stateless, just taking a `&str` as parameter would be the way to go.

I played around with this a while back. In an ideal world, I think we would replace the `Parseable` trait with a `Parser` trait that looked something like this:

```
pub trait Parser {
    /// Parse the given lockfile
    fn parse<R: std::io::Read>(&self, reader: R) -> ParseResult;
    /// Indicate the parser name.
    fn parser_name(&self) -> &'static str;
    /// Indicate the type of packages parsed by this parser
    fn package_type(&self) -> PackageType;
    /// Check if a lockfile is supported by filename
    fn file_name_match(&self, file_name: &std::ffi::OsStr);
}
```

And then all of the parsers would be unit structs that implement those functions.

cli/src/commands/parse.rs
```diff
@@ -21,7 +21,6 @@ const LOCKFILE_PARSERS: &[(&str, &dyn Fn(&Path) -> ParserResult)] = &[
     ("mvn", &parse::<Pom>),
     ("gradle", &parse::<GradleDeps>),
     ("nuget", &parse::<CSProj>),
-    ("auto", &get_packages_from_lockfile),
```
kylewillmon, line 24: I think you will need to update the `lockfile_types()` function to still return `"auto"` in its list so that `clap` will accept `-t auto`

cli/src/commands/parse.rs
```diff
@@ -33,46 +32,33 @@ pub fn handle_parse(matches: &clap::ArgMatches) -> CommandResult {
...
-    macro_rules! try_format {
-        ($lock:ident, $ty:literal) => {{
-            let packages = parse::<$lock>(path).ok();
-            if let Some(packages) = packages.filter(|(packages, _)| !packages.is_empty()) {
-                log::debug!("Submitting file as type {}", $ty);
-                return Ok(packages);
-            }
-        }};
+pub fn try_get_packages(lockfile_content: String) -> Result<(Vec<PackageDescriptor>, PackageType)> {
+    for (ty, parser) in LOCKFILE_PARSERS.iter().filter(|(ty, _)| ty != &"auto") {
```
kylewillmon, line 54: There's no need to call `.filter()` here now that `auto` has been removed from the list of parsers.

cli/src/lockfiles/mod.rs
```diff
@@ -22,6 +22,13 @@ pub type ParseResult = anyhow::Result<Vec<PackageDescriptor>>;
 
 pub trait Parseable {
     fn new(filename: &Path) -> Result<Self, io::Error>
+    where
+        Self: Sized,
+    {
+        Ok(Self::from_string(std::fs::read_to_string(filename)?))
+    }
```
kylewillmon, line 29: Is the `new()` function still called anywhere?

cli/src/lockfiles/mod.rs
```diff
@@ -22,6 +22,13 @@ pub type ParseResult = anyhow::Result<Vec<PackageDescriptor>>;
...
+
+    fn from_string(text: String) -> Self
```
kylewillmon, line 31: With a `parse()` and a `from_string()` method, we're kind of dancing around the [`FromStr` trait](https://doc.rust-lang.org/std/str/trait.FromStr.html). We could almost implement this as:

```
pub trait Parseable: FromStr<Err=anyhow::Error> + Into<Vec<PackageDescriptor>> { }
```

kylewillmon (commented):
I think this change would be an improvement

cli/src/lockfiles/mod.rs
cd-work: Only tests. I've considered removing it, but then you have a bunch of tests that all have to manually do this. So I thought I might as well keep it. I'd probably rename it though, to not favor it over `from_string`?

cli/src/lockfiles/mod.rs
cd-work: `FromStr` is fallible though, while this `from_string` method is not. The ownership is also significant here since we otherwise would need to clone it internally.

kylewillmon: Right. This would only work in combination with your other comment about making the parsers take `&str`. And then `T::from_string(foo).parse()` becomes `foo.parse::<T>()`

cd-work: I feel like the `FromStr` trait often gets in the way more than it actually benefits people. Especially because trait discovery is often harder than finding `::new` or similar methods.

cd-work:
Resolved in https://github.com/phylum-dev/cli/pull/423.

PR #399, opened 2022-05-20
Use new API endpoint for OIDC redirect

### LINKED ISSUE #373: Consume the new `/.well-known/openid-configuration` redirect endpoint.
# Overview
I recently [implemented a new API endpoint](https://github.com/phylum-dev/api/pull/222) for redirecting to the Keycloak OpenID configuration. We need to update the CLI to consume this instead of using the `oidc_discovery_url` config value.

# Acceptance Criteria
- [ ] All usages of the `oidc_discovery_url` config value should be replaced by the API endpoint `/.well-known/openid-configuration`.
- [ ] The CLI user should not have to specify the `oidc_discovery_url` config value anymore.

cd-work:
Previously the `oidc_discovery_url` had to be specified in the
configuration file to handle the user authentication.

This patch makes use of the new `/.well-known/openid-configuration`
endpoint of the API, which automatically redirects to the correct URL.

Closes #373.

### DIFF
--- a/cli/src/api/endpoints.rs
+++ b/cli/src/api/endpoints.rs
@@ -77,3 +77,8 @@ pub(crate) fn group_create(api_uri: &str) -> String {
 pub fn group_project_summary(api_uri: &str, group: &str) -> String {
     format!("{api_uri}/{API_PATH}/groups/{group}/projects")
 }
+
+/// GET /.well-known/openid-configuration
+pub fn oidc_discovery(api_uri: &str) -> String {
+    format!("{api_uri}/{API_PATH}/.well-known/openid-configuration")
+}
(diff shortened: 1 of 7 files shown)


maxrake (changes requested):
The code _seems_ good...but trying the code...from a fresh install fails:

image

*EDIT*: Just realized this **won't** work until the changes from API make it into prod. I retract my previous "needs work" ...but still think this should be confirmed after the API updates are in prod and before releasing the feature with CLI.

maxrake:
Update to the update: I tried again, with the staging infrastructure, and got the same errors. So...now I don't know what is going on.

image

maxrake:
After this issue is merged and before it is included in a CLI release, the internal developer documentation for [Setting up CLI for localdev](https://www.notion.so/phylum/Setting-up-CLI-for-localdev-3312fddd8fa040c7bba6d9f1dc16f6fe) should be updated.

cd-work:
Should work now @maxrake, there was an issue on the API side.

PR #401, opened 2022-05-20
Don't use streaming parsers

### LINKED ISSUE #402: Analyze command fails to parse unknown lockfiles
# Overview
The `phylum analyze` command takes a lockfile and will attempt to parse it based on type. This works when the lockfile name is one of the "well known" names, but fails with a return code of 101 when a non-standard name is provided.

# How To Reproduce
1. Using Phylum CLI v3.4.0
2. Create a Python lockfile named `requirements.txt` and put at least one pinned dependency in it
3. Run the command `phylum analyze requirements.txt` and see that it passes
4. Rename the lockfile to `requirements-dev.txt`
5. Run the command `phylum analyze requirements-dev.txt` and see that it fails with return code of 101
6. See error
```sh
[2022-05-20T19:02:33Z WARN  phylum_cli::commands::parse] Attempting to obtain packages from unrecognized lockfile type: requirements-dev.txt
thread 'main' panicked at 'Cannot call `finish()` on `Err(Err::Incomplete(_))`: this result means that the parser does not have enough data to decide, you should gather more data and try to reapply  the parser instead', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/nom-7.1.1/src/internal.rs:41:9
```

# Expected Behavior
The `analyze` command is able to parse the lockfile, regardless of it's name.

# Additional Context
This was discovered with Phylum CLI v3.4.0 but unclear if it affects other versions.

kylewillmon:
The use of streaming parsers in combination with nom::Finish will
occasionally cause a panic...

Fixes #402

### DIFF
--- a/cli/src/lockfiles/parsers/mod.rs
+++ b/cli/src/lockfiles/parsers/mod.rs
@@ -1,10 +1,7 @@
 use nom::{
     branch::alt,
     bytes::complete::{tag, take, take_until},
-    character::{
-        complete::{line_ending, none_of, not_line_ending, space0},
-        streaming::multispace0,
-    },
+    character::complete::{line_ending, multispace0, none_of, not_line_ending, space0},
     combinator::{eof, opt, recognize},
     error::{context, ParseError, VerboseError},
     multi::{count, many1, many_till},


matt-phylum (approved):

maxrake (approved):
Tested locally.

LGTM

PR #405, opened 2022-05-23
Add support for `gradle.lockfile`

### LINKED ISSUE #372: Add support for `gradle.lockfile`
# Overview

Currently the Phylum CLI only supports parsing the output of `gradle dependencies` as lockfile, however this format is made for consumption by humans. Instead, support for the `gradle.lockfile` format should be added, which is a proper lockfile generated by Gradle itself.

Since it seems like there's no user of the existing gradle "lockfile" parser, I'd also strongly suggest removing it before people start relying on the Phylum CLI being able to parse this rather arbitrary and prone-to-change format.

# Acceptance Criteria

- [x] Phylum CLI correctly identifies `gradle.lockfile`
- [x] All packages from the `gradle.lockfile` and their version are submitted to the API as Maven dependencies
- [x] Support for `gradle dependencies` parsing is removed

cd-work:
Before this patch, only preliminary Gradle support was present in the
Phylum CLI by using some of Gradle's output as pseudo-lockfile. This
patch adds proper Gradle support by adding a parser for the
gradle.lockfile file format.

Since the previous format was unreliable and based on human-readable
output, the new Gradle parser replaces the old one instead of being
offered as an alternative.

Closes #372.

### DIFF
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -19,7 +19,7 @@ const LOCKFILE_PARSERS: &[(&str, &dyn Fn(&Path) -> ParserResult)] = &[
     ("pipenv", &parse::<PipFile>),
     ("poetry", &parse::<Poetry>),
     ("mvn", &parse::<Pom>),
-    ("gradle", &parse::<GradleDeps>),
+    ("gradle", &parse::<GradleLock>),
     ("nuget", &parse::<CSProj>),
     ("auto", &get_packages_from_lockfile),
 ];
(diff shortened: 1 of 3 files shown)


maxrake (approved):
It all looks good. The only thing found, and it's a small find at that, is the possibility of changing this line in `cli/src/commands/parse.rs`:

`try_format!(GradleDeps, "gradle dependencies");`

...to say "gradle lock" or even "gradle lockfile" to match the other entries there.

cd-work:
I've decided to also rename `GradleDeps` to `GradleLock`. That should be a little more accurate and in line with the other structs.

maxrake (approved):

PR #406, opened 2022-05-23
Emit unique exit code when failing thresholds

### LINKED ISSUE #403: Emit unique exit code when "failing the build"
# Overview
The `phylum analyze` command will exit with an error `Project failed threshold requirements, failing the build!` and an exit code of 1 when the established project threshold values are not met. It would be great if this exit code was something other than 1 since 1 is a very common exit code and it can be hard to differentiate between (1) this was a failure in the sense that standards were not met and (2) this was a failure due to anything else. 

# Acceptance Criteria
- [ ] With project thresholds set and a `phylum analyze` command issued, a result that does not meet established thresholds should "fail the build" with an exit code other than 1
- [ ] The exit code for "failing the build" is unique amongst all other CLI exit codes
- [ ] Documentation is updated so users can know to look for this unique code in their own CI/QA use cases

cd-work:
Previously a failure to meet project thresholds would result in the
generic exit code of 1, making it difficult to identify the type of
error that occurred.

Instead of returning a generic exit code of 1, a special exit code of
100 has been introduced to identify failure to meet the project
thresholds.

Closes #403.

### DIFF
--- a/cli/src/bin/phylum.rs
+++ b/cli/src/bin/phylum.rs
@@ -1,5 +1,4 @@
 use std::path::PathBuf;
-use std::process;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use anyhow::{anyhow, Context};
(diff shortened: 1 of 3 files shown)


cli/src/commands/mod.rs
```diff
@@ -27,10 +29,20 @@ impl From<ExitCode> for CommandValue {
...
     SetThresholdsFailure = 13,
+    FailedThresholds = 100,
+}
+
+impl ExitCode {
+    /// Terminate the application with this exit code.
+    pub fn exit(&self) -> ! {
+        process::exit(*self as i32);
+    }
 }
```
cd-work, line 48: I think this method provides a nice abstraction to ensure that we never accidentally exit the application with an exit code already used for anything else. This way a simple grep for `process::exit` shows locations of potential violators.

cli/src/bin/phylum.rs
```diff
@@ -27,22 +26,22 @@ use phylum_types::types::job::Action;
 pub fn exit_warn(message: impl AsRef<str>) -> ! {
     warn!("{}", message.as_ref());
     print_user_warning!("Warning: {}", message.as_ref());
-    process::exit(0)
+    ExitCode::Ok.exit()
 }
 
 /// Print an error to the user before exiting with exit code 1.
```
maxrake, line 32: This comment should be updated to match the change.

maxrake (approved):
Thanks for updating the documentation.
LGTM

maxrake (approved):

PR #384, opened 2022-05-18
Use new format for package analysis endpoint

cd-work:
Fixes #376.

PR #408, opened 2022-05-23
Revert "Use new format for package analysis endpoint (#384)"

### LINKED ISSUE #376: `phylum package` command does not work with new API
After the latest API updates, the `phylum package` command does not work

```
❯ RUST_LOG=debug phylum package -t pypi six 0.16.0
[ ... SNIP ... ]
[2022-05-14T01:51:37Z DEBUG reqwest::async_impl::client] response '200 OK' for https://api.staging.phylum.io/api/v0/job/packages/pypi/six/0.16.0
[2022-05-14T01:51:37Z DEBUG phylum_cli::commands::packages] ==> Err(Other(data did not match any variant of untagged enum APIResult))
[2022-05-14T01:51:37Z DEBUG phylum_cli::print] ==> Err(Other(data did not match any variant of untagged enum APIResult))
❗ Response error:
data did not match any variant of untagged enum APIResult
```

kylewillmon:
This reverts commit df2130b5c81e0e05958f380a8f715ae70b2c0c73.

The change in #384 fixed #376, but accidentally caused phylum-dev/api#254

Doing this revert is the quickest way to resolve this mistake. But it will require us to re-open #376 and find a solution that does not cause this breakage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1235,7 +1235,7 @@ dependencies = [
 [[package]]
 name = "phylum_types"
 version = "0.1.0"
-source = "git+https://github.com/phylum-dev/phylum-types?branch=development#fb64032119e160a4b7719dd58ba1e10454d6af0a"
+source = "git+https://github.com/phylum-dev/phylum-types?branch=development#3f596a1924f640b70b1fd8aa7611aff29bc2010b"
 dependencies = [
  "chrono",
  "log",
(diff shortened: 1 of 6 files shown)


cd-work (commented):
This will break the development version in favor of being able to release without any problems, right? To me this seems like a reason why I generally prefer releases from separate branches since it seems like the actual goal here is not to revert it on the development branch, but to revert it in prod?

kylewillmon:
I think we have found a different approach

Issue #337, opened 2022-05-02
Consider use of `rust-cache` for CI


kylewillmon:
After reading [this blog post](https://www.reillywood.com/blog/rust-faster-ci/) about results in [nushell](https://github.com/nushell/nushell), I believe we could benefit from using `rust-cache` in our CI builds.

cd-work:
One thing I'd like to mention is that with Rust caching in CI one has to be careful not to cache too much. I've seen in multiple of my own repos that caching can sometimes initially boost build times but over the long term slow them down when the caches build up to become unwieldy.

kylewillmon:
> One thing I'd like to mention is that with Rust caching in CI one has to be careful not to cache too much. I've seen in multiple of my own repos that caching can sometimes initially boost build times but over the long term slow them down when the caches build up to become unwieldy.

It looks like `nushell` is using a version number in their [cache key](https://github.com/nushell/nushell/blob/ac48f5a318adc7204700dd10cc8092d5d864659d/.github/workflows/ci.yml#L44) to avoid this. If/when the build slows down, they can simply bump that version and bust the cache.

We could do something like that. Or we could even automatically bump the key once a month or so by using something like this:

```
  - name: Set cache version
    run: echo 'cache_version=v'$(date '+%Y.%m') >> $GITHUB_ENV
  - uses: Swatinem/rust-cache@v1
    with:
      key: ${{ env.cache_version }}
```

PR #411, opened 2022-05-23
Use rust-cache for test.yml

kylewillmon:
Update test.yml to use rust-cache in order to speed up builds.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,32 +13,6 @@ on:
     - cron: '30 5 * * 1'
 
 jobs:
-  lint:
-    # It is possible for lints to break after merge (e.g., new clippy checks), but it's not worth
-    # breaking the build if that happens. So we'll skip this job for scheduled runs.
-    if: github.event_name != 'schedule'
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout the repo
-        uses: actions/checkout@v3
-
-      - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-
-      - name: Format check
-        uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: --manifest-path cli/Cargo.toml -- --check
-
-      - name: Lint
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -13,52 +13,50 @@ on:
...
+        uses: actions/checkout@master
 
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
+          profile: minimal
+          components: rustfmt, clippy
+
+      - uses: Swatinem/rust-cache@v1
```
kylewillmon, line 33: I considered adding a special key here to address [this concern](https://github.com/phylum-dev/cli/issues/337#issuecomment-1116270098) expressed by @cd-work, but after reading the [details](https://github.com/Swatinem/rust-cache#cache-details) for this Github action, it looks like it is already using a very specific key (even including a hash of `Cargo.lock`), so I don't think it is necessary to add any additional keying until/unless we see problems.

louislang: We should apply this change elsewhere if we can.

maxrake (approved):
Great idea and great improvement in CI times! One small recommendation and one question, but approving now.

.github/workflows/test.yml
```diff
@@ -13,52 +13,50 @@ on:
...
-      - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
+      - name: Clippy
+        if: github.event_name != 'schedule'
+        uses: actions-rs/cargo@v1
         with:
-          toolchain: stable
-          profile: minimal
+          command: clippy
+          args: --locked -- -D warnings
```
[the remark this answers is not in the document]
kylewillmon: I just wanted these to be consistent. And testing against the default feature set is most important. We may also want to add some other feature set (such as all features or no features). But that should be added to the matrix and I leave that as future work

cd-work (approved):

maxrake (approved):

cd-work (approved):

PR #415, opened 2022-05-24
Refactored if/else chains to a non-exhaustive match statement

andreaphylum:


cli/src/bin/phylum.rs
```diff
@@ -58,16 +58,6 @@ async fn handle_commands() -> CommandResult {
...
-    #[cfg(feature = "selfmanage")]
-    if let Some(matches) = matches.subcommand_matches("uninstall") {
-        return handle_uninstall(matches);
-    }
-
-    #[cfg(feature = "extensions")]
-    if let Some(matches) = matches.subcommand_matches("extension") {
-        return handle_extensions(matches).await;
-    }
-
```
kylewillmon, line 70: These commands were at this point in the function to avoid requiring authentication to run them.

kylewillmon (changes requested):
We need to figure out a better way to deal with commands that require authentication and those that don't. This change makes all commands require authentication, which isn't good

cli/src/bin/phylum.rs
```diff
@@ -173,15 +142,6 @@ async fn handle_commands() -> CommandResult {
...
-        &mut config.auth_info,
-        &config.connection.uri,
-        timeout,
-        ignore_certs,
-    )
-    .await
-    .context("Error creating client")?;
-
     // PhylumApi may have had to log in, updating the auth info so we should save the config
     save_config(&config_path, &config).with_context(|| {
```
kylewillmon, line 146: As noted in the comment, `save_config` needs to happen after `PhylumApi::new()` in order to save new credentials if a login occurred.

cli/src/bin/phylum.rs
kylewillmon: It took a few hours for this to finally hit me... But now I see that you've solved this with the `async` block around `PhylumApi::new()`...

andreaphylum: Authentication will not happen until the `api` future is polled; this way, only the functions calling `api.await?` (see [this](https://github.com/phylum-dev/cli/pull/415/files#diff-ad5eecad8cf0e108661eb7654555c2e7b0fb3aa2f4b0d01549853615fd27dd34R182) for an example) will actually trigger the authentication process. I feel like this is a better pattern to use in general, and we could further refactor this module to that end.

cli/src/bin/phylum.rs
```diff
@@ -190,29 +150,58 @@ async fn handle_commands() -> CommandResult {
...
+    let api = async {
+        PhylumApi::new(
+            &mut config.auth_info,
+            &config.connection.uri,
+            timeout,
+            ignore_certs,
+        )
+        .await
+        .context("Error creating client")
+    };
```
kylewillmon, line 162: `async` blocks are weird... I fear this makes the code a bit too hard to read.

kylewillmon (approved):

cli/src/bin/phylum.rs
kylewillmon: I've thought a bit about making more use out of the `config` object so that we can pass that around instead of passing `ignore_certs`, `api`, and `timeout` all separately. But that can wait for a future PR.

cli/src/bin/phylum.rs
andreaphylum: We should then probably factor all of this behavior out into a separate `async fn`, in order to not have the logic split over many different points.

cli/src/bin/phylum.rs
```diff
@@ -46,45 +47,56 @@ pub fn exit_error(error: Box<dyn std::error::Error>, message: impl AsRef<str>) -
...
     let app = phylum_cli::app::app()
         .arg_required_else_help(true)
         .subcommand_required(true);
     let app_name = app.get_name().to_string();
-    let ver = app.get_version().unwrap();
-
     // Required for printing help messages since `get_matches()` consumes `App`
     let app_helper = &mut app.clone();
-
+    let ver = app.get_version().unwrap();
```
cd-work, line 89: Why was this moved?

cli/src/bin/phylum.rs
```diff
@@ -93,6 +105,20 @@ async fn handle_commands() -> CommandResult {
         )
     })?;
 
+    // `timeout` and `ignore_certs` are arguments to the API factory.
```
cd-work, line 108: This comment seems somewhat strange to me, since it documents how variables are used which I can usually just check with my IDE's functionality.

If the goal is to document that we're initializing everything required to create the API, that should be what is stated in the comment. This also comes with the advantage of being more generic and thus being more robust to changes.

cli/src/bin/phylum.rs
```diff
@@ -180,45 +181,68 @@ async fn handle_commands() -> CommandResult {
...
+async fn handle_extension_run(extension_subcmd: &str) -> CommandResult {
+    for extension in extensions::installed_extensions().unwrap_or_default() {
+        if extension_subcmd == extension.name() {
+            return extension.run().await;
+        }
     }
 
-    let should_submit = matches.subcommand_matches("analyze").is_some()
-        || matches.subcommand_matches("batch").is_some();
+    Err(anyhow!("Extension {extension_subcmd} not found"))
```
kylewillmon, line 223: I think this is unreachable. And if we do get here, it's probably because somebody added a new command to `app.rs` and forgot to handle it in `handle_commands()`. In which case, this error message would be confusing. I think a panic is appropriate.

```suggestion
    unreachable!()
```

cli/src/bin/phylum.rs
```diff
@@ -137,37 +167,8 @@ async fn handle_commands() -> CommandResult {
...
-        .and_then(|t| t.parse::<u64>().ok());
-
-    let ignore_certs =
-        matches.is_present("no-check-certificate") || config.ignore_certs.unwrap_or_default();
-    if ignore_certs {
-        log::warn!("Ignoring TLS server certificate verification per user request.");
-    }
-
+    // `auth` consumes `config` which prevents the API factory future from borrowing it, so we
+    // return early in order to guarantee that it has not been consumed after this point.
```
cd-work, line 171: We return early after every subcommand anyway, right? It's not like we wouldn't return if the `config` was borrowed?

cli/src/bin/phylum.rs
```diff
@@ -180,45 +181,68 @@ async fn handle_commands() -> CommandResult {
...
+        #[cfg(feature = "extensions")]
+        "extensions" => handle_extensions(matches).await,
+
+        #[cfg(feature = "extensions")]
+        extension_subcmd => handle_extension_run(extension_subcmd).await,
+
+        #[cfg(not(feature = "extensions"))]
+        _ => unreachable!(),
+    }
+}
```
cd-work, line 213: This is super clean now, love it.

cd-work (changes requested):

cli/src/bin/phylum.rs
```diff
@@ -180,45 +181,68 @@ async fn handle_commands() -> CommandResult {
...
+
+    let (subcommand, matches) = matches.subcommand().unwrap();
+    match subcommand {
+        "version" => handle_version(&app_name, ver),
+        "update" => handle_update(matches).await,
+        "parse" => handle_parse(matches),
+        "ping" => handle_ping(api.await?).await,
+        "project" => handle_project(&mut api.await?, matches).await,
+        "package" => handle_get_package(&mut api.await?, &config.request_type, matches).await,
+        "history" => handle_submission(&mut api.await?, config, matches).await,
```
kylewillmon, line 197: _Nothing to change here. I'm just intrigued by this_

I'm amazed that the borrow checker is okay with this. `api` holds a mutable reference to `config` until the `.await` happens. Which is just barely in time because this function consumes `config`...

If the arguments were in the opposite order, I'd guess you'd have to split it onto two lines to get it to compile.

kylewillmon (dismissed):
Splitting the code into a separate function definitely helps the readability

cli/src/bin/phylum.rs
```diff
@@ -180,45 +181,68 @@ async fn handle_commands() -> CommandResult {
...
-    // PhylumApi may have had to log in, updating the auth info so we should save the config
-    save_config(&config_path, &config).with_context(|| {
-        format!(
-            "Failed to save configuration to '{}'",
-            config_path.to_string_lossy()
-        )
-    })?;
+    // This is `Future<Output = PhylumApi>`. Commands that require access to the API
+    // will await on this, so that the API is not instantiated ahead of time for subcommands
+    // that don't require it.
```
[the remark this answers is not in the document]
kylewillmon: I think the comment is mentioning the type in an effort to point out that we are deliberately using the future here rather than immediately awaiting. Something like this might be more clear.

```suggestion
    // Creating the PhylumAPI object requires that the user be authenticated. But not
    // every command needs authentication. So here we intentionally delay awaiting on
    // the future in order to also delay the authentication requirement.
```

cli/src/bin/phylum.rs
cd-work: Either way the application will error out. I see no harm in printing an error message that at least attempts to provide some detail.

kylewillmon: That's fair. But as you [note elsewhere](https://github.com/phylum-dev/cli/pull/415/files/9cf4b7420f079ac9e9017e00b8d58fa11ce2a5a2#r881990505), the error message here is misleading. Using `unreachable!()` is just my way of avoiding writing a better error message 😄

cli/src/bin/phylum.rs
andreaphylum: Just to keep initialization-related/`app` unpacking things closer together.

cli/src/bin/phylum.rs
andreaphylum: Indeed, those two variables are used only a bit later but I didn't find an appropriate spot for them. I will change the comment, but the best thing imo would be to refactor the data dependencies between `Config` and `PhylumApi` in such a way that these bindings can be removed altogether. I am afraid it might not be super trivial, though.

cli/src/bin/phylum.rs
```diff
@@ -123,6 +149,10 @@ async fn handle_commands() -> CommandResult {
         print_update_message();
     }
```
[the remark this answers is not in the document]
andreaphylum: Agreed -- I didn't change anything around this spot as it wasn't a priority and I didn't want to unpack the logic and risk making a mistake, as I'm not familiar with this part's mechanic and I didn't find tests covering it.

cli/src/bin/phylum.rs
andreaphylum: We do return after every command, but that "early" means "earlier than `PhylumApi` has a chance of borrowing it" (which would be, anywhere in the closing `match` block).

cli/src/bin/phylum.rs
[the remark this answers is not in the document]
andreaphylum: I agree on both points, but on one hand we can't make the type explicit AFAIK (as it's an opaque `impl Future<...>`) and on the other there may be a real need to address this in some way, as I interpreted from @kylewillmon's observation that the `async` block felt weird in the sense that it's an unawaited `Future`. I feel like having the comment is a decent compromise, but if there was another way of expressing the type inline that's tolerable by Rust I'd go that way instead.

cli/src/bin/phylum.rs
```diff
@@ -180,45 +181,68 @@ async fn handle_commands() -> CommandResult {
...
+        _ => unreachable!(),
+    }
+}
 
-    if matches.subcommand_matches("ping").is_some() {
-        let resp = api.ping().await;
-        print_response(&resp, true, None);
-        return Ok(ExitCode::Ok.into());
+#[cfg(feature = "extensions")]
+async fn handle_extension_run(extension_subcmd: &str) -> CommandResult {
```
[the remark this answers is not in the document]
andreaphylum: Agreed -- though in this case we could also just make it a one-liner like `Extension::load(extension_subcmd)?.run(matches).await`.

cli/src/bin/phylum.rs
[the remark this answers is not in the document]
andreaphylum: Using `Extension::load` would probably take care of this at the root.

cli/src/bin/phylum.rs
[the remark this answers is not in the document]
kylewillmon: I think his point is only that mentioning the type is unnecessary. The comment is definitely helpful. But it's just as helpful without the type:

```suggestion
    // Commands that require access to the API will await on this, so that the API is not
    // instantiated ahead of time for subcommands that don't require it.
```

Or if you prefer, just replace the mention of the type with something like this:

```suggestion
    // Get the future, but don't await. Commands that require access to the API will await on this,
    // so that the API is not instantiated ahead of time for subcommands that don't require it.
```

cli/src/bin/phylum.rs
kylewillmon: `handle_submission()` also consumes `config`, but it gets to be in the `match` block... The difference, of course, is that (as we've discussed in another comment), the borrow ends when `api.await` occurs.

For `handle_auth()`, we don't need the API object, so the borrow can be ended with `drop(api)`. And then this can join the others in the `match` block.

kylewillmon (approved):

kylewillmon (approved):

cli/src/bin/phylum.rs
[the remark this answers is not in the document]
kylewillmon: Created #428 to track this

PR #420, opened 2022-05-24
docs: update image URL

furi0us333:
This fixes the broken link that resulted from the website migration

image


louislang:
That's a great idea @cd-work

maxrake (approved):

mathew-horner:
I see in your last commit you added the image to the repo but don't we also need to update the HTML to point to the local file?

furi0us333:
> I see in your last commit you added the image to the repo but don't we also need to update the HTML to point to the local file?

Yes, sorry, got distracted doing something else and didn't finish this.

README.md
```diff
@@ -1,5 +1,5 @@
 <p align="center">
-  <img height="100" src="https://phylum.io/logo/dark-bckg.svg">
+  <img height="100" src="https://app.phylum.io/static/media/phylum-blue.d9e09fd4ac5774237f0c44f3173cb783.svg">
```
kylewillmon, line 2: The hash in the filename means we have to change this link anytime the logo changes....

docs/dark-bckg.svg
```diff
@@ -0,0 +1,30 @@
+<svg width="344" height="126" viewBox="0 0 344 126" fill="none" xmlns="http://www.w3.org/2000/svg">
```
kylewillmon, line 1: Is this file being used? Also, this same file seems to be named `phylum-blue.svg` elsewhere, so why is it `dark-bckg.svg` here?

README.md
```diff
@@ -1,5 +1,5 @@
 <p align="center">
-  <img height="100" src="https://phylum.io/logo/dark-bckg.svg">
+  <img height="100" src="https://raw.githubusercontent.com/phylum-dev/cli/main/assets/dark-bckg.svg">
```
maxrake, line 2: Does this work?
```suggestion
  <img height="100" src="assets/dark-bckg.svg">
```

docs/quickstart.md
```diff
@@ -6,7 +6,7 @@ hidden: false
 ---
 
 <p align="center">
-  <img height="100" src="https://phylum.io/logo/dark-bckg.svg">
+  <img height="100" src="https://raw.githubusercontent.com/phylum-dev/cli/main/assets/dark-bckg.svg">
```
maxrake, line 9: Does this work?
```suggestion
  <img height="100" src="../assets/dark-bckg.svg">
```

kylewillmon (approved):

maxrake (approved):
The changes ~work~, but there is a question about using relative source references instead of a raw.githubusercontent.com link.

**EDIT:** It looks like the changes do not work...at least not from the GitHub view:

---
image

---

**UPDATE:** The link is for when this branch gets merged to `main`...retracting previous failure comment/image...

furi0us333:
> The changes work, but there is a question about using relative source references instead of a raw.githubusercontent.com link.

I could do this for the `README.md`, but not for the `quickstart.md` that is for readme.com

README.md
furi0us333: this will work, but keeping it in sync with `quickstart.md`

docs/quickstart.md
furi0us333: I am not confident this will work, will consider this after I get a chance to test the behavior on `readme.com`

PR #421, opened 2022-05-24
Require slashes for installing local extensions

### LINKED ISSUE #356: Extensions: Require slashes for local installs
Currently, `phylum extension add` will install extensions from a local directory.

In the future, we may have a marketplace of extensions and it would make more sense that `phylum extension add <extension-name>` adds an extension from there.

To ensure that we can distinguish between local and marketplace installations in the future, the `phylum extension add` should enforce a specific format for local installs.

After discussion, we have decided that a local install will be required to use a slash in the `phylum extension add` command. For example:

```
phylum extension add foo          # Install `foo` from marketplace
phylum extension add ./foo        # Install from local directory `foo`
phylum extension add foo/bar      # Install from local directory `foo/bar`
phylum extension add /path/to/foo # Install from local directory `/path/to/foo`
phylum extension add C:\dev\foo   # Install from local directory `C:\dev\foo`
```

## Error message
Since we do not currently support marketplace installs, any `phylum extension add` command that attempt to install from the marketplace will fail with a descriptive error message:

```
❗ Error: Attempting to install from marketplace, which is not yet supported. Did you mean `phylum extension add ./foo`
```

cd-work:
This reserves installation of extensions without slash, allowing us to
later use this syntax to install extensions from the marketplace without
risking users accidentally installing these remote extensions when
trying to install local ones.

Closes #356.

### DIFF
--- a/cli/src/commands/extensions/mod.rs
+++ b/cli/src/commands/extensions/mod.rs
@@ -5,7 +5,7 @@ use std::{collections::HashSet, convert::TryFrom, fs, io::ErrorKind, path::PathB
 use crate::commands::{CommandResult, CommandValue, ExitCode};
 pub use extension::*;
 
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use clap::{arg, ArgMatches, Command, ValueHint};
 use log::{error, warn};
 
@@ -62,7 +62,7 @@ pub fn add_extensions_subcommands(command: Command<'_>) -> Command<'_> {
 /// Entry point for the `extensions` subcommand.
 pub async fn handle_extensions(matches: &ArgMatches) -> CommandResult {
     match matches.subcommand() {
-        Some(("add", matches)) => handle_add_extension(matches.value_of_t("PATH")?).await,
+        Some(("add", matches)) => handle_add_extension(matches.value_of("PATH").unwrap()).await,
         Some(("remove", matches)) => {
             handle_remove_extension(matches.value_of("NAME").unwrap()).await
         }
@@ -73,8 +73,17 @@ pub async fn handle_extensions(matches: &ArgMatches) -> CommandResult {
 
 /// Handle the `extension add` subcommand path.
 /// Add the extension from the specified path.
-async fn handle_add_extension(path: PathBuf) -> CommandResult {
-    let extension = Extension::try_from(path)?;
+async fn handle_add_extension(path: &str) -> CommandResult {
+    // NOTE: Extension installation without slashes is reserved for the marketplace.
+    if !path.contains('/') && !path.contains('\\') {
+        return Err(anyhow!(
+            "Ambiguous extension URI '{}', use './{0}' instead",
+            path
+        ));
+    }
+
+    let extension_path = PathBuf::from(path);
+    let extension = Extension::try_from(extension_path)?;
 
     extension.install()?;
 


cli/src/commands/extensions/mod.rs
```diff
@@ -73,8 +73,14 @@ pub async fn handle_extensions(matches: &ArgMatches) -> CommandResult {
 
 /// Handle the `extension add` subcommand path.
 /// Add the extension from the specified path.
-async fn handle_add_extension(path: PathBuf) -> CommandResult {
-    let extension = Extension::try_from(path)?;
+async fn handle_add_extension(path: &str) -> CommandResult {
+    // NOTE: Extension installation without slashes is reserved for the marketplace.
+    if !path.contains('/') {
```
maxrake, line 78: #356 includes an example for Windows:
```
phylum extension add C:\dev\foo   # Install from local directory `C:\dev\foo`
```
Will this work for Windows' backslash path separator? I'm still learning Rust, but my guess is that since the type of `path` is a `&str`, it means a literal backslash. Python has the notion of [`os.sep`](https://devdocs.io/python~3.7/library/os#os.sep) to represent
> The character used by the operating system to separate pathname components. This is '/' for POSIX and '\\' for Windows.

Maybe there is something similar for Rust?

maxrake (changes requested):
One question and one `rustfmt` adjustment to make CI happy.

cli/src/commands/extensions/extension.rs
```diff
@@ -114,8 +114,8 @@ impl Extension {
 
     /// Execute an extension subcommand.
     pub async fn run(&self) -> CommandResult {
-        let script_path = self.path.join(&self.manifest.entry_point);
-        let deno = DenoRuntime::new();
+        let entry_point = self.path.join(&self.manifest.entry_point);
+        let mut deno = DenoRuntime::new();
         deno.run(&entry_point.to_string_lossy()).await?;
```
mathew-horner, line 119: How did this code compile before when the variable was named `script_path` and deno needed to be mutable?

cli/src/commands/extensions/mod.rs
mathew-horner, line 78: Are we supporting windows? If so, wouldn't we need to check for backslashes as well?

kylewillmon: Path types in rust support both `/` and `\` on Windows, so it's not enough to use `MAIN_SEPARATOR`.

Officially, I think what you're looking for is [`is_separator`](https://doc.rust-lang.org/std/path/fn.is_separator.html) which returns `true` for both `/` and `\` on Windows.

Practically, it's probably fine to just check for both `/` and `\`.

kylewillmon:
```suggestion
    if !path.contains('/') && !path.contains('\\') {
```

cli/src/commands/extensions/extension.rs
cd-work: It didn't.

maxrake (dismissed):

mathew-horner (dismissed):

cli/src/commands/extensions/mod.rs
```diff
@@ -73,8 +73,17 @@ pub async fn handle_extensions(matches: &ArgMatches) -> CommandResult {
...
```
maxrake, line 78: With the force pushes to this PR, it is not possible to see the history...but didn't this line get changed to:
`if !path.contains('/') && !path.contains('\\') {`
at one point? Did it go back?

Granted, the error message will tell the user the expected format to use, but this error message is _much_ more likely to present itself to Windows users...which may be considered unwelcoming. Perhaps a more nuanced discussion has already taken place with @kylewillmon, but he is out this week and any decisions regarding path separator support should be documented here and on the issue (#356)...which currently only says:

> After discussion, we have decided to require slashes in the argument for local installs. I have updated the title and description accordingly.

...but not which slashes or if only one type of slash will be accepted.

maxrake (commented):
One remaining question to address before approval.

cli/src/commands/extensions/mod.rs
cd-work: Yeah, good catch. This is what happens when you use the GitHub "apply suggestion" feature once every other year and then completely forget about it when you rebase the commit.

PR #423, opened 2022-05-25
Use unit structs for parsers

### LINKED ISSUE #347: Unhelpful `phylum analyze` output for nonexistent lock file

kylewillmon:
This changes our parsers to be strategy objects so that they are a bit
easier to use. This allows us to simplify the LOCKFILE_PARSERS array in
commands/parser.rs and use that array in try_get_packages instead of a
macro.

Closes #347

As @cd-work suggested in #398:

> Maybe having the parsers completely stateless, just taking a &str as parameter would be the way to go.

This patch takes that idea and runs with it.

### DIFF
--- a/cli/src/lockfiles/mod.rs
+++ b/cli/src/lockfiles/mod.rs
@@ -1,5 +1,4 @@
-use std::io;
-use std::marker::Sized;
+use std::fs::read_to_string;
 use std::path::Path;
 
 use phylum_types::types::package::PackageDescriptor;
@@ -20,12 +19,15 @@ pub use ruby::GemLock;
 
 pub type ParseResult = anyhow::Result<Vec<PackageDescriptor>>;
 
-pub trait Parseable {
-    fn new(filename: &Path) -> Result<Self, io::Error>
-    where
-        Self: Sized;
+pub trait Parser {
+    /// Parse from a string
+    fn parse(&self, data: &str) -> ParseResult;
+    /// Indicate the type of packages parsed by this parser
+    fn package_type(&self) -> PackageType;
+}
 
-    fn parse(&self) -> ParseResult;
-
-    fn package_type() -> PackageType;
+/// Parse a file with the given parser.
+pub fn parse_file<T: Parser, P: AsRef<Path>>(parser: T, path: P) -> ParseResult {
+    let data = read_to_string(path)?;
+    parser.parse(&data)
 }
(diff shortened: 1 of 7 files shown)


cli/src/lockfiles/mod.rs
```diff
@@ -20,12 +19,15 @@ pub use ruby::GemLock;
...
-    fn new(filename: &Path) -> Result<Self, io::Error>
-    where
-        Self: Sized;
-
-    fn parse(&self) -> ParseResult;
+pub trait Parse {
+    /// Parse from a string
+    fn parse(&self, data: &str) -> ParseResult;
+    /// Indicate the type of packages parsed by this parser
+    fn package_type(&self) -> PackageType;
```
mathew-horner, line 26: Could this still just be a static function? Not a big deal either way, but I think it makes more sense semantically to not have the reference to self. This function only cares about the implementing type, not the instance.

mathew-horner (dismissed):
Look good 👍

cli/src/lockfiles/mod.rs
```diff
@@ -20,12 +19,15 @@ pub use ruby::GemLock;
...
+}
 
-    fn package_type() -> PackageType;
+/// Parse a file with the given parser.
+pub fn parse_file<T: Parse, P: AsRef<Path>>(parser: T, path: P) -> ParseResult {
+    let data = read_to_string(path)?;
+    parser.parse(&data)
 }
```
cd-work, line 33: This might as well be a method on the trait, right? Like one that is already implemented by default.

cd-work (approved):
Approving so I don't block merge, but I'd suggest considering my suggestion before merge.

cli/src/lockfiles/mod.rs
kylewillmon: Non-methods cannot be dispatched from trait objects (see [object safety](https://doc.rust-lang.org/reference/items/traits.html#object-safety)). So if this was a static function and we have something of type `&dyn Parse`, we would have no way to call this function.

mathew-horner: Right, my mistake

cli/src/lockfiles/mod.rs
kylewillmon: Yep. I think that makes sense

mathew-horner (approved):

PR #425, opened 2022-05-25
Fix stack overflow on Windows

### LINKED ISSUE #424: Fix running CLI on Windows
Currently when trying to execute the Phylum CLI on Windows, a stack overflow will occur with any of the arguments.

cd-work:
This resolves an issue when trying to run the CLI on Windows, which
would cause a stack overflow due to the clap CLI parsing.

To reduce the amount of stack necessary to parse the CLI, all instances
of the `arg!` macro have been replaced with manually built `clap::Arg`
arguments.

Closes #424.

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -30,7 +30,7 @@ pub fn app<'a>() -> clap::Command<'a> {
         .author("Phylum, Inc.")
         .about("Client interface to the Phylum system")
         .args(&[
-            Arg::new("config").short('c').long("config").value_name("FILE").value_hint(ValueHint::FilePath),
+            Arg::new("config").short('c').long("config").value_name("FILE").help("Sets a custom config file").value_hint(ValueHint::FilePath),
             Arg::new("timeout").short('t').long("timeout").value_name("TIMEOUT").help("Set the timeout (in seconds) for requests to the Phylum api"),
             Arg::new("no-check-certificate").long("no-check-certificate").help("Don't validate the server certificate when performing api requests"),
         ])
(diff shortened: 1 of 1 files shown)


cd-work (commented):
Since this was somewhat monotonous work, I've reviewed everything myself to ensure there are no major issues. However since this changes the default of `required` for each parameter to `false`, regardless of the parameter type, I would suggest reviewers to pay special attention to that.

cli/src/app.rs
```diff
@@ -22,40 +22,39 @@ and 'engineering' domains
...
     let mut app = Command::new("phylum")
         .bin_name("phylum")
         .version(VERSION)
         .author("Phylum, Inc.")
         .about("Client interface to the Phylum system")
         .args(&[
-            arg!(-c --config <FILE> "Sets a custom config file").required(false).value_hint(ValueHint::FilePath),
-            arg!(-t --timeout <TIMEOUT> "Set the timeout (in seconds) for requests to the Phylum api").required(false),
-            arg!(--"no-check-certificate" "Don't validate the server certificate when performing api requests"),
+            Arg::new("config").short('c').long("config").value_name("FILE").value_hint(ValueHint::FilePath),
```
mathew-horner, line 33: Apologies if this was done on purpose, which I'm sure it was since it seems a little redundant, but the "Sets a custom config file" help text doesn't seem to have been copied over here.

cli/src/app.rs
```diff
@@ -64,51 +63,51 @@ pub fn app<'a>() -> clap::Command<'a> {
...
+                    Arg::new("group").short('g').long("group").value_name("group_name").help("Group to list projects for"),
                 ])
                 .aliases(&["projects"])
                 .subcommand(
                     Command::new("create")
                         .about("Create a new project")
                         .args(&[
-                            arg!(<name> "Name of the project"),
-                            arg!(-g --group <group_name> "Group which will be the owner of the project").required(false),
+                            Arg::new("name").value_name("name").help("Name of the project"),
```
mathew-horner, line 74: Does this one need `.required(true)`?

cli/src/app.rs
cd-work: Good catch. The intention was to make no functional changes so I just missed it.

cli/src/app.rs
cd-work: It does.

mathew-horner (approved):
Looks good 👍

Issue #426, opened 2022-05-25
Cannot use deno on x86_64-unknown-linux-musl


kylewillmon:
As seen in [this CI build](https://github.com/phylum-dev/cli/runs/6579718604?check_suite_focus=true), building deno fails for `x86_64-unknown-linux-musl` targets. This is because [rusty_v8](https://github.com/denoland/rusty_v8) does not support this target (see denoland/rusty_v8#49).

PR #427, opened 2022-05-25
Only use deno with the extensions feature

kylewillmon:
Unfortunately, it looks like the deno crates we are using fail to build
on x86_64-unknown-linux-musl... That is now recorded in #426, but in the
meantime, this change avoids including deno in our default builds

While making this change, I decided to use the `dep:` syntax for features which was introduced in Rust 1.60. I have updated our MSRV and docs accordingly.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,14 +1,15 @@
 [package]
 name = "phylum-cli"
 version = "3.5.0"
-authors = ["Eric Freitag <eric@phylum.io>"]
-edition = "2018"
+authors = ["Phylum, Inc. <engineering@phylum.io>"]
+edition = "2021"
+rust-version = "1.60"
 
 [features]
 default = ["selfmanage"]
 phylum-online = []
 selfmanage = []
-extensions = []
+extensions = ["dep:deno_ast", "dep:deno_core"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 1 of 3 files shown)


kylewillmon:
This should fix the Release workflow. To check this, [here is a manually kicked off run](https://github.com/phylum-dev/cli/actions/runs/2387384796)

cli/Cargo.toml
```diff
@@ -1,14 +1,15 @@
 [package]
 name = "phylum-cli"
 version = "3.5.0"
-authors = ["Eric Freitag <eric@phylum.io>"]
-edition = "2018"
+authors = ["Phylum, Inc. <engineering@phylum.io>"]
```
maxrake, line 4: 👍 
Nothing against Eric, but this is the correct move...especially for a public repo.

cli/Cargo.toml
```diff
@@ -1,14 +1,15 @@
...
+edition = "2021"
+rust-version = "1.60"
```
maxrake, line 6: Does this change warrant a reminder to devs to rustup to this version?

docs/building.md
```diff
@@ -4,7 +4,7 @@ category: 6255e67693d5200013b1fa41
 hidden: false
 ---
 
-Phylum is written in Rust, so you'll need a recent Rust installation to build it (we tested with v1.58.0). [Install Rust](https://www.rust-lang.org/tools/install)
+Phylum is written in Rust, so you'll need a recent Rust installation to build it (we tested with v1.61.0). [Install Rust](https://www.rust-lang.org/tools/install)
```
maxrake, line 7: v1.61.0 is slightly newer than the v1.60 specified in `cli/cargo.toml`. The guess is that this is intentional (testing was done on a newer version than the minimum one *required*), but pointing it out anyway.

maxrake (approved):
Sorry for the delay...broken builds are serious...was out for a bit...
Some questions are posed but nothing important enough to hold things up.
LGTM.

docs/building.md
kylewillmon: Yep. 1.61 is what I've got installed, so it's what I tested with.

cli/Cargo.toml
kylewillmon: The hope of adding it to Cargo.toml is that devs will get a helpful error message if they have not updated

PR #433, opened 2022-05-30
Refactor update check

### LINKED ISSUE #428: Refactor check for updates
We could probably move this into the previous conditional, right? At least I see no way for `check_for_updates` to be true unless `subcommand_matches("update")` is none. I think that would make things a little less convoluted when we're refactoring stuff already anyway.

I'd suggest something like this:

```rust
    if matches.subcommand_matches("update").is_none() {
        let start = SystemTime::now();
        let now = start
            .duration_since(UNIX_EPOCH)
            .expect("Time went backwards")
            .as_secs() as usize;

        let mut check_for_updates = false;
        let check_for_updates = config.last_update.map_or(true, |last_update| {
            const SECS_IN_DAY: usize = 24 * 60 * 60;
            now - last_update > SECS_IN_DAY
        });

        if check_for_updates {
            log::debug!("Checking for updates...");

            config.last_update = Some(now);
            save_config(&config_path, &config)
                .unwrap_or_else(|e| log::error!("Failed to save config: {}", e));

            if update::neds_update(false).await {
                print_update_message();
            }
        }
    }
```

I might be missing some important detail though, just something I noticed from the portion of this PR that shows up on the GitHub review.

_Originally posted by @cd-work in https://github.com/phylum-dev/cli/pull/415#discussion_r881984953_

cd-work:
This is a small refactor to the update check when the user is not
explicitly asking for an update.

It moves around some code to more accurately represent the intention of
what the code is suppose to do, while no functional changes are
introduced.

Closes #428.

### DIFF
--- a/cli/src/bin/phylum.rs
+++ b/cli/src/bin/phylum.rs
@@ -1,5 +1,5 @@
 use std::path::{Path, PathBuf};
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::time::UNIX_EPOCH;
 
 use anyhow::{anyhow, Context, Result};
 use clap::ArgMatches;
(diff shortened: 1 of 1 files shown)


maxrake (dismissed):

maxrake (approved):

mathew-horner (approved):
Just left a few non-critical comments. Feel free to merge regardless!

PR #437, opened 2022-05-31
Unify XDG directory methods

### LINKED ISSUE #414: Uniform the style of the `config_dir` and `data_dir` functions
`data_dir` is a function that resolves the XDG data directory and appears in the [extensions](https://github.com/phylum-dev/cli/blob/andrea/extensions-api/cli/src/config.rs#L193) code. As it is pretty similar to the `config_dir` function, we want to refactor the latter to look more similar to the former, and also correct its call sites to take that into account.

cd-work:
This merges together the two separate `data_dir` methods and streamlines
all the other XDG-related utility functions.

Since the only error that is possible with these methods is the very
uncommon scenario where the user's home directory does not exist, error
handling of this scenario has been made less explicit and prominent.

Due to the current order of operations, the errors should still be the
same. However this would make it more likely that in the future an
uninstall is only partially attempted due to a missing home directory.

Closes #414.

### DIFF
--- a/cli/src/commands/uninstall.rs
+++ b/cli/src/commands/uninstall.rs
@@ -1,31 +1,29 @@
 //! Phylum CLI removal.
 
+use std::fs;
 use std::io::Write;
 use std::path::{Path, PathBuf};
-use std::{env, fs};
 
 use anyhow::{anyhow, Result};
 use clap::ArgMatches;
 use tempfile::NamedTempFile;
 
 use crate::commands::{CommandResult, ExitCode};
-use crate::{config, print_user_success, print_user_warning};
+use crate::{dirs, print_user_success, print_user_warning};
 
 /// Handle the `uninstall` subcommand.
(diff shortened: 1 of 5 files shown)


cli/src/commands/uninstall.rs
```diff
@@ -1,53 +1,51 @@
...
 
-    if let Err(err) = Shell::Zsh.cleanup(&home_dir) {
+    if let Err(err) = Shell::Zsh.cleanup() {
         print_user_warning!("ZSH config error: {}", err);
     }
 
     Ok(ExitCode::Ok.into())
 }
 
 /// Remove the entire `~/.config/phylum` directory.
```
maxrake, line 33: Not due to the changes in this review...just noticed now...but this comment is too specific. Instead, it could be:

`/// Remove the entire phylum config directory.`

maxrake (approved):
There was only one superficial suggestion for a comment change...but LGTM.

maxrake (approved):

PR #441, opened 2022-06-02
Add alpine musl build test

cd-work:


entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+export V8_FROM_SOURCE=yes
+export GN="$(pwd)/gn/out/gn"
+export CLANG_BASE_PATH=/usr
+export GN_ARGS='use_custom_libcxx=false use_lld=false v8_enable_backtrace=false v8_enable_debugging_features=false'
+
+# Bulid GN
+if [ ! -d "./gn" ]; then
+    git clone https://gn.googlesource.com/gn
+    cd gn
+    python3 build/gen.py
```
maxrake, line 26:
```suggestion
    python build/gen.py
```
`python` was created as a link on line 14 above.
Another option is to remove that link and keep `python3` here.

maxrake (commented):
Submitting w/o approval since this is still draft, but here are some general comments:
* Is there an issue for this work? More background on the reason for this change would be helpful.
* There are some `shellcheck` findings to fix
* Using the `Dockerfile` file name and placing the file in the root of the repo might lead people to believe there is a Docker image available for the Phylum CLI
  * Since this is a test file used to confirm alpine musl builds, it might be more appropriate to rename the file (e.g., `alpine_musl_build.dockerfile`), move it to a subdirectory (e.g., `cli/tests/`), or both
* Similarly, `entrypoint.sh` is a misleading name for a file in the root of the repo
  * It might go better in the `scripts` directory and/or be renamed

cd-work:
> Is there an issue for this work? More background on the reason for this change would be helpful.
https://github.com/phylum-dev/cli/issues/426

> Submitting w/o approval since this is still draft, but here are some general comments

This isn't even close to being something I would suggest to merge. The sole purpose of this PR was to test how long it would take to build v8 in CI.

entrypoint.sh
cd-work: It is not another option, because other parts of the build rely on the python symlink to be in place.

entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+apk add binutils-gold # Required only for 3.15?
+ln -s /usr/bin/python3 /usr/bin/python
+
+# Env
+export V8_FROM_SOURCE=yes
+export GN="$(pwd)/gn/out/gn"
+export CLANG_BASE_PATH=/usr
+export GN_ARGS='use_custom_libcxx=false use_lld=false v8_enable_backtrace=false v8_enable_debugging_features=false'
+
+# Bulid GN
```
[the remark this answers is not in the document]
cd-work: Because `GN` is required to build `v8`, which is required to build `rusty_v8`, which is required to build `deno`.

cd-work:
Just to keep everyone in the loop:
 - I've submitted [a PR to docker-rust], since the latest rusty_v8 version requires llvm13
 - I've submitted [a PR to rusty_v8], to get some feedback on upstreaming this

Since the extensions work is currently not planned to be released yet, I do not think it's necessary to rush a resolution in this PR. I think we can definitely wait for a `rust:alpine13.6` docker container to be available since I do not expect much friction with that PR.

I'd assume the rusty_v8 PR is probably going to take a bit longer, so while I think it makes sense to wait for feedback it will probably be sensible to add a temporary musl release build step once extensions are ready. It should be simple to do this once we have concrete plans to release extensions.

[a PR to docker-rust]: https://github.com/rust-lang/docker-rust/pull/104
[a PR to rusty_v8]: https://github.com/denoland/rusty_v8/pull/999

PR #453, opened 2022-06-09
Migrate from `deno_core` to `deno_runtime`

### LINKED ISSUE #452: Provide extensions standard library
# Overview

By default, deno_core only provides an embedded JavaScript runtime. This allows executing JavaScript code but provides no other utility functions besides the bare minimum. To ensure people can use this runtime to write their applications, the deno CLI also provides various interfaces like the filesystem and network, similar to what nodejs/browser runtimes provide.

To make sure our extensions can do meaningful work, we should provide some of this functionality too. Most of which can likely just be used from the deno CLI project directly, so no extra development effort is necessary.

Integration with the deno runtime can be seen in the CLI [here](https://github.com/denoland/deno/blob/ba13b8e2a9efff13e4a7c0659bb2ed184cc4e683/runtime/worker.rs#L116-L174). Most of the different extensions are provided as separate crates that can be included on demand.

Additionally the [Deno standard library](https://deno.land/std) also provides bindings to this functionality, allowing for easy use without having to go through Deno abstractions. Providing this by default might be a nice thing to have, though loading these modules over the network is a possibility as long as the deno extensions necessary were added to the runtime.

Much of this functionality is not immediately necessary for our extensions, so the best approach is likely to incrementally add what is necessary, starting with basic functionality like filesystem/client networking.

# Acceptance Criteria

- [ ] Load runtime extensions for at least file read/write, HTTP requests and console I/O
- [ ] Write examples/tests using these APIs with web/local deno STD
- [ ] Ensure binary size increase is within reasonable bounds

cd-work:
The deno project provides a couple of different binaries and libraries,
the main entry point being its CLI. This CLI depends on `deno_runtime`
which in turn depends on `deno_core`.

Previously we were using `deno_core`, the lowest level library for
embedding deno. However this does not come with any utility functions
and requires including all of these manually.

This patch switches to the `deno_runtime` library instead, which is one
level above `deno_core`. While it still allows us all necessary
flexibility of `deno_core`, we now automatically get access to the Deno
API (https://doc.deno.land/deno/stable).

This change does not increase the output size or compilation time.

Closes #452.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2804,6 +2804,7 @@ dependencies = [
  "cidr",
  "clap",
  "deno_ast",
+ "deno_core",
  "deno_runtime",
  "dialoguer",
  "env_logger",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,10 +13,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "abort_on_panic"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "955f37ac58af2416bac687c8ab66a4ccba282229bd7422a28d2281a5e66a6116"
+
+[[package]]
 name = "adler"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
+name = "aead"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b613b8e1e3cf911a086f53f03bf286f52fd7a7258e4fa606f0ef220d39d8877"
+dependencies = [
+ "generic-array",
+]
 
 [[package]]
 name = "aes"
@@ -28,6 +43,30 @@ dependencies = [
  "cipher",
  "cpufeatures",
  "opaque-debug",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.9.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "df5f85a83a7d8b0442b6aa7b504b8212c1733da07b98aae43d4bc21b2cb3cdf6"
+dependencies = [
+ "aead",
+ "aes",
+ "cipher",
+ "ctr",
+ "ghash",
+ "subtle",
+]
+
+[[package]]
+name = "aes-kw"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "da2cfe3156f99bc2978b812bcc9c98c6850810e2c0cd48366abb7163a1beeed0"
+dependencies = [
+ "aes",
+ "generic-array",
 ]
 
 [[package]]
@@ -48,6 +87,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
 dependencies = [
  "memchr",
+]
+
+[[package]]
+name = "alloc-no-stdlib"
+version = "2.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "35ef4730490ad1c4eae5c4325b2a95f521d023e5c885853ff7aca0a6a1631db3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "697ed7edc0f1711de49ce108c541623a0af97c6c60b2f6e2b65229847ac843c2"
+dependencies = [
+ "alloc-no-stdlib",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
```
kylewillmon, line 2: A ton of new packages were added to `Cargo.lock`. I am curious if that is mostly because of the resolver change or because of the switch to `deno_runtime`.

cli/src/deno.rs
```diff
@@ -1,55 +1,71 @@
...
+        format_js_error_fn: Default::default(),
+        get_error_class_fn: Default::default(),
+        origin_storage_dir: Default::default(),
+        broadcast_channel: Default::default(),
+        source_map_getter: Default::default(),
+        root_cert_store: Default::default(),
+        blob_store: Default::default(),
+        extensions: Default::default(),
+        stdio: Default::default(),
+        seed: Default::default(),
```
kylewillmon, line 57: I think the use `Default::default()` makes this code harder to read... Here are a few examples to explain:

* `should_break_on_first_statement` is a `bool`, so it gets the default value of `false`. But `should_break_on_first_statment: false` would be easier to read.
* `seed` is `Option<u64>`, so it gets a value of `None` which makes sense because we don't care about the seed. But if `seed` were just `u64`, then that would be a seed of `0` which would be bad. Writing `seed: None` makes it clear without having to look at the type.

If `WorkerOptions` implemented `Default`, then we could use `..Default::default()` and trust that the developers thought about the default for each field when implementing that trait. Without that, we have to do the thinking ourselves. And if we have to think about the actual values, we might as well write out the values (i.e., mostly `None`).

cli/src/deno.rs
```diff
@@ -1,55 +1,71 @@
...
+        .unwrap_or(1);
+    let bootstrap = BootstrapOptions {
+        cpu_count,
+        runtime_version: env!("CARGO_PKG_VERSION").into(),
+        user_agent: "phylum-cli/extension".into(),
+        no_color: !colors::use_color(),
+        is_tty: colors::is_tty(),
+        enable_testing_features: Default::default(),
+        debug_flag: Default::default(),
+        ts_version: Default::default(),
```
kylewillmon, line 32: > Sets `Deno.version.typescript` in JS runtime.

Setting this to an empty string doesn't seem right... But I guess we don't have anything better to set it to...

kylewillmon (approved):

Cargo.toml
cd-work: It's because of the change to `deno_runtime`. The resolver change is just necessary to make things build.

Cargo.lock
```diff
@@ -3476,9 +5297,24 @@ dependencies = [
 
 [[package]]
 name = "zeroize"
-version = "1.5.5"
+version = "1.4.3"
```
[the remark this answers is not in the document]
cd-work: It's an upgrade if you look at it upside down.

cli/src/deno.rs
cd-work: I'd argue using `Default::default()` here makes things **significantly** easier to understand. The goal is to make my intentions here clear and obvious, none of these fields are set to any specific value. I just wanted to set these values to **any** value that wouldn't hinder us. I didn't set `should_break_on_first_statement` to `false` with this, I just tried not setting it to anything else and see if it does the trick for us.

> seed is Option<u64>, so it gets a value of None which makes sense because we don't care about the seed. But if seed were just u64, then that would be a seed of 0 which would be bad. Writing seed: None makes it clear without having to look at the type.

I mean a seed of 0 is still a valid seed and it would be just as valid as any other fixed seed we might set this to. Again, the decision was to **NOT** worry about the seed and have it set to just some random value and move on. I think default here exactly proves the point that I did not worry about setting any seed, while setting the seed to `None` might look like I intentionally set it to this value.

> Without that, we have to do the thinking ourselves. And if we have to think about the actual values, we might as well write out the values (i.e., mostly None).

The writer has to do the thinking, but the reader doesn't. And if we add explicit options for everything the reader has to do the thinking.

If there are specific options you want to assign certain values, like `seed` for example, we can do that. But I don't think it provides a value to set fields we do not care about to a specific value, since that misleads the reader more than it helps.

cli/src/deno.rs
```diff
@@ -1,55 +1,71 @@
...
+        location: Default::default(),
+        unstable: Default::default(),
+        args: Default::default(),
```
mathew-horner, line 35: Can you not do `..Default::default()` instead?

mathew-horner (approved):

cli/src/deno.rs
cd-work: Nope, because the struct doesn't implement `Default`. And `..Default::default()` uses the struct's default implementation, not the default implementation of each field.

cd-work:
Waiting for @andreaphylum to merge the API part before this, since rebasing this PR should be much simpler.

cli/src/deno.rs
mathew-horner: Gotcha, unfortunate 👍

kylewillmon (dismissed):

kylewillmon (dismissed):

kylewillmon (approved):

PR #455, opened 2022-06-09
use url crate to build URLs

### LINKED ISSUE #454: Package names in API requests need to be percent encoded
# Overview
Sometimes a package name will contain characters that are not allowed in a URL path component, for example `@angular/material`. When making API calls, the slash needs to be escaped for the API call to arrive at the correct route.

# How To Reproduce

```
❯ phylum package -j @angular/material 13.2.2 | jq .name,.complete
"@angular~material"
false
```

# Expected Behavior

```
❯ phylum package -j @angular%2Fmaterial 13.2.2 | jq .name,.complete
"@angular/material"
true
```

(This example with the name already percent encoded on the command line should only work until this bug is fixed. It is not a sustainable workaround.)

# Additional Context

It looks like this used to work, at least in some cases, with an older version of the API. The current CLI code is replacing slashes with tildes and the old implementation of the API call used to replace tildes with slashes. Percent encoding is the standard, it's already working in the API, and it's what the UI is doing, so it's what CLI should do as well.

matt-phylum:
This PR changes all the functions in `api::endpoints` to construct the endpoint address using the url crate. The url crate takes care of escaping the path components for us.

Using the url crate here moves the time when URL parse errors occur to earlier in the process of making a request. Previously, reqwest was doing this for us and combining any URL parse errors into the reqwest error, but now it is a distinct error type. Now we can pass the already parsed Url into reqwest.

We also handle the case where the base address in the config file contains a query string or hash, silently dropping them instead of creating invalid URLs (which would be different invalid URLs than previously when we used string concatenation).

I added at least one test for every path function. I don't know if we need all of them, but it helped me check for regressions.

closes #454 

## Checklist

### DIFF
--- a/cli/src/api/endpoints.rs
+++ b/cli/src/api/endpoints.rs
@@ -1,84 +1,299 @@
 /// API endpoint paths
+use thiserror::Error as ThisError;
+use url::{ParseError, Url};
+
 use super::{JobId, PackageDescriptor};
 
-const API_PATH: &str = "api/v0";
+const API_PATH: &str = "api/v0/";
+
+// This wrapper provides important context to the user when their configuration has a bad URL.
+// Without it, the message can be something like "Error creating client" caused by "empty host".
+#[derive(Debug, ThisError)]
+#[error("invalid API URL")]
+pub struct BaseUriError(#[from] pub ParseError);
 
 /// PUT /job
-pub fn put_submit_package(api_uri: &str) -> String {
-    format!("{api_uri}/{API_PATH}/job")
+pub fn put_submit_package(api_uri: &str) -> Result<Url, BaseUriError> {
+    Ok(get_api_path(api_uri)?.join("job")?)
 }
 
 /// GET /health
-pub fn get_ping(api_uri: &str) -> String {
-    format!("{api_uri}/{API_PATH}/health")
+pub fn get_ping(api_uri: &str) -> Result<Url, BaseUriError> {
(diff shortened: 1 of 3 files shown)


cli/src/api/endpoints.rs
```diff
@@ -1,84 +1,299 @@
...
+    Ok(get_api_path(api_uri)?.join(".well-known/openid-configuration")?)
+}
+
+fn parse_base_url(api_uri: &str) -> Result<Url, BaseUriError> {
+    let mut url = Url::parse(api_uri)?;
+
+    // Ensure the path can be a base and ends with a slash so it can be safely joined to.
+    // If we don't do this, https://example.com/a and https://example.com/a/ are different.
+    url.path_segments_mut()
+        .map_err(|_| ParseError::RelativeUrlWithCannotBeABaseBase)?
```
mathew-horner, line 129: `ParseError::RelativeUrlWithCannotBeABaseBase` what does this error variant mean?

mathew-horner (approved):
👍

cli/src/api/endpoints.rs
matt-phylum: It's if the URL is something like `data:text…` or `mailto:user@example.com` etc. https://docs.rs/url/2.2.2/url/struct.Url.html#method.cannot_be_a_base

mathew-horner (dismissed):

mathew-horner (approved):

PR #456, opened 2022-06-09
Use phylum-bot to create cargo update PRs

### LINKED ISSUE #432: GitHub Actions dependency updates should automatically run CI tests
Currently when the automatic PR for minor/patch dependency updates is sent, it will not automatically start running tests and because of that it will be impossible to merge the PR. As a resolution it is necessary to manually push to the branch again to force a build.

I'm not entirely certain if this is due to missing signatures or permissions that need to be granted to the specific bot user account, but it should not be necessary to push to the commit if there were no changes on master since its creation.

kylewillmon:
Using a PAT for @phylum-bot should allow the created PRs to kick off workflows.

Fix #432

### DIFF
--- a/.github/workflows/cargo-update.yml
+++ b/.github/workflows/cargo-update.yml
@@ -33,10 +33,10 @@ jobs:
           github-token: ${{ secrets.GH_RELEASE_PAT }}
           script: |
             github.rest.pulls.create({
-              owner: "phylum-dev",
-              repo: "cli",
+              owner: context.repo.owner,
+              repo: context.repo.repo,
               head: "auto-cargo-update",
-              base: "main",
+              base: context.ref,
               title: "build: Bump Cargo.lock dependencies",
               body: "Bump dependencies in Cargo.lock for all SemVer-compatible updates.",
             });
--- a/.github/workflows/cargo-update.yml
+++ b/.github/workflows/cargo-update.yml
@@ -17,12 +17,26 @@ jobs:
       - name: Cargo update
         run: cargo update
 
+      - name: Commit changes
+        id: commit
+        continue-on-error: true
+        run: |
+          git config user.name 'Phylum Bot'
+          git config user.email 'phylum-bot@users.noreply.github.com'
+          git commit -a -m "build: Bump Cargo.lock dependencies"
+          git push --force origin HEAD:auto-cargo-update
+
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v4
+        if: ${{ steps.commit.outcome == 'success' }}
+        uses: actions/github-script@v6
         with:
-          token: ${{ secrets.GH_RELEASE_PAT }}
-          commit-message: "build: Bump Cargo.lock dependencies"
-          title: "build: Bump Cargo.lock dependencies"
-          body: |
-            Bump dependencies in Cargo.lock for all SemVer-compatible updates.
-          branch: auto-cargo-update
+          github-token: ${{ secrets.GH_RELEASE_PAT }}
+          script: |
+            github.rest.pulls.create({
+              owner: "phylum-dev",
+              repo: "cli",
+              head: "auto-cargo-update",
+              base: "main",
+              title: "build: Bump Cargo.lock dependencies",
+              body: "Bump dependencies in Cargo.lock for all SemVer-compatible updates.",
+            });
--- a/.github/workflows/cargo-update.yml
+++ b/.github/workflows/cargo-update.yml
@@ -7,10 +7,6 @@ on:
   schedule:
     - cron: '30 5 * * 1'
 
-permissions:
-  contents: write
-  pull-requests: write
-
 jobs:
   cargo-update:
     runs-on: ubuntu-latest
@@ -24,7 +20,7 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v4
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.GH_RELEASE_PAT }}
           commit-message: "build: Bump Cargo.lock dependencies"
           title: "build: Bump Cargo.lock dependencies"
           body: |


kylewillmon:
Confirmed functionality by triggering the workflow [on this branch](https://github.com/phylum-dev/cli/actions/runs/2471334597), which opened #457 and kicked off [more workflows](https://github.com/phylum-dev/cli/actions/runs/2471339037)

.github/workflows/cargo-update.yml
```diff
@@ -24,7 +20,7 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v4
```
maxrake, line 21: It might be time to replace this externally controlled action with something else. Options include, but may not be limited to:
* Fork the `peter-evans/create-pull-request` repo
* Reference the `peter-evans/create-pull-request` repo by SHA instead of `v4` so it can't change out from under the tag
* Use the [`actions/github-script`](https://github.com/actions/github-script) instead
  * It's still an action, but at least it's maintained by GitHub
  * The [octokit documentation for "Pulls -> Create"](https://octokit.github.io/rest.js/v18#pulls-create) is the source of what would go into that action's "asynchronous function call"
* Use the [GitHub Pulls REST API](https://docs.github.com/en/rest/pulls/pulls#create-a-pull-request) directly
  * Examples there use `curl` and the `gh` CLI, both of which are available [on the runners](https://github.com/actions/virtual-environments/blob/ubuntu20/20220605.1/images/linux/Ubuntu2004-Readme.md)

maxrake (approved):
It works...but using the `peter-evans/create-pull-request` action with a token other than the `GITHUB_TOKEN` (which only lives as long as the workflow does) feels risky. Further, re-using a long-lived personal access token (PAT) with an externally controlled GitHub action could be considered bad security practice.

.github/workflows/cargo-update.yml
kylewillmon: Replaced with `actions/github-script`. Thanks for the recommendation.

kylewillmon:
I've updated the code to use a simple script instead of the third party action. I tested it by manually running [this action](https://github.com/phylum-dev/cli/actions/runs/2477109683).

.github/workflows/cargo-update.yml
```diff
@@ -21,12 +17,26 @@ jobs:
...
-          body: |
-            Bump dependencies in Cargo.lock for all SemVer-compatible updates.
-          branch: auto-cargo-update
+          github-token: ${{ secrets.GH_RELEASE_PAT }}
+          script: |
+            github.rest.pulls.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              head: "auto-cargo-update",
+              base: context.ref,
```
kylewillmon, line 39: For simplicity of the script, I use the ref from this action as the base branch. That means that running this workflow from a branch gets you a PR to that branch (as can be seen in #460)

cd-work (approved):

.github/workflows/cargo-update.yml
```diff
@@ -21,12 +17,26 @@ jobs:
       - name: Cargo update
         run: cargo update
 
+      - name: Commit changes
+        id: commit
+        continue-on-error: true
+        run: |
+          git config user.name 'Phylum Bot'
+          git config user.email 'phylum-bot@users.noreply.github.com'
```
maxrake, line 25: Screen Shot 2022-06-13 at 9 14 57 AM

To get the verified commits for `phylum-bot`, the `phylum-dev/phylum-ci` repository had to:
* Set the user name/email to `phylum-bot <69485888+phylum-bot@users.noreply.github.com>`
* Import the matching GPG key for signing commits:

```yml
      - name: Import GPG key for bot account
        uses: crazy-max/ghaction-import-gpg@v4
        with:
          gpg_private_key: ${{ secrets.PHYLUM_BOT_GPG_PRIVATE_KEY }}
          passphrase: ${{ secrets.PHYLUM_BOT_GPG_PASSPHRASE }}
          git_user_signingkey: true
          git_commit_gpgsign: true
          git_tag_gpgsign: true
```
It may not be absolutely necessary to sign these commits. However, if the signing option is pursued, it is worth also considering how to properly protect the secrets. For instance, in the `phylum-ci` repo, the bot related secrets are protected with the `Production` environment...which has limited access.

maxrake (approved):
Looks good...just one remaining question about whether signed commits are necessary and, if so, how the GPG keys could/should be protected and imported in this context.

.github/workflows/cargo-update.yml
```diff
@@ -21,12 +17,26 @@ jobs:
...
+          github-token: ${{ secrets.GH_RELEASE_PAT }}
+          script: |
+            github.rest.pulls.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              head: "auto-cargo-update",
+              base: context.ref,
+              title: "build: Bump Cargo.lock dependencies",
+              body: "Bump dependencies in Cargo.lock for all SemVer-compatible updates.",
+            });
```
maxrake, line 42:
```suggestion
            })
```
works with the semi-colon, but not necessary.

.github/workflows/cargo-update.yml
kylewillmon: It's not at all necessary to sign these commits. They can and should be signed when the merge occurs. And just like when this was first discussed ([here](https://github.com/phylum-dev/cli/pull/339#issuecomment-1116305063)), we still do not enforce a signature requirement on any branch.

On a philosophical level, this can be treated just like a PR coming in from a forked repository. There is no need to assign any trust to @phylum-bot because an authorized human will review and merge the PR.

.github/workflows/cargo-update.yml
cd-work: I'd encourage using a semicolon, even if not necessary. That way adding a new line doesn't require changing the previous one.

maxrake: I'll preface by saying that I didn't actually *try* any of this (with or without semicolons) so there is a chance this is just bad advice...but the comment originated from reviewing the examples on the [`actions/github-script` documentation](https://github.com/actions/github-script). In all but one case there, semicolons are not used for multiline examples.

That said, this conversation will be resolved as it doesn't cause any trouble to use the semicolon and there are no other changes to make in this review.

PR #462, opened 2022-06-13
Added build script as workaround for Window debug builds

andreaphylum:
The build script added in this PR exists as a workaround for [this issue](https://github.com/rust-lang/rust/issues/34283).

This issue is present under Windows in debug builds.

The CLI has a long `match` statement in the body of a function. LLVM makes it so that the stack space required by the `match` statement is proportional to the sum of the stack space requirements for each branch, rather than to the maximum for all of the branches (which is what happens on higher optimization levels and on different targets). As a result, Windows debug
builds have too high a stack utilization, and will result in a stack overflow when run. By expanding the available stack space at link time, we prevent this from happening. We want to make sure this variation is only applied to the affected target(s) as there are no advantages to a larger stack space other than preventing this issue in this specific configuration.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -4,7 +4,6 @@ version = "3.5.0"
 authors = ["Phylum, Inc. <engineering@phylum.io>"]
 edition = "2021"
 rust-version = "1.60"
-build = "build.rs"
 
 [features]
 default = ["selfmanage"]
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -4,6 +4,7 @@ version = "3.5.0"
 authors = ["Phylum, Inc. <engineering@phylum.io>"]
 edition = "2021"
 rust-version = "1.60"
+build = "build.rs"
 
 [features]
 default = ["selfmanage"]
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -27,4 +27,3 @@ fn main() {
         println!("cargo:rustc-link-arg=/STACK:0x800000");
     }
 }
-
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -8,16 +8,15 @@ use std::env;
 // The CLI has a long `match` statement in the body of a function. LLVM makes it
 // so that the stack space required by the `match` statement is proportional to
 // the sum of the stack space requirements for each branch, rather than to the
-// maximum for all of the branches (which is what happens on higher optimization
+// maximum of all of the branches (which is what happens on higher optimization
 // levels and on different targets).
 //
-// As a result, Windows debug builds have too high a stack utilization, and will
-// result in a stack overflow when run. By expanding the available stack space
-// at link time, we prevent this from happening.
+// As a result, Windows debug builds will result in a stack overflow when run,
+// because of too high a stack utilization. We can prevent this by expanding the
+// available stack space at link time.
 //
-// We want to make sure this variation is only applied to the affected target(s)
-// as there are no advantages to a larger stack space other than preventing this
-// issue in this specific configuration.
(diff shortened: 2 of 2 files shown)


cli/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "3.5.0"
 authors = ["Phylum, Inc. <engineering@phylum.io>"]
 edition = "2021"
 rust-version = "1.60"
+build = "build.rs"
```
cd-work, line 7: This is not required, right? The file should be picked up by default.

cli/build.rs
```diff
@@ -0,0 +1,27 @@
...
+// to the maximum for all of the branches (which is what happens on higher
+// optimization levels and on different targets). As a result, Windows debug
+// builds have too high a stack utilization, and will result in a stack overflow
+// when run. By expanding the available stack space at link time, we prevent
+// this from happening. We want to make sure this variation is only applied to
+// the affected target(s) as there are no advantages to a larger stack space
+// other than preventing this issue in this specific configuration.
+
+fn main() {
+    let os = env::var("CARGO_CFG_TARGET_OS").unwrap();
```
cd-work, line 20: I've mentioned this on slack, but I don't see why we wouldn't use cfg for this?

cd-work (changes requested):

cli/Cargo.toml
cd-work: Why? It serves no purpose as far as I can tell and I rarely see it added.

cli/build.rs
andreaphylum: For the same reason we may not use `env!` -- `cfg!` is also a compile time macro and the inputs to the build scripts are explicitly stated to be environment variables. We want to make sure it is the build script to process those inputs, not the process that's in charge of building the build script.

cli/Cargo.toml
andreaphylum: To be explicit about the fact that a build script is, in fact, present and required; this way, if the file gets inadvertently deleted, Cargo will output an error:

```
> cargo check
   Compiling phylum-cli v3.5.0 (F:\Phylum\repos\cli\cli)
error: couldn't read cli\build.rs: The system cannot find the file specified. (os error 2)
```

cli/build.rs
```diff
@@ -0,0 +1,27 @@
...
+// The CLI has a long `match` statement in the body of a function. LLVM makes
+// it so that the stack space required by the `match` statement is proportional
+// to the sum of the stack space requirements for each branch, rather than
+// to the maximum for all of the branches (which is what happens on higher
+// optimization levels and on different targets). As a result, Windows debug
+// builds have too high a stack utilization, and will result in a stack overflow
+// when run. By expanding the available stack space at link time, we prevent
+// this from happening. We want to make sure this variation is only applied to
+// the affected target(s) as there are no advantages to a larger stack space
+// other than preventing this issue in this specific configuration.
```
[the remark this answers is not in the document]
cd-work: > I think it is better to be a bit more thorough at the expense of verbosity

I can take a look, but generally I think being overly verbose is often misleading and confusing, rather than helpful. It's certainly an art form but stating things multiple times in different variations doesn't help with understanding. Choosing the most concise phrasing which contains all relevant information is usually best for understanding in my experience.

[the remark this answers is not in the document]
andreaphylum: Agreed -- after all, having such a clear and well-communicating documentation is one of the reasons I like Rust so much in the first place. Considering the topic at hand is quite complex, it's a good idea to find a formula that makes everything both easy to understand and complete.

cd-work (changes requested):

cd-work (approved):

cd-work (approved):

PR #466, opened 2022-06-13
feat: coordinate phylum-ci Docker image releases with new CLI releases

maxrake:
This PR adds a repository dispatch event trigger to the release workflow. It triggers the Docker workflow in the `phylum-dev/phylum-ci` repo to build and push Docker images containing the latest `phylum` Python package and the Phylum CLI version from the tag that caused the Release workflow to run. A `curl` command that POSTs to the `dispatches` endpoint with the expected payload is used instead of the [`actions/github-script` action](https://github.com/actions/github-script) because this is a simple trigger. The trigger is further limited so that it will not fire for pre-releases of the CLI.

## Checklist

  - https://github.com/phylum-dev/phylum-ci/issues/50
  - It won't be possible to truly test this change until the corresponding work in `phylum-dev/phylum-ci` is merged to `main`
  - https://github.com/phylum-dev/phylum-ci/pull/63

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -156,6 +156,19 @@ jobs:
             phylum-*.zip
             phylum-*.zip.minisig
 
+      - name: Trigger phylum-ci Docker image creation
+        # Don't trigger for pre-releases
+        if: ! contains(github.ref, 'rc')
+        # Reference: https://docs.github.com/en/rest/repos/repos#create-a-repository-dispatch-event
+        run: >
+          curl
+          -X POST
+          --fail-with-body
+          -H "Accept: application/vnd.github.v3+json"
+          -H "Authorization: token ${{ secrets.GH_RELEASE_PAT }}"
+          -d '{"event_type":"build-push-docker-images","client_payload":{"CLI_version":"$GITHUB_REF_NAME"}}'
+          https://api.github.com/repos/phylum-dev/phylum-ci/dispatches
+
   Update-Documentation:
     name: Update the documentation
     needs: Release


.github/workflows/release.yml
```diff
@@ -156,6 +156,19 @@ jobs:
...
+        if: ! contains(github.ref, 'rc')
+        # Reference: https://docs.github.com/en/rest/repos/repos#create-a-repository-dispatch-event
+        run: >
+          curl
+          -X POST
+          --fail-with-body
+          -H "Accept: application/vnd.github.v3+json"
+          -H "Authorization: token ${{ secrets.GH_RELEASE_PAT }}"
+          -d '{"event_type":"build-push-docker-images","client_payload":{"CLI_version":"$GITHUB_REF_NAME"}}'
+          https://api.github.com/repos/phylum-dev/phylum-ci/dispatches
```
cd-work, line 170: Can you indent the arguments without it breaking? Having it all on the same level as `curl` kinda makes it difficult to distinguish from multiple commands in the same script.

cd-work (changes requested):

.github/workflows/release.yml
maxrake: Yes, it is possible. Some [test cases were added to a throwaway branch](https://github.com/phylum-dev/cli/blob/yaml_case/.github/workflows/test.yml#L76-L148) and [manually run in CI](https://github.com/phylum-dev/cli/runs/6884226132?check_suite_focus=true), yielding the following results:

* Simply indenting the lines does not work
* Adding trailing backslash line continuation characters works to indent the command line
  * It works for *both* folded (`>`) and literal (`|`) yaml block styles
* Using the `--no-progress-meter` option is helpful to reduce output noise
* The `--fail-with-body` option was added in `curl` v7.76.0
  * This version is not available in `ubuntu-latest`
  * This version is available in `ubuntu-22.04`

It comes down to a style choice in whether it looks better to have indentation *with* trailing backslashes or no indentation but also no trailing backslashes. The assumption is that indentation is still the preferred option even if it means adding a different kind of cognitive load to future readers of this code. @cd-work, Is that correct?

This review will be updated with those findings shortly.

kylewillmon (approved):

cd-work (dismissed):

.github/workflows/release.yml
```diff
@@ -156,6 +156,29 @@ jobs:
...
+    needs: Release
+    # The `--fail-with-body` option in `curl` was added in v7.76.0, which is too
+    # new for the `ubuntu-latest` runner that currently makes use of Ubuntu 20.04.
+    # TODO: Update to `ubuntu-latest` once `ubuntu-22.04` support is stabilized.
+    #       https://github.com/phylum-dev/cli/issues/467
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Trigger phylum-ci Docker image creation
+        # Don't trigger for pre-releases
+        if: contains(github.ref, 'rc') != true
```
kylewillmon, line 170: As mentioned in phylum-dev/phylum-ci#65, I think this option is more readable:

```suggestion
        if: ${{ !contains(github.ref, 'rc') }}
```

kylewillmon (approved):

.github/workflows/release.yml
cd-work: `!= true` might look kinda silly, but if it works, it works. Only suggestion I could make is that `== false` looks less like a bug.

cd-work (approved):

PR #474, opened 2022-06-17
Inject TypeScript API interface module

### LINKED ISSUE #449: Provide TypeScript shim for Extension API
# Overview

The API functions exposed by Deno can be used from consumers using `Deno.core.opAsync(..)`. This is somewhat awkward to use, doesn't provide any type safety and lacks solid documentation.

To make writing extensions as simple as possible, we should provide our own shim that translates the generic, unchecked Deno API to our custom TypeScript wrapper. Consumers should be able to just import this API and then use the functions provided by us, with a sensible module structure and method naming.

# Acceptance Criteria

- [ ] Phylum API TypeScript module is injected into the Deno runtime, for any extension to use
- [ ] TypeScript shim maps each Extensions API function to a TypeScript function
- [ ] Functions are documented, so the injected module can be used as preliminary documentation for writing extensions

cd-work:
This adds a new shim module to Deno's module loader under the name of
`phylum`, which allows interacting with Phylum's API through a
pre-defined TypeScript interface.

Fixes #449.


cli/src/extension_api.ts
```diff
@@ -0,0 +1,176 @@
+// TODO: Figure out how to make API async-safe.
+// TODO: Always use String in API, since &str can't easily cross JS boundary.
+// TODO: Auto-generate interface types?
+// TODO: Auto-generate the entire file?
+// TODO: Document methods.
```
cd-work, line 5: These are the remaining things that should be considered before merging this PR, the main points are the following:

Currently our API isn't really async safe, as soon as one async block is held across another, a mutable borrow error will show up. Of course it's possible to await everything immediately, but that kinda defeats the purpose of making it async. I believe we'll have to reconsider our PhylumApi initialization approach (ping @andreaphylum).

Some of our APIs take a `&str`. It might be possible to share string references between JavaScript and Rust, but even if it is that doesn't sound like a great idea and would likely be very convoluted. So I'd recommend defining APIs with only types that can be easily shared (I've fixed all existing APIs in this PR).

Lastly, this file probably shouldn't have to be written by hand. I've attempted to write out all the interface types by hand but quickly gave up after considering the maintenance overhead. Generating this entire file with a proc macro should be possible if we play our hands right, but at the very least I'd like to generate the interface types by hand.

That said, the interface types aren't strictly necessary, so technically we could also just point users to some API documentation where these types are publicly documented.

andreaphylum: I think immutably shared `&str` may not be a concern. I assume those strings are owned by v8, and IIRC strings in JS are an immutable primitive, so we have a guarantee that a `&str` is not mutated across async boundaries.

Our current implementation of API initialization is more complex than I'd like, although I do have a few ideas about how to get rid of the "lazy initialized-once future" abstraction. This still doesn't solve the fundamental problem of holding guards across async boundaries. Even if we were to hold a mutex guard from an immutable `Ref` borrow, we can still trigger a deadlock in as simple a way as

```typescript
await Promise.all([Deno.core.opAsync('first_api'), Deno.core.opAsync('second_api')])
```

As soon as I have some spare cycles, I want to do a deep dive in `deno_core`'s code to understand how its concurrency model interacts with v8, and the invariants it upholds.

cd-work: > I think immutably shared &str may not be a concern. I assume those strings are owned by v8, and IIRC strings in JS are an immutable primitive, so we have a guarantee that a &str is not mutated across async boundaries.

The concern is that `&str` does not exist as a type in JavaScript. All you have is `string` and that is not the same.

> As soon as I have some spare cycles, I want to do a deep dive in deno_core's code to understand how its concurrency model interacts with v8, and the invariants it upholds.

I'm not sure how this is related to `v8`'s concurrency model? Fundamentally this issue is all Rust, it's all about being able to call two API functions without resolving the first one before calling the second.

andreaphylum: > The concern is that &str does not exist as a type in JavaScript. All you have is string and that is not the same.

We should investigate what exactly this distinction is. Per my (limited) understanding, there is a serialization layer between v8 and Rust, and data gets in/comes out of Deno extensions via this layer. If Deno is able to zero-copy serialize a v8-owned JS `string` into a `&str` reference (would be sufficient to create a slice out of the length and the underlying pointer to the utf-8 string, if it's available...), we do not need to create owned values.

> I'm not sure how this is related to v8's concurrency model? Fundamentally this issue is all Rust, it's all about being able to call two API functions without resolving the first one before calling the second.

Having a `PhylumApi` object behind an immutable reference would solve this at the root; all of its methods currently bind to `&self`, and we may have as many immutable `Ref`s as we want. I'll open an issue and work on a fix for this later.

cd-work: For prosperity from DMs:

The issue is here is specifically that Deno does not accept using a `&str` here and doing so will result in an error. While deno might be able to handle a `&str`, it will always deserialize the JavaScript `string` as `String` and does not do any implicit conversions:

```
[2022-06-20T17:56:31Z ERROR phylum] Execution failed: TypeError: Error parsing args at position 1: invalid type: string "f16e99b2-cfdc-4293-ac6a-1bb040b594f1", expected a borrowed string
        at Object.opAsync (deno:core/01_core.js:152:35)
        at Function.getJobStatus (deno:phylum:18:32)
        at file:///home/chris/.local/share/phylum/extensions/test/main.ts:6:35
```

maxrake:
Removing myself as unqualified to review this PR...

andreaphylum (dismissed):

cli/src/commands/extensions/api.rs
```diff
@@ -233,13 +235,14 @@ async fn get_package_details(
...
 #[op]
-fn parse_lockfile(lockfile: &str, lockfile_type: &str) -> Result<Vec<PackageDescriptor>> {
+fn parse_lockfile(lockfile: String, lockfile_type: String) -> Result<Vec<PackageDescriptor>> {
     let parser = LOCKFILE_PARSERS
         .iter()
         .find_map(|(name, parser)| (*name == lockfile_type).then(|| *parser))
         .ok_or_else(|| anyhow!("Unrecognized lockfile type: `{lockfile_type}`"))?;
 
-    let lockfile_data = std::fs::read_to_string(Path::new(lockfile))?;
+    let lockfile_data = fs::read_to_string(&lockfile)
```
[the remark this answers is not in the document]
cd-work: Hm I suppose we could, but that's fundamentally unrelated to this PR.

Did you have any reason for initially designing it as synchronous? Was it an oversight or just because of the existing implementation in use by subcommands?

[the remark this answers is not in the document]
andreaphylum: Yeah, it is unrelated, just something I noticed while reviewing. I guess I just did not see an immediate need for it at the time and it did not occur to me to have this be `async`.

andreaphylum (approved):

PR #476, opened 2022-06-18
Show a spinner while waiting for API

### LINKED ISSUE #487: Improve the UX for laggy subcommands
# Overview
It takes a few seconds to build the API client and perform auth. Showing something like a spinner would make this less awkward for the end user.

# Acceptance Criteria
- [ ] CLI response to the user is snappy, even if it is just a loading animation.
- [ ] Loading animation doesn't infect machine readable output or piping, etc.

samtay:
The pause for any subcommand that has to hit the API is a little grating; this at least makes it less awkward.

- Stopped using the `spinners` lib, as that only prints to stdout hence ruins any machine readable output.
- Currently we only used that library to show a spinner when downloading the new phylum binary; just swapped out the usage with the new one.
- Added minimal spinners to each subcommand that waits for API client creation & auth.

I verified stdout is still fine for piping, redirection, etc:
```
❯ phs project -j | jq
[
  {
    "name": "samtay.github.io",
    "id": "9903dfb2-c4b2-484a-92c8-c403dbbf9062",
    "updated_at": "2022-04-20T04:09:10.235346Z",
    "ecosystem": null
  }
]
```

I considered gating this to Linux/Mac only, but I _think_ windows supports these ANSI codes. And either way, we are already using such codes for coloring pretty liberally in `cli`, so if it's something we need to check on, it's probably a separate issue.

Closes #487

## Checklist


cli/src/spinner.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+mod ansi {
+    pub const CURSOR_POSITION_SAVE: &str = "\x1B7";
+    pub const CURSOR_POSITION_RESTORE: &str = "\x1B8";
+    pub const CURSOR_POSITION_0: &str = "\x1B[0G";
+    pub const CURSOR_HIDE: &str = "\x1B[?25l";
+    pub const CURSOR_SHOW: &str = "\x1B[?25h";
+    pub const CLEAR_LINE: &str = "\x1B[2K";
+}
```
cd-work, line 136: I'm not sure how this worked in the spinners library, but having all these hardcoded **will** break on certain environments. Proper resolution of the `CURSOR_HIDE` escape sequence would involve resolving the `civis` terminfo capability for example.

cd-work (changes requested):
I haven't reviewed any other part of the code, but I couldn't help noticing the hardcoded escape sequences.

cli/src/bin/phylum.rs
```diff
@@ -174,12 +175,12 @@ async fn handle_commands() -> CommandResult {
         "version" => handle_version(&app_name, ver),
         "update" => handle_update(sub_matches).await,
         "parse" => handle_parse(sub_matches),
-        "ping" => handle_ping(api.await?).await,
-        "project" => handle_project(&mut api.await?, sub_matches).await,
-        "package" => handle_get_package(&mut api.await?, sub_matches).await,
-        "history" => handle_history(&mut api.await?, sub_matches).await,
-        "group" => handle_group(&mut api.await?, sub_matches).await,
-        "analyze" | "batch" => handle_submission(&mut api.await?, &matches).await,
+        "ping" => handle_ping(Spinner::wrap_with_message(api, "Pinging server...").await?).await,
```
kylewillmon, line 178: The message `"Pinging server..."` wraps the `api` future, but that future only includes reaching out to the OIDC provider. The "ping" to the actual API server is separate and won't be wrapped by this spinner.

kylewillmon (approved):

cli/src/spinner.rs
samtay: Spinners lib does the [same kinda thing](https://github.com/FGRibreau/spinners/blob/16ed2c6f5ec95c5cb9edca3e0f6debf144c7d78c/src/lib.rs#L192).

In another project I have used [crossterm](https://github.com/crossterm-rs/crossterm/blob/21155716e2eedd5ba8ab97168e5eed7cd50d2ad8/src/cursor.rs#L311-L320) which aims to have a much more varied support across terminals / OSes, but I was not sure if bringing in this dependency would be worthwhile.

Another note is that, while these now go to stderr, which is an improvement, another optimization would be to use [atty](https://docs.rs/atty/latest/atty/) and gate all prettifying ANSI codes based on tty status. For example `phs project > out` with no `-j` will currently print ANSI codes to the output file. Similarly `phs project > out 2>&1` will print the these spinner codes to the out file.

cli/src/bin/phylum.rs
samtay: Yea, mostly just changed this line to demonstrate usage, I can revert it.

cli/src/spinner.rs
```diff
@@ -0,0 +1,136 @@
...
+        self.handle.await.ok(); // of no consequence if the spinner thread panics, squash it
+    }
+
+    /// Spin until receiver sees `Command::Stop` or the channel is closed.
+    async fn spin(mut rx: Receiver<Command>, msg: Option<String>) {
+        let mut msg = msg;
+        let mut dots = SPINNER_DOTS.iter().cycle();
+        eprint!("{}", ansi::CURSOR_POSITION_SAVE);
+        eprint!("{}", ansi::CURSOR_HIDE);
+        eprint!("{}", ansi::CLEAR_LINE);
```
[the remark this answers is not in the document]
samtay: Good point! I never considered that.

FYSA, there are a number of such blocking writes via `print_user_warning!` etc. sprinkled throughout other async functions. Like [here for example](https://github.com/phylum-dev/cli/blob/8b401b3cfdd8115f5593661321e6097cc8526b31/cli/src/commands/jobs.rs#L89-L106). But, they're admittedly less concerning than this separate loading animation thread with repeated writes.

I don't _think_ it's something we need to worry about, unless for some reason we expect slow writes to these handles 🤔 And if it ever does become a problem, we could just swap out with `tokio::io::Stderr` and throw some `await`s around. But do let me know if that's something you'd like in this PR 👍

cli/src/spinner.rs
cd-work: I'd say checking for a TTY definitely makes sense, in fact this will likely be added to STD in a future release (https://github.com/rust-lang/rust/pull/98033).

With escape sequences there's unfortunately probably more people that do it wrong than people that do it right. I'm not sure it matters much for us since we have no goal to support a wide range of terminals, but it's certainly something to consider.

kylewillmon (approved):

cli/src/spinner.rs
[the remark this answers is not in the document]
cd-work: Technically pretty much every function is a blocking function. Nothing executes **instantly**.

The important bit should just be that we're aware of these limitations, so if we ever need to write **a bunch** of data we don't suddenly get confused why things are behaving strangely.

cli/src/spinner.rs
samtay: So are there any changes you'd like me to make as part of this PR?

cd-work: I think a TTY check should be pretty trivial and provide the biggest benefit. Escapes I'd be okay with for now and put in a mental reminder should anyone ever complain that it looks like crap on the linux console or something.

cd-work (changes requested):
I feel like the spinner API would be way simpler if the atty handling wasn't externalized? I doubt consumers of the spinner care about it not actually doing anything when the output isn't a tty.

kylewillmon (dismissed):

samtay:
@cd-work let me know if there's any other changes you'd like to see, or if you'd rather I close it for now

cd-work:
@samtay I was waiting for @kylewillmon to just approve it again so you can merge it.

I personally would have recommended further code cleanup, I think the tty handling shouldn't have to significantly impact our code and we can just skip it at the printing step, since performance is irrelevant. But if @kylewillmon is fine with the current version we can go with that too.

kylewillmon (approved):

PR #478, opened 2022-06-20
Deno standard library module loader

andreaphylum:
This PR modifies our custom module loader, adding:
- Support for importing remote files, provided they are under the `deno.land` domain and use the `https://` scheme.
- Support for importing local files, provided they are under the Phylum extensions directory.
- Tests to validate the above mentioned restrictions

A note wrt local relative imports: with this implementation, if we have `extension-a` and `extension-b`, we can achieve a rudimentary dependency system by doing

```javascript
// extension-a/main.ts
import * as B from '../extension-b/main.ts'
```

It is desirable that we find an alternative formula for this so that the dependency is explicitly stated, and prevent traversals of this nature, but I don't think it is an urgent requirement. Example:

```javascript
// extension-a/main.ts
import * as B from 'ext://extension-b/main.ts'
```


cli/src/deno.rs
```diff
@@ -81,9 +84,29 @@ pub async fn run(extension_state: ExtensionState, entry_point: &str) -> Result<(
...
+        }
+
+        Ok(fs::read_to_string(path).await?)
+    }
 
-impl ModuleLoader for TypescriptModuleLoader {
+    async fn load_from_deno_std(path: Url) -> Result<String> {
+        let response = reqwest::get(path).await?;
+        Ok(response.text().await?)
+    }
```
cd-work, line 106: The function makes it sound like this could only load from deno std, but it seems like there's no restriction on the remote?

cli/src/deno.rs
```diff
@@ -96,12 +119,10 @@ impl ModuleLoader for TypescriptModuleLoader {
...
         let module_specifier = module_specifier.clone();
         Box::pin(async move {
-            let path = module_specifier
-                .to_file_path()
-                .map_err(|_| anyhow!("Invalid module path"))?;
-
             // Determine source file type.
-            let media_type = MediaType::from(&path);
+            // We do not care about invalid URLs yet: this match statement is inexpensive, bears
+            // no risk and does not do I/O -- it operates fully off of the contents of the URL.
```
cd-work, line 124:
```suggestion
            // We do not care about invalid URLs yet: This match statement is inexpensive, bears
            // no risk and does not do I/O -- it operates fully off of the contents of the URL.
```

cd-work (changes requested):
> Support for importing remote files, provided they are under the deno.land domain and use the https:// scheme.

Have you considered how to communicate this to our users? I think one of the issues with allowing remote extensions is that users will expect *any* remote extension to work. Do you aim to just inform them through error messages, or is there also a way we might be able to document this?

> A note wrt local relative imports: with this implementation, if we have extension-a and extension-b, we can achieve a rudimentary dependency system by doing

Honestly I think it would be better if different extensions didn't depend on each other like this. It should be trivial for them to ship the desired code within their existing directory and duplication should be irrelevant. Anything else is just going to be super prone for breakage.

cli/src/deno.rs
```diff
@@ -81,9 +84,29 @@ pub async fn run(extension_state: ExtensionState, entry_point: &str) -> Result<(
 }
 
 /// See https://github.com/denoland/deno/blob/main/core/examples/ts_module_loader.rs.
-struct TypescriptModuleLoader;
+struct PhylumExtensionsModuleLoader;
+
+impl PhylumExtensionsModuleLoader {
+    async fn load_from_filesystem(path: &Path) -> Result<String> {
```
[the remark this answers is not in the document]
andreaphylum: From testing, it appears that it gets normalized into an absolute path by `deno_core`. Not sure how strong of a guarantee we have on that front, but otoh I think it would be pretty inconvenient for an implementor of the trait to have to deal with relative URLs anyway, so I believe we can rely on this.

Added note: not only local paths are passed in as absolute, but also remote ones.

cli/src/deno.rs
andreaphylum: Agreed, same as with the related comment below.

cli/src/deno.rs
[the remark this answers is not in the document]
cd-work: I feel like it's probably safer to check/redo normalization ourselves, since this is somewhat of a security check and we wouldn't want people to be able to bypass it with symlinks?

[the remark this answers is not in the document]
cd-work: > Wrt renormalization: are we confident about the fact that a relative path is relative to the current working directory?

I'm not sure what you mean? Are you saying that we don't necessarily know that the relative path was intended to be relative to our applications current binary, instead of relative to the loaded module?

Honestly I'd say it doesn't even matter that much, if deno normalizes things already, just doing `!relative && !symlink` would be enough for me. My goal is just that we're safe from the worst-case scenario, I don't really care if we just explode in those situations.

> Anyway, I found a guarantee that the path is absolute here

That's no guarantee, it's inside of `deno_core` and thus might change at any time.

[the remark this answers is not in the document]
andreaphylum: > Honestly I'd say it doesn't even matter that much, if deno normalizes things already, just doing !relative && !symlink would be enough for me. My goal is just that we're safe from the worst-case scenario, I don't really care if we just explode in those situations.

Agreed 100%

> That's no guarantee, it's inside of deno_core and thus might change at any time.

The type is an alias to `url::Url` which is from a separate crate. Any change to that behaviour would most definitely break our build. If we want to be extra sure, we could just add tests for it.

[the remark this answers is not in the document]
cd-work: > The type is an alias to url::Url which is from a separate crate. Any change to that behaviour would most definitely break our build. If we want to be extra sure, we could just add tests for it.

They could easily change that to use `Url::parse` and `Path::as_str` or whatever. Just advocating to better be safe than sorry.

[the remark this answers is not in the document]
andreaphylum: Being better safe than sorry is something I fully agree with, although I'm not sure how we could hedge our bets in this specific case -- can't find a way of determining if a [`Url`](https://docs.rs/url/latest/url/struct.Url.html) object is relative or absolute, and the documentation in general seems to suggest there is no state where relative paths are valid. We do use methods from `Url` so if they changed the type we'd get a build error.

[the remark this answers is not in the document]
cd-work: We get a path though, not a url, right? All we have to do is check that path?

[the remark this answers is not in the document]
andreaphylum: No, the other way around -- we get a `Path` from the `Url` [here](https://github.com/phylum-dev/cli/blob/andrea/module-loader/cli/src/deno.rs#L153) which is stated to be [absolute](https://docs.rs/url/latest/url/struct.Url.html#method.to_file_path).

andreaphylum:
> Honestly I think it would be better if different extensions didn't depend on each other like this. It should be trivial for them to ship the desired code within their existing directory and duplication should be irrelevant. Anything else is just going to be super prone for breakage.

I missed this bit. I agree with this, but at the moment we don't have a simple way of sharing extension related data with the Deno runner. This would be necessary in order for the module loader to be aware of the extension's path, so that it can filter imports according to it.

I think this deserves its own, separate PR. I would like to add that having the Deno runtime code in `src/deno.rs` completely separated from the rest of the extensions code in `commands/extensions` is a bit confusing, and we should probably plan a refactoring that brings the two together and makes it possible for `deno::run` to accept `Extension` objects instead of a single entry point path.

On that note, do you think `commands/extensions` is a bit too nested in the source tree? On one hand, it represents indeed a subcommand tree; on the other, it is an entire major feature of the CLI and could maybe be better placed in the top level and have a shim module under `commands/`.

cli/src/deno.rs
```diff
@@ -81,9 +83,45 @@ pub async fn run(extension_state: ExtensionState, entry_point: &str) -> Result<(
...
+            .to_file_path()
+            .map_err(|_| anyhow!("{path:?}: is not a path"))?;
+
+        let extensions_path = extensions_path()?;
+        if !path.starts_with(&extensions_path) {
+            return Err(anyhow!(
+                "`{}`: importing from paths outside of the extension's directory is not allowed",
+                path.to_string_lossy(),
+            ));
+        }
```
cd-work, line 100: Do you have an idea how we could cleanly prevent users from specifying dependencies between extensions here? I feel like we should better be safe than sorry and not allow this.

cli/src/deno.rs
```diff
@@ -81,9 +83,45 @@ pub async fn run(extension_state: ExtensionState, entry_point: &str) -> Result<(
...
+
+        if path.is_symlink() {
+            return Err(anyhow!(
+                "`{}`: importing from symlinks is not allowed",
+                path.to_string_lossy(),
+            ));
+        }
```
cd-work, line 107: Have you actually tested this? I'd be curious if deno resolves symlinks internally or not.

cli/src/deno.rs
```diff
@@ -118,8 +154,23 @@ impl ModuleLoader for TypescriptModuleLoader {
...
-            // Read the source and transpile it if necessary.
-            let mut code = fs::read_to_string(&path)?;
+            // Load either a local file under the extensions directory, or a Deno standard library
+            // module. Reject all URLs that do not fit these two use cases.
+            let mut code = match (module_specifier.scheme(), module_specifier.host()) {
+                ("file", None) => {
+                    ExtensionsModuleLoader::load_from_filesystem(&module_specifier).await?
+                }
+                ("https", Some(Host::Domain(domain))) => {
+                    ExtensionsModuleLoader::load_from_deno_std(&module_specifier, domain).await?
```
cd-work, line 164: We're passing the `module_specifier` anyway. Why not handle the domain stuff internally?

cd-work (changes requested):

cd-work:
> On that note, do you think commands/extensions is a bit too nested in the source tree? On one hand, it represents indeed a subcommand tree; on the other, it is an entire major feature of the CLI and could maybe be better placed in the top level and have a shim module under commands/.

I think the structure is problematic mainly because of its redundancy. The only way to interact with the CLI is through its subcommands, thus everything will inevitably land inside `cli/src/commands`.

The deno code for example has absolutely nothing to do with commands, but if we move all the code used by command handlers to the `commands` directory it would rest there too. I'm not sure what the "perfect" structure would be, but maybe having just `commands/extension.rs` and then having a toplevel `extensions/` module would make more sense. So the module in `commands` is responsible **ONLY** for handling the command.

cli/src/deno.rs
andreaphylum: It would require another `match` statement and another couple of lines of error management in the `load_*` function. If we use the current solution instead, other invalid states are handled by the catch-all arm. Neither solution is perfectly clean looking, though.

cli/src/deno.rs
andreaphylum: I can test it manually; it would be nice to have this be automated, but I'm not sure how we can do that cleanly -- I don't like the idea of having a symlink commited in the repo, if that's something which is at all possible.

cli/src/deno.rs
cd-work: You already have a conditional anyway though? All you have to do is change your `== "deno.land` to a pattern match?

I don't think it's a big deal to have `None` and `!= deno.land` handled by the same error handler. As long as there's a generic "only deno.land is supported as domain" error message that applies perfectly well to any potential domain.

andreaphylum: That makes sense.

cli/src/deno.rs
```diff
@@ -109,13 +109,14 @@ impl ExtensionsModuleLoader {
         Ok(fs::read_to_string(path).await?)
     }
 
-    async fn load_from_deno_std(path: &Url, domain: &str) -> Result<String> {
-        if domain == "deno.land" {
+    async fn load_from_deno_std(path: &Url) -> Result<String> {
+        if let Some(Host::Domain("deno.land")) = path.host() {
```
cd-work, line 113: I *feel* like match would be the way to go here? though this certainly works.

cli/src/deno.rs
```diff
@@ -109,13 +109,14 @@ impl ExtensionsModuleLoader {
...
             let response = reqwest::get(path.clone()).await?;
             Ok(response.text().await?)
         } else {
             Err(anyhow!(
-                "`{domain}`: importing from domains except `deno.land` is not allowed"
+                "`{}`: importing from domains except `deno.land` is not allowed",
```
cd-work, line 118:
```suggestion
                "`{}`: importing from domains other than `deno.land` is not allowed",
```

cli/src/deno.rs
cd-work: I'm not gonna die on this hill, fine with both. But `_ =>` would work perfectly fine too of course.

I just generally feel like `if let` + `else` is somewhat rare, but if clippy doesn't warn on it I don't mind it either.

cli/src/deno.rs
```diff
@@ -13,9 +13,11 @@ use deno_runtime::deno_core::{
 use deno_runtime::permissions::Permissions;
 use deno_runtime::worker::{MainWorker, WorkerOptions};
 use deno_runtime::{colors, BootstrapOptions};
+use tokio::fs;
+use url::{Host, Url};
 
 use crate::commands::extensions::api;
-use crate::commands::extensions::extension::ExtensionState;
+use crate::commands::extensions::extension::{extensions_path, ExtensionState};
```
cd-work, line 20:
```suggestion
use crate::commands::extensions::extension::ExtensionState;
```

Should import on the module level here.

cli/tests/extensions.rs
```diff
@@ -282,6 +284,97 @@ fn conflicting_extension_name_is_filtered() {
...
 
+// Test that the rules of the module loader are obeyed:
+// - Both TypeScript and JavaScript are supported.
+// - Files under $XDG_DATA_HOME/phylum/extensions may be imported.
+// - Remote URLs under https://deno.land are supported -- i.e., the Deno's standard library.
+// - No other URLs are supported.
+//   - We explicitly test that a https:// url which is not under `deno.land` is rejected.
+//   - We explicitly test that a directory traversal attempt is rejected.
+//
+// These tests are based on the fixtures under `fixtures/module-import-extension`.
```
cd-work, line 295: Wouldn't it be easier to have individual tests for all these? That would also mean we don't need a massive comment explaining things.

cd-work (changes requested):

cli/tests/extensions.rs
```diff
@@ -5,6 +5,8 @@ use std::env;
 use std::path::{Path, PathBuf};
 use std::sync::Mutex;
 
+use predicates::prelude::*;
```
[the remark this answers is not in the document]
andreaphylum: It is used in the new test, and it is what `assert_cmd` [uses](https://docs.rs/assert_cmd/latest/assert_cmd/assert/struct.Assert.html#method.stdout), alongside having a prelude module as well. It is now a direct dependency but we had that as transitive all along, it's just not reexported for some reason.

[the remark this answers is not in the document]
cd-work: Oh I see, you're using it for string contains in one instances. But it should either be used consistently or not at all, it seems like in some cases you're collecting into a string and just using that.

Also would be nice if we didn't include two different versions of it. Seems like there are some Cargo.lock changes introduced.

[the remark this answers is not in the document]
andreaphylum: Yes, I suggest we migrate all the other changes to that, either in this or in a different PR.

I don't think we are including two different versions? If anything, I believe adding it as a development dependency might have changed the resolved version. Let me check whether `assert_cmd` actually reexports it, or whether I can just put the same semver in there.

cli/tests/extensions.rs
cd-work: This is one thing I've considered anyway, but shouldn't we maybe have some abstractions to just run generic extension code? Having an abstraction to install & run from a path would be nice, but I feel like just having a helper that builds extensions from a string would clean things up significantly.

> Furthermore, I think having the complete set of rules there as a comment makes it clear what are the boundaries that we are testing, and I feel the current solution is a good "logical box".

I think each individual rule would be a good test, combining all of them into one, doesn't make sense to me. This is clearly running multiple integration tests, we're running the binary multiple times and testing their output, so why throw it into one test case?

andreaphylum: > This is one thing I've considered anyway, but shouldn't we maybe have some abstractions to just run generic extension code? Having an abstraction to install & run from a path would be nice, but I feel like just having a helper that builds extensions from a string would clean things up significantly.

Indeed, there is a lot of repetition. Let's create an issue for it and address it separately, as it is a big enough diff to deserve its own PR.

> I think each individual rule would be a good test, combining all of them into one, doesn't make sense to me. This is clearly running multiple integration tests, we're running the binary multiple times and testing their output, so why throw it into one test case?

I think those assertions are conceptually tightly coupled, don't really stand out on their own even if the form in which they are tested is the execution of a binary, and should be dealt with in bulk -- i.e. if I change one assertion in isolation I want to be aware of the fact that there is a bigger set of behaviors I need to be aware of, and I feel like multiple asserts in the same function is what makes the most sense.

cli/tests/extensions.rs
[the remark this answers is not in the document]
andreaphylum: Changes to the lockfile have been reverted by specifying the same dependency as `assert_cmd`.

cli/tests/extensions.rs
andreaphylum: That's kind of a slippery slope argument -- it's not like our only options are as extreme as _one assert per test_ vs _one single test for everything_, and the current solution is the middle ground grouping that makes the most sense to me on a logical standpoint.

We're not testing _the module loader_, but _the ruleset of the module loader_ whose rules don't make any sense in isolation: asserting that we accept `deno.land` hosts on its own and then asserting, in a separate logical context, that we don't accept some other remote URLs would be outright confusing to me.

At the same time, filesystem paths and remote URLs are still `Url` objects and we're testing for rules on `Url`s.

If we incur into a bug that's either related to URLs or local files, then we should rather introduce a new, separate test for controlling that individual behavior, _or_ add a stricter formal rule that all `Url`s have to abide to, within this test.

A compromise I'm willing to make would be breaking up the test, and grouping the resulting tests in a dedicated `mod {}` block. The important thing is that the tight interdependence of those assertions is made evident.

cd-work (approved):

cd-work (approved):

cd-work (approved):

PR #484, opened 2022-06-20
Remove custom tempdir implementation

cd-work:


cli/tests/extensions.rs
```diff
@@ -29,18 +27,18 @@ lazy_static! {
 // working directory should be installed.
 #[test]
 fn extension_is_installed_correctly() {
-    let tmp_dir = TmpDir::new();
+    let tempdir = TempDir::new().unwrap().into_path();
```
kylewillmon, line 30: The use of `.into_path()` intentionally leaks the directory so that it won't be automatically deleted. Is this desired?

If we keep the `TempDir` object around and use `tempdir.path()` when the path is needed, then we can rely on the `Drop` implementation to delete the directory (if possible) when the test is done.

cd-work: Probably a good idea, yeah. Just went for the minimal plausible implementation. I'll try and fix it up.

kylewillmon (approved):

PR #489, opened 2022-06-20
fix: phylum-ci Docker image trigger

### LINKED ISSUE #488: `phylum-ci` Docker image trigger broken
# Overview
The CLI v3.6.0 release triggered a `phylum-ci` Docker image build in the `phylum-dev/phylum-ci` repository. That build failed with the error:

`phylum-init: error: argument -r/--phylum-release: an invalid version was provided`

# How To Reproduce
It is not desirable to reproduce this error since that would create extra requests for approval on the image creation. However, the existing error can be seen in the [build logs](https://github.com/phylum-dev/phylum-ci/runs/6974685371?check_suite_focus=true)

# Expected Behavior
The trigger from a CLI release is expected to cause a `phylum-ci` Docker image to be created without errors.

# Additional Context
The following trigger from the CLI repository:

```sh
curl \
    -X POST \
    --fail-with-body \
    --no-progress-meter \
    -H "Accept: application/vnd.github.v[3](https://github.com/phylum-dev/cli/runs/6974683314?check_suite_focus=true#step:2:3)+json" \
    -H "Authorization: token ***" \
    -d '{"event_type":"build-push-docker-images","client_payload":{"CLI_version":"$GITHUB_REF_NAME"}}' \
   [ https://api.github.com/repos/phylum-dev/phylum-ci/dispatch](https://api.github.com/repos/phylum-dev/phylum-ci/dispatches)
```

...causes this `phylum-ci` Docker image build command to run:

```sh
  docker build \
    --tag phylum-ci \
    --build-arg PKG_SRC=phylum-*.whl \
    --build-arg PKG_NAME=phylum-*.whl \
    --build-arg CLI_VER=$GITHUB_REF_NAME \
    --build-arg BUILDKIT_INLINE_CACHE=1 \
    .
```

when it was meant to result in a build command like this:

```sh
  docker build \
    --tag phylum-ci \
    --build-arg PKG_SRC=phylum-*.whl \
    --build-arg PKG_NAME=phylum-*.whl \
    --build-arg CLI_VER=v3.6.0 \
    --build-arg BUILDKIT_INLINE_CACHE=1 \
    .
```

The difference is that `$GITHUB_REF_NAME`, taken literally, is not a valid version. However, if the variable is properly substituted for a valid value, like `v3.6.0`, then the build succeeds. The reason the variable was taken literally is because in bash, expressions don't expand in single quotes. The reason single quotes were used here is because the data payload includes a JSON string that *must* use double quotes.

maxrake:
The change in this PR is to encapsulate the `curl` based trigger's data
payload in double quotes since expressions do not expand in single
quotes. This means that the inner double quotes have to be escaped
because they are still needed since the payload includes a JSON string.

Promote the `if` condition for triggers from step to job level.
This change will ensure the Trigger job does not run for pre-releases.
Previously, the conditional was one level lower, on a step within the
job. That was confusing because the UI would show that the Job ran, but
it wasn't until drilling into the results that it would be clear that
the trigger step was skipped.

Closes #488

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -159,6 +159,10 @@ jobs:
   Trigger:
     name: Trigger phylum-ci Docker image creation
     needs: Release
+    # Don't trigger for pre-releases
+    # NOTE: This is an instance where the expression syntax (`${{ }}`) is required for the `if` conditional,
+    #       contrary to the GitHub workflow syntax documentation. Do not remove the expression syntax.
+    if: ${{ !contains(github.ref, 'rc') }}
     # The `--fail-with-body` option in `curl` was added in v7.76.0, which is too
     # new for the `ubuntu-latest` runner that currently makes use of Ubuntu 20.04.
     # TODO: Update to `ubuntu-latest` once `ubuntu-22.04` support is stabilized.
@@ -166,10 +170,6 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Trigger phylum-ci Docker image creation
-        # Don't trigger for pre-releases
-        # NOTE: This is an instance where the expression syntax (`${{ }}`) is required for the `if` conditional,
-        #       contrary to the GitHub workflow syntax documentation. Do not remove the expression syntax.
-        if: ${{ !contains(github.ref, 'rc') }}
         # Reference: https://docs.github.com/en/rest/repos/repos#create-a-repository-dispatch-event
         run: |
           curl \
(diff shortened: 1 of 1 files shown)


maxrake:
There might be a cleaner way to represent this without having to escape the double quotes...like with `jq`, maybe...but there was not enough time to find a working alternative. The proposed solution was tested and works.

cd-work (approved):

PR #490, opened 2022-06-21
Async safe API borrows

### LINKED ISSUE #444: Report errors on, or circumvent, multiple `RefCell` borrows in Deno ops
Currently, Deno ops (i.e. our Rust-side API) require a `Rc<RefCell<OpState>>` parameter for handling state. All of our ops are asynchronous, and a user not awaiting on them immediately will cause the `RefCell` to be mutably borrowed multiple times, which will result in a panic:

```typescript
Deno.core.print(JSON.stringify(await PhylumApi.get_user_info())) // this will not panic
Deno.core.print(JSON.stringify(PhylumApi.get_user_info())) // this will panic
```

We need to either leverage some concurrency primitive (e.g. `Mutex`/`RwLock`), and make sure it won't ever result in a deadlock, or provide the user with actionable feedback (e.g. "please immediately await Phylum API calls"). Allowing concurrency and favoring more async-friendly interior mutability structures in place of `Rc<RefCell<T>>` should in general be preferred, but Deno's documentation is currently not super clear on the mechanisms that could cleanly complement that.

andreaphylum:
In this PR, Deno's `OpState` is borrowed immutably and access to the underlying `OnceFuture` is synchronized via a `tokio::sync::Mutex`.

This may introduce contention on first access and a very minor overhead on subsequent accesses, when retrieving a reference to the `PhylumApi` object. API calls themselves can be executed entirely concurrently.

Closes #444.

## Checklist

- [ ] ~Have you ensured that you have met the expected acceptance criteria?~
- [ ] ~Have you updated all affected documentation?~


cli/src/commands/extensions/api.rs
```diff
@@ -47,35 +48,41 @@ impl<T: Unpin> OnceFuture<T> {
...
-        })))
+        }))))
+    }
+}
+
+impl ExtensionState {
+    async fn get(&self) -> Result<Rc<PhylumApi>> {
+        // The mutex guard is only useful for synchronizing access to the encapsulated future.
+        // Once a `Result<Rc<PhylumApi>>` is obtained, the guard is dropped: subsequent awaits on
+        // `PhylumApi` methods are not synchronized via this mutex, and can happen concurrently.
```
cd-work, line 65: Saying this mutex is only useful for synchronizing access is kinda misleading. I was going to suggest you'd use a pure synchronization primitive instead until I realized it's not just for synchronization, it's also necessary because we need interior mutability.

cli/tests/fixtures/extensions/async-borrows/main.ts
```diff
@@ -0,0 +1,13 @@
...
+await Promise.all([
+  PhylumApi.get_user_info(),
+  PhylumApi.get_user_info(),
+  PhylumApi.get_user_info(),
+  PhylumApi.get_user_info(),
+  PhylumApi.get_user_info(),
+  PhylumApi.get_user_info(),
+  PhylumApi.get_user_info(),
+  PhylumApi.get_user_info(),
+  PhylumApi.get_user_info(),
```
cd-work, line 12: Maybe a tiny bit excessive? Also this sends actual web requests, I don't think we want that?

cd-work (changes requested):

cli/tests/fixtures/extensions/async-borrows/main.ts
andreaphylum: The point of the test was to create a complex concurrent situation where there is some contention/risk of deadlocking.

This is still far from perfect as we can't _guarantee_ deterministic behavior, i.e. reliably reproducing a situation where a concurrency issue may occur in the first place.

I think actual calls are fine as long as we're in the context of an integration test -- the actual network access is part of the behavior under test. If we don't want that, we can mark this test as `#[ignore]` and make it so it only runs locally, but I do think having coverage on this is important.

cli/src/commands/extensions/api.rs
andreaphylum: Wording these comments is kind of tricky. Interior mutability is needed because we need to synchronize the mutable access to `OnceFuture`. I'll try and reword this.

cli/tests/fixtures/extensions/async-borrows/main.ts
cd-work: > The current one doesn't, but that doesn't mean that it is 100% correct and there is no corner case where concurrency issues happen. A shotgun approach is not elegant, I agree; what would be a good test here to have in its place?

There is no "100% guarantee" unless you want to start formally verifying our code. Tests should be purpose-driven and if you have an edge-case beyond our original problem that you think is uncovered, we cover it explicitly.

cd-work (approved):

cd-work (approved):

PR #491, opened 2022-06-21
Add detailed messages for HTTP conflicts

### LINKED ISSUE #473: Simplify error messages
# Overview
The error messages displayed by the CLI are not very user-friendly and should be simplified and more direct.

# How To Reproduce
1. Take an action that generates an error (e.g., try and create a project that already exists)
2. See error messaging

# Expected Behavior
A clear and concise error message is displayed to the user, not JSON.

# Additional Context
image

cd-work:
Fixes #473.

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -47,6 +47,8 @@ pub enum PhylumApiError {
     #[error(transparent)]
     BaseUri(#[from] BaseUriError),
     #[error(transparent)]
+    Response(#[from] ResponseError),
+    #[error(transparent)]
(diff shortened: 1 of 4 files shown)


cd-work:
I think this should be the most straight-forward solution to this problem, enabling us to arbitrarily improve error messages based on return code, while still returning helpful HTTP errors when we cannot.

It might be possible to create additional abstractions that automatically generate more human-friendly messages for common status codes like `409`, but with only two instances being present to my knowledge at this point I don't think that would be worth it at this stage.

andreaphylum (approved):

PR #494, opened 2022-06-22
Add style guideline's rustfmt.toml

cd-work:


andreaphylum (dismissed):
It's odd that `fmt` has not been applied correctly to the `selfmanage` feature gated part as the CI failure shows -- maybe some rustfmt configuration conflict? Otherwise, looks good to me.

cd-work:
> It's odd that fmt has not been applied correctly to the selfmanage feature gated part as the CI failure shows -- maybe some rustfmt configuration conflict? Otherwise, looks good to me.

Just had a really old nightly installed.

PR #498, opened 2022-06-23
Add phylum extension new command

### LINKED ISSUE #412: Extensions: `new` command
# Overview

When users want to create their own extensions, we should make the friction to do so as slim as possible. One way that has gotten very common in other tools is to provide a command that automatically sets up a workspace with all necessary files.

A `phylum extension new <project>` command could thus automatically create the extension directory, the manifest with all required keys and an entrypoint in the desired language (maybe default to TypeScript with `--js` or `--wasm` as options in the future?).

Ideally when running `phylum extension new test` -> `phylum install ./test`, I should get a fully working extension that prints something simple like `Hello, World!` when executed with `phylum test`.

The subcommand name is subject to discussion, `new` is short and used by `cargo` and probably many other tools already. But something like `phylum extension create <project>` might read a little easier since it's closer to being proper english.

# Acceptance Criteria

- [ ] New subcommand for creating extensions
- [ ] Subcommand takes new name as parameter
- [ ] Execution creates an extension skeleton with at least a directory, the manifest and an entrypoint
- [ ] (Optional) Extension is ready to be used right after creation, printing `Hello, World!` by default
- [ ] (Optional) Extension language can be specified as CLI parameters

cd-work:
This patch adds a new command for automatically creating extension
skeletons in the specified repository, making the initial steps for
getting started with writing your own extensions much simpler.

Closes #412.

### DIFF
--- a/cli/src/commands/extensions/mod.rs
+++ b/cli/src/commands/extensions/mod.rs
@@ -1,19 +1,27 @@
 use std::collections::HashSet;
+use std::ffi::OsStr;
 use std::fs;
 use std::io::ErrorKind;
 use std::path::PathBuf;
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use clap::{arg, ArgMatches, Command, ValueHint};
-use extension::Extension;
 use futures::future::BoxFuture;
 use log::{error, warn};
 
 use crate::api::PhylumApi;
+use crate::commands::extensions::extension::{Extension, ExtensionManifest};
 use crate::commands::{CommandResult, CommandValue, ExitCode};
+use crate::print_user_success;
 
 pub mod api;
 pub mod extension;
+
+const EXTENSION_SKELETON: &[u8] = b"\
+import { PhylumApi } from 'phylum';
+
+console.log('Hello, World!');
+";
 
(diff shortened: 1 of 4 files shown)


cli/src/commands/extensions/mod.rs
```diff
@@ -26,6 +34,9 @@ pub fn command<'a>() -> Command<'a> {
         .subcommand(
             Command::new("remove").about("Uninstall extension").arg(arg!([NAME]).required(true)),
         )
+        .subcommand(
+            Command::new("new").about("Create a new extension").arg(arg!([PATH]).required(true)),
```
andreaphylum, line 38: I think it would be good UX to also allow users to launch `phylum extension new` without an argument and have that create an extension in the current directory, similarly to what other ecosystem tools do.

But I'd say it's not super urgent and we could tackle this in a separate issue/PR.

andreaphylum (approved):

cli/src/commands/extensions/mod.rs
cd-work: I've decided to mimic `cargo new` in behavior and honestly think that's a good idea. There's already enough invariants to ensure and I think creating one in the current directory would just come with unnecessary ambiguity.

PR #500, opened 2022-06-24
Extensions: permission support

### LINKED ISSUE #448: Extensions: implement `permissions` in manifest
We need to grant extensions the possibility of requesting permissions. For example, `hash-check` may require read-only access to the files in a project. These permissions should be requested in the manifest, and a user must grant explicit consent at install time. Deno has CLI flags for granting permissions; we should translate the version serialized in the package manifest in the equivalent Deno runtime setup option.

andreaphylum:
Closes #448.

## Checklist


cli/src/commands/extensions/extension.rs
```diff
@@ -55,6 +64,14 @@ impl Extension {
         &self.manifest.entry_point
     }
 
+    pub fn permissions(&self) -> &Permissions {
+        &self.manifest.permissions
+    }
+
+    pub fn requires_permissions(&self) -> bool {
+        !self.manifest.permissions.is_empty()
+    }
```
cd-work, line 73: Considering that `permissions` is already a public method, I'm not sure we're gaining much with this extra one.

cli/src/commands/extensions/extension.rs
```diff
@@ -122,10 +139,13 @@ impl Extension {
         Extension::try_from(extension_path(name)?)
     }
 
+    pub fn path(&self) -> Result<PathBuf, std::io::Error> {
```
cd-work, line 142:
```suggestion
    pub fn path(&self) -> Result<PathBuf, io::Error> {
```

Maybe we'd also like to add a doc comment that canonicalization is performed, just in case someone calls it and is curious.

cli/src/commands/extensions/mod.rs
```diff
@@ -30,6 +34,10 @@ pub fn command<'a>() -> Command<'a> {
         .subcommand(
             Command::new("add")
                 .about("Install extension")
+                .arg(
+                    arg!(-y --yes "Automatically accept requested permissions")
+                        .action(ArgAction::SetTrue),
+                )
```
cd-work, line 40: Is `action` required here? Flags should be by default `SetTrue`, right?

Also we could consider hiding this option, since "normal" people shouldn't be using it. But I think the documentation appropriately shows already that this is not something usually desired.

cli/src/commands/extensions/mod.rs
```diff
@@ -70,7 +78,13 @@ pub fn add_extensions_subcommands(command: Command<'_>) -> Command<'_> {
 /// Entry point for the `extensions` subcommand.
 pub async fn handle_extensions(matches: &ArgMatches) -> CommandResult {
     match matches.subcommand() {
-        Some(("add", matches)) => handle_add_extension(matches.value_of("PATH").unwrap()).await,
+        Some(("add", matches)) => {
+            handle_add_extension(
+                matches.value_of("PATH").unwrap(),
+                matches.get_one::<bool>("yes").copied().unwrap_or(false),
```
cd-work, line 84: Pretty sure there's an easier way to check for flags. I'd have to lookup clap docs myself again but I'm like 90% certain and there's probably a couple examples in our codebase too.

cli/src/commands/extensions/mod.rs
```diff
@@ -106,6 +120,52 @@ async fn handle_add_extension(path: &str) -> CommandResult {
...
+
+    if !accept_permissions && extension.requires_permissions() {
+        println!("The `{}` extension requires the following permissions:", extension.name());
+
+        if let Some(read_paths) = permissions.read() {
+            println!("  Read from the following paths:");
+            for path in read_paths {
+                println!("    {}", path);
+            }
+        }
```
cd-work, line 137: I'm sure this is one of the reasons why the PR is still a draft, but I'm confident we can reduce the repetition here.

cli/src/commands/extensions/mod.rs
```diff
@@ -106,6 +120,52 @@ async fn handle_add_extension(path: &str) -> CommandResult {
...
+        if let Some(access_urls) = permissions.net() {
+            println!("  Access the following URLs:");
+            for url in access_urls {
+                println!("    {}", url);
+            }
+        }
+
+        if !Term::stdout().is_term() {
+            return Err(anyhow!("can't ask for permissions: not a terminal"));
+        }
```
cd-work, line 162: Should we hint at the `-y` flag?

cli/src/commands/extensions/permissions.rs
```diff
@@ -0,0 +1,159 @@
...
+
+#[derive(Clone, Default, Deserialize, Debug, Serialize)]
+pub struct Permissions {
+    read: Option<Vec<String>>,
+    write: Option<Vec<String>>,
+    run: Option<Vec<String>>,
+    net: Option<Vec<String>>,
+}
+
+// XXX In Deno, `Some(vec![])` actually means "allow all". We do not ever
```
cd-work, line 15: I'd recommend changing `do not ever` to `never`, wouldn't want to confuse people in such an important comment.

cli/src/commands/extensions/permissions.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub fn net(&self) -> Option<&Vec<String>> {
+        self.net.as_ref().and_then(|v| if v.is_empty() { None } else { Some(v) })
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.read().is_none()
+            && self.write().is_none()
+            && self.run().is_none()
+            && self.net().is_none()
+    }
```
cd-work, line 41: The concept of "empty" for permissions seems strange to me. I'd probably go more the route of deno's `allow_all` method and call this `is_allow_none` or something like that. "Empty" could mean either everything allowed, or nothing allowed, it's not exactly clear.

cli/src/commands/extensions/permissions.rs
```diff
@@ -0,0 +1,159 @@
...
+    fn resolve_path<S: AsRef<str> + std::fmt::Display>(src_path: S) -> Result<PathBuf> {
+        let path = PathBuf::from(src_path.as_ref());
+
+        if !path.is_relative() {
+            return Err(anyhow!("`{src_path}`: absolute paths are not allowed"));
+        }
+
+        if path.components().into_iter().any(|c| c == Component::ParentDir) {
+            return Err(anyhow!("`{src_path}`: directory traversals are not allowed"));
+        }
```
cd-work, line 52: I'm not sure these exceptions are necessary? They don't really provide any security advantage so they just kinda seem arbitrary. Especially because a single symlink would circumvent all of these restrictions.

cd-work (commented):
Just took a preliminary look because I was curious.

cli/src/commands/extensions/mod.rs
andreaphylum: Not sure we want to hide that: that option would most likely be used when setting up a CI environment in an unsupervised way, so I'd say that's most likely a desired use case.

cli/src/commands/extensions/mod.rs
cd-work: I think `is_present` should exist? That's what my memory is of it anyway.

cli/src/commands/extensions/permissions.rs
andreaphylum: I think they are very much necessary, as they are what prevents extensions from requiring permissions to files which are absolute paths or are attempting a directory traversal.

This does not make everything 100% safe, but at least it guarantees that in the event of passing a `-y` and granting unintended permissions the extension still won't be allowed to escape the current working directory or access absolute paths.

Having a symlink in the working directory which could be exploited is outside of our scope of influence, as an attacker who's able to do that already doesn't need to use an extension as an entry point.

cd-work: > I think they are very much necessary, as they are what prevents extensions from requiring permissions to files which are absolute paths or are attempting a directory traversal.

What is the purpose of preventing this? I see no benefit in doing so.

> This does not make everything 100% safe, but at least it guarantees that in the event of passing a -y and granting unintended permissions the extension still won't be allowed to escape the current working directory or access absolute paths.

What if the application wants to store things in a user config directory, or some cache directory? What is the benefit of locking them to the "current working directory" when that working directory is not restricted to anything specific?

Is there any actual real-world attack scenario this would prevent? It just seems like security theater to me, pretending like it improves security while only annoying extension developers.

> Having a symlink in the working directory which could be exploited is outside of our scope of influence, as an attacker who's able to do that already doesn't need to use an extension as an entry point.

You seem like you want to lock users down to their local directory. Then writing a symlink to the local directory is something perfectly within the abilities of an extension author with write permissions. This is absolutely not outside of our scope of influence.

andreaphylum: > What is the purpose of preventing this? I see no benefit in doing so.

This is my opinion here, but I see no need to allow as much freedom as possible to the extension developers when the use cases aren't meant to be broad scoped, but rather very very specific (i.e. accessing a package's directory and accessing Phylum's API).
 
> What if the application wants to store things in a user config directory, or some cache directory?

I don't think this is necessarily a use case we may want to support, and if we do, this calls at a minimum for further discussion and planning.

> What is the benefit of locking them to the "current working directory" when that working directory is not restricted to anything specific?

The benefit is in giving a clear scope to the _user_ of the extension, who will know that if they execute any given extension in any given current directory context, nothing else will be accessed outside of that.

> Is there any actual real-world attack scenario this would prevent? It just seems like security theater to me, pretending like it improves security while only annoying extension developers.

Having a malicious extension ask for unintended permissions which may be inadvertently granted via `-y` (i.e. in a CI environment) and then having no bounds on what those permissions may be looks to me like the very use case Phylum is working to prevent, as it would allow complete control of the environment the extension runs in.

Again, I don't think extensions developers' experience and freedom of action is an explicit target use case here, certainly not enough to allow for no limitations in their behavior; the research I conducted in the scope of this feature assumed as much from the get go.

If we want for it to be, we need to go back to the drawing board and decide what's allowed, what is not, and why.

Furthermore, I'm not saying the solution as it is today in this draft PR is 100% secure, I'm saying we should strive for it to be by keeping in mind that the safety of the end user comes first, and everything else comes next, and having the extension developers' scope of action be restricted by a whitelist is one way of achieving that.

Whether, or to which extent, that qualifies as "security theater", I do not know, and we can freely discuss.

> You seem like you want to lock users down to their local directory. Then writing a symlink to the local directory is something perfectly within the abilities of an extension author with write permissions. This is absolutely not outside of our scope of influence.

Allowing the creation of a symlink would definitely be something I'd argue against. I am not even sure how that could be achieved in Deno currently, and how it plays out with the permissions system, but I'd most certainly want for that not to be a possibility at all unless we find a valid and compelling use case.

cd-work: > Having a malicious extension ask for unintended permissions which may be inadvertently granted via -y (i.e. in a CI environment) and then having no bounds on what those permissions may be looks to me like the very use case Phylum is working to prevent, as it would allow complete control of the environment the extension runs in.

If you make the assumption that someone might just accept random permissions without reading it, then you're screwed anyway. If I can run `bash`, it doesn't matter at all what paths you're limiting me to.

> The benefit is in giving a clear scope to the user of the extension, who will know that if they execute any given extension in any given current directory context, nothing else will be accessed outside of that.

The scope isn't clearly limited though. The user will only know what permissions are granted during the installation process. If he runs the extension 6 weeks later he is never prompted again and has zero clue what access the extension has. In that scenario it's much easier for the user to confirm access to `/var/run` than confirm access to `./test` and remember 6 weeks later that when running it in a directory with a `test` child that it will have access to that specific directory. I honestly don't think absolute paths are a problem here, relative paths are **MUCH** more dangerous but there's just no way around allowing them.

cd-work: > Yes, and that's why we should not allow the user to run bash. I get that there may always be ways to escape a sandbox, but just allowing everything indiscriminately is hardly a solution to that.

That's what will happen when you just `-y` the `bash` run permission though. So in the case that the user just blindly accepts permissions, the point is moot anyway.

> How so? Can we not just prevent upwards directory traversal in any way and be sure we're only ever going to access present working directory?

Present working directory during execution can be literally **any** directory on the developer's machine. Accidentally running things in the wrong directory isn't exactly a super rare occasion. And the confirmation during installation is basically pointless, because the user will have no reference point to what a relative path like `./node_modules` will actually mean when the extension is executed.

andreaphylum: > Present working directory during execution can be literally **any** directory on the developer's machine. Accidentally running things in the wrong directory isn't exactly a super rare occasion. And the confirmation during installation is basically pointless, because the user will have no reference point to what a relative path like `./node_modules` will actually mean when the extension is executed.

We can't really have a relative-path-based permissions system (which isn't even ours; it's Deno's, we only add filtering to it) and also account for user errors. `./node_modules` is always going to refer to "the `node_modules` folder in the current working directory" and it's pretty unambiguous given the semantics of what our extensions are meant to do.

The typical use case is going to be like this: user navigates to their project's directory, and launches `phylum ext-name` which is going to enact behavior relative to the current working directory, like most ecosystems CLIs do.

On the other hand, absolute paths wouldn't possibly make sense since they're going to be different for each machine, save for system files that are likely to be found in the same spot; this is not a use case that we may want to even support in the first place. And neither is directory traversal: what would a permission to access `../../something.yml` or `node_modules/../package.json` even mean?

While this filtering doesn't certainly solve the problem of a user launching a malicious extension in the wrong directory, not having it allows for things that we definitely do not want, and it's the latter situation that this is intended to account for, not the former.

I strongly believe we should keep this in.

cli/src/commands/extensions/mod.rs
```diff
@@ -108,6 +115,41 @@ async fn handle_install_extension(path: &str) -> CommandResult {
     let extension_path = PathBuf::from(path);
     let extension = Extension::try_from(extension_path)?;
 
+    let permissions = extension.permissions();
+
+    // Attempt to construct a `PermissionsOptions` from the `Permissions`
+    // object in order to validate the permissions.
+    let _ = PermissionsOptions::try_from(permissions)?;
+
+    if !accept_permissions && !extension.permissions().is_allow_none() {
```
cd-work, line 124: Wonder if an early return would be better here? It's kinda far down in the method but it would allow us to remove the indentation of a pretty big block of code.

Otherwise my best recommendation would probably be to move that code to a separate method?

cli/src/commands/extensions/mod.rs
```diff
@@ -108,6 +115,41 @@ async fn handle_install_extension(path: &str) -> CommandResult {
...
+        if !Term::stdout().is_term() {
+            return Err(anyhow!(
+                "Can't ask for permissions: not a terminal. For unsupervised usage, consider \
+                 using the -y / --yes flag."
+            ));
+        }
+
+        if !Confirm::new().with_prompt("Do you accept?").interact()? {
+            return Err(anyhow!("permissions not granted, aborting"));
+        }
```
cd-work, line 150: I wonder if this should show up like a hard error, considering the user didn't do anything wrong and everything worked perfectly fine on our end?

cli/src/commands/extensions/permissions.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        let allow_net = value.net().cloned();
+        let allow_run = value.run().cloned();
+
+        Ok(PermissionsOptions {
+            allow_read,
+            allow_write,
+            allow_net,
+            allow_run,
+            ..Default::default()
```
cd-work, line 62: Other options fall back to default, but... there are no other options? This just seems like a footgun waiting to go off.

cli/src/commands/extensions/permissions.rs
```diff
@@ -0,0 +1,112 @@
...
+                "yarnpkg".to_string(),
+                "pip".to_string(),
+                "poetry".to_string(),
+            ]),
+            net: None,
+        };
+        let permissions_options = PermissionsOptions::try_from(&permissions);
+
+        println!("{:?}", permissions_options);
+        assert!(permissions_options.is_ok());
```
cd-work, line 92: This isn't exactly a detailed test that conversion went right, I'd assume it's because pulling the paths back out of `PermissionsOptions` is difficult?

cli/tests/extensions.rs
```diff
@@ -280,6 +280,135 @@ fn conflicting_extension_name_is_filtered() {
     assert!(output.contains("extension was filtered out"));
 }
 
+mod permissions {
```
cd-work, line 283: This file is hundreds of lines long already anyway, why not move this into an actual separate module?

cli/tests/extensions.rs
```diff
@@ -280,6 +280,135 @@ fn conflicting_extension_name_is_filtered() {
...
+    use std::ffi::OsStr;
+
+    use assert_cmd::assert::Assert;
+    use predicates::prelude::*;
+
+    use super::*;
+
+    struct TestCLI {
```
cd-work, line 291:
```suggestion
    struct TestCli {
```

Pretty sure clippy should warn about this too.

cli/tests/extensions.rs
```diff
@@ -280,6 +280,135 @@ fn conflicting_extension_name_is_filtered() {
...
+            Self { tempdir: TempDir::new().unwrap(), cwd: None }
+        }
+    }
+
+    impl TestCLI {
+        fn new() -> Self {
+            Default::default()
+        }
+
+        fn cwd(mut self, cwd: PathBuf) -> Self {
```
cd-work, line 307:
```suggestion
        fn cwd(&mut self, cwd: PathBuf) -> &mut Self {
```

Pretty sure this is the more common way to do this in Rust, this way the API can be used either way.

cli/tests/extensions.rs
```diff
@@ -280,6 +280,135 @@ fn conflicting_extension_name_is_filtered() {
...
+            let mut cmd = Command::cargo_bin("phylum").unwrap();
+
+            cmd.env("XDG_DATA_HOME", self.tempdir.path()).args(args);
+
+            if let Some(cwd) = self.cwd.as_ref() {
+                cmd.current_dir(cwd);
+            }
+
+            cmd.assert()
+        }
```
cd-work, line 326: This is exactly the kinda stuff I'd love to share between normal and end-to-end integration tests. I feel like we'll probably need this a bunch anyway.

cli/tests/extensions.rs
```diff
@@ -280,6 +280,135 @@ fn conflicting_extension_name_is_filtered() {
...
+        let test_cli = TestCLI::new().cwd(fixtures_path().join("permissions"));
+
+        test_cli
+            .install_extension(&fixtures_path().join("permissions").join("correct-read-perms"))
+            .success();
+
+        test_cli
+            .run(&["correct-read-perms"])
+            .success()
+            .stdout(predicate::str::contains("await Deno.readFile"));
```
cd-work, line 355: I don't understand, why should this output `await Deno.readFile`?

cli/tests/fixtures/extensions/permissions/correct-run-perms/PhylumExt.toml
```diff
@@ -0,0 +1,6 @@
+name = "correct-run-perms"
+description = "An extension with correctly specified permissions"
+entry_point = "main.ts"
```
cd-work, line 3: Just a curious idea, we could default to `main.ts`, right? Similar to how Rust loads `src/main.rs` by default. Completely unrelated to this PR, sorry.

cli/tests/fixtures/extensions/permissions/correct-read-perms/main.ts
```diff
@@ -0,0 +1,4 @@
+import { walk, exists } from "https://deno.land/std@0.143.0/fs/mod.ts"
```
cd-work, line 1: I don't see **any** reason why we'd have separate test cases for correct/incorrect. Can't we just merge all of these together? The only difference should be the permissions, not the actual test extension. Anything else is just confusing.

cli/src/commands/extensions/permissions.rs
andreaphylum: Might be a remnant of the early attempts with only a couple of those permissions.

cli/tests/extensions.rs
andreaphylum: It does not. I personally prefer this style but I see the official API guidelines also suggest using the `TestCli` form.

cli/tests/extensions.rs
andreaphylum: Because `main.ts` is reading from `main.ts` and printing it out. 😅

cli/tests/fixtures/extensions/permissions/correct-read-perms/main.ts
andreaphylum: I agree it feels a bit unwieldy to have many fixtures; but these are integration tests, and I want to make sure to test the whole process of installing and running that behaves as expected. If I were to change the permissions in-flight, I could have parts of the code uncovered.

For example, someone could go touch the mechanism for deserializing the manifest (could be as simple as adding `#[serde(rename)]`) and no test would break if we are building the permissions manually.

cli/src/commands/extensions/mod.rs
```diff
@@ -108,6 +115,41 @@ async fn handle_install_extension(path: &str) -> CommandResult {
...
+        print_permissions_list("Write to the following paths:", permissions.write());
+        print_permissions_list("Run the following commands:", permissions.run());
+        print_permissions_list("Access the following domains:", permissions.net());
+
+        if !Term::stdout().is_term() {
+            return Err(anyhow!(
+                "Can't ask for permissions: not a terminal. For unsupervised usage, consider \
+                 using the -y / --yes flag."
+            ));
+        }
```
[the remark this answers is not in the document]
cd-work: Yeah, just meant we can do this before actually asking.

cli/src/commands/extensions/mod.rs
cd-work: We could also exit with a non-zero exit code without producing an explicit error.

cli/src/commands/extensions/permissions.rs
cd-work: The conversion itself is done by us though, not Deno. Here you're only testing that the result is not an error, but it might as well just always create `PermissionsOptions::allow_all`.

I'm not sure we need a dedicated test for this, I doubt we'll ever screw it up. But if we do it would make sense to test that the conversion actually is correct?

cli/tests/extensions.rs
cd-work: Yes, I know clippy will warn on this under certain scenarios, might just be because it's a test that it does not. So we should definitely go with `TestCli`.

cli/tests/extensions.rs
cd-work: Oh wow... That just confused me, got it.

cd-work (changes requested):

cd-work (approved):

cli/tests/extensions/permissions.rs
```diff
@@ -0,0 +1,125 @@
...
+    let test_cli = TestCli::new().cwd(fixtures_path().join("permissions"));
+
+    test_cli
+        .install_extension(&fixtures_path().join("permissions").join("incorrect-net-perms"))
+        .success();
+
+    test_cli
+        .run(&["incorrect-net-perms"])
+        .failure()
+        .stderr(predicate::str::contains(r#"Error: Requires net access to "api.github.com""#));
```
cd-work, line 113: Should probably also have phylum.io for this, just for consistency and so we never hit someone else's stuff during our tests.

cd-work (changes requested):

cd-work (approved):

PR #506, opened 2022-06-28
Add extensions integration tests

### LINKED ISSUE #429: CLI Integration tests
The CLI needs integration tests to confirm the functionality of our tool over time. To start, these can test some basic functionality, but we can add more over time.

# Acceptance Criteria
- [ ] Integration tests are run in CI
- [ ] The API URL and token are configurable (to allow running against prod, staging, or localdev)
- [ ] The tests do not run by default when running `cargo test`
- [ ] The tests do not use the installed `settings.yaml`

cd-work:
This adds a set of extension integration tests, which run against the
staging environment.

Since these tests can take some time, they're hidden behind the
`integration-tests` feature flag, which can also be used for future
long-running integration tests.

To ensure these don't break, they're still run on every commit in CI,
but if execution time is excessive that can be changed in the future
with only changes to the CI.

Closes #429.


integration/src/main.rs
```diff
@@ -0,0 +1,57 @@
+// TODO: Things to figure out:
+//  - Tests all have to be called manually, that sucks!
+//  - Tests run after each other, one failure prevents next one from running
+//  - No status output about success
```
andreaphylum, line 4: I've never needed a custom test harness so far, but I assume someone might have developed one at some point? If that was me, test concurrency would be one of my first concerns, so maybe we can look for something in the wild.

.github/workflows/test.yml
```diff
@@ -70,6 +70,8 @@ jobs:
         with:
           command: test
           args: --all-features
+        env:
+          PHYLUM_API_KEY: ${{ secrets.PHYLUM_TOKEN }}
```
maxrake, line 74: This token is for the _production_ system. The "invalid refresh token" errors in the tests now might be fixed if a different token, for the _staging_ environment is used instead. @furi0us333 added the `PHYLUM_TOKEN` as an organizational secret...and might be the right person to add the token for staging. It appears a relevant account already exists in the password manager.

NOTE: If the organizational secret names are changed (e.g., to differentiate between a PRODUCTION and STAGING token), then all the repos that make use of the current `PHYLUM_TOKEN` org secret will also need to be updated.

integration/src/main.rs
cd-work: I've definitely seen at least one advertised on /r/rust recently. But I think we should be able to go without.

maxrake: I've also seen this bit on page 86 of the "Rust for Rustaceans" book:

> A test harness is generated for each file in `tests/`. Test harnesses are not generated for files in subdirectories under `tests/` to allow you to have shared submodules for your tests. NOTE: If you explicitly want a test harness for a file in a subdirectory, you can opt in to that by calling the file `main.rs`.

I don't know if that helps or not for this review, but it did seem like a useful tidbit.

maxrake:
I have to punt on this review...it just contains too much material I'm not familiar with. I did attempt to understand it, but _still_ not knowing Rust caused me to quickly get in over my head. My apologies to @andreaphylum and/or @kylewillmon for adding this to your plate.

I did find these resources during the review, which may be of help:

* [Blog Post: How to Build a Custom Test Harness in Rust](https://www.infinyon.com/blog/2021/04/rust-custom-test-harness/)
* [The Book: Test Organization](https://doc.rust-lang.org/book/ch11-03-test-organization.html)
* [Test Runner code for `Fluvio` project](https://github.com/infinyon/fluvio/tree/master/tests)

cli/tests/end_to_end/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+use reqwest::StatusCode;
+
+mod extension;
+
+// TODO: Hardcode this?
+const API_URL: &str = "https://api.staging.phylum.io";
+const PROJECT_NAME: &str = "integration-tests";
+
+/// Create config file for the desired environment.
+pub fn create_config(dir: &Path) -> PathBuf {
```
andreaphylum, line 16: I'd also add this as a component of the builder mentioned above.

cli/tests/end_to_end/extension.rs
```diff
@@ -0,0 +1,128 @@
...
+use assert_cmd::Command;
+use tempfile::TempDir;
+
+use crate::end_to_end;
+
+const HEADER: &str = "import { PhylumApi } from 'phylum';";
+
+/// Test Phylum API functions.
+#[tokio::test]
+pub async fn api() {
```
andreaphylum, line 14: I'd use more verbose function names for tests, e.g. api_called_correctly_from_extensions or something along these lines. I've started doing that recently and I feel it makes things a lot clearer, whereas I had to read the body to understand what api meant at first.

integration/src/extension.rs
```diff
@@ -0,0 +1,126 @@
...
+
+use assert_cmd::assert::Assert;
+use assert_cmd::Command;
+use tempfile::TempDir;
+
+const HEADER: &str = "import { PhylumApi } from 'phylum';";
+
+/// Test Phylum API functions.
+#[cfg(test)]
+pub fn api(config: &Path) {
```
andreaphylum, line 12: This file looks like a duplicate of `end_to_end/extension.rs`, is this intended?

andreaphylum (changes requested):

cli/tests/end_to_end/extension.rs
```diff
@@ -0,0 +1,128 @@
...
+        promises.push(PhylumApi.getUserInfo());
+        await Promise.all(promises);
+    "#,
+        |assert| {
+            assert.success();
+        },
+    );
+}
+
+fn with_extension<F>(config: &Path, code: &str, mut f: F)
```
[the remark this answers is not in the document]
cd-work: I mean you'll always need the code, so you probably don't want that in the builder pattern. So in that case I feel like there's no point to change it right away, but we might as well just wait until we have testcases that require it? Keep it simple?

cli/tests/end_to_end/extension.rs
cd-work: I think the key is to avoid redundancy. We're already in the extensions module so this test is already `extensions::api`, making it `extensions::api_called_correctly_from_extensions` doesn't provide much additional benefit, of course you're checking that it works correctly because that's the default you'd expect for any test that doesn't point out otherwise, you know it's an extension because that's the module we're in and already part of the test name.

So I don't mind making the function names more explicit, but it should provide additional information that doesn't exist yet, rather than just making it harder to decipher the concrete purpose of the test from its name.

integration/src/extension.rs
cd-work: No.

cli/tests/end_to_end/mod.rs
andreaphylum: Indeed -- we probably want overall improvements to the structure of the test helpers, but I'd say this is fine for now.

cli/tests/end_to_end/extension.rs
andreaphylum: That specific one was a random suggestion. The core concept is that `extensions::api` is also not immediately clear to me and anything that makes this clearer would be better, really.

kylewillmon (approved):

PR #507, opened 2022-06-29
Add args support for deno extensions

### LINKED ISSUE #416: Extensions: Allow reading CLI arguments from extensions
# Overview

There should be an API for extensions to query for the parameters passed after the extension subcommand. This way they can have their own CLI handling.

# Acceptance Criteria
- [ ] Extensions can read any CLI parameter after their subcommand
- [ ] Extensions cannot access CLI parameters before their subcommand

cd-work:
This makes it possible for extensions to read all parameters passed
after their subcommand using `Deno.args`.

Closes #416.


cd-work:
Hoping that we can maybe get https://github.com/clap-rs/clap/pull/3885 pushed through relatively quickly.

cli/src/commands/extensions/extension.rs
```diff
@@ -124,8 +125,18 @@ impl Extension {
 
     /// Execute an extension subcommand.
     pub async fn run(&self, api: BoxFuture<'static, Result<PhylumApi>>) -> CommandResult {
+        let args = env::args()
+            .skip_while(|arg| arg != self.name())
+            .skip(1)
+            .collect();
```
cd-work, line 131: This approach is a little unfortunate, theoretically this means that an extension with the name `phylum` will get all arguments and other similar conflicts with existing parameters. But I'm not sure we have much better options?

kylewillmon: The issue of an extension named `phylum` can be resolved with an extra `.skip(1)`. But that doesn't solve other problems...

```suggestion
        let args = env::args()
            .skip(1)
            .skip_while(|arg| arg != self.name())
            .skip(1)
            .collect();
```

cd-work: > The issue of an extension named phylum can be resolved with an extra .skip(1). But that doesn't solve other problems...

Yeah, was considering to mention that but it's really irrelevant. The problem persists either way.

kylewillmon (approved):

kylewillmon (approved):

PR #510, opened 2022-06-30
Remove redundant -extension suffix from tests

cd-work:
There's no need to repeat over and over again that these test fixtures
are extensions, considering they're already in the extension directory.
To ensure we don't have to repeat ourselves excessively and make naming
more clear and simple, the -extension suffix is removed from all test
extensions.


cd-work:
It would probably possible to throw this in with the other extension test refactorings that are outstanding, but I think keeping things small is probably a good thing and this seems like a nice little step in the right direction.

kylewillmon (approved):

maxrake (approved):

PR #516, opened 2022-07-04
[WIP] Add extension documentation

cd-work:


docs/extensions/extensions.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Phylum CLI Extensions
+category: TODO
```
cd-work, line 3: @furi0us333 I think it makes sense to put the documentation for writing extensions in a different section. You can generate a new category ID for that once this is settled, right?

docs/extensions/example.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+When replacing the `console.log` with this function call and executing our
+extension, you'll run into the following error:
+
+TODO: This error likely needs to be updated on release.
+
+```
+Execution failed: Error: Requires write access to "./duplicates.txt"
+```
```
cd-work, line 219: I have no idea yet how we should document this since I don't know what error we'll end up with. I would recommend opening an issue and then updating it once the error message is fixed?

docs/extensions/api.md
```diff
@@ -0,0 +1,30 @@
...
+Deno's API is built into the Deno runtime, providing access to all external
+interfaces like Network, Disk, or the terminal console. All available
+functionality is documented in [Deno's API docs]. This functionality is
+available to all extensions without any imports.
+
+Additionally, Deno also provides a complementary standard library. This includes
+utility functions for several commonly used structures like collections, http,
+and async. These modules can be imported using the URLs documented in Deno's STD
+documentation or by downloading them and including individual modules as files.
+All STD functionality is documented in [Deno's STD docs]
```
andreaphylum, line 21:
```suggestion
All STD functionality is documented in [Deno's STD docs].
```

In general I'd use an explicit "Deno's standard library" unless the uppercase STD has a special meaning.

docs/extensions/example.md
andreaphylum: Agreed. In the meantime we may want to add a note that mentions that that error is coming straight from Deno and/or is temporary and will be changed, just in case upstream doesn't get the changes we need in time.

docs/extensions/manifest_format.md
```diff
@@ -0,0 +1,93 @@
...
+The paths also need to match **exactly** with the process executed by the
+extension. `/usr/bin/curl` cannot be executed when `curl` was requested as
+permission and vice versa.
+
+### Net
+
+Net permissions list domains which can be accessed by the extension.
+
+Network permissions only describe the domains and subdomains an extension has
+access to, regardless of its path segments. Access to a domain does not
```
andreaphylum, line 88:
```suggestion
access to, regardless of its path segments or protocol specifiers. Access to a domain does not
```

docs/extensions/manifest_format.md
```diff
@@ -0,0 +1,93 @@
...
+automatically grant access to all of its subdomains.
+
+If the requested domain requests a redirect, you'll also require permissions to
```
andreaphylum, line 91:
```suggestion
If the requested domain requests a redirect, you'll also need to require permissions to
```

docs/extensions/quickstart.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Quickstart
+category: TODO
+hidden: true
+---
+
+## Creating an extension
+
+The best way to get started with writing your own Phylum CLI extension, is to
```
andreaphylum, line 9:
```suggestion
The best way to get started with writing your own Phylum CLI extension is to
```

extensions/duplicates/PhylumExt.toml
```diff
@@ -0,0 +1,5 @@
+name = "duplicates"
+entry_point = "main.ts"
+
+# [permissions]
+# write = ["./duplicates.txt"]
```
andreaphylum, line 5: Do we plan on enabling/using these permissions in the future? If we do, then I'd say it's fine to keep this.

andreaphylum (changes requested):

docs/extensions/manifest_format.md
cd-work: I believe the existing wording is correct here and superior to your suggestion.

extensions/duplicates/PhylumExt.toml
cd-work: If you look at the example, you'll see that the commented-out lines here are variations to the example introduced at the end of the documentation to introduce the concept of permissions.

The goal is to provide all the necessary code to ensure users can trace back their problems if they fail to follow along.

docs/extensions/manifest_format.md
```diff
@@ -0,0 +1,93 @@
...
+access the redirect target. It's easiest to just specify the redirect target
+directly when making requests.
```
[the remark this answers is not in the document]
cd-work: > where making requests directly to the redirect target is impractical, the user will have to request permissions for both domains.

That makes no sense though. You **MUST** specify both URLs, so you know the redirect target to begin with. It's completely pointless to specify both when you're fully aware of the redirect target anyway. There's no scenario where it's impractical to specify the redirect target, when the alternative is to specify the redirect target plus the original domain.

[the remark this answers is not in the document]
cd-work: Fair enough, since permissions only focus on domains I was disregarding the path elements, but I suppose since redirects might append both path elements and subdomains that is a valid usecase.

docs/extensions/extensions.md
furi0us333: I added a new category in the docs, please use `62c5cb137dbdad00536291a6` to ensure relevant documentation lands under the correct category.

docs/extensions/example.md
```diff
@@ -0,0 +1,234 @@
...
+
+When replacing the `console.log` with this function call and executing our
+extension, you'll run into the following error:
+
+```
+❗ Error: Execution failed caused by: Error: Requires write access to "./duplicates.txt", run again with the --allow-write flag
+```
+
+The `--allow-write` flag here comes directly from Deno and is **not** supported
+by Phylum's CLI extensions.
```
kylewillmon, line 220: If we're going to document this, it might be worth adding a link to #520 so that users can track progress of the fix (or maybe even submit a PR).

kylewillmon (approved):
These docs look great!

Cc: @furi0us333 for a heads up that we'll be adding a section to the docs for extensions. This will likely require some manual manipulation in readme.com to fixup the page layout

docs/extensions/example.md
cd-work: I'm not sure our documentation is the right place for that? Would seem kinda strange to basically have "TODOs" in our documentation?

That's why I decided to remove it and instead move it to the #520 description.

cd-work: Sure, sounds good.

kylewillmon (approved):

PR #526, opened 2022-07-06
Prevent extension install with invalid permissions

### LINKED ISSUE #519: Path segment should be removed when prompting for permissions
Currently when an extension asks for a URL permission which includes path elements, the entire URL will be printed (e.g. `phylum.io/path/segments`). However deno's extension system is solely based on exactly matching subdomain and domain, while ignoring any path segments.

To prevent extensions from tricking the user into granting unexpected permissions, we should strip the path segments when prompting for permissions to clearly communicate that everything under the domain will be accessible, regardless of path segments.

cd-work:
Since Deno's check for networking permissions only considers the domain,
the scheme and path can be disregarded and should be stripped to prevent
confusing users when prompting for permissions.

While it would be possible to strip this and continue while warning the
user, preventing an install ensures that no extensions with invalid net
permissions are ever published. It also avoids printing warnings when
deserializing existing extensions during startup for clap's CLI.

Closes #519.

### DIFF
--- a/test/PhylumExt.toml
+++ b/test/PhylumExt.toml
@@ -1,6 +1,0 @@
-name = "test"
-description = "API injection test"
-entry_point = "main.ts"
-
-[permissions]
-net = ["http://blog.phylum.io/test/ing"]
--- a/test/PhylumExt.toml
+++ b/test/PhylumExt.toml
@@ -1,0 +1,6 @@
+name = "test"
+description = "API injection test"
+entry_point = "main.ts"
+
+[permissions]
+net = ["http://blog.phylum.io/test/ing"]
--- a/cli/src/commands/extensions/permissions.rs
+++ b/cli/src/commands/extensions/permissions.rs
@@ -1,7 +1,9 @@
 use std::path::PathBuf;
 
 use deno_runtime::permissions::PermissionsOptions;
-use serde::{Deserialize, Serialize};
+use serde::de::Error as _;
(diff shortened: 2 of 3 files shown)


test/PhylumExt.toml
```diff
@@ -0,0 +1,6 @@
+name = "test"
+description = "API injection test"
+entry_point = "main.ts"
+
+[permissions]
+net = ["http://blog.phylum.io/test/ing"]
```
kylewillmon, line 6: Was this just for your local testing? It's not used in any test and isn't in the normal location.

andreaphylum:
I'd add either a simple unit test with a successful and a failing test case, deserializing a hardcoded toml string, or an integration test with a fixtured extension (we can do that in a separate issue + PR). The rest looks good to me.

test/PhylumExt.toml
cd-work: Correct.

cd-work:
> I'd add either a simple unit test with a successful and a failing test case, deserializing a hardcoded toml string, or an integration test with a fixtured extension (we can do that in a separate issue + PR). The rest looks good to me.

Completely agree, I just thought of the same thing. Will add a test.

andreaphylum (approved):

kylewillmon (dismissed):

kylewillmon (approved):

Issue #413, opened 2022-05-24
Extension tests: refactor `TmpDir` behavior


andreaphylum:
We want to refactor some commonly repeated patterns in the extensions tests. Sample code as per @matt-phylum 's suggestion:

```rust
struct TestEnvironment {
    data_home: TmpDir,
}

impl TestEnvironment {
    fn new() -> Self {
        Self {
            data_home: TmpDir::new()
        }
    }

    fn phylum_command(&self) -> Command {
        Command::cargo_bin("phylum")
            .unwrap()
            .env("XDG_DATA_HOME", &self.data_home)
    }

    fn install_test_extension(&self, name: &str) -> Assert {
        self.phylum_command()
            .args(&["extension", "add"])
            .arg(fixtures_path().join(name))
            .assert()
    }

    fn extension_home(&self) -> PathBuf {
        PathBuf::from(&self.tmp_dir)
            .join("phylum")
            .join("extensions")
    }
}
```

```rust
let test_env = TestEnvironment::new();
test_env.install_test_extension("sample-extension").success();

let extension_path = test_env.extension_home().join("sample-extension");
assert!(WalkDir::new(&extension_path).into_iter().count() > 1);

test_env.phylum_command()
    .args(&["extensions", "remove", "sample-extension"])
    .assert()
    .success();

assert!(!extension_path.exists());
```

cd-work:
This can probable be merged with https://github.com/phylum-dev/cli/issues/492?

PR #530, opened 2022-07-07
Extensions tests refactor

### LINKED ISSUE #492: Extension tests refactoring
Extension tests in `cli/tests/extensions.rs` currently show a lot of similar/repeated code.

- [x] Write helper for running the CLI (see #413)
- [x] Switch to the `predicates` crate everywhere for matching outputs
- [x] Group tests into modules

andreaphylum:
Closes #492 and #413.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -663,9 +663,9 @@ dependencies = [
 
 [[package]]
 name = "crypto-common"
-version = "0.1.4"
+version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5999502d32b9c48d492abe66392408144895020ec4709e549e840799f3bb74c0"
+checksum = "2ccfd8c0ee4cce11e45b3fd6f9d5e69e0cc62912aa6a0cb1bf4617b0eba5a12f"
 dependencies = [
  "generic-array",
  "typenum",
@@ -771,7 +771,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3495912c9c1ccf2e18976439f4443f3fee0fd61f424ff99fde6a66b15ecb448f"
 dependencies = [
  "cfg-if",
- "hashbrown 0.12.1",
+ "hashbrown 0.12.2",
  "lock_api",
  "parking_lot_core 0.9.3",
 ]
@@ -1800,9 +1800,9 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
-version = "0.12.1"
+version = "0.12.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "db0d4cf898abf0081f964436dc980e96670a0f36863e4b83aaacdb65c9d7ccc3"
+checksum = "607c8a29735385251a339424dd462993c0fed8fa09d378f259377df08c126022"
 
 [[package]]
 name = "hashlink"
@@ -1936,9 +1936,9 @@ checksum = "9a3a5bfb195931eeb336b2a7b4d761daec841b97f947d34394601737a7bba5e4"
 
 [[package]]
 name = "hyper"
-version = "0.14.19"
+version = "0.14.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "42dc3c131584288d375f2d07f822b0cb012d8c6fb899a5b9fdb3cb7eb9b6004f"
+checksum = "02c929dc5c39e335a03c405292728118860721b10190d98c2a0f0efd5baafbac"
 dependencies = [
  "bytes",
  "futures-channel",
@@ -2001,7 +2001,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "10a35a97730320ffe8e2d410b5d3b69279b98d2c14bdb8b70ea89ecf7888d41e"
 dependencies = [
  "autocfg 1.1.0",
- "hashbrown 0.12.1",
+ "hashbrown 0.12.2",
  "serde",
 ]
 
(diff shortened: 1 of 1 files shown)


cli/Cargo.toml
```diff
@@ -11,6 +11,14 @@ selfmanage = []
...
 extensions = ["dep:deno_runtime", "dep:deno_core", "dep:deno_ast"]
 
+[[test]]
+name = "extensions"
+path = "tests/extensions/mod.rs"
+
+[[test]]
+name = "end_to_end"
+path = "tests/end_to_end/mod.rs"
+
```
kylewillmon, line 21: @cd-work and I discussed this and many other options when the `end_to_end` integration test was first added. Using this instead of auto-discovery is confusing. Also, this makes it more difficult for the tests to share code since you can no longer just put `mod common;` in the top-level file and then `use crate::common;` where necessary.

kylewillmon (changes requested):
The move of integration tests from `tests/extensions.rs -> tests/extensions/mod.rs` and `tests/end_to_end.rs -> tests/end_to_end/mod.rs` is unrelated to the refactor at issue here and needs a larger discussion. It should be moved to its own PR or issue where it can be further discussed.

cli/Cargo.toml
cd-work: I mean at the very least with this approach there wouldn't be a reason to separate end-to-end and other integration tests? That should make sharing a little easier at least.

andreaphylum: What would be the ideal structure in place of this? A couple options I could see:

- Single top-level integration tests file which includes test sub-modules (has the drawback of having everything together, if this is undesirable)
- Separate but discoverable `extensions.rs` + `extensions/<module>.rs` and `end_to_end.rs` + `end_to_end/<module.rs>` (has the drawback of warnings in common code)

Cargo.toml
```diff
@@ -2,5 +2,6 @@
 resolver = "2"
 members = [
   "cli",
+  "cli/test-utils",
```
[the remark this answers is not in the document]
andreaphylum: I reverted this. CI succeeds but it does [produce these warnings](https://github.com/phylum-dev/cli/runs/7250429294?check_suite_focus=true#step:10:14) on compilation of the test crates with all features enabled.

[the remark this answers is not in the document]
kylewillmon: Ok. So it's the `dead_code` lint that's triggering. That only triggers on unexported items, which means you can get rid of all of those by simply changing `mod common;` to `pub mod common;` (which is exactly how you're getting rid of them when you move the code to a separate crate)

kylewillmon (dismissed):
This looks much better.

It is still generating dead code warnings when `cargo test` is run (without `--all-features`). The fix for this is either to use the same `pub mod common` trick or to add a bunch of `#[cfg(feature = "end-to-end-tests")]` guards

cd-work:
> It is still generating dead code warnings when cargo test is run (without --all-features). The fix for this is either to use the same pub mod common trick or to add a bunch of #[cfg(feature = "end-to-end-tests")] guards

I think this should definitely be addressed. It's a common module, so if there's code in there that isn't shared it should either be marked as `#[allow(unused)]` explicitly, or moved to where it is actually used. I don't think we should ever have any warnings emitted.

cli/tests/common/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+pub use assert_cmd::assert::Assert;
+pub use assert_cmd::Command;
+use phylum_cli::api::{PhylumApi, PhylumApiError, ResponseError};
+use phylum_cli::config::{AuthInfo, Config, ConnectionInfo};
+use phylum_types::types::auth::RefreshToken;
+pub use predicates::prelude::*;
+use reqwest::StatusCode;
+use tempfile::TempDir;
+
+const HEADER: &str = "import { PhylumApi } from 'phylum';";
```
cd-work, line 14: Being a bit more descriptive on the name here is probably a good idea.

Just a really rough idea:

```suggestion
const TS_PHYLUM_IMPORT: &str = "import { PhylumApi } from 'phylum';";
```

cli/tests/end_to_end/extension.rs
```diff
@@ -1,128 +1,97 @@
...
+    let lockfile = create_lockfile(test_cli.temp_path());
+    let project = create_project().await;
+
+    {
+        test_cli
+            .create_extension("console.log(await PhylumApi.getUserInfo())")
+            .run()
+            .success()
+            .stdout(predicates::str::contains("email"));
+    }
```
cd-work, line 17: What's the point of the scopes here?

We already have each individual API separated by whitespace, if separation is so important we need to have useless scopes, they should probably be moved into separate functions.

Unless the scopes here are necessary to make things compile?

cli/tests/integration.rs
```diff
@@ -0,0 +1,10 @@
+#[path = "common/mod.rs"]
+mod common;
+
+#[cfg(feature = "end-to-end-tests")]
+#[path = "end_to_end/mod.rs"]
+mod end_to_end;
+
+#[cfg(feature = "extensions")]
+#[path = "extensions/mod.rs"]
+mod extensions;
```
cd-work, line 10: Are these path annotations necessary? I feel like they should load properly automatically?

cd-work (changes requested):
Was looking at it anyway, so I thought I'd chime in with my 2 cents too.

cli/tests/end_to_end/extension.rs
andreaphylum: The `create_extension` method returns a RAII guard that uninstalls and deletes the extension once it goes out of scope. We could give a randomized name to the extension instead of the hardcoded `sample` like we had in the previous `end_to_end` tests but I felt the RAII guard and scoping were a viable solution as well, plus we don't have to worry about collisions and other similar concerns. We may want to document this, though.

cli/tests/common/mod.rs
andreaphylum: I was thinking to move this in `TestExtension::new`, maybe get rid of the constant altogether and hardcode it in the call to `fs::write` as it's not being used anywhere else. What do you think?

cd-work: I don't see a problem with that.

cd-work (approved):

PR #533, opened 2022-07-07
Use POST for job submission instead of PUT

kylewillmon:
The `/api/v0/data/jobs` endpoint is adding support for the more correct `POST` method. The CLI should update to match as soon as this method is available in prod.


cd-work (approved):
Confirmed this works with `phylum analyze`.

PR #539, opened 2022-07-11
Validate file permissions in Phylum extension API

### LINKED ISSUE #523: Extensions should check permissions for phylum API functions
The current implementation for CLI extensions does not restrict its API based on the permissions of the extensions. This could lead to some permission bypass scenarios.

For example, the `parse_lockfile()` function can be used on files that the extension does not have permission to read. With the right parser and file combination, this may allow the caller to read some contents of the file. As we add support for more types of lock files, we open up more potential for this attack or others like it.

We should review our API and add permission checks as necessary to avoid this sort of potential attack.

cd-work:
Since some of our own APIs access user-defined files and return parts of
these files, it's necessary to validate the access against the
extension's permissions.

Closes #523.

### DIFF
--- a/cli/src/commands/extensions/permissions.rs
+++ b/cli/src/commands/extensions/permissions.rs
@@ -6,8 +6,15 @@ use serde::de::Error as _;
 use serde::{Deserialize, Deserializer, Serialize};
 
 /// Resource permissions.
-#[derive(Serialize, Deserialize, Clone, Default, Debug)]
+#[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, Eq)]
+#[serde(transparent)]
 pub struct Permission(Option<Vec<String>>);
+
+impl From<Option<Vec<String>>> for Permission {
+    fn from(raw: Option<Vec<String>>) -> Self {
+        Permission(raw)
+    }
+}
 
 impl Permission {
     // XXX In Deno, `Some(vec![])` actually means "allow all". We never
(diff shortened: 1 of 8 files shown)


cli/src/commands/extensions/permissions.rs
```diff
@@ -1,29 +1,53 @@
...
+    }
+
+    /// Check if access to resource is permitted.
+    pub fn validate(&self, resource: &String, resource_type: &str) -> Result<()> {
+        if self.get().map_or(false, |allowed| allowed.contains(resource)) {
+            Ok(())
+        } else {
+            Err(anyhow!("Requires {resource_type} access to {resource:?}"))
+        }
+    }
```
cd-work, line 28: I pass the `resource_type` in here just for the error, but I'd love to hear about ideas how this could be automatically inferred from the `Permission` struct.

I've considered adding like a trait bound with a phantomdata on it, then specify permissions as `Permission<Net>` where `Net` has a function to return `"net"`, but that seems like **a lot** of additional boilerplate.

cli/tests/common/mod.rs
```diff
@@ -1,10 +1,13 @@
...
 use std::ffi::OsStr;
+use std::fs::{self, File};
+use std::io::Write;
 use std::path::{Path, PathBuf};
-use std::{env, fs};
 
 pub use assert_cmd::assert::Assert;
 pub use assert_cmd::Command;
 use phylum_cli::api::{PhylumApi, PhylumApiError, ResponseError};
+use phylum_cli::commands::extensions::permissions::Permissions;
```
andreaphylum, line 10: I'm not sure whether it would be better to condition the whole module on `#[cfg(feature = "extensions")]` or to put this only where applicable. We aren't really running integration tests that don't have this feature yet, but might in the future, unless we plan on removing the feature eventually.

andreaphylum (changes requested):

cli/tests/common/mod.rs
cd-work: I've just added a bunch of attributes, should hopefully be fine until the feature is published.

andreaphylum (approved):

cli/src/commands/extensions/permissions.rs
andreaphylum: I think it is not necessary to have this right now, but going forward if we realize we want some form of validation having specialized permission types could be an option there. Those could implement some trait with default methods. It is also quite a bit of boilerplate, so I'd say we're still in the same league.

Another option could be to just add an enum field to the permissions object, which maybe could be a bit less heavy handed on boilerplate than using phantom data but is also non-zero cost.

PR #546, opened 2022-07-13
Fix clippy ref warning in extension API

cd-work:
This fixes an issue where clippy would complain about references being
held across await points in our async extension APIs.

The underlying issue is that mutable access to our `RefCell` state would
panic when another asynchronous method held a reference to it across
await points.

This patch does not fix the underlying issue, but instead provides
guidelines to help developers avoid the pitfalls with mutable references
in asynchronous code.

### DIFF
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -33,7 +33,7 @@ async fn analyze(
 ) -> Result<ProjectId> {
     // Ensure extension has file read-access.
     let state = ExtensionState::from(op_state);
-    state.permissions().read.validate(&lockfile, "read")?;
+    state.permissions.read.validate(&lockfile, "read")?;
 
     let api = state.api().await?;
 
@@ -168,7 +168,7 @@ async fn parse_lockfile(
 ) -> Result<Vec<PackageDescriptor>> {
     // Ensure extension has file read-access.
     let state = ExtensionState::from(op_state);
-    state.permissions().read.validate(&lockfile, "read")?;
+    state.permissions.read.validate(&lockfile, "read")?;
 
     // Fallback to automatic parser without lockfile type specified.
     let lockfile_type = match lockfile_type {
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -1,12 +1,12 @@
-use std::cell::{Ref, RefCell};
-use std::ops::Deref;
+//! Extension API functions.
+
+use std::cell::RefCell;
 use std::path::Path;
 use std::rc::Rc;
 use std::str::FromStr;
 
 use anyhow::{anyhow, Context, Error, Result};
 use deno_runtime::deno_core::{op, OpDecl, OpState};
-use futures::future::BoxFuture;
 use phylum_types::types::auth::{AccessToken, RefreshToken};
 use phylum_types::types::common::{JobId, ProjectId};
 use phylum_types::types::job::JobStatusResponse;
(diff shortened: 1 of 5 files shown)


cli/src/commands/extensions/api.rs
```diff
@@ -50,6 +49,23 @@ impl<T: Unpin> OnceFuture<T> {
...
+// which could be held across await points.
+//
+// When making a field of `ExtensionState` mutably accessible, the mutation
+// should occur internally through SYNCHRONOUS methods (e.g. `with_x(|x| ...)`),
+// so holding a mutable reference is impossible.
+//
+// If a mutable reference MUST be held across await points, like `PhylumApi`,
+// its synchronization should force blocking through an async-safe `Mutex`. This
+// will stall all extension API methods trying to access this field, so ensure
+// the blocking duration is minimal.
```
cd-work, line 67: I've discussed this with @andreaphylum in Slack and I think this is a good summary of how to deal with the potential pitfalls here.

Since @kylewillmon hasn't been part of this conversation, I think it would be good to get the first review of this explanation from you, just to ensure this documentation properly explains the underlying issue even without me or @andreaphylum elaborating in this PRs comments.

cd-work: Hm maybe I should clarify the section about what you can do when you **need** mutable references.

The idea with the `Ref` is fine is basically that the `Rc<RefCell<OpState>>` reference we use to access the `ExtensionState` can be held across await points because we never want to mutably access `ExtensionState`, which is also why I pointed out it should only ever be accessed through `ExtensionStateRef` (rather than bypassing it and calling `borrow_mut` or something).

So in that sense `Ref` is always perfectly safe, as long as there never will be mutable access, you can hold and access however many immutable references you want, because that's not an issue.

The problem only arises when you need to access things mutably. In that scenario you should **never** hold a reference to that data point (mutable or immutable) across await points. And methods like `with_x(|x| ...)` are one way you can try to enforce this.

@kylewillmon So I think you got the gist of it appropriately, but the comment should probably be updated when it comes to creating safe mutable APIs. Does this explanation make sense to you and if so, do you have suggestions for how to put this into one paragraph?

kylewillmon: > So in that sense `Ref` is always perfectly safe, as long as there never will be mutable access, you can hold and access however many immutable references you want, because that's not an issue.

This is where I think you are mistaken. The `RefCell<OpState>` will panic if any code tries to create a `RefMut` while there are any outstanding `Ref` objects. Even if we never need a `RefMut` ourselves for the `ExtensionState` object contained within, the `RefCell` has no way to distinguish that from any `RefMut` created by `deno_core` to mutate other pieces of `OpState`. Therefore, we are risking a panic anytime we hold a `Ref` across an await point since there may be some `deno_core` code that runs in that time.

Hiding the `Ref` inside of `ExtensionStateRef` just hides the clippy lint and prevents us from being alerted about this potential problem in the future.

cd-work: > This is where I think you are mistaken. The RefCell<OpState> will panic if any code tries to create a RefMut while there are any outstanding Ref objects. Even if we never need a RefMut ourselves for the ExtensionState object contained within, the RefCell has no way to distinguish that from any RefMut created by deno_core to mutate other pieces of OpState. Therefore, we are risking a panic anytime we hold a Ref across an await point since there may be some deno_core code that runs in that time.

This is making the assumption that deno core might mutably access OpState while we're awaiting. As far as I know that doesn't happen, but we could just add another layer of `Rc` in between to resolve that. The issue stays the same either way though. I just decided not to add an `Rc` because it seems to work fine right now.

> Hiding the Ref inside of ExtensionStateRef just hides the clippy lint and prevents us from being alerted about this potential problem in the future.

You say that, but even in your PR we hold a ref across an await point. And there's not much we can do about this. That's the entire reason behind my attempt to document it. We cannot avoid this issue and Clippy will never catch all instances of it (some of which might be justified). So we need some way to document things.

cli/src/commands/extensions/state.rs
```diff
@@ -0,0 +1,94 @@
+//! Shared extension state.
```
cd-work, line 1: A separate module here allows us to control access through visibility modifiers.

cli/src/commands/extensions/state.rs
```diff
@@ -0,0 +1,94 @@
...
+//
+// Using `RefCell` to avoid locking should only be done when the value is
+// private and never held across await points internally. Prefer [`Cell`] if the
+// inner value only needs to be replaced.
+//
+/// Extension state the APIs have access to.
+struct ExtensionStateInner {
+    api: Mutex<OnceFuture<Result<Rc<PhylumApi>>>>,
+    permissions: Permissions,
+}
```
cd-work, line 61: This is intentionally private without `Deref`. Unfortunately this means that access to things like `Permissions` which are not public will require a plumbing method in `ExtensionState`, but this way it's impossible to directly extract the `ExtensionStateInner` from the `OpState` without going through `ExtensionState`.

TODO(chris): Is this actually necessary? If `ExtensionState` is stored on `OpState`, typesystem shouldn't allow extracting `Rc<ExtensionStateInner>` directly.

cli/src/commands/extensions/state.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    pub fn permissions(&self) -> &Permissions {
+        &self.0.permissions
+    }
+}
+
+impl From<Rc<RefCell<OpState>>> for ExtensionState {
+    fn from(op_state: Rc<RefCell<OpState>>) -> Self {
+        op_state.borrow().borrow::<ExtensionState>().clone()
```
cd-work, line 92: Unfortunately there's nothing stopping someone from doing this outside this module, potentially holding a ref to `OpState` across await points.

I'm not sure if this can be avoided, but I'm open for suggestions.

andreaphylum: Tracking refs to `OpState` is probably outside of our scope of concern -- we cannot control what Deno does with it in the first place.

I believe that as long as we have a top-level way of getting an immutable reference that can be held across await points (could be `Rc` or even `Arc`, but not `Ref`/`RefMut`), and we know that reference is valid, we can control the rest of the flow via interior mutability and do our own synchronization.

cd-work: Yeah I agree, someone could always hold a borrow of some other random opstate, there's no way we can prevent that.

cli/src/commands/extensions/state.rs
```diff
@@ -0,0 +1,102 @@
...
+    api: Mutex<OnceFuture<Result<Rc<PhylumApi>>>>,
+}
+
+impl ExtensionStateInner {
+    pub async fn api(&self) -> Result<Rc<PhylumApi>> {
+        // This mutex guard is only useful for synchronizing mutable access while
+        // awaiting the PhylumApi future. Subsequent access to the API is
+        // immediate and will not hold the Mutex for extended periods of time.
+        let mut guard = self.api.lock().await;
+        Ok(guard.get().await.as_ref().map_err(|e| anyhow!("{:?}", e))?.clone())
```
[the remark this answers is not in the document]
kylewillmon: It would probably involve changing the return type here to be `Result<MutexGuard<&PhylumApi>>` or something crazy.... `Rc` is definitely simpler.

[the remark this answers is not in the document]
cd-work: You don't want to hold a mutex guard though, because having a mutex guard while interacting with the API will mean that no other function can interact with the API for that entire period.

So that's a really bad idea.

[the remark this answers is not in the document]
cd-work: Honestly just `Cell` might work?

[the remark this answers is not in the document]
cd-work: Nvm it definitely doesn't. There's a reason we have `Mutex` here and I'm pretty sure it would be difficult to avoid. We need some barrier to synchronize on and the `Mutex` lock provides exactly that.

kylewillmon (approved):

PR #547, opened 2022-07-13
Avoid holding a RefCell reference across await points

kylewillmon:
This patch fixes the clippy warning about holding references across await points. Here is an example of the warnings currently being emitted by clippy:

```
warning: this `RefCell` reference is held across an `await` point
   --> cli/src/commands/extensions/api.rs:114:9
    |
114 |     let op_state = state.borrow();
    |         ^^^^^^^^
    |
    = note: `#[warn(clippy::await_holding_refcell_ref)]` on by default
    = help: ensure the reference is dropped before calling `await`
note: these are all the `await` points this reference is held through
   --> cli/src/commands/extensions/api.rs:114:5
    |
114 | /     let op_state = state.borrow();
115 | |     let extension_state = op_state.borrow::<ExtensionState>();
116 | |     extension_state.permissions.read.validate(&lockfile, "read")?;
117 | |     drop(op_state);
...   |
139 | |     Ok(job_id)
140 | | }
    | |_^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#await_holding_refcell_ref
```

I am posting this as an alternative to #546 to start a discussion. I believe this is a simpler solution to the problem that clippy is warning about, but it is possible that I have missed some other issue addressed by that PR.

### DIFF
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -110,11 +110,15 @@ async fn analyze(
     project: Option<String>,
     group: Option<String>,
 ) -> Result<ProjectId> {
-    // Ensure extension has file read-access.
-    let op_state = state.borrow();
-    let extension_state = op_state.borrow::<ExtensionState>();
-    extension_state.permissions.read.validate(&lockfile, "read")?;
-    drop(op_state);
+    // The block ensures that `op_state` is not held across an await. This could
+    // also be handled with `drop(op_state);`, but that leads to a false
+    // positive clippy lint (see rust-lang/rust-clippy#6353)
+    {
+        // Ensure extension has file read-access.
+        let op_state = state.borrow();
+        let extension_state = op_state.borrow::<ExtensionState>();
+        extension_state.permissions.read.validate(&lockfile, "read")?;
+    }
 
     let api = ApiRef::from(state).await?;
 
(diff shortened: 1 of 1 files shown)


cli/src/commands/extensions/api.rs
```diff
@@ -110,11 +110,14 @@ async fn analyze(
...
-    extension_state.permissions.read.validate(&lockfile, "read")?;
-    drop(op_state);
+    // The block ensures that `op_state` is not held across an await. This could also be handled
+    // with `drop(op_state);`, but that leads to a false positive clippy lint (see rust-lang/rust-clippy#6353)
+    {
+        // Ensure extension has file read-access.
+        let op_state = state.borrow();
+        let extension_state = op_state.borrow::<ExtensionState>();
+        extension_state.permissions.read.validate(&lockfile, "read")?;
+    }
```
kylewillmon, line 120: It's likely cleaner to simply extract this code to a separate function (e.g., `check_permissions(op_state, &lockfile, "read")?;`), which would avoid the odd-looking block and the comment explaining why the odd-looking block is necessary.

cd-work (changes requested):
I don't think this appropriately addresses the underlying issue at all. The problem is never with borrowing from the `OpState`, since that is only ever immutably borrowed. So all you're doing here is fixing a clippy lint which is a false-positive.

The real underlying issue is much more severe and affects our extension API to a degree that Clippy cannot detect. As such any wrapper types that are used or APIs accidentally exposing things incorrectly will lead to the underlying issue being actually triggered, without Clippy warning about it.

#546 is an attempt at making these issues clear and obvious, because Clippy will never catch all of them, so developers have the power to work around them. Additionally it introduces an abstraction with appropriate documentation that clears up the lint and ensures incorrect usage on the top level cannot be introduced (which is checked by the compiler).

kylewillmon:
Closing in favor of #546

Issue #535, opened 2022-07-08
Investigate deno on aarch64-unknown-linux-musl


kylewillmon:
Now that #528 has been merged, we may have introduced additional barriers for releasing CLI extensions because of the ability to compile deno (See #426).

We need to investigate compiling deno for `aarch64-unknown-linux-musl` before releasing a CLI with these binaries. If it's not feasible, we may need to revert #528

PR #548, opened 2022-07-13
Bump to v3.7.0 - ARM64 Linux support and Maven workspaces

kylewillmon:
The next release!

Release-Version: v3.7.0
Release-Summary: ARM64 Linux support and Maven workspaces


maxrake (approved):
So the possibility of backing out the new target triple is lower now (#535)?

kylewillmon:
> So the possibility of backing out the new target triple is lower now (#535)?

Yes. Thank you. I have added [a comment](https://github.com/phylum-dev/cli/issues/535#issuecomment-1183619988) and closed that issue now. I should have done that before putting up this PR.

PR #550, opened 2022-07-14
Support effective-pom files with site information

ejortega:
See https://github.com/phylum-dev/phylum-types/issues/44

Fixed an issue when parsing an effective pom generated from https://github.com/apache/maven/blob/master/pom.xml.

This PR is to bump the phylum-types dependency and update the tests and test file.


Cargo.lock
```diff
@@ -844,7 +844,7 @@ dependencies = [
  "async-trait",
  "deno_core",
  "tokio",
- "uuid 1.1.2",
+ "uuid",
```
louislang, line 847: Was a specific version specified here for a reason? `1.1.2` might be required for compatibility somewhere.

ejortega: The dependency on uuid 0.8.2 was removed. There's a PR for that as well https://github.com/phylum-dev/cli/pull/549

cd-work (approved):

PR #553, opened 2022-07-14
Fix CI release readme release process

cd-work:
The `docs` command is expecting a directory as parameter, but we were
incorrectly passing all documentation files.


maxrake (approved):
🤞

PR #558, opened 2022-07-18
Remove warnings from generic lockfile parser

### LINKED ISSUE #556: Parser `WARN` statements
# Overview
Given an `effective-pom.xml` file that has been renamed to `pom.xml` the CLI seems to chose the `pypi` parser and spits out a bunch of `WARN` statements. Note: If the file is not renamed, it processes cleanly.

# How To Reproduce
1. Rename an `effective-pom.xml` file to `pom.xml` and submit for analysis
2. See `WARN` messages with "parsers::pypi"

# Expected Behavior
`WARN` message for "unrecognized lockfile type," but correctly determine the maven ecosystem and use that parser.

# Additional Context
image
...snip...
image

What is interesting is that it seems to have processed correctly and the listed `Ecosystem` is `Maven`

cd-work:
Since the generic lockfile parser attempts to parse every single
lockfile, its output would be rather verbose and misleading, complaining
about invalid packages when parsing an invalid lockfile.

This patch removes all warnings during lockfile parsing and provides
some additional output informing the user about the generic lockfile
parsing status.

Closes #556.

### DIFF
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -11,6 +11,7 @@ use super::{CommandResult, ExitCode};
 use crate::lockfiles::{
     CSProj, GemLock, GradleLock, PackageLock, Parse, PipFile, Poetry, Pom, PyRequirements, YarnLock,
 };
+use crate::{print_user_success, print_user_warning};
 
 pub const LOCKFILE_PARSERS: &[(&str, &dyn Parse)] = &[
     ("yarn", &YarnLock),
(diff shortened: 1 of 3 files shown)


maxrake (approved):
Verified with a local build and testing:

image

---

The `effective-pom.xml` file used was the one from this repository's test fixtures: `cli/tests/fixtures/effective-pom.xml`

Interestingly, there was only one warning emitted when using v3.7.1 of the CLI, instead of the stream of them shown in issue #556 

Regardless, the changes here are an improvement. LGTM.

Issue #551, opened 2022-07-14
Add `extension run` subcommand


andreaphylum:
# Overview
For ease of testing extensions, we could add a `phylum extension run <path>` subcommand that allows users to run an extension in-place without previously installing it.

PR #560, opened 2022-07-19
Poetry ecosystem extension

### LINKED ISSUE #544: `poetry` ecosystem extension
## Associated user story

https://github.com/phylum-dev/roadmap/issues/16

andreaphylum:
Closes #544.

## Checklist


extensions/poetry/tests.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    const status = await process.status()
+    await process.close()
+
+    return status
+  }
+
+  async installExtension (extension: string) {
+    await this.run(['extension', 'install', '-y', extension])
+  }
```
cd-work, line 42: I wonder, do we actually need to test the extension through `phylum` here? If the goal is just to ensure the extension itself is working, maybe executing the extension's typescript code directly without `phylum` makes more sense? It probably would have less boilerplate at least.

Generally as an extension developer I would expect the installation and arg passing process to be tested by our CLI itself, so no reason to test it with the extension?

andreaphylum: We do. We cannot import `phylum` as a module as it is injected by our own runtime and not available anywhere else.

This is a lot of boilerplate indeed, but it could be trimmed down significantly by #551.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+} else {
+  let status = await Deno.run({
+    cmd: ['poetry', ...Deno.args],
+    stdout: 'inherit',
+    stderr: 'inherit',
+  }).status()
+
+  Deno.exit(status.code)
+}
```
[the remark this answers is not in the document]
andreaphylum: This is intentional. Not having the `else` clause would result in a double execution of `poetry add`, which outputs errors as the packages had already been added to the lockfile. I wonder if we should interpret the other arguments, too, and run `poetry install` except in the cases where `--dry-run` or `--lock` were already specified.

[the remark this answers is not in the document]
cd-work: Huh it surprises me that running `poetry add` when the package is already in the lockfile causes an error.

But the goal of `phylum poetry add` should be to actually install the package. So at the very least we need to add an install step.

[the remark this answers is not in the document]
cd-work: Yeah that doesn't seem desirable.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,53 @@
+import { PhylumApi } from "phylum"
+import { parse } from "https://deno.land/std/flags/mod.ts"
+
+// Add packages to the lockfile, and upload it to Phylum for analysis.
+// Return `true` if the packages pass the project's thresholds.
+// Return `false` if the packages don't pass the thresholds, or the processing
+// job is not finished.
+async function poetryCheck(args: string[]) {
+    console.log("Updating package lock…")
```
maxrake, line 9: This language does not quite line up with Python or `poetry`. Maybe `Attempting to add a dependency ...`

Generically, changing `lock` to `lockfile` could work but it has the downside of not being precise here since _both_ the `pyproject.toml` and `poetry.lock` file are updated with a `poetry add` command.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    let process = Deno.run({ cmd: ["poetry", "add", '--lock', ...args] })
```
maxrake, line 11:
```suggestion
    let process = Deno.run({ cmd: ["poetry", "add", ...args] })
```
It would be surprising to a user to run `poetry add`, expecting to have a specific package added to _both_ the `pyproject.toml` and `poetry.lock` files, only to have this `--lock` argument added...which says to update the `poetry.lock` file and not the `pyproject.toml` file.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,53 @@
...
+    await process.status()
+    await process.close()
+
+    console.log("Package lock updated.\n")
```
maxrake, line 15: Similar comment on wording here.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,53 @@
...
+    console.log("Analyzing packages…")
+
+    const jobId = await PhylumApi.analyze("./poetry.lock")
```
maxrake, line 18: So...the entire lockfile is analyzed instead of just the package(s) that were added by the user-supplied `poetry add` or `poetry update` command? That might make it hard to use this extension for any project that does not already meet the risk threshold values.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,53 @@
...
+    const jobStatus = await PhylumApi.getJobStatus(jobId)
+
+    if (jobStatus.pass && jobStatus.status === "complete") {
+        console.log("All packages pass project thresholds.\n")
+        return true
+    } else if (jobStatus.pass) {
+        console.warn("Unknown packages were submitted for analysis, please check again later.")
+        return false
+    } else {
+        let packages = pkgs.join("', '")
```
maxrake, line 28: `pkgs` is not defined. Indeed, running this extension with a known bad package resulted in an error:

```
❯ /Users/maxrake/dev/phylum/localdev/cli/target/debug/phylum extension run /Users/maxrake/dev/phylum/localdev/cli/extensions/poetry add "pyyaml==5.3.1"
The `poetry` extension requires the following permissions:

  Read from the following paths:
    './poetry.lock'

  Run the following commands:
    'poetry'


Do you accept? yes
Updating package lock…

Updating dependencies
Resolving dependencies... (16.9s)

Writing lock file
Package lock updated.

Analyzing packages…
❗ Error: Execution failed caused by: ReferenceError: pkgs is not defined
    at poetryCheck (file:///Users/maxrake/dev/phylum/localdev/cli/extensions/poetry/main.ts:29:24)
    at async file:///Users/maxrake/dev/phylum/localdev/cli/extensions/poetry/main.ts:39:29
```

An unexpected side effect of this error was that the `pyproject.toml` and `poetry.lock` files were updated anyway. If this "failure" path is reached, it is expected that the bad package(s) would not be allowed and the files would not be updated.

maxrake (changes requested):
There are some failures that have been identified and need to be corrected. There are also some larger design questions that might be better suited for an offline discussion.

cli/src/commands/extensions/mod.rs
```diff
@@ -47,6 +47,8 @@ pub fn command<'a>() -> Command<'a> {
         .subcommand(
             Command::new("run")
                 .about("Run an extension from a directory")
+                .allow_hyphen_values(true)
+                .disable_help_flag(true)
```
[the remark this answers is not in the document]
andreaphylum: So that `--help` can be forwarded to the extensions that may want to print their own help.

extensions/poetry/PhylumExt.toml
```diff
@@ -0,0 +1,7 @@
+name = "poetry"
+description = "poetry package manager hooks"
+entry_point = "main.ts"
+
+[permissions]
+run = ["poetry"]
+read = ["./poetry.lock"]
```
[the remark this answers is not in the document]
andreaphylum: Permissions only contemplate what the extension can access through the Deno API.

Once the control is yielded to an external process, i.e. `poetry`, we no longer have control over what it accesses, and that is intrinsic into the concept of allowing execution of an external program via Deno. The same is true for the `ffi` permissions which we intentionally are not supporting.

There is fundamentally no workaround to this, barring building actual OS level sandboxes.

Not sure about `write` also covering `read`; my hunch is that one would have to request both of them, as Deno's permission system is still relatively new and requiring the user to do so would be simpler to maintain at first.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,53 @@
+import { PhylumApi } from "phylum"
+import { parse } from "https://deno.land/std/flags/mod.ts"
```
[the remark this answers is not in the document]
andreaphylum: This was not intentional, but I don't have a strong opinion over which option is better.

I see Deno's std is in constant flux; pinning the version would allow for less maintenance but it would expose us to vulnerabilities/bugs, using latest on the other hand would require us to regularly run and check the extensions for breaking changes.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,53 @@
...
+
+// Add packages to the lockfile, and upload it to Phylum for analysis.
+// Return `true` if the packages pass the project's thresholds.
+// Return `false` if the packages don't pass the thresholds, or the processing
+// job is not finished.
```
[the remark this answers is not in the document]
andreaphylum: Having this kind of feedback from the user side is very useful, as there is no immediately apparent "right way" to do this. I like your suggestions, but definitely let's discuss this.

extensions/poetry/main.ts
andreaphylum: We need the `--lock` argument because we don't actually want to install the package until we have validated it passes thresholds.

I probably missed adding `poetry install`, or an equivalent method, in case of success. What would be the appropriate action to take, in this case? Is `poetry install` ok, even though it installs _all_ packages in the lockfile?

extensions/poetry/main.ts
andreaphylum: I'd just remove this log line altogether, honestly. `poetry` already prints messages, the last one being `Writing lockfile`.

extensions/poetry/main.ts
andreaphylum: The point of processing the whole lockfile would be addressing the transitive dependencies that arise from `poetry add`, which the user can't know about beforehand, and for which it is not trivial to compute a diff (e.g. new transitive dependencies are easy: just process those; but bumped versions of other transitive dependencies could make it so that the project passes, or stops passing, the thresholds).

If one project does not meet the risk thresholds, that is something the developers need to address, no? Otherwise there is no point in gating `poetry` behind `phylum` in the first place.

If forcing an installation is a desirable use case, we could either add a `--force` flag or tell the user to not run poetry through `phylum`.

This latter option is what I'd gravitate towards, as it "feels wrong" to do and thus may push the user towards doing the right thing instead.

This is not unlike Rust does in some places -- i.e. using `unwrap()` or `unsafe` feels wrong and automatically makes you aware of the fact that you need to be extra careful, which in turn will prompt you to manage errors correctly or put more energy in having the safety invariants hold.

extensions/poetry/main.ts
andreaphylum: Good catch -- this is a leftover from previous commits where the arguments were parsed and the list of packages was extracted.

I would like to find an idiomatic way of rolling back a lockfile.
I'm no expert on poetry; can we rely on the following assumptions?
- `poetry add --lock` only updates `pyproject.toml` and `poetry.lock`
- `poetry add --lock` is not influenced by any files other than the above two

If that's the case, we could just simply copy those files in a temporary directory, run `poetry` in there, upload the updated lockfile, and then run `poetry add` again afterwards. This feels a bit wasteful because we run `poetry` twice, but I'm not sure whether there are better paths. Any ideas?

extensions/poetry/main.ts
```diff
@@ -0,0 +1,53 @@
...
+        return false
+    } else {
+        let packages = pkgs.join("', '")
+        console.error(`Installing '${packages}' caused threshold failure.`)
+        return false
+    }
+}
+
+// If the subcommand is `add`, update the lockfile and process it through
+// Phylum. Otherwise, pass the arguments through to `poetry`.
```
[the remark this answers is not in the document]
andreaphylum: Good to know -- are there any other packages that write to the lockfile that we may have to support?

extensions/poetry/tests.ts
```diff
@@ -0,0 +1,93 @@
...
+
+describe("Poetry extension", async () => {
+  // These tests may fail if the packages aren't processed on staging.
+
+  it("correctly handles the `--dry-run` argument", async () => {
+    let status = await phylum.runExt(['add', '--dry-run', 'pandas'], phylum.fixturesDir)
+    assert(status.code === 0)
+  })
+
+  it("correctly allows a valid package", async () => {
```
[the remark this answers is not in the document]
andreaphylum: Absolutely, good point.

extensions/poetry/tests.ts
```diff
@@ -0,0 +1,93 @@
...
+    let status = await phylum.runExt(['add', 'pandas'], phylum.fixturesDir)
+    assert(status.code === 0)
+  })
+
+  it("allows duplicating the `--lock` flag", async () => {
```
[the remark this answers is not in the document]
andreaphylum: This is more to check whether `poetry` will ever disallow doing something like `poetry add --lock --lock`. In which case we would need to manually filter out the arguments before passing them to our `--lock` run, which I wanted to avoid if possible.

cli/src/commands/extensions/mod.rs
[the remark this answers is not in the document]
maxrake: Roger that...understood. It might be surprising/confusing to have a `phylum` command without help, but an alternative solution is not coming to mind. The current documentation for this command ([in review](https://github.com/phylum-dev/cli/pull/562/files#diff-0f156a4fb7cba9f87e2a86038a1b13e9f6a8a8890c6955c95234ea9715ab5c5b)) says:

```sh
phylum extension run [OPTIONS] <PATH> [OPTIONS]...

[---TRIMMED---]

The first set of options are for the `run` command. The second set of options
are for the extension.
```

Is that statement about options false? It's not possible to have something like `phylum extension run --help` to print the `phylum` help message and also have something like `phylum extension run -y /path/to/ext --help` to print the extension's help message?

extensions/poetry/main.ts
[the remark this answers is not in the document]
maxrake: The belief is that `poetry` is the only package/tool that modifies the `poetry.lock` lockfile and therefore the only one to support.

cli/src/commands/extensions/mod.rs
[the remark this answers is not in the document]
andreaphylum: I _think_ currently the `clap` crate does not have that capability, unless the `--help` comes after `--`. One thing we _could_ do, though quite a bit hacky, is manually matching the `PATH` parameter against `-h` / `--help` and print a help instead of running an extension.

extensions/poetry/main.ts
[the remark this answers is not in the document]
andreaphylum: Sorry, I meant _subcommands_ rather than _packages_. 😅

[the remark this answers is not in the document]
maxrake: Ah. A review of the `poetry` subcommands resulted in just two that **directly** add dependencies to the lockfile: `add` and `update`. There are other subcommands that touch/modify the lockfile, but they do so **indirectly** (i.e., not because a user supplied a specific package). These commands are: `install` and `lock`. There is also the possibility that `remove` will update the lockfile in a way that not only removes the specified package, but re-resolves the dependencies, resulting in a lockfile that has new/different packages than from before the command was run. Documentation for these commands is [here](https://python-poetry.org/docs/cli) and I'd be more than willing to try any specific scenario for you (or screen share to see it together).

extensions/poetry/main.ts
maxrake: Yes, `poetry install` would work here. It updates the active/current poetry virtual environment with the packages from the `poetry.lock` file. If an environment already exists, only the new/different packages will be added/modified.

extensions/poetry/main.ts
maxrake: Understood. For other readers, this was discussed a bit offline, during an integrations team sync, and the decision was made to continue using the results from analyzing the entire lockfile when making the pass/fail decision.

FWIW, an alternative solution would be to do what the `phylum-ci` tool does, and [get the new dependencies](https://github.com/phylum-dev/phylum-ci/blob/main/src/phylum/ci/ci_base.py#L256-L272) by parsing the lockfile both before and after changes are made to it and then taking the set difference to get the list of new packages. Then, each of those packages' analyze results could be compared to the individual risk domain thresholds set for the project to determine if any single package violates the thresholds.

extensions/poetry/main.ts
maxrake: > I would like to find an idiomatic way of rolling back a lockfile. I'm no expert on poetry; can we rely on the following assumptions?
> 
> * `poetry add --lock` only updates `pyproject.toml` and `poetry.lock`
> * `poetry add --lock` is not influenced by any files other than the above two

Yes, the belief is that these assumptions can be relied upon.

For other readers, this was discussed a bit offline, during an integrations team sync, and the idea of using temporary lock files (as it were) seemed to win the day. The original files would be "locked" while the Phylum extension runs and then either removed or atomically returned to their previous state, depending on the outcome and any error conditions of the extension logic.

cli/src/commands/extensions/mod.rs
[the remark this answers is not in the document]
cd-work: Actually you shouldn't disable the help flag on `run`, you should have it on the extension subcommands. And it should be on those already from my previous PR?

extensions/poetry/main.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  // If the analysis failed, exit with an error.
+  if (!analysisOutcome) {
+    Deno.exit(1)
+  }
+} else {
+  let status = await Deno.run({
+    cmd: ['poetry', ...Deno.args],
+    stdout: 'inherit',
+    stderr: 'inherit',
```
cd-work, line 48: Inherit is the default, no need to specify it manually.

cli/src/commands/extensions/mod.rs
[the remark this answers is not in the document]
cd-work: I guess you could maybe add a custom help flag, but I feel like that would be more trouble than it's worth.

extensions/poetry/PhylumExt.toml
```diff
@@ -0,0 +1,8 @@
+name = "poetry"
+description = "poetry package manager hooks"
```
cd-work, line 2:
```suggestion
description = "Poetry package manager hooks"
```

extensions/poetry/fixtures/.phylum_project
```diff
@@ -0,0 +1,5 @@
+---
+id: 7f78e2f6-9833-40d1-95f7-432f8d48a963
+name: test-poetry
+created_at: "2022-07-27T15:21:23.503757423+02:00"
+group_name: ~
```
cd-work, line 5: Why add a `.phylum_project` for a fixture/test? Feels like that should just always be specified in the test as parameters. That way parameters are directly obvious when looking at a test case and there's no auxiliary files necessary.

extensions/poetry/parse.ts
```diff
@@ -0,0 +1,36 @@
...
+
+type PackageSpec = {
+  name: string,
+  version: string,
+}
+
+// Parse a package name and version from a line of a poetry dry run output.
+// Returns `null` if no match was found.
+function parseDryRunLine(line: string): PackageSpec | null {
+  const installingRegexp = /: selecting\s+([^\s]+)\s+\(([^\)]+)\)/
```
cd-work, line 11: The following command seems to print `selecting` for something that doesn't actually get installed. So I'm not sure your approach works:

```
$ poetry add --dry-run -vvv jinja
Creating virtualenv test-CHLv_kh8-py3.10 in /home/chris/.cache/pypoetry/virtualenvs
Using virtualenv: /home/chris/.cache/pypoetry/virtualenvs/test-CHLv_kh8-py3.10
PyPI: No release information found for jinja-0.5beta, skipping
PyPI: No release information found for jinja-0.6, skipping
PyPI: No release information found for jinja-0.7, skipping
PyPI: 5 packages found for jinja *
Using version ^1.2 for Jinja

Updating dependencies
Resolving dependencies...
   1: fact: test is 0.1.0
   1: derived: test
   1: fact: test depends on Jinja (^1.2)
   1: selecting test (0.1.0)
   1: derived: Jinja (>=1.2,<2.0)
PyPI: No release information found for jinja-0.5beta, skipping
PyPI: No release information found for jinja-0.6, skipping
PyPI: No release information found for jinja-0.7, skipping
PyPI: 1 packages found for jinja >=1.2,<2.0
   1: selecting jinja (1.2)
   1: Version solving took 0.008 seconds.
   1: Tried 1 solutions.

Writing lock file

Finding the necessary packages for the current system

Package operations: 1 install, 0 updates, 0 removals

  • Installing jinja (1.2)
  • Installing jinja (1.2)
```

extensions/poetry/main.ts
```diff
@@ -0,0 +1,108 @@
...
+  const manifestBackup = new FileBackup('pyproject.toml')
+
+  await lockfileBackup.backup()
+  await manifestBackup.backup()
+
+  let process = Deno.run({
+    cmd: ['poetry', subcommand, '-vvv', '-n', '--dry-run', ...args.map(s => s.toString())],
+    stdout: 'piped',
+    stderr: 'piped',
+  })
```
cd-work, line 50: The whole point is that this shouldn't update the lockfile, right? Why do you store and revert it?
If you need to store and revert it anyway, what's the point of doing a dry run?

There's just so much complexity in such a simple extension, it seems somewhat difficult to get a reasonable grasp of all the invariants and how they play together.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  await process.status()
+  await process.close()
+
+  // If it existed before, restore the previous contents of the lockfile;
+  // otherwise, delete it. This is a workaround to the fact that in poetry
+  // 1.1.x, the `--dry-run` argument does not prevent the lockfile from 
+  // being modified. This is not fixed as of poetry 1.1.14.
+  // Prudently, do the same for the manifest (pyproject.toml).
```
cd-work, line 59: >   // Prudently, do the same for the manifest (pyproject.toml).

You do it just for fun? I'm not sure I understand the benefit of doing it for a file that is never affected.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,108 @@
...
+  await manifestBackup.restoreOrDelete()
+
+  const output = new TextDecoder().decode(await process.output())
+  const packages = parseDryRun(output)
+
+  console.log(`Analyzing packages:`)
+  for (const { name, version } of packages) {
+    console.log(`  - ${name} ${version}`)
+  }
+  console.log()
```
cd-work, line 72: Does the user want to know this? I probably wouldn't mind but I'm not sure if this is a benefit or just annoying. Imagine this adds a billion packages, it would get extremely verbose.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  if (jobStatus.pass && jobStatus.status === "complete") {
+    console.log(`[${green("phylum")}] All packages pass project thresholds.\n`)
+    return true
+  } else if (jobStatus.pass) {
+    console.warn(`[${yellow("phylum")}] Unknown packages were submitted for analysis, please check again later.\n`)
+    return false
+  } else {
+    console.error(`[${red("phylum")}] The operation caused a threshold failure.\n`)
+    return false
```
cd-work, line 85: Do we want the same exit code in both instances?

extensions/poetry/tests.ts
```diff
@@ -0,0 +1,152 @@
...
+    let status = await phylum.runExt(['add', 'pandas'], phylum.fixturesDir)
+    assert(status.code === 0)
+  })
+
+  // TODO tqdm is artificially marked as not passing; replace with an actual
+  // known-bad package.
+  it("correctly denies an invalid package", async () => {
+    let status = await phylum.runExt(['add', 'tqdm'], phylum.fixturesDir)
+    assert(status.code === 1)
+  })
```
cd-work, line 128: I'd like to explicitly point this out as something that should **NOT** be merged until it is resolved.

cd-work (changes requested):

extensions/poetry/fixtures/.phylum_project
andreaphylum: This was actually an oversight. It is not meant to be there.

extensions/poetry/main.ts
andreaphylum: I mentioned this elsewhere: `poetry` actually applies changes to the lockfile and `pyproject.toml` _regardless_ of the `--dry-run` and `--lock` flags. It is a [known bug](https://github.com/python-poetry/poetry/issues/3666) that should be addressed in 1.2, which has been delayed for quite a while and is not going to be released anytime soon afaik. Forcing backup and restore is the only workaround to that.

I agree that the extension is more complex than I'd like but the workarounds are definitely necessary.

extensions/poetry/main.ts
andreaphylum: It is affected with at least one combination of parameters, though I don't remember exactly which one.

We cannot control Poetry's invariants or the way they change things over time, but we do know we want our first `poetry` command not to affect those two files specifically and backing both up guarantees to us that no undesired change will be ever made.

The cost of this code is approximately zero, and the benefit is a guarantee over something that is not under our control and about which we'd have otherwise to make assumptions about, crossing fingers that no PR gets into Poetry that violates those assumptions.

extensions/poetry/main.ts
andreaphylum: Yes, as we discussed -- I'd go with either passthrough or exit(1) unless there are other compelling reasons or the user could benefit from that, and in that case we need to make sure our exit codes don't clash with Poetry's.

extensions/poetry/tests.ts
andreaphylum: Other tests are in a similar situation, where they are explicitly relying on the state of our service; those are not much better than this in my opinion.

Even once we have an explicitly failing Python package in staging or prod, we have no guarantee that that is going to be its permanent state.

The alternative would be bundling a mock API server but that's preposterous to even write down.

This test was explicitly requested in this same review and I agree it should be there.

Any suggestions?

extensions/poetry/parse.ts
andreaphylum: I tried to reproduce this with a project named `a` and I discovered that the first selected package is indeed `a`, so chances are your test package was named `test`:

```
Updating dependencies
Resolving dependencies...
   1: fact: a is 0.1.0
   1: derived: a
   1: fact: a depends on Jinja (^1.2)
   1: selecting a (0.1.0)
   1: derived: Jinja (>=1.2,<2.0)
PyPI: No release information found for jinja-0.5beta, skipping
PyPI: No release information found for jinja-0.6, skipping
PyPI: No release information found for jinja-0.7, skipping
PyPI: 1 packages found for jinja >=1.2,<2.0
   1: selecting jinja (1.2)
   1: Version solving took 0.004 seconds.
   1: Tried 1 solutions.
```

I repeat what I mentioned via DMs here. Parsing `Installing` lines is unreliable, because none is output with the `--lock` flag, but the lockfile is updated nonetheless:

```
$ poetry add --dry-run --lock -vvv jinja                                                                                     andrea phylum-ubuntu fsh 2022-08-02 13:35:25
Using virtualenv: /home/andrea/.cache/pypoetry/virtualenvs/a-2bdB6j1N-py3.10
PyPI: No release information found for jinja-0.5beta, skipping
PyPI: No release information found for jinja-0.6, skipping
PyPI: No release information found for jinja-0.7, skipping
PyPI: 5 packages found for jinja *
Using version ^1.2 for Jinja

Updating dependencies
Resolving dependencies...
   1: fact: a is 0.1.0
   1: derived: a
   1: fact: a depends on Jinja (^1.2)
   1: selecting a (0.1.0)
   1: derived: Jinja (>=1.2,<2.0)
PyPI: No release information found for jinja-0.5beta, skipping
PyPI: No release information found for jinja-0.6, skipping
PyPI: No release information found for jinja-0.7, skipping
PyPI: 1 packages found for jinja >=1.2,<2.0
   1: selecting jinja (1.2)
   1: Version solving took 0.003 seconds.
   1: Tried 1 solutions.
$
```

extensions/poetry/tests.ts
cd-work: Which other test is relying on a temporarily manually set thing in our backend? If there are others, we should get rid of those too.

Honestly I don't particularly care about this test anyway, since it never gets executed, but really that's just one more reason why it shouldn't be kept in some garbage state indefinitely when nobody will ever touch it again.

extensions/poetry/main.ts
cd-work: I don't remember any discussion regarding this? Generally I think the user very much could benefit from having different error codes here.

And I think the error code you've selected isn't exactly safe from crashing with poetry. Quite the opposite I'd say?

extensions/poetry/tests.ts
maxrake: How about using `pyyaml==5.3.1` instead? That is my go-to known bad Python package.

extensions/poetry/parse.ts
andreaphylum: This file was removed due to a reversion to submitting the whole package lockfile.

extensions/poetry/tests.ts
andreaphylum: > Which other test is relying on a temporarily manually set thing in our backend? If there are others, we should get rid of those too.

I didn't say that. I said that both rely on the state of our service whose consistency we can't really guarantee. Besides, `tqdm` is the "bad dependency" in the mock API script I was using, not even on staging. Between a mock API script one can control and the wider service on staging, the latter is probably more reliable long term, provided we include that somewhere; I don't think we need to do that, as staging will be more than enough anyway for the purpose of these tests (see next paragraph) and reliability is not a concern.

> Honestly I don't particularly care about this test anyway, since it never gets executed, but really that's just one more reason why it shouldn't be kept in some garbage state indefinitely when nobody will ever touch it again.

These tests are meant for development workflow and I have been running them regularly. The fact that they aren't run in CI doesn't mean nobody runs them. If we ever want to put extensions test through CI, nobody is stopping us, though.

> How about using `pyyaml==5.3.1` instead? That is my go-to known bad Python package.

That's good to know going forward -- I did find one `yaml` package on Phylum but Poetry couldn't `add` that. Went with `cffi` in the meantime, which carries risk in Phylum due to the fact that it involves FFI.

extensions/poetry/main.ts
andreaphylum: > I'd say the cost of creating a file in the user's project directory is always **significantly** bigger than zero. 

We don't do that, though? We store the file's contents in memory, run `poetry`, and then immediately either restore the contents of the file (if the file was present before) or delete the file (if it wasn't). The only destructive action is thus writing data to a file that was read from the same file, and the action is limited to the files we have requested permissions for. The failure cases are: the user having their files changed by Poetry (which we can't avoid anyway without a backup) because our extension has crashed before restoring the files, and OS-I/O level errors that corrupt the file (over which we will never have any control whatsoever).

> There's also a possibility with every file we touch that we will do irrecoverable damage to that file.

At some point I suggested enforcing a clean git tree / a `--force` flag, but that is way overkill for this use case and probably annoying to the user who wants to add a number of packages in multiple commands. What are some examples of failure cases in which we could do irrecoverable damage to a file?

extensions/poetry/main.ts
andreaphylum: Changed this to print the first 10 items, and then a line that states `...and X more` in case the list exceeds 10 items. This way, the data is incomplete but not jarring to read, and the user can be made aware of the size of their dependency tree.

extensions/poetry/main.ts
andreaphylum: Yeah, having `exit(1)` is more of a catch-all in the event that we don't care about specific meaning. If we do care, we need to come up with two numbers neither of which is 1. IIRC you did use some custom exit values at some point; what was your rationale for choosing them?

extensions/poetry/main.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+}
+
+// If the subcommand is `add`, update the lockfile and process it through
+// Phylum. Otherwise, pass the arguments through to `poetry`.
+if (Deno.args.length >= 1 && ['add', 'update', 'install'].includes(Deno.args[0])) {
+  // Skip the `add` string. Pass the rest of the arguments as-is.
```
maxrake, line 94: Comment does not match code...and can likely be removed entirely.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,110 @@
...
+    console.warn(`[${yellow("phylum")}] Unknown packages were submitted for analysis, please check again later.\n`)
+    return false
+  } else {
+    console.error(`[${red("phylum")}] The operation caused a threshold failure.\n`)
+    return false
+  }
+}
+
+// If the subcommand is `add`, update the lockfile and process it through
+// Phylum. Otherwise, pass the arguments through to `poetry`.
```
maxrake, line 92:
```suggestion
// If the subcommand modifies the lockfile, process it through Phylum.
```
Comment does not match code.

extensions/poetry/tests.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  it("correctly allows a valid package", async () => {
+    let status = await phylum.runExt(['add', 'numpy'], phylum.fixturesDir)
+    assert(status.code === 0)
+  })
+
+  // TODO tqdm is artificially marked as not passing; replace with an actual
+  // known-bad package.
+  it("correctly denies an invalid package", async () => {
```
maxrake, line 78:
```suggestion
  it("correctly denies a known bad package", async () => {
```
This is just a suggestion...but perhaps makes the distinction between known bad (expected to fail based on thresholds) vs. invalid (which sounds more like there is a structural flaw with the package...like in the backend system or how it was processed by Phylum).

maxrake (changes requested):
Mostly suggested changes at this point...take them or leave them 😜. The "requested changes" is mostly for handling exceptions more cleanly and returning a non-zero exit code when there are exceptions.

extensions/poetry/tests.ts
cd-work: > These tests are meant for development workflow and I have been running them regularly. The fact that they aren't run in CI doesn't mean nobody runs them.

I strongly disagree with that. I certainly won't run them, you might but even that I'd be surprised about when you make changes to the CLI instead of the extension itself. And even then it's only one extension now, that would become unsustainable with more extensions.

So realistically nobody will ensure these extensions work consistently on every commit in master and it can potentially take a very long time before anyone tries to run them again.

> Besides, tqdm is the "bad dependency" in the mock API script I was using, not even on staging.

So that sounds even if I wanted to try running your tests, I couldn't? That sounds like even more of an issue. Having tests only a single developer knows how to execute sounds like a **TERRIBLE** idea.

extensions/poetry/main.ts
cd-work: Tested a bunch of `npm` exit codes, noted the approximate ranges, increased by an order of magnitude and picked a number unusual enough to where it wouldn't be easily picked by anyone else trying to do something similar.

extensions/poetry/main.ts
cd-work: Honestly I think this is the worst of both worlds. Now you still get your output spammed, but it's not even useful because there's no way to see **everything** that was installed.

I say either we go the `cargo` route and be very verbose, listing everything we got by default.

Or we stick to the ecosystem's own standards and hide the packages like `poetry` does.

I think I'd opt for the latter, to be consistent with the ecosystem. But I don't mind it much either way. I definitely don't like the half-way point with only showing a limited amount though.

extensions/poetry/main.ts
cd-work: > We don't do that, though? We store the file's contents in memory, run poetry, and then immediately either restore the contents of the file (if the file was present before) or delete the file (if it wasn't). The only destructive action is thus writing data to a file that was read from the same file, and the action is limited to the files we have requested permissions for. The failure cases are: the user having their files changed by Poetry (which we can't avoid anyway without a backup) because our extension has crashed before restoring the files, and OS-I/O level errors that corrupt the file (over which we will never have any control whatsoever).

We do have control over corruption though, that's why I previously suggested to atomically swap the file out after writing to a new one.

extensions/poetry/tests.ts
andreaphylum: > I strongly disagree with that. I certainly won't run them, you might but even that I'd be surprised about when you make changes to the CLI instead of the extension itself.

That is not a good argument for throwing away test code that has proven useful for development and will prove useful during maintenance of the extension. To reiterate, I **do need** and **will need** those tests for continued maintenance, regardless of who runs them, and not having them would actively hamper my ability to work on the extension. If this PR gets blocked because of the presence of that test file, I will remove it and then simply proceed to waste valuable time by copying it outside of the repo and then back inside at every commit.

> And even then it's only one extension now, that would become unsustainable with more extensions.
>
> So realistically nobody will ensure these extensions work consistently on every commit in master and it can potentially take a very long time before anyone tries to run them again.

The unsustainability is easily addressed by setting up proper test execution in CI if and when we so desire (which, in my opinion, is not right here and right now, but that should be discussed elsewhere).

At some point we may even consider just moving the extensions directory to its own repository with its own CI pipeline, as extensions are not "code added to the CLI" but rather their own thing.

> So that sounds even if I wanted to try running your tests, I couldn't? That sounds like even more of an issue. Having tests only a single developer knows how to execute sounds like a **TERRIBLE** idea.

You can now, against staging, as I changed that package to a known failing package that's in staging.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,110 @@
...
+// process. This way, we can track the actual lockfile changes, as specifying
+// `--dry-run` only would not output anything in combination with flags such
+// as `--lock` that do not perform the actual operations.
+async function poetryCheckDryRun(subcommand: string, args: string[]): boolean {
+  // Read and backup the current poetry lockfile contents.
+  const lockfileBackup = new FileBackup('poetry.lock')
+  const manifestBackup = new FileBackup('pyproject.toml')
+
+  await lockfileBackup.backup()
+  await manifestBackup.backup()
```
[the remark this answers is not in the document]
andreaphylum: Good catch!

Unfortunately, our permissions system doesn't allow us upwards traversal, we can only request permissions for a path in the form `./poetry.lock`.

I would suggest erroring out and asking the user to move to the top level of their git tree for the subcommands that we hijack. I would definitely prefer if we could mimick `poetry`'s behavior, but I don't think we can right now.

What do you think?

extensions/poetry/tests.ts
andreaphylum: 100% agree.

extensions/poetry/main.ts
andreaphylum: I will go for the latter.

extensions/poetry/main.ts
andreaphylum: > We do have control over corruption though, that's why I previously suggested to atomically swap the file out after writing to a new one.

The concept is that a corruption can only happen in a way the extension cannot be held responsible for -- i.e. bad hard drive or another `poetry` command (or worse, another random one) writing to the file at the same time. Both _aren't_ avoided by writing to a new file and atomically renaming: in the first case, we just write bad data to a different file, and in the second case, we could get the file overwritten _after_ having moved it, which is also problematic at the outset because then we can't guarantee we uploaded the right packages.

extensions/poetry/main.ts
andreaphylum: I was not able to reproduce any error condition in `poetry` that would return anything but `1`. For the sake of consistency, I'd reuse the same exit codes as `npm`.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  let process = Deno.run({
+    cmd: ['poetry', subcommand, '-vvv', '-n', '--dry-run', ...args.map(s => s.toString())],
+    stdout: 'piped',
+    stderr: 'piped',
+  })
+
+  await process.status()
+  await process.close()
```
cd-work, line 52: Is it necessary to await close when already awaiting status? A process cannot return a status without closing, right? IIRC Deno's API docs also don't call close after status.

extensions/poetry/main.ts
```diff
@@ -0,0 +1,98 @@
...
+  // otherwise, delete it. This is a workaround to the fact that in poetry
+  // 1.1.x, the `--dry-run` argument does not prevent the lockfile from 
+  // being modified. This is not fixed as of poetry 1.1.14.
+  // Prudently, do the same for the manifest (pyproject.toml).
+  //
+  // See: https://github.com/python-poetry/poetry/pull/5718
+  await lockfileBackup.restoreOrDelete()
+  await manifestBackup.restoreOrDelete()
+
+  const jobId = await PhylumApi.analyze(lockfileData['package_type'], lockfileData['packages'])
```
cd-work, line 66: Out of curiosity, why did you index `lockfileData` using the map-type brackets, rather than just `lockfileData.package_type`. Did the latter give you a warning or something?

cd-work (changes requested):

extensions/poetry/main.ts
andreaphylum: Let me check that out -- I added this because I observed some process hanging at some point, but I'm not confident I haven't solved that by other means.

extensions/poetry/main.ts
cd-work: Oh that's actually a good point. Maybe we should change this? The type is constructed specifically for our extension API, so maybe it should be `packageType` and `packages` in Rust instead?

extensions/poetry/main.ts
cd-work: Considering how processes generally work, that would really surprise me. So please do let me know since it would be relevant for our other ecosystem extensions too.

extensions/poetry/main.ts
andreaphylum: If we can do it easily (e.g. `#[serde(rename = "packageType"]`), yeah, otherwise we are going to add camel case variable names and Clippy is not going to be happy about it. I _think_ Deno uses `serde` for v8 types so this should be feasible.

extensions/poetry/main.ts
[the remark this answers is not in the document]
cd-work: 100% agree that asking the user to go to the root is the best choice.

extensions/poetry/main.ts
```diff
@@ -42,6 +42,12 @@ async function poetryCheckDryRun(subcommand: string, args: string[]): number {
   await lockfileBackup.backup()
   await manifestBackup.backup()
 
+  if (manifestBackup.fileContent === null) {
+    console.log(`\`pyproject.toml\` was not found in the current directory.`)
+    console.log(`Please move to the Poetry project's top level directory and try again.`)
+    return 127
+  }
```
cd-work, line 49: Can we not check for these files to exist **before** trying to back them up? This seems like a strange order of operations.

Or are you trying to also cover issues like permission errors? In that case the error message isn't clear.

Since this is an error, I'd probably also use `console.error` or `console.warn`, right? To get it to STDERR maybe?

cd-work (changes requested):

cd-work (approved):

extensions/poetry/main.ts
maxrake: I also vote for displaying like poetry does. If the idea was that _something_ should be said then maybe a simple one-line output, at the correct spot, to say that Phylum is analyzing the dependencies would help to explain why the process is taking longer than when just using `poetry` directly.

maxrake (approved):
Tested locally. LGTM.

PR #561, opened 2022-07-19
`extensions run` subcommand

### LINKED ISSUE #551: Add `extension run` subcommand

andreaphylum:
Closes #551.

## Checklist

### DIFF
--- a/cli/src/commands/extensions/extension.rs
+++ b/cli/src/commands/extensions/extension.rs
@@ -141,8 +141,13 @@ impl Extension {
         Extension::try_from(extension_path(name)?)
     }
 
-    /// Return the path to this extension's entry point.
+    /// Return the path to this extension.
     pub fn path(&self) -> PathBuf {
+        self.path.clone()
+    }
+
+    /// Return the path to this extension's entry point.
+    pub fn entry_point_path(&self) -> PathBuf {
         self.path.join(self.entry_point())
     }
 
--- a/cli/src/commands/extensions/extension.rs
+++ b/cli/src/commands/extensions/extension.rs
@@ -63,10 +63,6 @@ impl Extension {
         self.manifest.description.as_deref()
     }
 
-    pub fn entry_point(&self) -> &str {
-        self.manifest.entry_point()
-    }
-
     pub fn permissions(&self) -> Cow<'_, Permissions> {
         match self.manifest.permissions.as_ref() {
             Some(permissions) => Cow::Borrowed(permissions),
@@ -147,8 +143,8 @@ impl Extension {
     }
 
     /// Return the path to this extension's entry point.
-    pub fn entry_point_path(&self) -> PathBuf {
-        self.path.join(self.entry_point())
+    pub fn entry_point(&self) -> PathBuf {
+        self.path.join(self.manifest.entry_point())
     }
 
     /// Execute an extension subcommand.
--- a/cli/src/commands/extensions/mod.rs
+++ b/cli/src/commands/extensions/mod.rs
@@ -1,4 +1,3 @@
-use std::borrow::Borrow;
 use std::collections::HashSet;
 use std::ffi::OsStr;
 use std::fmt::Display;
@@ -9,7 +8,6 @@ use std::path::PathBuf;
 use ansi_term::Color;
 use anyhow::{anyhow, Context, Result};
 use clap::{arg, ArgMatches, Command, ValueHint};
-use deno_runtime::permissions::PermissionsOptions;
 use dialoguer::console::Term;
 use dialoguer::Confirm;
 use futures::future::BoxFuture;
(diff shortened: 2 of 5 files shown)


cli/src/commands/extensions/mod.rs
```diff
@@ -114,6 +125,35 @@ pub async fn handle_run_extension(
     Ok(CommandValue::Code(ExitCode::Ok))
 }
 
+/// Handle the `extension run <PATH>` command path.
+///
+/// Run the extension located at the given path.
+pub async fn handle_run_local_extension(
```
cd-work, line 131: I don't really like the name of `run_local_extension`, kinda makes it sound like there's a way to run remote extensions too.

cli/src/commands/extensions/mod.rs
```diff
@@ -114,6 +125,35 @@ pub async fn handle_run_extension(
...
+    api: BoxFuture<'static, Result<PhylumApi>>,
+    matches: &ArgMatches,
+) -> CommandResult {
+    let path = matches.value_of("PATH").unwrap();
+    let options = matches.get_many("OPTIONS").map(|options| options.cloned().collect());
+
+    let extension_path = PathBuf::from(path);
+    let extension = Extension::try_from(extension_path)?;
+
+    println!("{:?}", extension);
```
cd-work, line 141: Leftover from debugging?

cli/src/commands/extensions/mod.rs
```diff
@@ -114,6 +125,35 @@ pub async fn handle_run_extension(
...
+
+    // Attempt to construct a `PermissionsOptions` from the `Permissions`
+    // object in order to validate the permissions.
+    let permissions = extension.permissions();
+    let _ = PermissionsOptions::try_from(permissions.borrow())?;
```
cd-work, line 146: This seems like a really hacky way to "validate" permissions.

cli/src/commands/extensions/extension.rs
```diff
@@ -141,8 +141,13 @@ impl Extension {
...
     }
 
-    /// Return the path to this extension's entry point.
+    /// Return the path to this extension.
     pub fn path(&self) -> PathBuf {
+        self.path.clone()
+    }
+
+    /// Return the path to this extension's entry point.
+    pub fn entry_point_path(&self) -> PathBuf {
```
cd-work, line 150:
```suggestion
    pub fn entry_point(&self) -> PathBuf {
```

Probably be fine to omit the type here? Path is the only representation we can return for entrypoint here anyway and that just makes things a little more concise.

cli/src/deno.rs
```diff
@@ -91,14 +92,20 @@ pub async fn run(
...
+        Self { extension_path: Rc::new(extension_path) }
+    }
+
+    async fn load_from_filesystem(extension_path: &Path, path: &Url) -> Result<String> {
         let path = path.to_file_path().map_err(|_| anyhow!("{path:?}: is not a path"))?;
 
-        let extensions_path = extension::extensions_path()?;
-        if !path.starts_with(&extensions_path) {
+        if !path.starts_with(extension_path) {
+            eprintln!("{:?}", extension_path);
```
cd-work, line 108: Want to remove this too probably.

cd-work (changes requested):

cli/src/commands/extensions/mod.rs
andreaphylum: I copied it over verbatim from `handle_install_extension`. What could be a better way?

cli/src/commands/extensions/extension.rs
andreaphylum: It would conflict with the other `entry_point` function at line 66, which returns the filename of the entry point (i.e. what is in the manifest). I agree the name is not the best.

cli/src/commands/extensions/mod.rs
andreaphylum: Changed it to `handle_run_extension_from_path`.

cli/src/commands/extensions/mod.rs
cd-work: There's a `From<&Permissions> for PermissionsOptions` impl, so isn't this basically just a no-op that will **always** succeed anyway? I think this code might be pointless.

andreaphylum: Indeed.

cli/src/commands/extensions/extension.rs
andreaphylum: We could make both more explicit: `entry_point_directory` and `entry_point_filename`. What do you think?

cd-work: It seems like `entry_point` is only ever used to `join` with the manifest path? In that case having the `entry_point` method as the full path might actually clean things up?

cd-work (changes requested):

cd-work (approved):

PR #562, opened 2022-07-19
docs: update extension documentation

maxrake:
## Completed Actions

* Rename extensions docs to have unique page slugs
  * Readme.com does not namespace the slugs by category, so all pages on <https://docs.phylum.io/> need unique page slugs, which are created from the markdown filename
* Add man pages for extension subcommands `new` and `run`
  * These were missing
  * Check for proper understanding
* Change instances of phylum extension `add` to `install`
* Add a full example manifest file to `docs/extensions/extension_manifest.md`
  * The TOML format can be tricky if unseen before and showing an example with all the possible fields populated helps
* Update extension documentation for style, format, and content
  * General updates to make things more consistent and understandable, at least IMO

## Possible Actions and Remaining Questions

* For `cli/src/extension_api.ts`
  * `getJobStatus` has comments/docs that say "Get job results", which sounds like something else
    * Are they meant to be worded differently
* Update the asciinema link in the main `README.md` file
  * Not _really_ related to extensions work...but noticed during this review
* How is the `run` subcommand expected to work?
  * The documentation may need to be updated and/or confirmed
  * Running the following was expected to work, but did not:

  ```console
  ❯ cargo build --all-features
  [---TRIMMED---]
  ❯ target/debug/phylum extension run --yes ./extensions/duplicates ./cli/tests/fixtures/package-lock.json
  ❗ Error: Execution failed caused by: `/Users/maxrake/dev/phylum/localdev/cli/extensions/duplicates/main.ts`: importing from paths outside of the extension's directory is not allowed
  ```


docs/extensions/extensions_manifest.md
```diff
@@ -1,37 +1,40 @@
...
-     - [`[[run]]`] — Required process execution permissions
-     - [`[[net]]`] — Required network domain permissions
+- [`name`] — Extension name
+- [`entry_point`] — Execution entry point
+- [`permissions`] — Permissions required for execution
+  - [`[[read]]`][read] — Required read path permissions
+  - [`[[write]]`][write] — Required write path permissions
+  - [`[[env]]`][env] — Required environment variable permissions
+  - [`[[run]]`][run] — Required process execution permissions
+  - [`[[net]]`][net] — Required network domain permissions
```
maxrake, line 19: This still may not be the best way to represent the TOML tables, but the old way was definitely wrong (see the current formatting here - <https://docs.phylum.io/docs/manifest_format>). The changes here create the desired links and looks like this:

---

image

cd-work: I've followed [cargo manifest format](https://doc.rust-lang.org/cargo/reference/manifest.html) for documenting things here. I find their methods to be best at getting the point across in list form.

docs/extensions/extensions_example.md
```diff
@@ -149,9 +152,9 @@ for each dependency.
 
 This results in a simple array with all dependencies and their versions:
 
-```
+```json
 {
-  accepts: ["1.3.8", "1.0.0"],
+  "accepts": ["1.3.8", "1.0.0"],
```
cd-work, line 157: This quote also doesn't match `console.log` output.

docs/extensions/extensions_example.md
```diff
@@ -185,14 +188,14 @@ more than one version in our lockfile.
 
 For our example, the output looks like this:
 
-```
+```console
```
cd-work, line 191: What language is this even supposed to be? I don't think we need syntax highlighting for every little code block.

docs/extensions/extensions_overview.md
```diff
@@ -4,13 +4,17 @@ category: 62c5cb137dbdad00536291a6
 hidden: true
 ---
 
+## Overview
```
cd-work, line 7: Same argument as the previous overview heading applies here too.

cd-work (changes requested):

docs/command_line_tool/phylum_extension_run.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+### Options
+
+`-y`, `--yes`
+&emsp; Automatically accept requested permissions
+
+### Details
+
+The extension will be run without it having been installed first.
```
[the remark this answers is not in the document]
maxrake: How about:
`The provided extension will be run without prior installation. The provided extension will not be installed. If the extension is already installed, the provided path to the extension is the one that will be run.`

It is wordier, but perhaps more clear. More/other suggestions and tweaks to this language are encouraged.

docs/extensions/extensions_api.md
```diff
@@ -1,9 +1,11 @@
 ---
-title: Extension API
+title: Extensions API
```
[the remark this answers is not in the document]
maxrake: A coin flip on which way sounded better. There is no real reason to make it plural and it will be changed back to singular in the next commit.

docs/extensions/extensions_api.md
```diff
@@ -1,9 +1,11 @@
...
 category: 62c5cb137dbdad00536291a6
 hidden: true
 ---
 
+## Overview
```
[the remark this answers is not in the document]
maxrake: The reason for this change is b/c the Readme.com settings for Phylum were recently changed to include an auto-generated Table of Contents along the right hand side of pages. For pages that have more than just a line or two to start with, it seemed wrong to not have a clickable link in the ToC to the "top" of the page. See the current [api](https://docs.phylum.io/docs/api) or [manifest_format](https://docs.phylum.io/docs/manifest_format) pages for that view.

That said, it is just a style choice and not one that is applied consistently. So it can be changed back, if desired.

docs/extensions/extensions_example.md
```diff
@@ -108,7 +111,7 @@ function is asynchronous, we need to `await` it.
 
 The list of packages will look something like this:
 
-```
+```json
```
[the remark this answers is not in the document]
maxrake: Roger that. The code was not run to confirm. It was just a guess...but a wrong one. It will be changed.

docs/extensions/extensions_example.md
```diff
@@ -128,9 +131,9 @@ tell it which field to group by using `dep => dep.name`.
 
 This will transform our package list into the following:
 
-```
+```json
 {
-  accepts: [
+  "accepts": [
```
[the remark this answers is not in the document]
maxrake: Roger that. The code was not run to confirm. It was just a guess...but a wrong one. It will be changed.

docs/extensions/extensions_example.md
maxrake: It is meant to represent code formatting for a shell session. [GitHub markdown](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks) supports this code fence "language" via [linguist](https://github.com/github/linguist/blob/59492635f1f7c156289aa38be14f56037a1c6784/lib/linguist/languages.yml#L6090), but apparently [Readme.com does not](https://rdmd.readme.io/docs/code-blocks).

The only reason for ensuring a language is specified for all fenced code blocks is to make [markdownlint (rule MD040)](https://github.com/DavidAnson/markdownlint/blob/v0.25.1/doc/Rules.md#md040) happy...which is enabled in my IDE and may be included in future QA checks across Phylum's repos. The rule's rationale is:

> Specifying a language improves content rendering by using the correct syntax highlighting for code.

That said, it may not make sense for _every_ code block, but doing so makes it clear that it was not accidentally forgotten.

docs/command_line_tool/phylum_extension_run.md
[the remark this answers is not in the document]
cd-work: I'm not sure this is more clear, to me it kinda just sounds even more confusing.

The `The provided extension will not be installed` seems somewhat redundant, it reads like you just said the same thing twice.

The last sentence has me all confused about which path of the extension will actually be executed, when in reality it really just doesn't matter if it's installed or not.

I think any focus on installed or not will just confuse people, because the installation status is ultimately irrelevant.

docs/extensions/extensions_example.md
cd-work: > That said, it may not make sense for every code block, but doing so makes it clear that it was not accidentally forgotten.

That's precisely the issue though, just making stuff colorful with no rhyme or reason doesn't actually improve anything. For normal text the `console` thing will just color everything blue, even on GitHub. That's not actually an improvement, quite the opposite actually.

Having just plain text codeblocks shouldn't be an issue.

docs/extensions/extensions_manifest.md
maxrake: ah...that cargo format is nice.

If the understanding is correct, then the only "table" that can exist in an extension manifest is `[permissions]`. The permission types (`read`, `write`, `env`, `run`, and `net`) are all just key-value pairs, correct? If the `[permissions]` table is allowed to show up multiple times in a manifest, then it can be represented as an array of tables (e.g, `[[permissions]]`).

maxrake: The most recent changes cause the output to look like this:

---

image

cd-work: Why did you remove the array double brackets?

docs/extensions/extension_api.md
```diff
@@ -19,13 +21,13 @@ utility functions for several commonly used structures like collections, http,
...
 [Deno's API docs]: https://doc.deno.land/deno/stable
 [Deno's standard library docs]: https://deno.land/std
 
 ## Phylum API
 
-The Phylum extension API is documented in the [TypeScript module file].
+The Phylum extension API is documented in the [TypeScript module file][api].
 
-[TypeScript module file]: ./cli/src/extension_api.ts
+[api]: https://github.com/phylum-dev/cli/blob/main/cli/src/extension_api.ts
```
cd-work, line 33: If you're going to assign the URL a specific key, I'd recommend using one that's going to be unique. Having it called "api" in a file full of documentation about different APIs is going to be confusing and potentially cause issues in the future.

docs/extensions/extension_example.md
```diff
@@ -212,7 +215,7 @@ await Deno.writeTextFile(
 When replacing the `console.log` with this function call and executing our
 extension, you'll run into the following error:
 
-```
+```console
 ❗ Error: Execution failed caused by: Error: Requires write access to "./duplicates.txt"
 ```
```
cd-work, line 220: I think this should be text too, no reason to color it blue (at least that's what GitHub does).

docs/extensions/extension_manifest.md
```diff
@@ -1,37 +1,47 @@
...
+
+- [`name`] — Extension name
+- [`description`] - Description of the extension
+- [`entry_point`] — Execution entry point
+- [`[permissions]`][perms] — Permissions required for execution
+  - [`read`][read] — Required read path permissions
+  - [`write`][write] — Required write path permissions
+  - [`env`][env] — Required environment variable permissions
+  - [`run`][run] — Required process execution permissions
+  - [`net`][net] — Required network domain permissions
```
cd-work, line 20: These are arrays, so double brackets were correct.

docs/extensions/extension_manifest.md
```diff
@@ -92,3 +102,27 @@ If the requested domain requests a redirect, you'll also require permissions to
...
+read = [
+    "./path/to/file.txt",
+    "./path/to/directory",
+    "./config_file.yaml",
+]
+write = ["./output_file.txt"]
+env = ["PHYLUM_API_KEY"]
+run = ["npm", "yarn"]
+net = ["www.phylum.io", "phylum.io"]
+```
```
cd-work, line 128: I wonder how much value this actually provides? This is **a lot** of work to maintain, having to constantly update it whenever the manifest is changed. Meanwhile it doesn't really provide much benefit for people that know how toml works (which is and should be documented on the toml GitHub).

I feel like it would be more beneficial to document individual sections, but combining it all should really be something any developer is capable of.

cd-work (changes requested):

docs/extensions/extensions_manifest.md
maxrake: `permissions` appears to be a [table](https://toml.io/en/v1.0.0#table), specified only one time in the manifest. If `permissions` can be...or is expected to be...specified more than one time in the manifest, then it would be an [array of tables](https://toml.io/en/v1.0.0#array-of-tables).

docs/extensions/extension_manifest.md
maxrake: These are [**key-value pairs**](https://toml.io/en/v1.0.0#keyvalue-pair)...which have no extra notation (even if the value is an array).
Single brackets are used to represent [**tables**](https://toml.io/en/v1.0.0#table).
Double brackets are used to represent [**arrays of tables**](https://toml.io/en/v1.0.0#array-of-tables).

The [cargo manifest documentation](https://doc.rust-lang.org/cargo/reference/manifest.html), already pointed to for reference, makes use of this same format.

docs/extensions/extension_manifest.md
maxrake: The ambiguity already exposed in this review, for expressing the manifest file format in documentation, is proof that an example is helpful. Seeing an example is worth a thousand words (not quite the maxim...but close 😉).

Putting an example in each section is also good. The key-value pairs are ambiguous in what the value is expected to be. Saying something like "...where the value is an array of strings." helps, but a concrete example really drives the point home.

docs/extensions/extension_example.md
maxrake: Roger that...will do.

docs/extensions/extension_api.md
maxrake: Roger that...it will be changed. The choice here was due to line length considerations. It appears that the file was using 80 characters for the width. Numbers are also possible...kind of like a footnotes style. Would that be preferable? Or, would a more descriptive key that causes the line to exceed 80 be acceptable. FWIW...the readme.com rendering allows for lines up to 118 characters before wrapping. GitHub doesn't care about the length and will softwrap anything until it sees 2+ newlines in a row.

docs/extensions/extensions_manifest.md
cd-work: Good point, the distinction between "array" and "array of tables" makes sense.

docs/extensions/extension_api.md
cd-work: First of all I don't separated URL links should take line length into consideration. That's kinda the entire point of putting them outside of the block of text. If the URL is longer than 80 columns, you wouldn't be able to use it at all otherwise.

I don't think footnotes like `[1]` would be superior, that would just make things more difficult to follow along. This also would likely only work if you put all the links at the bottom of the entire page. I personally think it's nicer to have them after each paragraph with more descriptive IDs.

If you still want to give it an ID and it **must** absolutely be below 80 characters total, I'd go with `ext_api` at least. But I think we shouldn't care about the column limitation.

docs/extensions/extension_manifest.md
cd-work: You can just as well put examples under individual sections though, thus documenting what people are actually interesting in.

That will also make it very natural for people that add new documentation, giving them the opportunity to decide if they want their description accompanied by an example.

I'm not against examples in general, I just don't think a big blob of the entire manifest (however big it might get) is doing us any good.

docs/extensions/extension_api.md
maxrake: Agreed...column limitation shouldn't matter. The thought was not that the output would exceed 80 characters in the rendered form, but that the view in the IDE would...and didn't conform to the existing (although unspoken) limit of 80. It will be put back to the original form:

```
[TypeScript module file]: https://github.com/phylum-dev/cli/blob/main/cli/src/extension_api.ts
```

docs/extensions/extension_manifest.md
maxrake: Roger that...change coming.

docs/extensions/extension_api.md
cd-work: > The thought was not that the output would exceed 80 characters in the rendered form, but that the view in the IDE would...and didn't conform to the existing (although unspoken) limit of 80.

Honestly I don't find this relevant at all. IDEs should soft-wrap if the line is too long and even if it was completely cut off it's just a random URL with the important description of what it links to being right at the start; clearly visible.

So ultimately even if it's a thousand columns wide, it doesn't matter, because nobody actually is expected to read the URL. But you're expected to read the paragraphs.

docs/extensions/extension_manifest.md
```diff
@@ -35,14 +35,23 @@ format. Manifest files consist of the following sections:
...
+name = "hello-world_1"
+```
 
 ## Description
 
 The description is an optional short blurb about the extension. This should be
 plain text (not Markdown).
 
+```toml
+name = "hello-world_1"
```
cd-work, line 51: I don't think it's necessary to always provide the name? The goal should be to to have a subsection of the config documented, not to provide a fully working example.

cd-work (changes requested):

cd-work (approved):

PR #564, opened 2022-07-21
Refactor CLI output formatting

### LINKED ISSUE #106: Swap custom 'Renderable' implementation for a serde serializer
Renderable is used to provide human readable output when querying Projects/Packages by the cli.

It could instead be replaced with implementing a custom serializer to make use of the existing serde infrastructure.

See https://github.com/serde-rs/example-format/blob/master/src/ser.rs for an example

cd-work:
This patch combines the two `Summarize` and `Renderable` traits into a
single `Format` trait, to simplify the CLI outputting process.

Instead of filtering issues as part of the `Summarize` trait, there's
now also a separate `FilterIssues` trait which allows filtering issues
on demand.

The `print_response` method responsible for handling errors and
outputing JSON/summarize based on pretty flag was removed in favor of
anyhow error propagation in combination with `Format::write_stdout`.

The `--filter` flag was also added to the `phylum package` subcommand,
since this was the only subcommand that outputs issues which didn't
allow for issue filtering.

Closes #106.


cli/src/app.rs
```diff
@@ -174,6 +174,7 @@ pub fn app<'a>() -> clap::Command<'a> {
                     .short('j')
                     .long("json")
                     .help("Produce output in json format (default: false)"),
+                Arg::new("filter").long("filter").value_name("filter").help(FILTER_ABOUT),
```
cd-work, line 177: I added the `--filter` flag to `phylum package`, because I see no reason why it wouldn't have this option. It lists issues so those can be easily filtered.

cli/src/histogram.rs
```diff
@@ -0,0 +1,85 @@
...
+        buckets
+    }
+}
+
+impl fmt::Display for Histogram {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let scale = |s| {
+            let max = *self.values.iter().max().unwrap_or(&1) as f32;
+            56.0 * f32::log2(s) / f32::log2(max)
+        };
```
maxrake, line 59: There is a bug here. When `s` is `1`, the return value is `0`...which manifests as a "missing" histogram bar. This can be seen by viewing the same results from both the UI and the CLI:

---

Screen Shot 2022-07-22 at 5 18 04 PM


---

This is not new code here (just moved)...so feel free to dismiss this as out of scope. If so, a new issue should be filed.

maxrake (changes requested):
This was built and tested locally. It appears to work as intended. However, there are too many Rust changes for me to feel comfortable marking this review as approved w/o someone else taking a look first. Even still, a few issues were found, one of which (documentation formatting) should be addressed in this review.

docs/command_line_tool/phylum_package.md
```diff
@@ -3,19 +3,33 @@ title: phylum package
...
+`--filter`
+&emsp; Provide a filter used to limit the issues displayed
+
+EXAMPLES
+# Show only issues with severity of at least 'high'
+    --filter=high
+
+# Show issues with severity of 'critical' in the 'author'
+and 'engineering' domains
+    --filter=crit,aut,eng
```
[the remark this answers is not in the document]
cd-work: I've just copied the output from the `--help`, where it is also documented on the individual flag. I'm not sure if we should move it?

The `phylum_history.md` file just removes this stuff completely.

cli/src/histogram.rs
cd-work: Great catch, this should be an absolutely trivial fix and I see no reason why we'd keep carrying this bug around with us.

docs/command_line_tool/phylum_package.md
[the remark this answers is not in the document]
maxrake: Flipping a coin now...

...came up tails...remove it to match `phylum_history.md`.

cli/src/histogram.rs
```diff
@@ -0,0 +1,83 @@
...
+
+            let bar = "█".repeat(scale(*count));
+
+            histogram += &format!("\n{:>4} - {:<4} [{:>5}] {}", bar_min, bar_max, count, bar);
+        }
+
+        write!(f, "{:^10} {:>8}", "Score", "Count")?;
+        write!(f, "{}", histogram)
+    }
+}
```
cd-work, line 83: I basically went for a complete rewrite of this function, since (imo) the previous version was not maintenance friendly.

I hope the expanded verbosity with more descriptive intermediate variables and comments makes it much easier to understand even for people unfamiliar with Rust what is going on here.

I'd be curious what you think @maxrake.

maxrake: This re-write is much easier to follow. The descriptive names and better spacing/layout of the code sections does help readability. Ensuring that non-zero package counts always show at least one bar takes care of the bug where _no_ bar was displayed for a bucket package count of 1.

There are still some edge cases in the scaling output that might look odd, but likely not worth worrying about in this review. For instance...an extreme case is when each bucket with a package has _exactly_ 1 package:

---

image

---

The bars are all the same length...which is expected, but they are 1 bar wide instead of 56 bars wide. Similarly, the "scaling" of a bucket with a single package in it will _always_ be 1 bar wide, which is most often going to be incorrect when considering a logarithmic scale:

---

image

image

---

Again, this shouldn't be taken as reason to make further changes...just that there are some edge cases that might still be unaccounted for.

cd-work: Thanks for pointing this out. If someone has a suggestion for changing the scaling algorithm completely I'm happy to do so, but I think with this PR it's best to just focus on fixing the critical issue and keeping the underlying algorithm the same.

maxrake: Nothing comes to mind for me...and I also like keeping this PR focused on the critical issue.

That said, @kylewillmon has a history of enjoying these kinds of math/logic/algorithm problems...and usually finding elegant solutions!

kylewillmon:
```
let scaled = 56.0 * f32::log2(count as f32) / f32::log2(max);
```

This line presents a big problem when `max` is 1 because `log2(1)` is 0.... So that becomes `NaN` and then it becomes 0 after a cast to `usize` (which I had to figure out [on the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f83e604449ce310654f6e737cd5b7771)).

The simple solution to this is to just double the numbers before the call to `log2()`

```
let scaled = 56.0 * f32::log2(count as f32 * 2.0) / f32::log2(max * 2.0);
```

maxrake: Simple...elegant...love it!

The divide by zero issue was also observed in code earlier, but running it did not result in an error (see screenshot above) so the thought was there was some sort of short circuit in play with the numerator being 0. That playground link is cool and really helps in understanding what is really going on.

maxrake: The change works:

image

kylewillmon (approved):

cli/src/filter.rs
```diff
@@ -2,7 +2,50 @@ use std::collections::HashSet;
...
+}
+
+impl FilterIssues for PackageStatusExtended {
+    fn filter(&mut self, filter: &Filter) {
+        self.issues.retain(|issue| !should_filter_issue(filter, issue.severity, issue.domain));
+    }
+}
+
+/// Check if a package should be filtered out.
+fn should_filter_issue(filter: &Filter, level: RiskLevel, risk_type: impl Into<RiskType>) -> bool {
```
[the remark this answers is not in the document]
cd-work: How is it unusual? The feature exists specifically for this purpose.

cli/src/commands/jobs.rs
```diff
@@ -8,43 +8,39 @@ use phylum_types::types::common::{JobId, ProjectId};
...
+    let resp = match resp {
+        Ok(resp) => resp,
+        Err(err) => {
+            if err.status() == Some(StatusCode::NOT_FOUND) {
+                print_user_warning!(
+                    "No results found. Submit a lockfile for processing:\n\n\t{}\n",
+                    Blue.paint("phylum analyze <lock_file>")
+                );
+            }
+            return Action::None;
```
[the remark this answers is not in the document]
cd-work: The errors get propagated properly now.

cli/src/filter.rs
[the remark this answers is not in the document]
kylewillmon: Fair enough. [The book](https://doc.rust-lang.org/book/ch10-02-traits.html#trait-bound-syntax) agrees with you.

> The `impl Trait` syntax is convenient and makes for more concise code in simple cases, while the fuller trait bound syntax can express more complexity in other cases.

I guess I'm just not used to it for some reason... Ignore this comment.

kylewillmon (approved):

cli/src/filter.rs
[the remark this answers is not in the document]
cd-work: I mean it was introduced fairly recently, just a couple years back. But the goal is specifically to avoid having to add type parameters every time when it is not necessary.

There's still good reasons to prefer generics in many cases, `impl Trait` can't be used to specify the exact same type twice for example (`(x: T, y: T)`) and the type can't be manually specified by the caller (`foo::<Xyz>(x)`). But since none of them are relevant here, there's no reason not to use the shorthand.

kylewillmon (approved):

PR #565, opened 2022-07-22
Update NPM extension with passthrough

cd-work:
This adds pathrough of commands other than `npm install` to the `phylum
npm` extension.

### DIFF
--- a/extensions/npm/main.ts
+++ b/extensions/npm/main.ts
@@ -1,17 +1,18 @@
 import { PhylumApi } from "phylum";
 
-// Parse CLI args.
-const args = Deno.args;
-if (args.length != 2 || args[0] !== "install") {
-    console.error("Usage: phylum npm install <PKG>");
+if (Deno.args.length >= 1 && 'install'.startsWith(Deno.args[0])) {
+    // Analyze new dependencies with phylum before install.
+    await install(Deno.args);
 } else {
-    await install(args[1]);
+  let status = await Deno.run({ cmd: ['npm', ...Deno.args] }).status()
+
+  Deno.exit(status.code)
 }
 
 // Analyze and install package.
-async function install(pkg: string) {
+async function install(args: string[]) {
     console.log("Updating package lock…");
-    await Deno.run({ cmd: ["npm", "i", "--package-lock-only", pkg] }).status();
+    await Deno.run({ cmd: ["npm", "i", "--package-lock-only", ...args] }).status();
     console.log("Package lock updated.\n");
 
     console.log("Analyzing packages…");
@@ -22,11 +23,13 @@ async function install(pkg: string) {
         console.log("All packages pass project thresholds.\n");
 
         console.log(`Installing '${pkg}'…`);
-        await Deno.run({ cmd: ["npm", "i", pkg] }).status();
+        await Deno.run({ cmd: ["npm", "i", ...args] }).status();
         console.log("Package install complete.");
     } else if (jobStatus.pass) {
         console.warn("Unknown packages were submitted for analysis, please check again later.");
+        Deno.exit(9990);
     } else {
         console.error(`Installing '${pkg}' caused threshold failure.`);
+        Deno.exit(9991);
     }
 }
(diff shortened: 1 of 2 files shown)


maxrake (approved):
This extension appears to have the same behavior as the one for `poetry` ([under review](https://github.com/phylum-dev/cli/pull/560)). That is, it updates the lockfile but does not revert it in cases of failure. Marking approved anyway since the direct goal of this PR appears to have been met.

extensions/npm/main.ts
```diff
@@ -22,11 +23,13 @@ async function install(pkg: string) {
         console.log("All packages pass project thresholds.\n");
 
         console.log(`Installing '${pkg}'…`);
-        await Deno.run({ cmd: ["npm", "i", pkg] }).status();
+        await Deno.run({ cmd: ["npm", "i", ...args] }).status();
         console.log("Package install complete.");
     } else if (jobStatus.pass) {
         console.warn("Unknown packages were submitted for analysis, please check again later.");
+        Deno.exit(9990);
```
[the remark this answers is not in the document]
cd-work: You've mentioned this in @andreaphylum's PR and I'll just join into the conversation there. I don't think it's a good idea to talk in two different places about the same thing, but thanks for mentioning it for anyone who hasn't followed the other PR.

extensions/npm/main.ts
```diff
@@ -22,11 +23,13 @@ async function install(pkg: string) {
...
     } else {
         console.error(`Installing '${pkg}' caused threshold failure.`);
+        Deno.exit(9991);
```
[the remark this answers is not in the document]
cd-work: I checked what error codes npm omits, found some stuff in the ~100 range and increased it by a magnitude to hopefully stay clear of npm's own exit codes.

extensions/npm/main.ts
```diff
@@ -1,17 +1,18 @@
 import { PhylumApi } from "phylum";
 
-// Parse CLI args.
-const args = Deno.args;
-if (args.length != 2 || args[0] !== "install") {
-    console.error("Usage: phylum npm install <PKG>");
+if (Deno.args.length >= 1 && 'install'.startsWith(Deno.args[0])) {
```
andreaphylum, line 3: Is the usage of `startsWith` here used to only cover `install` and the `i` shorthand, or are there other possible cases? Or does `npm` itself use this strategy?

extensions/npm/main.ts
```diff
@@ -22,11 +23,13 @@ async function install(pkg: string) {
         console.log("All packages pass project thresholds.\n");
 
         console.log(`Installing '${pkg}'…`);
```
andreaphylum, line 25: Is `pkg` here still defined?

extensions/npm/main.ts
cd-work: Npm uses the strategy. i, in, ins, inst, insta, instal, install are all valid. It's absolutely terrible, but someone must have spent too much time configuring cisco routers I guess.

kylewillmon: This doesn't cover all the other aliases: `add`, `isnt`, `isnta`, `isntal`, and `isntall`

cd-work: What the... why do those even exist?! I mean `add` sure, but `isntall`? Really?

cd-work:
This should be up to date with @andreaphylum's poetry extension now, making sure our ecosystem isn't all over the place. I'd appreciate a review from @andreaphylum just to make sure I've captured his intentions appropriately, then I can quickly throw up another patch to do the same with yarn.

extensions/npm/main.ts
```diff
@@ -1,32 +1,85 @@
...
+    await checkDryRun(Deno.args[0], Deno.args.slice(1));
+}
 
-    console.log("Analyzing packages…");
-    const jobId = await PhylumApi.analyze("./package-lock.json");
-    const jobStatus = await PhylumApi.getJobStatus(jobId);
+// Run the command with side effects.
+console.log('Applying changes…');
+let status = await Deno.run({ cmd: ['npm', ...Deno.args] }).status();
+Deno.exit(status.code);
```
andreaphylum, line 44: I know JS has hoisting, but I'd move this at the bottom of the file. Having this part here between `checkDryRun` and `FileBackup` feels a bit inconsistent semantically.

andreaphylum (approved):
Looks good to me. Approving because the suggested change is minor and optional.

andreaphylum (approved):

PR #568, opened 2022-07-26
Use package list parameter for analyze API

cd-work:
This patch changes the `analyze` extension API to take in the package
type and a list of packages, rather than a lockfile path. This comes
with several advantages.

Since we don't do any file I/O anymore, we don't have to worry about
checking for permissions. The lockfile type also doesn't need to be
known to us, instead people can just freely submit any packages for
their ecosystem.

This new API provides a much more composable interface which allows
future extension developers to easily integrate the CLI into their
custom workflows even with custom proprietary lockfiles.

The previous usecase of just analyzing a known lockfile without the
extension having to parse it themselves is still covered by plumbing the
`parseLockfile` API's output into `analyze`.

### DIFF
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -199,8 +199,7 @@ async fn parse_lockfile(
     let lockfile_type = match lockfile_type {
         Some(lockfile_type) => lockfile_type,
         None => {
-            let (packages, package_type) =
-                parse::get_packages_from_lockfile(Path::new(&lockfile))?;
+            let (packages, package_type) = parse::get_packages_from_lockfile(Path::new(&lockfile))?;
             return Ok(PackageLock {
                 package_type,
                 packages: packages.into_iter().map(PackageSpecifier::from).collect(),
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -49,7 +49,7 @@ struct PackageLock {
 async fn analyze(
     op_state: Rc<RefCell<OpState>>,
     package_type: PackageType,
-    mut packages: Vec<Package>,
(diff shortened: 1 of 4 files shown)


cli/src/commands/extensions/api.rs
```diff
@@ -8,38 +8,54 @@ use std::str::FromStr;
...
+        Self { name: descriptor.name, version: descriptor.version }
+    }
+}
+
+/// Parsed lockfile content.
+#[derive(Serialize, Deserialize, Debug)]
+struct PackageLock {
+    packages: Vec<PackageSpecifier>,
+    package_type: PackageType,
+}
```
cd-work, line 43: I've created custom types here for making our extension API a little easier to use, but all this typesystem repackaging can be somewhat exhausting for our own codebase.

While I don't care that much about the lockfile parsing function just spitting out an array with each package containing the same information over and over again, I think for specifying the packages for analysis this makes things much less confusing to the API consumer. And updating `parseLockfile` just helps with plumbing at that point.

andreaphylum:
I like this idea -- in general, the more we can isolate I/O flows in the API, the better.

I'd also gradually review other API functions and see if similar splits could be worth it.

cli/src/commands/extensions/api.rs
```diff
@@ -51,8 +67,17 @@ async fn analyze(
         },
     };
 
+    let packages = packages
+        .drain(..)
```
kylewillmon, line 71: I think using `.into_iter()` here would allow us to avoid the `mut` on the `packages` parameter.

This comment also applies to the code for the reverse of this conversion in `parse_lockfile()` which is using `mut` and `drain(..)` instead of `into_iter()`

kylewillmon (approved):

andreaphylum (approved):

PR #571, opened 2022-07-27
Improve extension error formatting

### LINKED ISSUE #570: Improve extension JavaScript errors
When running a JavaScript module through deno, errors in the code have rich information attached to it:

```
error: Uncaught ReferenceError: ftch is not defined
ftch("https://example.org");
^
    at file:///tmp/testing/main.ts:3:1
```

While our extensions print just the error itself:

```
❗ Error: Execution failed caused by: ReferenceError: ftch is not defined
    at file:///tmp/testing/main.ts:2:1
```

Since Deno can print advanced error details like the exact location the error occurred, this is likely attached directly to the Deno JavaScript error. So ideally it should be fairly easy for us to improve our error messages using the information exposed by Deno.

cd-work:
This patch takes the error formatting from Deno's CLI and applies it to
our extension errors, allowing for a richer JavaScript error responds
which hopefully will help developers with debugging their extensions.

Closes #570.

### DIFF
--- a/cli/src/fmt_deno_error.rs
+++ b/cli/src/fmt_deno_error.rs
@@ -1,0 +1,174 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+// https://github.com/denoland/deno/blob/main/cli/fmt_errors.rs
+
+//! This mod provides DenoError to unify errors across Deno.
+
+use std::fmt::Write as _;
+
+use deno_runtime::colors::{cyan, italic_bold, red, yellow};
+use deno_runtime::deno_core::error::{format_file_name, JsError, JsStackFrame};
+
+// Keep in sync with `/core/error.js`.
+pub fn format_location(frame: &JsStackFrame) -> String {
+    let _internal = frame.file_name.as_ref().map_or(false, |f| f.starts_with("deno:"));
+    if frame.is_native {
+        return cyan("native").to_string();
+    }
+    let mut result = String::new();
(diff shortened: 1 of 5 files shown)


cli/src/fmt_deno_error.rs
```diff
@@ -0,0 +1,174 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+// https://github.com/denoland/deno/blob/main/cli/fmt_errors.rs
```
cd-work, line 2: I initially began writing all of this myself by hand, but after getting somewhat stumped by all the edgecases one has to handle to print a simple error message, I've looked at the Deno CLI's source code and decided it might be smarter to just take their error formatter.

As far as I can tell, this formatter is completely independent from the actual Deno CLI, so there shouldn't be any issues with just copying their output. If this PR is accepted I'd maybe also follow up with an upstream PR to expose this functionality, since it clearly seems at least somewhat useful to other people.

kylewillmon (approved):
Getting deno to expose error formatting upstream would definitely be preferable to us trying to keep `fmt_deno_error.rs` up-to-date. But I think this approach is fine for now.

Also fortunately, deno is MIT licensed (matching our CLI), so there shouldn't be any problems even if we needed to keep this copy/paste approach long term.

cd-work:
Wanted to make sure there's at least a PR upstream before merging, so here it is:
https://github.com/denoland/deno/pull/15345

PR #573, opened 2022-07-28
Switch to deno-like permission system

cd-work:
This removes the permissions from the extension manifest, instead going
for a more Deno-inspired permissions approach.

Permission requests are now interactively presented to the user while
executing the extension. Alternatively `phylum <extension>` and `phylum
extension run` also have the `--allow-all/read/write/net/env/run` flags
which allow for overriding the permissions from the CLI.

## TODO

- [ ] Tests
- [ ] Documentation


cd-work:
@kylewillmon / @andreaphylum The extension system should now be fully working, behaving identically to what `deno` does. Let me know if you think this makes sense.

I've considered following @kylewillmon suggestion to cache the permissions, but I think that's better suited for a separate follow-up PR. This should be fully functional on its own.

cli/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 rust-version = "1.60"
 
 [features]
-default = ["selfmanage"]
+default = ["selfmanage", "extensions"]
```
andreaphylum, line 9: When is the next CLI release planned? Do we want to enable extensions already?

cli/src/commands/extensions/api.rs
```diff
@@ -187,13 +187,14 @@ async fn get_package_details(
 /// Equivalent to `phylum parse`.
 #[op]
 async fn parse_lockfile(
-    op_state: Rc<RefCell<OpState>>,
+    _op_state: Rc<RefCell<OpState>>,
```
andreaphylum, line 190: I _think_ we could get away with just omitting `op_state` altogether here. IIRC it is passed in only if the `#[op]` proc macro finds it among the arguments to the op.

cli/src/commands/extensions/mod.rs
```diff
@@ -44,16 +40,67 @@ pub fn command<'a>() -> Command<'a> {
         .subcommand(
             Command::new("new").about("Create a new extension").arg(arg!([PATH]).required(true)),
         )
-        .subcommand(
+        .subcommand(permission_args(
```
andreaphylum, line 43: I don't love the semantics of passing a `Command` to a function that returns `Command` -- it kind of breaks the flow of the builder pattern. 

I suggest we could create an internal trait `CommandExt` with a `with_permission_args()` method that is implemented `for Command`. It is a bit more convoluted but probably also more idiomatic, and other modules requiring permission arguments could just import the trait.

(I won't block the PR on this, as this is just a minor style thing, the current solution is acceptable)

cli/Cargo.toml
cd-work: Just for my personal tests because I was getting sick of `--all-features`.

cli/src/commands/extensions/api.rs
cd-work: The permissions still need to be validated here, so that's an open TODO. That's why I haven't removed it yet, it'll likely come back.

cli/src/commands/extensions/mod.rs
cd-work: I don't see how that would be more idiomatic? And under the hood it would be exactly the same? I really don't see what you'd gain here, it doesn't make any sense to me.

andreaphylum: Clarity -- I don't understand what `permission_args(command)` means unless I spend some time reading the code.

cli/src/commands/extensions/api.rs
```diff
@@ -187,13 +187,14 @@ async fn get_package_details(
...
     lockfile: String,
     lockfile_type: Option<String>,
 ) -> Result<PackageLock> {
+    // TODO: Validate read permissions
```
kylewillmon, line 194: This seems to be the biggest question left to answer. What will this permission check look like?

cli/src/commands/extensions/mod.rs
```diff
@@ -311,3 +313,36 @@ pub fn installed_extensions() -> Result<Vec<Extension>> {
...
+        return Permissions {
+            read: Permissions::new_read(&Some(Vec::new()), true),
+            write: Permissions::new_write(&Some(Vec::new()), true),
+            net: Permissions::new_net(&Some(Vec::new()), true),
+            env: Permissions::new_env(&Some(Vec::new()), true),
+            run: Permissions::new_run(&Some(Vec::new()), true),
+            ffi: Permissions::new_ffi(&None, false),
+            hrtime: Permissions::new_hrtime(false),
+        };
+    }
```
kylewillmon, line 329: Two comments:

1. Can we just use [`Permissions::allow_all()`](https://docs.rs/deno_runtime/0.71.0/deno_runtime/permissions/struct.Permissions.html#method.allow_all) here?
2. If they user sets `--allow-all`, we can imply they don't want to be prompted. So we should probably set `prompt` to `false`. (In theory this shouldn't matter... I just don't want to hang a script if something goes wrong)

kylewillmon (commented):
The approach here looks good to me. Looking forward to seeing it in action.

cli/src/commands/extensions/mod.rs
kylewillmon: Perhaps a middle ground here would be to use an [`ArgGroup`](https://docs.rs/clap/3.2.15/clap/struct.ArgGroup.html) and add these arguments to each command with `.group(permissions_args())`?

cli/src/commands/extensions/mod.rs
andreaphylum: I'd say no to 1. because we want to explicitly deny the users from doing FFI. Dealing with that would be very tricky and the Deno team also warns the users about it in their [docs](https://deno.land/manual/getting_started/permissions):

> --allow-ffi Allow loading of dynamic libraries. Be aware that dynamic libraries are not run in a sandbox and therefore do not have the same security restrictions as the Deno process. Therefore, use with caution. Please note that --allow-ffi is an unstable feature.

cli/src/commands/extensions/mod.rs
andreaphylum: I was not aware of `ArgGroup` -- it looks like it's perfectly fit for this use case, I like it.

kylewillmon: I misunderstood `ArgGroup`. It works by referencing arguments that you have already added to the command. So it doesn't work for this use case.

However, the behavior I was hoping for can be accomplished with `.args(permission_args())` where `permission_args()` returns `Vec<clap::Arg>`

cd-work:
Since Deno does in fact support recursive read/write permissions, we've decided in our sync meeting to go for a different permissions approach based on the existing manifest version.

PR #574, opened 2022-07-29
Remove Deno's built-in permission prompts

cd-work:
This removes the JavaScript `Deno.permissions.request` operation to
prevent extension developers from sidestepping our manifest permission
system.

### DIFF
--- a/cli/tests/extensions/permissions.rs
+++ b/cli/tests/extensions/permissions.rs
@@ -87,10 +87,12 @@ pub async fn get_package_details() {
     let test_cli = TestCli::builder().with_config(true).build();
 
     test_cli
-        .extension("\
+        .extension(
+            "\
         await Deno.permissions.request({ name: 'net' });
-        await fetch('https://phylum.io');
-        ")
+        await fetch('https://phylum.io');\
+             ",
+        )
         .build()
         .run()
         .failure()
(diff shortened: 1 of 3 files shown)


andreaphylum (dismissed):

cli/tests/extensions/permissions.rs
```diff
@@ -81,3 +81,20 @@ fn correct_run_permission_successful_install_and_run() {
...
+        .extension(
+            "\
+        await Deno.permissions.request({ name: 'net' });
+        await fetch('https://phylum.io');\
+             ",
+        )
+        .build()
+        .run()
+        .failure()
+        .stderr("❗ Error: Execution failed caused by: Requires net access to \"phylum.io\"\n");
```
kylewillmon, line 99: So the `request` function is now just a silent no-op? I think that might confuse extension developers.

In a perfect world, I'd prefer to have `request` throw an exception. But if that's not possible, we should at least add a note in the documentation.

cd-work: Technically it's not just a silent noop, it returns the following:

```
PermissionStatus { state: undefined, onchange: null }
```

kylewillmon: That's fair. And it's similar to what would be returned if the request were denied (i.e., `PermissionStatus { state: "denied", ... }`)...

So a Deno developer would probably be checking the return value to see if the permission was granted (i.e., `res.state === "granted"`). That test will fail, but that's probably fine.

Either way, I do think we should mention in the docs that we have disabled this functionality.

kylewillmon (approved):

PR #577, opened 2022-08-01
Bump dependency major versions

cd-work:
This fixes the issue in our automated weekly dependency updates which
caused them not to build without intervention.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -41,7 +41,7 @@ reqwest = { version = "0.11.3", features = ["blocking", "json", "rustls-tls"], d
 routerify = { version = "3.0.0", features =["all"] }
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = "^1.0"
-serde_yaml = "0.8.14"
+serde_yaml = "0.9.2"
 serde-xml-rs = "0.5.1"
 sha2 = "0.10.2"
 shellexpand = "2.0.0"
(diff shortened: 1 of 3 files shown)


cli/Cargo.toml
```diff
@@ -52,13 +52,13 @@ tokio = { version = "^1.0", features = ["full"] }
 toml = "0.5.8"
 unicode-width = "0.1.9"
 url = { version = "2", features = ["serde"] }
-zip = "0.6.2"
+zip = { version = "0.6.2", default-features = false, features = [] }
```
cd-work, line 55: The latest version of zip doesn't actually build without disabling its password support, since the rust crypto folks have a habit of yanking old versions and zip seems rather poorly maintained.

The master is up to date, but unfortunately no release has been made since then. But it seems like we don't need all those extra features anyway, since we just want to extract zips.

maxrake (approved):
The updates look good and build now. The remaining failures appear to be due to the Phylum account associated with the `secrets.PHYLUM_TOKEN_STAGING` token not being designated as a Pro account. This might have to be changed by @furi0us333 or @louislang

cd-work:
@maxrake Yeah, currently all the PRs are failing due to this. @andreaphylum is working with @louislang to resolve this.

maxrake (approved):
Tests are passing again. LGTM.

Issue #582, opened 2022-08-02
Config file overwritten with temporary values


kylewillmon:
The CLI writes to the config file almost every time it is run (e.g., to update the last update check timestamp). But the values it writes are not always meant to be written to the config file.

For example, a user with a permanent `offline_access` key set in their config file may decide to use the `PHYLUM_API_KEY` environment variable to change the key for a single run of the CLI. But the CLI will write this value to the config file (even if it's completely invalid like in the example below). This is unexpected.

# How To Reproduce
```
❯ cat ~/.config/phylum/settings.yaml | grep offline_access                
  offline_access: [---REDACTED---]

❯ PHYLUM_API_KEY="" phylum auth status                                    
[2022-08-01T17:10:43Z ERROR phylum] Execution failed: Error creating client
    
    Caused by:
        0: HTTP status client error (400 Bad Request) for url (https://login.staging.phylum.io/auth/realms/phylum/protocol/openid-connect/token)
        1: {"error":"invalid_grant","error_description":"Invalid refresh token"}
❗ Error: Execution failed caused by: Error creating client

❯ cat ~/.config/phylum/settings.yaml | grep offline_access
  offline_access: ""
```

# Expected Behavior
Temporary values set by environment variable or command-line argument should not be automatically saved to the user's config file.

PR #584, opened 2022-08-02
ignore empty refresh token from environment

### LINKED ISSUE #572: A message should be printed when attempting to authenticate using an empty token
# Overview
Empty refresh tokens are obviously invalid, and are likely the most common type of invalid refresh token we're going to get. Without needing to assume the refresh token is a certain format and decoding it, CLI could simply check if the refresh token is empty and print a warning or error.

This would be particularly useful when running the CLI in automation. Currently, if your refresh token is empty you get a message `{"error":"invalid_grant","error_description":"Invalid refresh token"}`. If it's running in automation, the environment will often be gone and you won't be able to go back to verify what the token was.

# How to reproduce
```
❯ PHYLUM_API_KEY="" phylum auth status
[2022-08-01T17:03:47Z ERROR phylum] Execution failed: Error creating client
    
    Caused by:
        0: HTTP status client error (400 Bad Request) for url (https://login.staging.phylum.io/auth/realms/phylum/protocol/openid-connect/token)
        1: {"error":"invalid_grant","error_description":"Invalid refresh token"}
❗ Error: Execution failed caused by: Error creating client
```

# Acceptance Criteria
- [ ] Outputs a helpful message informing the user that there is no refresh token when the user attempts to authenticate but the refresh token is set to an empty string.

matt-phylum:
This PR updates the configuration loading code such that if the offline access token is an empty string it gets replaced with `None`. There is already code in a few places that checks whether we have an access token by checking if the access token is `None`, but in certain situations it's possible to set the access token to `Some("")` (I think commonly this happens by setting the environment variable to an empty string), which gets past all the checks and results in an "invalid refresh token" error from the authentication server. After this PR, if the refresh token is empty string, CLI will behave as if the refresh token were not configured at all, displaying a message in some situations and beginning authentication in others.

I think this resolves will resolve the "invalid refresh token" error we see sometimes when running the CI integration, but the integration would still fail in those cases, just with a different behavior. It doesn't seem like we currently have a notion of running noninteractively, so the CLI will likely try to start interactive authentication from the build agent and fail. I'm not sure exactly what that failure will look like. There might need to be some further enhancement there.

Potential issue: If the token is set via an environment variable, it looks like we currently still renew the refresh token and save it back to the config file. If the user is for some reason temporarily using a different refresh token by overriding via environment variables, this may cause their saved credentials to be unexpectedly overwritten. Also, if the environment variable is set to an empty string, every command will prompt for authentication because the environment variable in the calling process scope cannot be updated.

Closes #572 

## Checklist

I don't think there's documentation to update for this change, but I'm not that familiar with the documentation for this area.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1,3 +1,4 @@
+use std::env::VarError;
 #[cfg(unix)]
 use std::fs::{DirBuilder, Permissions};
 use std::io::{self, Write};
@@ -13,7 +14,7 @@ use phylum_types::types::common::ProjectId;
 use phylum_types::types::package::PackageType;
 use serde::{Deserialize, Deserializer, Serialize};
 
-use crate::dirs;
+use crate::{dirs, print_user_warning};
 
 pub const PROJ_CONF_FILE: &str = ".phylum_project";
 
@@ -122,12 +123,17 @@ pub fn read_configuration(path: &Path) -> Result<Config> {
     };
 
     // If an api token has been set in the environment, prefer that
-    if let Ok(key) = env::var("PHYLUM_API_KEY") {
-        config.auth_info.offline_access = Some(RefreshToken::new(key));
-    }
-
-    if config.auth_info.offline_access.as_ref().map(|t| t.as_str().is_empty()).unwrap_or_default() {
-        config.auth_info.offline_access = None;
+    match env::var("PHYLUM_API_KEY") {
+        Ok(key) if !key.is_empty() => {
+            config.auth_info.offline_access = Some(RefreshToken::new(key));
+        },
+        Ok(_) => {
+            print_user_warning!("Ignoring empty PHYLUM_API_KEY");
+        },
+        Err(VarError::NotUnicode(_)) => {
+            print_user_warning!("Ignoring invalid PHYLUM_API_KEY");
+        },
+        Err(VarError::NotPresent) => {},
     }
 
     Ok(config)
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -136,6 +136,12 @@ pub fn read_configuration(path: &Path) -> Result<Config> {
         Err(VarError::NotPresent) => {},
     }
 
+    // The code that checks if we have a token expects `Some(token)` to always be a
+    // usable token.
+    if config.auth_info.offline_access.as_ref().map(|t| t.as_str().is_empty()).unwrap_or_default() {
+        config.auth_info.offline_access = None;
+    }
+
     Ok(config)
 }
 
@@ -197,10 +203,12 @@ mod tests {
 
     use super::*;
 
+    const CONFIG_TOKEN: &str = "FAKE TOKEN";
+
     fn write_test_config(path: &Path) {
         let con = ConnectionInfo { uri: "http://127.0.0.1".into() };
 
-        let auth = AuthInfo { offline_access: Some(RefreshToken::new("FAKE TOKEN")) };
+        let auth = AuthInfo { offline_access: Some(RefreshToken::new(CONFIG_TOKEN)) };
 
         let config = Config {
             connection: con,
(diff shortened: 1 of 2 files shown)


matt-phylum:
Should there be a specific warning emitted in this case, or is the usual "you're not authenticated" good enough?

Is the handling of the environment variable right? Or should we handle empty environment variables separately, possibly ignoring them before overwriting the value from the configuration file?

kylewillmon:
> Potential issue: If the token is set via an environment variable, it looks like we currently still renew the refresh token and save it back to the config file. If the user is for some reason temporarily using a different refresh token by overriding via environment variables, this may cause their saved credentials to be unexpectedly overwritten.

This is #582 and will be fixed soon.

> Also, if the environment variable is set to an empty string, every command will prompt for authentication because the environment variable in the calling process scope cannot be updated.

This is only true if the user sets the environment variable permanently instead of just for one CLI run (e.g., `export PHYLUM_API_KEY=""; phylum auth status` vs `PHYLUM_API_KEY="" phylum auth status`)

But that gets at the root question here. What does it mean if the `PHYLUM_API_KEY` environment variable is set to empty string?

If we accept this PR, then we are interpreting an empty string environment variable as a request to ignore the token in the config file. Which means the CLI will prompt for authentication and overwrite the config file. If that is the user's intention, they should instead just run `phylum auth login`, which will do the same thing.

Instead, I think we should be ignoring empty string environment variables and simply using the variable from the config file.

cli/src/config.rs
```diff
@@ -126,6 +126,10 @@ pub fn read_configuration(path: &Path) -> Result<Config> {
         config.auth_info.offline_access = Some(RefreshToken::new(key));
```
kylewillmon, line 126: This is the line where it is getting set to `Some("")` when an empty string environment variable is used.

If we prevent that assignment (by adding an `if !key.is_empty()` clause around it), then we will use the token from the config, which will probably fix more of the issues that are occurring in the phylum-ci project.

kylewillmon (changes requested):
We need to prevent the `Some("")` assignment from happening in the first place rather than just changing it to `None` after the fact.

cli/src/config.rs
matt-phylum: I've kept the current change to `None` code for the unlikely case that the configuration file contains `offline_access: ""` due to manual editing. Now if the environment variable is set to empty there should be a warning and the environment value will be ignored.

cli/src/config.rs
```diff
@@ -122,8 +123,23 @@ pub fn read_configuration(path: &Path) -> Result<Config> {
...
     // If an api token has been set in the environment, prefer that
-    if let Ok(key) = env::var("PHYLUM_API_KEY") {
-        config.auth_info.offline_access = Some(RefreshToken::new(key));
+    match env::var("PHYLUM_API_KEY") {
+        Ok(key) if !key.is_empty() => {
+            config.auth_info.offline_access = Some(RefreshToken::new(key));
+        },
+        Ok(_) => {
+            print_user_warning!("Ignoring empty PHYLUM_API_KEY");
+        },
```
kylewillmon, line 132: It might be personal preference (and feel free to ignore this comment). But since this is a match statement, you can use `Ok("")` explicitly rather than using a guard.

```suggestion
        Ok("") => {
            print_user_warning!("Ignoring empty PHYLUM_API_KEY");
        },
        Ok(key) => {
            config.auth_info.offline_access = Some(RefreshToken::new(key));
        },
```

cli/src/config.rs
```diff
@@ -122,8 +123,23 @@ pub fn read_configuration(path: &Path) -> Result<Config> {
...
+        Err(VarError::NotUnicode(_)) => {
+            print_user_warning!("Ignoring invalid PHYLUM_API_KEY");
+        },
+        Err(VarError::NotPresent) => {},
+    }
+
+    // The code that checks if we have a token expects `Some(token)` to always be a
+    // usable token.
+    if config.auth_info.offline_access.as_ref().map(|t| t.as_str().is_empty()).unwrap_or_default() {
```
kylewillmon, line 141: There are many shorter ways to write this condition. Here is one possibility.

```suggestion
    if config.auth_info.offline_access.as_ref().map(|t| t.as_str()) == Some("") {
```

kylewillmon (approved):
Thanks for the PR and quick updates!

kylewillmon (approved):

PR #585, opened 2022-08-02
Update extension example documentation

### LINKED ISSUE #580: Fix duplicates extension example permissions
Currently the duplicates extension example used in our extension documentation will not work when following along with the documentation. Since `parseLockfile` requires access to the target lockfile, permissions must be granted before this point already.

Since this will show off our permission system already, it's also not necessary to have the second part of this where we write the duplicates to a file. This part of the example can be removed entirely.

cd-work:
This updates the existing extension example documentation to take the
new permission system into account.

It also updates to the latest `parseLockfile` return type.

Closes #580.

### DIFF
--- a/docs/extensions/extension_example.md
+++ b/docs/extensions/extension_example.md
@@ -215,7 +215,7 @@ the [extension manifest documentation].
 
 While it would be possible to request read access to just `./package-lock.json`,
 this would only work for `package-lock.json` files defeating the purpose of
-passing the lockfile as a paramater. Instead, we request read access to all
+passing the lockfile as a parameter. Instead, we request read access to all
 files in the working directory:
 
 ```toml
--- a/docs/extensions/extension_example.md
+++ b/docs/extensions/extension_example.md
@@ -13,36 +13,34 @@ than one version present in our lockfile.
 The full example looks like this:
 
 ```ts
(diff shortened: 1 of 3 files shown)


docs/extensions/extension_example.md
```diff
@@ -192,43 +190,48 @@ For our example, the output looks like this:
...
-```
+[extension manifest documentation]: https://docs.phylum.io/docs/extension_manifest#permissions
 
-This is exactly what should have happened, since Deno's sandbox doesn't allow us
-to interact with the outside world unless we've been granted permission to do
-so. To request permissions, you'll have to edit the `PhylumExt.toml` manifest
-and add the following:
+While it would be possible to request read access to just `./package-lock.json`,
+this would only work for `package-lock.json` files defeating the purpose of
+passing the lockfile as a paramater. Instead, we request read access to all
```
maxrake, line 218:
```suggestion
this would only work for `package-lock.json` files, defeating the purpose of
passing the lockfile as a parameter. Instead, we request read access to all
```

maxrake (dismissed):
I followed along with the updated example and tested it locally. It all seems to match up and work as advertised:

image

docs/extensions/extension_example.md
cd-work: Good catch, thanks.

maxrake (approved):

PR #588, opened 2022-08-02
Better error messages

kylewillmon:
Use the `description` from API as the error message when available. Also clean up double printing and excessive language.

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -17,7 +17,7 @@ use phylum_types::types::user_settings::UserSettings;
 use reqwest::header::{HeaderMap, HeaderValue};
 use reqwest::{Client, IntoUrl, Method, StatusCode};
 use serde::de::{DeserializeOwned, IgnoredAny};
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use thiserror::Error as ThisError;
 
 use self::endpoints::BaseUriError;
@@ -63,10 +63,23 @@ impl PhylumApiError {
 
 /// Non-successful request response.
 #[derive(ThisError, Debug)]
-#[error("HTTP request error ({code}):\n{body}")]
+#[error("HTTP request error ({code}):\n\n{details}")]
 pub struct ResponseError {
     pub code: StatusCode,
-    pub body: String,
+    pub details: String,
+}
+
+/// The guts of an API JSON error
+#[derive(Deserialize)]
+struct ApiJsonErrorInner {
+    error_id: String,
+    description: String,
+}
+
+/// A JSON error returned by the Phylum API
+#[derive(Deserialize)]
+struct ApiJsonError {
+    error: ApiJsonErrorInner,
 }
 
 impl PhylumApi {
@@ -110,7 +123,14 @@ impl PhylumApi {
         let body = response.text().await?;
 
         if !status_code.is_success() {
-            let err = ResponseError { body, code: status_code };
+            let details = if let Ok(err) = serde_json::from_str::<ApiJsonError>(&body) {
+                log::debug!("Error ID: {}", err.error.error_id);
+                err.error.description
+            } else {
+                body
+            };
+            let err = ResponseError { details, code: status_code };
+
             return Err(err.into());
         }
 
--- a/cli/src/bin/phylum.rs
+++ b/cli/src/bin/phylum.rs
@@ -1,10 +1,10 @@
+use std::fmt::Display;
 use std::path::PathBuf;
 use std::time::UNIX_EPOCH;
 
 use anyhow::{anyhow, Context, Result};
 use clap::ArgMatches;
 use env_logger::Env;
-use log::{error, warn};
 use phylum_cli::api::PhylumApi;
 #[cfg(feature = "extensions")]
 use phylum_cli::commands::extensions;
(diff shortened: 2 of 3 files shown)


cli/src/api/mod.rs
```diff
@@ -110,7 +123,14 @@ impl PhylumApi {
         let body = response.text().await?;
 
         if !status_code.is_success() {
-            let err = ResponseError { body, code: status_code };
+            let details = if let Ok(err) = serde_json::from_str::<ApiJsonError>(&body) {
+                log::debug!("Error ID: {}", err.error.error_id);
+                err.error.description
+            } else {
+                body
+            };
```
kylewillmon, line 131: I'm debating if this is more readable as `.map(...).unwrap_or(...)` 🤷‍♂️

```
            let details = serde_json::from_str::<ApiJsonError>(&body)
                .map(|err| {
                    log::debug!("Error ID: {}", err.error.error_id);
                    err.error.description
                })
                .unwrap_or(body);
```

cli/src/api/mod.rs
```diff
@@ -110,7 +123,14 @@ impl PhylumApi {
...
+            let err = ResponseError { details, code: status_code };
```
cd-work, line 132: Since you're re-wrapping the error here, I'd assume the previous work to have status-code based error message overrides still works?

cli/src/bin/phylum.rs
```diff
@@ -203,6 +193,6 @@ async fn main() {
             ),
         },
         Ok(CommandValue::Code(code)) => code.exit(),
-        Err(error) => exit_error(error.into(), "Execution failed"),
+        Err(error) => exit_fail(error, ExitCode::Generic),
```
cd-work, line 196: What's the motivation behind this change? I'd assume it just didn't look good with the new error messages?

cli/src/api/mod.rs
cd-work: I don't think so. I'd usually go for a match when having two branches anyway, but in this scenario it doesn't really matter and your solution is technically shorter I think.

cd-work (commented):
Do you have any before/after examples at the ready demonstrating how the error messages should be transformed by this patch? I can give it a test run myself, just curious what the expectation is.

kylewillmon:
> Do you have any before/after examples at the ready demonstrating how the error messages should be transformed by this patch?

I meant to add that to the description, sorry. Here it is:

### Before
```
❯ phylum project create foo2 
[2022-08-03T13:05:19Z ERROR phylum] Execution failed: HTTP request error (403 Forbidden):
    {
      "error": {
        "error_id": "c610736f-bc8b-43e4-8b85-8b8dcfe5179d",
        "code": 403,
        "reason": "Forbidden",
        "apiError": {
          "type": "TierExceeded",
          "details": {
            "type": "Projects",
            "details": {
              "count": 5,
              "limit": 5
            }
          }
        },
        "description": "Unlimited projects are a Phylum Pro feature\nPhylum's Community version is restricted to five (5) Phylum projects\nUpgrade your account to Phylum Pro to use unlimited projects https://app.phylum.io/upgrade-cli"
      }
    }
❗ Error: Execution failed caused by: HTTP request error (403 Forbidden):
{
  "error": {
    "error_id": "c610736f-bc8b-43e4-8b85-8b8dcfe5179d",
    "code": 403,
    "reason": "Forbidden",
    "apiError": {
      "type": "TierExceeded",
      "details": {
        "type": "Projects",
        "details": {
          "count": 5,
          "limit": 5
        }
      }
    },
    "description": "Unlimited projects are a Phylum Pro feature\nPhylum's Community version is restricted to five (5) Phylum projects\nUpgrade your account to Phylum Pro to use unlimited projects https://app.phylum.io/upgrade-cli"
  }
}
```
### After
```
❯ phylum project create foo2
❗ Error: HTTP request error (403 Forbidden):

Unlimited projects are a Phylum Pro feature
Phylum's Community version is restricted to five (5) Phylum projects
Upgrade your account to Phylum Pro to use unlimited projects https://app.phylum.io/upgrade-cli
```

cli/src/api/mod.rs
kylewillmon: That's correct:

```
❯ phylum project create test
❗ Project 'test' already exists
```

cli/src/api/mod.rs
```diff
@@ -110,7 +123,14 @@ impl PhylumApi {
         let body = response.text().await?;
 
         if !status_code.is_success() {
-            let err = ResponseError { body, code: status_code };
+            let details = if let Ok(err) = serde_json::from_str::<ApiJsonError>(&body) {
+                log::debug!("Error ID: {}", err.error.error_id);
```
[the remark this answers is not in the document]
kylewillmon: The error ID is a GUID. You can see it in the "before" example. It is unique to every query that generates an error. It's purpose is to be able to trace errors in the backend logs (i.e., if someone submits an error report, we can find that exact query instance in the backend logs). That's why I wanted to make sure we would still print the value.

cli/src/bin/phylum.rs
kylewillmon: There are two changes here:

1. `Error: Execution failed caused by:` is replaced with `Error:`. This was mostly to reduce clutter in the output. I don't think the longer message provided any more information.
2. The `exit_error` function goes away. This was motivated by the first change since `exit_fail` already prints `Error:`, so it was easy to combine the 2 functions.

cd-work (approved):

cd-work (approved):

PR #589, opened 2022-08-03
Combine conflicting test cases

kylewillmon:
These two test cases were fighting over the `PHYLUM_API_KEY` environment
variable causing failures. This patch fixes them in the dumbest way
possible: By merging them into a single test.


cd-work (changes requested):
Is this really a good idea? Sounds like it's just another issue waiting to happen.

Have you considered moving this to integration testing maybe, where things are executed as separate processes? We can still validate tokens using the `phylum token` subcommand.

kylewillmon:
> Is this really a good idea?

Nope. Definitely not. Just the quickest fix I could think of.

PR #592, opened 2022-08-03
Move token tests to integration

### LINKED ISSUE #590: Unit test race condition with environment variables
Multiple unit tests are using the same environment variable, which is causing a race condition and sporadic failures like this:

```
---- config::tests::test_pass_api_key_through_env stdout ----
⚠️  Ignoring empty PHYLUM_API_KEY
thread 'config::tests::test_pass_api_key_through_env' panicked at 'assertion failed: `(left == right)`
  left: `Some(RefreshToken("FAKE TOKEN"))`,
 right: `Some(RefreshToken("ENV VARIABLE TOKEN"))`', cli/src/config.rs:247:9
```

There are multiple ways to fix this, but here is a suggestion from @cd-work:

> Have you considered moving this to integration testing maybe, where things are executed as separate processes? We can still validate tokens using the `phylum token` subcommand.

_Originally posted by @cd-work in https://github.com/phylum-dev/cli/pull/589#pullrequestreview-1060907675_

cd-work:
To allow for easily creating new integration test modules, the
integration test setup has been reworked to work as a single big crate.

The config file is now passed to our CLI builder as an `Option<Config>`,
allowing us to override the configuration file with a different value.

The `run` method was also removed from `TestCli` since it just
unnecessarily restricted the capabilities of the returned type. Instead
the full `Command` is returned so consumers can modify args, env, etc
themselves.

Closes #590.

### DIFF
--- a/cli/tests/common.rs
+++ b/cli/tests/common.rs
@@ -1,4 +1,3 @@
-use std::ffi::OsStr;
 #[cfg(feature = "extensions")]
 use std::fs::File;
 #[cfg(feature = "extensions")]
@@ -17,7 +16,7 @@ pub use predicates::prelude::*;
 use reqwest::StatusCode;
 use tempfile::TempDir;
 
-const API_URL: &str = "https://api.staging.phylum.io";
+pub const API_URL: &str = "https://api.staging.phylum.io";
 const PROJECT_NAME: &str = "integration-tests";
 
 enum Cwd {
@@ -34,14 +33,14 @@ impl Default for Cwd {
 
 #[derive(Default)]
 pub struct TestCliBuilder {
+    config: Option<Config>,
     cwd: Cwd,
-    with_config: bool,
 }
 
 impl TestCliBuilder {
     pub fn build(self) -> TestCli {
         let tempdir = TempDir::new().unwrap();
-        let config_path = if self.with_config { Some(create_config(tempdir.path())) } else { None };
+        let config_path = self.config.map(|config| create_config(tempdir.path(), config));
 
         let cwd = match self.cwd {
             Cwd::Path(p) => Some(p),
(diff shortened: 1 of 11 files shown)


cli/tests/common.rs
```diff
@@ -95,15 +97,15 @@ impl TestCli {
...
+        self.cmd().args(&["extension", "install", "-y", &path.to_string_lossy()]).assert()
     }
 
     #[cfg(feature = "extensions")]
     pub fn extension<'a>(&'a self, code: &'a str) -> TestExtensionBuilder<'a> {
         TestExtensionBuilder::new(self, code)
     }
 
-    pub fn run<S: AsRef<str> + AsRef<OsStr>>(&self, args: &[S]) -> Assert {
+    pub fn cmd(&self) -> Command {
```
andreaphylum, line 108: I'm ok with this change but I'd _also_ add a `run` (or more appropriately named) shorthand method to avoid repeating `cmd().args(&[...]).assert()` everywhere.

cli/tests/end_to_end/extension.rs
```diff
@@ -5,7 +5,7 @@ use crate::common::{create_lockfile, create_project, TestCli};
 /// Test Phylum API functions.
 #[tokio::test]
 pub async fn get_user_info() {
-    let test_cli = TestCli::builder().with_config(true).build();
+    let test_cli = TestCli::builder().with_config(None).build();
```
andreaphylum, line 8: I would rather have something like `Default::default()` here, as `None` intuitively means "with no config", while the semantics of this call are "with a config, but don't specify which".

If we don't need to create a config, we can just skip the `.with_config()` invocation altogether.

andreaphylum (changes requested):

cli/tests/end_to_end/extension.rs
cd-work: > If we don't need to create a config, we can just skip the .with_config() invocation altogether.

Yeah that's exactly why I chose `None` as parameter. These are tests and brevity is somewhat important. So I think having to specify the config every time would be annoying, especially because `Config::default` will not work.

andreaphylum: How about adding `.with_default_config()` or similar name (maybe shorter)? It's an extra one-liner method but it would be used in most places.

cd-work: It wouldn't be the default config though? I'm not sure tbh, I don't like having two different methods like this.

andreaphylum (approved):

cli/tests/end_to_end/extension.rs
andreaphylum: I'd say this is not a dealbreaker anyway. We can keep the current solution.

Issue #593, opened 2022-08-04
Document the extensions that we provide


kylewillmon:
We need to add documentation for the extensions that we provide. These can be documented directly in the repository, with a link in our extensions docs on readme.com

## Acceptance Criteria

- [x] Add a `README.md` file to the top-level `extensions` folder that contains a list of our extensions with a brief description of each
- [x] Add a `README.md` to each extension folder that explains the extension in greater detail including examples of how to run the extension and any supported arguments, etc.
- [x] Add a section to `docs/extensions/extension_overview.md` that will tell users that we have written a small set of extensions and include a link to the folder on GitHub for more information.

PR #599, opened 2022-08-05
Removed extensions feature

andreaphylum:


maxrake:
Does it make sense to also change `hidden: true` to `hidden: false` in each of the files in `docs/extensions/*.md` here? Or, should that happen in #593 ?

cd-work (changes requested):
> Does it make sense to also change hidden: true to hidden: false in each of the files in docs/extensions/*.md here? Or, should that happen in #593 ?

I think that's a great idea, we should make the docs public too.

maxrake (commented):
Does it make sense to also change `hidden: true` to `hidden: false` in each of the files in `docs/extensions/*.md` here? Or, should that happen in #593 ?

cli/src/app.rs
```diff
@@ -312,11 +310,8 @@ pub fn app<'a>() -> clap::Command<'a> {
                 ),
         );
 
-    #[cfg(feature = "extensions")]
-    {
-        app = app.subcommand(extensions::command());
-        app = extensions::add_extensions_subcommands(app);
-    }
+    app = app.subcommand(extensions::command());
```
cd-work, line 313: We can move this directly into the builder, right?

andreaphylum: Yes -- though I'd leave the line after that as-is, outside of the builder.

cd-work (approved):

docs/extensions/extension_api.md
```diff
@@ -1,7 +1,6 @@
 ---
 title: Extension API
 category: 62c5cb137dbdad00536291a6
-hidden: true
```
maxrake, line 4: All the rest of the readme.com docs in this repo keep the attribute instead of removing it. Not sure if it matters b/c the default, when not present, might be `hidden: false`.

maxrake (changes requested):
My apologies...but I forgot to also mention the `docs/command_line_tool/phylum_extension*.md` files that also should be made visible.

docs/extensions/extension_api.md
andreaphylum: Good point, just realized the same by looking at the other files, and reverted this.

maxrake (approved):

PR #602, opened 2022-08-05
Poetry extension `README`

andreaphylum:


extensions/poetry/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## How it works
+
+When invoking `phylum poetry`, subcommands that would modify the
+`pyproject.toml` or `poetry.lock` files will trigger a Phylum analysis.
+
+- If the analysis is successful, the corresponding changes will be applied.
+- If the analysis is unsuccessful because some of the new dependencies don't
+  meet the required project thresholds, the command will exit with the error code
+  127.
```
cd-work, line 36: Not sure we should mention the error code explicitly here. I wouldn't want to nail myself down on that error code being 100% exactly this forever. Ultimately it shouldn't matter to users and if they need it in a script they can still check at their own risk.

extensions/poetry/README.md
```diff
@@ -0,0 +1,51 @@
...
+- If the analysis is waiting for Phylum to process one or more of the submitted
+  packages, the command will exit with the error code 126 and the changes will
+  _not_ be applied.
```
cd-work, line 39: Same exit code concerns here. I'd just use `the command will fail` as wording for these.

cd-work (changes requested):
I think that should be mostly good to go tbh.

extensions/poetry/README.md
```diff
@@ -0,0 +1,51 @@
+# Phylum Poetry extension
+
+A [Phylum CLI](phylum-cli) extension that checks your [Poetry][poetry]
```
maxrake, line 3:
```suggestion
A [Phylum CLI][phylum-cli] extension that checks your Python [Poetry][poetry]
```

extensions/poetry/README.md
andreaphylum: Agree -- though at some point we may want to document that there _is_ indeed a difference between the two failures, if/when we do intend to explicitly expose that to users.

maxrake (changes requested):
The wording gets the main points across and should be good for this first release. One suggested change in wording and one broken link reference to fix.

maxrake (approved):

cd-work (approved):

PR #606, opened 2022-08-05
Musl builds

cd-work:


entrypoint.sh
```diff
@@ -0,0 +1,36 @@
...
+    binutils-gold \
+    sccache
+ln -s /usr/bin/python3 /usr/bin/python
+
+export V8_FROM_SOURCE=yes
+GN="$(pwd)/gn/out/gn"
+export GN
+export CLANG_BASE_PATH=/usr
+export GN_ARGS='use_custom_libcxx=false use_lld=false v8_enable_backtrace=false v8_enable_debugging_features=false'
+# TODO
```
maxrake, line 22: Still TODO?

entrypoint.sh
```diff
@@ -0,0 +1,36 @@
...
+export SCCACHE_DIR="./sccache"
+export SCCACHE="/usr/bin/sccache"
+
+# Bulid GN
```
maxrake, line 26:
```suggestion
# Build GN
```

Cargo.toml
```diff
@@ -4,3 +4,6 @@ members = [
   "cli",
   "xtask",
 ]
+
+[patch.crates-io]
```
louislang, line 8: Might want to get the `rusty_v8` repo under `phylum-dev` at some point, if it's gonna be the long term solution!

louislang (approved):
lgtm! 🚀

maxrake (commented):
Looks like the Windows builders are not happy with a long path:

```
C:/Users/runneradmin/.cargo/git/checkouts/rusty_v8-34148fcc67bb1ca5/7a65f0f/buildtools/third_party/libc++/trunk/test/std/experimental/utilities/propagate_const/propagate_const.class/propagate_const.non-const_observers/explicit_operator_element_type_ptr.pass.cpp
```

Looks like it is about 2 characters too long. Stupid Windows. Not sure if any part of that path is under our control to make shorter. Maybe there is a way to tell Windows to work with longer paths...looking now...

.github/workflows/test.yml
```diff
@@ -82,6 +82,19 @@ jobs:
         if: github.event_name != 'schedule'
         run: shellcheck -o all -S style -s sh $(find . -iname "*.sh")
 
+  musl-test:
```
maxrake, line 85: This job will need to either:

1. Be set as a required status check in the `phylum-dev/cli` settings. This can be done by someone with permissions to the repo and _may_ have to wait until after this PR is merged to `main`.
2. Included in the `test` "rollup job" below...which is already set as a required status check. Doing it that way would look something like this:

```yaml
  # This job reports the results of the test matrix above
  test:
    if: always()
    needs: [test-matrix, musl-test]
    runs-on: ubuntu-latest
    steps:
      - if: needs.test-matrix.result != 'success' || needs.musl-test.result != 'success'
        name: Fail the build
        run: exit 1
```

cd-work:
Oh, sorry for opening this as a non-draft PR, my other one got closed and I couldn't re-open.

This is all still heavily WIP and nothing works.

PR #615, opened 2022-08-10
Build universal binary on macOS

### LINKED ISSUE #611: Publish universal binaries on macOS
Instead of providing the arm/x86 binaries separately, the macos CLI should be provided as a universal binary using `lipo`.

cd-work:
Instead of providing architecture-specific versions of the Phylum CLI,
this instead builds the two versions into a universal version using
`lipo`.

Closes #611.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,7 +86,7 @@ jobs:
   Build-Release-Artifacts:
     name: Build the release artifacts
     needs: [Generate-Completion-Files, Build-CLI]
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     steps:
       - name: Install host dependencies
         run: sudo apt install -yq zip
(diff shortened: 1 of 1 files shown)


cd-work:
We need to wait until aarch64 is enabled for macOS again before merging this, but the process should be the same.

maxrake:
Some more discussion/coordination might be needed before continuing with this PR. Removing the individual, specific, supported platforms in favor of a consolidated `apple-darwin` platform has some ramifications:

* It deviates from the established pattern of using [rust platform/target names](https://doc.rust-lang.org/nightly/rustc/platform-support.html)
* The `phylum-init` script entry point in the `phylum-ci` repo is based on the established artifact naming patterns
  * [constants](https://github.com/phylum-dev/phylum-ci/blob/main/src/phylum/constants.py#L6-L20) 
  * [`get_target_triple()`](https://github.com/phylum-dev/phylum-ci/blob/main/src/phylum/init/cli.py#L204-L208)
* The `phylum-init.sh` script in the `cli` repo makes assumptions about the target triple
  * [code here](https://github.com/phylum-dev/cli/blob/main/scripts/phylum-init.sh#L9-L40)
  * It is not clear if this script is still used for anything

The changes can still be made and synchronized across the various repos and documentation that relies on the current naming pattern...but there might be alternative solutions that don't require breaking the pattern.

cd-work:
We talked this over in our sync and didn't see any immediate advantage to us for adopting a universal macOS binary. So closing this PR.

PR #616, opened 2022-08-10
Improve uninstall with missing extension

### LINKED ISSUE #600: Improve error when uninstalling invalid extension
When uninstalling an extension, it should nicely print something like `Error: Extension 'whatever' does not exist`. However, right now it prints the following error:

```
[2022-08-05T15:25:39Z ERROR phylum] Execution failed: /home/chris/.local/share/phylum/extensions/yarn: not a directory
❗ Error: Execution failed caused by: /home/chris/.local/share/phylum/extensions/yarn: not a directory
```

This might confuse users as to what the actual issue is, thus it should be improved.

cd-work:
This changes the error when trying to uninstall an extension when it is
not already installed, to more clearly state why the process could not
be completed successfully.

Before:

```
❗ Error: /home/chris/.local/share/phylum/extensions/duplicates: not a directory
```

After:

```
❗ Error: No extension with name "duplicates" installed
```

Closes #600.


cd-work:
I'm not sure the new error message is optimal here, but it's the best I could come up with. Suggestions are always very welcome. But it should definitely be an improvement.

maxrake (approved):
The error message text makes sense to me...better wording is not coming to mind. LGTM!

PR #624, opened 2022-08-11
Transitioning away from `musl`

### LINKED ISSUE #426: Cannot use deno on x86_64-unknown-linux-musl

andreaphylum:
This PR changes the installation scripts and the `README` to refer to the new `gnu` based binaries.

Opening as draft PR to request feedback on wording and alternative strategies.

Closes #426.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@ The command line interface (CLI) allows users to submit their project package de
 
    | Target | Package |
    | --- | --- |
-   | x86_64-unknown-linux-musl | [phylum-x86_64-unknown-linux-musl.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-x86_64-unknown-linux-musl.zip) <br /> [phylum-x86_64-unknown-linux-musl.zip.minisig](https://github.com/phylum-dev/cli/releases/latest/download/phylum-x86_64-unknown-linux-musl.zip.minisig) |
-   | aarch64-unknown-linux-musl | [phylum-aarch64-unknown-linux-musl.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-unknown-linux-musl.zip) <br /> [phylum-aarch64-unknown-linux-musl.zip.minisig](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-unknown-linux-musl.zip.minisig) |
+   | x86_64-unknown-linux-gnu | [phylum-x86_64-unknown-linux-gnu.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-x86_64-unknown-linux-gnu.zip) <br /> [phylum-x86_64-unknown-linux-gnu.zip.minisig](https://github.com/phylum-dev/cli/releases/latest/download/phylum-x86_64-unknown-linux-gnu.zip.minisig) |
+   | aarch64-unknown-linux-gnu | [phylum-aarch64-unknown-linux-gnu.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-unknown-linux-gnu.zip) <br /> [phylum-aarch64-unknown-linux-gnu.zip.minisig](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-unknown-linux-gnu.zip.minisig) |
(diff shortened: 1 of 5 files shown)


cli/src/install.sh
```diff
@@ -95,6 +95,17 @@ source ${completions_dir}/phylum.bash" \
     success "Completions are enabled for bash."
 }
 
+check_glibc() {
+    # On systems with musl libc, running ldd on phylum will exit with an error.
+    # If that happens, report an explanation and exit.
+    ldd phylum >/dev/null 2>&1
+    if [ $? != 0 ]; then
```
kylewillmon, line 102: It's much more natural to put the call directly in the if

```suggestion
    if ldd phylum >/dev/null 2>&1; then
```

cli/src/install.sh
```diff
@@ -141,6 +152,7 @@ cleanup_pre_xdg() {
 
 cd "$(dirname "$0")"
 banner
+check_glibc
```
kylewillmon, line 155: We only rely on `glibc` on Linux. We need to guard this so that it doesn't run on macos installs.

README.md
```diff
@@ -80,6 +80,16 @@ extensions:
...
+### musl support
+
+Following the release of the extensions feature, as of version 3.8.0, Phylum CLI
+has a dependency on `glibc` and thus can no longer support `musl` compilation targets.
+
+This means that `phylum` won't be executable in environments such as Alpine Linux.
+
+If your use case requires a lightweight Docker base image, consider using
+[Debian slim](https://hub.docker.com/_/debian).
+
```
kylewillmon, line 92: I don't like the idea of saying that we "can no longer support" musl. Users can still build from source to get binaries that are dynamically-linked against musl. And some truly advanced user may even find a way to build a statically-linked version.

The change is the binaries that we provide. Here's a possible way to write this section:

```suggestion
### libc version

As of version 3.8.0, the provided Linux binaries of the Phylum CLI depend on
`glibc`. We no longer provide binaries that are statically compiled with `musl`
libc

This means the provided binaries won't be executable in environments such as
Alpine Linux.

If your use case requires a lightweight Docker base image, consider using
[Debian slim](https://hub.docker.com/_/debian) or [building from
source](https://docs.phylum.io/docs/alternate_install)
```

kylewillmon (changes requested):

README.md
```diff
@@ -80,6 +80,21 @@ extensions:
 
 https://github.com/phylum-dev/cli/tree/main/extensions
 
+### musl support
```
kylewillmon, line 83: I still think the heading here could be improved. This section isn't really talking about `musl support` because we can still build against musl libc as long as we dynamically link.

I think a heading like `libc version` or `libc requirement` would be more accurate.

kylewillmon (approved):
There are still other things to do. Specifically:

1. Update the release workflow to create the `*-gnu.zip` files
2. Update the code for `phylum update` to use `-gnu` instead of `-musl`

It should be simple enough to add the second item to this PR. The first item can be left for its own PR.

cli/src/install.sh
```diff
@@ -95,6 +95,24 @@ source ${completions_dir}/phylum.bash" \
...
+    if [ "${platform_str}" != "Linux" ]; then
+        return 0
+    fi
+
+    # On systems with musl libc, running ldd on phylum will exit with an error.
+    # If that happens, report an explanation and exit.
+    if ! ldd phylum >/dev/null 2>&1; then
+        error "The current operating system does not support running Phylum. Please use a system with glibc."
+        error "See: https://github.com/phylum-dev/cli#musl-support"
+        echo
```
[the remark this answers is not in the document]
cd-work: Empty line I'd assume?

kylewillmon (approved):

kylewillmon (approved):

README.md
```diff
@@ -80,6 +80,18 @@ extensions:
...
 
+### musl binaries
+
+As of version 3.8.0, the provided Linux binaries of the Phylum CLI depend on
+`glibc`. We no longer provide binaries that are statically compiled with the
+`musl` libc.
+
+This means the provided binaries won't be executable in environments such as
+Alpine Linux. If your use case requires a lightweight Docker base image,
+consider using [Debian slim](debian-slim) instead.
```
maxrake, line 91:
```suggestion
consider using [Debian slim][debian-slim] instead.
```
This is a broken link otherwise.

maxrake (commented):
The `Trigger` job in the `release` workflow can also be uncommented here...even though that trigger won't be approved for the v3.8.0 release.

PR #625, opened 2022-08-12
Handle `-h`, `--help` and `help` as first argument of `extension run`

### LINKED ISSUE #597: `phylum extension run` doesn't allow double-hyphen parameters
When running an installed extension using parameters like `phylum npm install --save typescript` it works without any problems, however when doing the same using `phylum extension run ./npm install --save typescryt` clap throws an error since it cannot recognize the `--save` flag.

To make sure people can fully test their extensions with just `phylum extension run` and provide a consistent user experience, we should support double-hyphen flags in both.

### LINKED ISSUE #605: Help output for `phylum extension run` causes error
# Overview

The `phylum extension run --help` command results in an error and no help message is displayed.

# How To Reproduce

1. With a CLI that includes extensions, run `phylum extension run --help`
2. See error

# Expected Behavior

Running `--help` on any CLI command results in a useful help message

# Additional Context

Screenshot:

image

andreaphylum:
Closes #605.

Adds a hard-coded handling of the `-h` and `--help` flags and `help` subcommand when those are supplied as first parameter to `phylum extension run`.

### DIFF
--- a/cli/src/bin/phylum.rs
+++ b/cli/src/bin/phylum.rs
@@ -147,7 +147,7 @@ async fn handle_commands() -> CommandResult {
         #[cfg(feature = "selfmanage")]
         "uninstall" => uninstall::handle_uninstall(sub_matches),
 
-        "extension" => extensions::handle_extensions(app_helper, Box::pin(api), sub_matches).await,
+        "extension" => extensions::handle_extensions(Box::pin(api), sub_matches, app_helper).await,
         extension_subcmd => {
             extensions::handle_run_extension(Box::pin(api), extension_subcmd, sub_matches).await
         },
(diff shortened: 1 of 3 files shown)


cli/src/bin/phylum.rs
```diff
@@ -147,7 +147,7 @@ async fn handle_commands() -> CommandResult {
         #[cfg(feature = "selfmanage")]
         "uninstall" => uninstall::handle_uninstall(sub_matches),
 
-        "extension" => extensions::handle_extensions(Box::pin(api), sub_matches).await,
+        "extension" => extensions::handle_extensions(app_helper, Box::pin(api), sub_matches).await,
```
cd-work, line 150: Not sure but we might want to keep API as first parameter just for consistency with everything else?

cd-work (changes requested):
Very clean and simple, nice.

cli/src/bin/phylum.rs
andreaphylum: Makes sense to me -- I put it as last argument to also preserve consistency with other methods having `sub_matches` as second argument.

cd-work (approved):

PR #635, opened 2022-08-17
Add macos arm64 runner

cd-work:


.github/workflows/release.yml
```diff
@@ -9,6 +9,7 @@ on:
   push:
     branches:
       - main
+      - eternal_ci
```
cd-work, line 12: This is solely for the purpose of demonstrating this PR works.

kylewillmon: That's what we've got the `workflow_dispatch` trigger for. You can always manually run this workflow on any branch [from the actions tab](https://github.com/phylum-dev/cli/actions/workflows/release.yml)

.github/workflows/release.yml
```diff
@@ -46,14 +47,14 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # - target: x86_64-unknown-linux-musl
-          #   os: ubuntu-latest
+          - target: x86_64-unknown-linux-musl
+            os: ubuntu-latest
           # - target: aarch64-unknown-linux-musl
           #   os: ubuntu-latest
```
kylewillmon, line 53: Any reason that we can't re-enable this one on a self-hosted runner as well?

.github/workflows/release.yml
```diff
@@ -46,14 +47,14 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # - target: x86_64-unknown-linux-musl
-          #   os: ubuntu-latest
+          - target: x86_64-unknown-linux-musl
+            os: ubuntu-latest
```
kylewillmon, line 51: This needs to be `-gnu` now, right?

cd-work: I actually didn't want to comment this out, I'll fix all that up in @andreaphylum's PR.

.github/workflows/release.yml
kylewillmon: Answered in other comment

PR #640, opened 2022-08-18
Fix incorrect help for `phylum extension run -h`

### LINKED ISSUE #639: `phylum extension run --help` shows `phylum` help, not `run` help
As stated in the title, the `--help`, `-h` and `help` output for `phylum extension run` is incorrect.

cd-work:
This resolves an issue for the `phylum extension run` subcommand where
invocations of `help`/`--help`/`-h` would cause it to print the `phylum`
help, rather than the `phylum extension run` help.

The patch also cleans up some unnecessary edge-case handling of
scenarios that are impossible to hit.

Closes #639.

### DIFF
--- a/cli/src/print.rs
+++ b/cli/src/print.rs
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 
 use ansi_term::Color::{Blue, Cyan};
+use anyhow::{anyhow, Result};
 use clap::Command;
 use prettytable::format;
 use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
@@ -37,16 +38,18 @@ pub fn print_update_message() {
     eprintln!("{:-^50}\n\n", "");
 }
 
-pub fn print_sc_help(mut app: &mut Command, subcommands: &[&str]) {
+pub fn print_sc_help(mut app: &mut Command, subcommands: &[&str]) -> Result<()> {
     for subcommand in subcommands {
-        match app.get_subcommands_mut().find(|sc| &sc.get_name() == subcommand) {
+        match app.find_subcommand_mut(*subcommand) {
             Some(subcommand) => app = subcommand,
             // Subcommand doesn't exist; don't print anything.
-            None => return,
+            None => return Err(anyhow!("Subcommand '{subcommand}' does not exist")),
         }
     }
 
-    let _ = app.print_help();
+    app.print_help()?;
+
+    Ok(())
 }
 
 /// Limit a string to a specific length, using an ellipsis to indicate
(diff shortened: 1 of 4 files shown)


cli/src/print.rs
```diff
@@ -37,14 +37,16 @@ pub fn print_update_message() {
...
-    for sc in app.get_subcommands_mut() {
-        if sc.get_name() == subcommand {
-            let _ = sc.print_help();
-            break;
+pub fn print_sc_help(mut app: &mut Command, subcommands: &[&str]) {
+    for subcommand in subcommands {
+        match app.get_subcommands_mut().find(|sc| &sc.get_name() == subcommand) {
+            Some(subcommand) => app = subcommand,
+            // Subcommand doesn't exist; don't print anything.
+            None => return,
```
cd-work, line 45: Currently this relies on us always passing the right subcommands in, otherwise it will just silently exit the program. I think that's fine, but maybe we should print an error or crash just in case?

kylewillmon: I think some sort of panic would be reasonable here. If we ever hit that case, it definitely means a bug in the code.

cli/src/print.rs
```diff
@@ -37,14 +37,16 @@ pub fn print_update_message() {
...
 }
 
-pub fn print_sc_help(app: &mut Command, subcommand: &str) {
-    for sc in app.get_subcommands_mut() {
-        if sc.get_name() == subcommand {
-            let _ = sc.print_help();
-            break;
+pub fn print_sc_help(mut app: &mut Command, subcommands: &[&str]) {
+    for subcommand in subcommands {
+        match app.get_subcommands_mut().find(|sc| &sc.get_name() == subcommand) {
```
kylewillmon, line 42: Is there a reason not to use the [`find_subcommand_mut`](https://docs.rs/clap/latest/clap/struct.App.html#method.find_subcommand_mut) method here? It is shorter and it also works with aliases.

```suggestion
        match app.find_subcommand_mut(subcommand) {
```

cli/src/print.rs
```diff
@@ -37,14 +37,16 @@ pub fn print_update_message() {
     eprintln!("{:-^50}\n\n", "");
 }
 
-pub fn print_sc_help(app: &mut Command, subcommand: &str) {
-    for sc in app.get_subcommands_mut() {
-        if sc.get_name() == subcommand {
-            let _ = sc.print_help();
-            break;
+pub fn print_sc_help(mut app: &mut Command, subcommands: &[&str]) {
```
kylewillmon, line 40: TIL what `sc` in this function name stands for.... I've breezed past this function many times and never realized that `sc == subcommand`....

kylewillmon (approved):

cli/src/print.rs
cd-work: I just used the same method that was used before without looking at the available API, find sounds good to me.

kylewillmon (approved):

Issue #645, opened 2022-08-22
Extensions: Add `getProjects()` function


kylewillmon:
# Overview
A `getProjects()` function would be useful in the extension API.

# Acceptance Criteria
- [ ] Takes an option group name
- [ ] Returns a list of projects including their name and ID

Issue #644, opened 2022-08-22
Extensions: Add a `getCurrentProject()` function


kylewillmon:
Many extensions operate within the current project. A `getCurrentProject()` function would help these extensions.

## Acceptance Criteria
- [ ] Function returns project name and group name details from the current `.phylum_project` file
- [ ] Function returns `null` if no `.phylum_project` file is found.

kylewillmon:
I am immediately questioning whether or not this is actually necessary....

The only thing that an extension can do currently with a project name and group name is run `analyze()`. But if an extension wants to run the analyze on the current project, they simply need to pass in `null` for the project name.

Am I missing a use case for this function?

cd-work:
This would be the only offline way to check if you've got a project linked at all for example, you wouldn't necessarily want to run into an API failure for that.

I could also see this being helpful to some projects that have very specific extensions they want to use for multiple of their own projects, using the phylum project name might be a good way to identify the repository they're in at the moment.

Generally this is pretty basic information that I can definitely see being useful to *some* people in the future. I'd much rather provide it here than seeing people start parsing the file themselves.

This will become especially useful if we decide to add the lockfile path to the project settings at some point.

PR #646, opened 2022-08-22
Remove getProjectDetails() function from extension API

### LINKED ISSUE #642: `PhylumApi.getProjectDetails` does not use groups
# Overview
Calling `getProjectDetails` in the extension API does not allow you to pass along the group.

# How To Reproduce
### `group-example/main.ts`
```
cat group-example/main.ts
import { PhylumApi } from 'phylum';

const project = await PhylumApi.getProjectDetails("foobarabc");
console.log(project);
```

### Create **group** project and Run Extension
```
❯  phs project create foobarabc --group extensionstest
✅ Successfully created new project, foobarabc

❯  phs extension run ./group-example
Extension error: Uncaught (in promise) Error: HTTP request error (404 Not Found):

could not find any matching resource for the given id
    at async Function.getProjectDetails (deno:phylum:127:16)
    at async file:///Users/louis/Development/PHYLUM/localdev-test/gradle-work/group-example/main.ts:2:17
```

# Expected Behavior
We should be able to query the API for group projects.

# Additional Context
The project appears to 404 due to the fact that we can't specify the group the project is associated with. 

Since projects can be created with groups, we should allow an optional group parameter to `PhylumApi.getProjectDetails`

kylewillmon:
None of our extensions use the `getProjectDetails()` function and it is based on a deprecated endpoint. This patch removes it from the extension API as a simple way to fix #642. We could deprecate it instead, but I am doubtful that it provides much, if any, value and I do not want the initial release of extensions to include a deprecated function.

New functions will be added to the extension API in #644 and #645 to handle any use cases that may exist involving "project details"

### DIFF
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -14,7 +14,6 @@ use phylum_types::types::job::JobStatusResponse;
 use phylum_types::types::package::{
     Package, PackageDescriptor, PackageStatusExtended, PackageType,
 };
-use phylum_types::types::project::ProjectDetailsResponse;
 use serde::{Deserialize, Serialize};
 use tokio::fs;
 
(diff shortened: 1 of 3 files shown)


cd-work (approved):
Should be good once CI is done.

cd-work (approved):

PR #654, opened 2022-08-24
Move error logging to API module

### LINKED ISSUE #653: Improve error message for non-existent project
# Overview
When attempting to submit an analysis job with a project that doesn't exist, the user receives a 404 error that is unlikely to clarify the issue. This case is more common now as a user can delete a project, but not delete the `.phylum_project` file and run the `analyze` command.

# How To Reproduce
1. Create a project
2. Submit a file for analysis
3. See success
4. Delete the project, but do not delete the `.phylum_project` file
5. Run the analysis command again
6. See the 404 error 

# Expected Behavior
A more helpful error message so the user can troubleshoot their situation - ideally, clearly stating that the project does not exist.

# Additional Context
image

Related API change: https://github.com/phylum-dev/api/issues/457
Closing this issue should be the last piece of this Story: https://github.com/phylum-dev/roadmap/issues/35

cd-work:
Instead of handling the addition of higher-level detail to error
messages in the locations where the API is used, this moves it to the
API instead. This way all locations calling these API functions will
implicitly get the refined error messages without requiring additional
handling at a higher level.

This change was inspired by the existing `get_project_id` function,
which already provides a friendlier error message for missing projects.
So this patch also helps with consistency of error handling

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -1,6 +1,6 @@
 use std::time::Duration;
 
-use anyhow::{anyhow, Context};
+use anyhow::{anyhow, Context, Error as AnyhowError};
 use phylum_types::types::auth::TokenResponse;
 use phylum_types::types::common::{JobId, ProjectId};
 use phylum_types::types::group::{CreateGroupRequest, CreateGroupResponse, ListUserGroupsResponse};
@@ -226,13 +226,21 @@ impl PhylumApi {
 
     /// Create a new project
     pub async fn create_project(&mut self, name: &str, group: Option<&str>) -> Result<ProjectId> {
-        let response: CreateProjectResponse = self
+        let result: Result<CreateProjectResponse> = self
             .post(
                 endpoints::post_create_project(&self.config.connection.uri)?,
                 CreateProjectRequest { name: name.to_owned(), group_name: group.map(String::from) },
             )
-            .await?;
-        Ok(response.id)
+            .await;
+
+        result
+            .map_err(|err| match err {
+                PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. }) => {
+                    AnyhowError::new(err).context(format!("Project '{name}' already exists")).into()
+                },
+                err => err,
+            })
+            .map(|response| response.id)
     }
 
     /// Delete a project
@@ -352,7 +360,15 @@ impl PhylumApi {
     /// Get all groups the user is part of.
     pub async fn create_group(&self, group_name: &str) -> Result<CreateGroupResponse> {
         let group = CreateGroupRequest { group_name: group_name.into() };
-        self.post(endpoints::group_create(&self.config.connection.uri)?, group).await
+        let uri = endpoints::group_create(&self.config.connection.uri)?;
+        let result = self.post(uri, group).await;
+
+        result.map_err(|err| match err {
+            PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. }) => {
+                AnyhowError::new(err).context(format!("Group '{group_name}' already exists")).into()
+            },
+            err => err,
+        })
     }
 
     pub fn config(&self) -> &Config {
--- a/cli/tests/common.rs
+++ b/cli/tests/common.rs
@@ -6,12 +6,11 @@ use std::{env, fs};
 
 pub use assert_cmd::assert::Assert;
 pub use assert_cmd::Command;
-use phylum_cli::api::{PhylumApi, PhylumApiError, ResponseError};
+use phylum_cli::api::{PhylumApi, PhylumApiError};
 use phylum_cli::commands::extensions::permissions::Permissions;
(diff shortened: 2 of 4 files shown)


cli/src/api/mod.rs
```diff
@@ -296,9 +304,23 @@ impl PhylumApi {
...
+        response
+            .map_err(|err| match err {
+                PhylumApiError::Response(ResponseError { code: StatusCode::NOT_FOUND, .. }) => {
+                    AnyhowError::new(err)
+                        .context(format!(
+                            "Project with ID {project} does not exist. This usually means Phylum \
+                             is still linked to a project that has been deleted."
+                        ))
+                        .into()
+                },
```
kylewillmon, line 320: This may be obsoleted by phylum-dev/api#478

cd-work: Looks like it would be, yeah. Seems like the requests were submitted to API and CLI simultaneously somewhat.

I'm fine with merging this and then just ripping it out again, but I'd personally prefer we just scrap this PR and move on without it, so we don't have useless code churn.

I'd still like to move the other checks to the API since I think that's a better place for them.

cd-work:
I've changed this PR to only move the existing error handling over, we can easily apply the other changes again should the API's change be insufficient.

I *probably* should have opened a new PR for this, but I think it should be fine to just merge it here.

cli/tests/common.rs
```diff
@@ -228,10 +227,9 @@ pub async fn create_project() -> &'static str {
     // Attempt to create the project, ignoring conflicts.
     let mut api = PhylumApi::new(config, None).await.unwrap();
     match api.create_project(PROJECT_NAME, None).await {
-        Ok(_) | Err(PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. })) => {
-        },
+        Ok(_) | Err(PhylumApiError::Other(_)) => (),
         err @ Err(_) => {
-            err.unwrap();
+            let _ = err.unwrap();
         },
```
kylewillmon, line 233: The code change here points out a big issue with this approach. The caller of `api.create_project()` can no longer distinguish a conflict from other errors. (Not easily at least. It's technically possible with a `downcast_ref()` and a second `match`).

There are a few ways that we could change this to allow a caller to continue to detect these errors:
1. We can add new variants to `PhylumApiError` for `ProjectAlreadyExists` and `GroupAlreadyExists` errors.
2. Instead of using anyhow context to improve the error message, we can just overwrite the `details` field in the `ResponseError` and continue to use the `PhylumApiError::Response` variant.

Of course, we could also accept that callers shouldn't be able to detect these cases and that the `PhylumApiError` type shouldn't exist (because we should be using `anyhow::Error` instead). But I'm not sure we're ready to take that step (or if that actually make sense).

I am leaning towards option 2 above since it is the simplest way to improve the error message and maintains the usage of `PhylumApiError::Response`. As an added bonus, it's also the easiest change to rip out later if the error message is improved on the API side.

cli/src/api/mod.rs
```diff
@@ -352,7 +360,15 @@ impl PhylumApi {
     /// Get all groups the user is part of.
     pub async fn create_group(&self, group_name: &str) -> Result<CreateGroupResponse> {
         let group = CreateGroupRequest { group_name: group_name.into() };
-        self.post(endpoints::group_create(&self.config.connection.uri)?, group).await
+        let uri = endpoints::group_create(&self.config.connection.uri)?;
+        let result = self.post(uri, group).await;
+
+        result.map_err(|err| match err {
+            PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. }) => {
+                AnyhowError::new(err).context(format!("Group '{group_name}' already exists")).into()
```
kylewillmon, line 368: `AnyhowError::new()` looks odd to me... Is that different than `anyhow!(err)`?

kylewillmon (changes requested):

cli/tests/common.rs
cd-work: I don't like overwriting details tbh, since that means we actually lose information.

Considering we always use anyhow for everything anyway, I don't think matching on errors is a big priority in the CLI anyway? I'm not sure I'd call that a good thing, but it's pretty consistent across our entire codebase anyway.

If we decide this shouldn't be a thing, we should probably nuke the anyhow enum variant completely?

kylewillmon: > I don't like overwriting details tbh, since that means we actually lose information.

Yes and no. The reason the code exists is to replace a poor error message with a better one. So while we do technically lose information, it's just the poor error message that is lost, which is expected.

cd-work: It's not poor in general though, it's just poor at the toplevel. As a detail it still might be interesting.

cd-work (commented):
There's probably improvements that could be made, but I've never been entirely convinced by this patch. So I'm just going to close it in favor of maybe trying to rework it from scratch in the future.

Issue #665, opened 2022-09-02
`phylum update` does not respect `--no-check-certificate`


kylewillmon:
# Overview
Running `phylum --no-check-certificate update` can still result in certificate errors.

# How To Reproduce
Set up `mitmproxy` or some other TLS intercepting proxy and run a command like the following:
```
❯ HTTPS_PROXY=http://localhost:8080 phylum --no-check-certificate update                    
[2022-09-02T15:27:32Z WARN  phylum] Ignoring TLS server certificate verification per user request.
[2022-09-02T15:27:35Z WARN  rustls::conn] Sending fatal alert BadCertificate
❗ Error: Failed to get the latest version

Caused by:
    0: error sending request for url (https://api.github.com/repos/phylum-dev/cli/releases/latest): error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
    1: error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
    2: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
```

# Expected Behavior
The `--no-check-certificate` flag should cause the invalid certificate to be accepted and operation should continue as normal.

PR #666, opened 2022-09-02
Send an appropriate User-Agent header

kylewillmon:
This patch adds a user-agent header to our HTTP requests. I chose to use the `git-version` (i.e., the same thing output by `phylum version`) for this in order to properly identify RC versions and local builds.

## Before
before

## After
after

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -21,6 +21,7 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error as ThisError;
 
 use self::endpoints::BaseUriError;
+use crate::app::USER_AGENT;
 use crate::auth::{
     fetch_oidc_server_settings, handle_auth_flow, handle_refresh_tokens, AuthAction, UserInfo,
 };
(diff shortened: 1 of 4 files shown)


cli/src/update/unix.rs
```diff
@@ -57,7 +58,7 @@ impl Default for ApplicationUpdater {
 
 /// Generic function for fetching data via HTTP GET.
 async fn http_get(url: &str) -> anyhow::Result<reqwest::Response> {
-    let client = Client::builder().user_agent("phylum-cli").build()?;
+    let client = Client::builder().user_agent(USER_AGENT.as_str()).build()?;
```
kylewillmon, line 61: I discovered #665 when modifying this line. We should fix that in the future, but I didn't want to crowd this PR.

andreaphylum (approved):

cli/src/api/mod.rs
```diff
@@ -171,9 +171,9 @@ impl PhylumApi {
             HeaderValue::from_str(&format!("Bearer {}", tokens.access_token)).unwrap(),
         );
         headers.insert("Accept", HeaderValue::from_str("application/json").unwrap());
-        headers.insert("version", HeaderValue::from_str(version).unwrap());
```
kylewillmon, line 174: I am fairly confident that this header is not being used or recorded anywhere and I think the `User-Agent` is a much more natural place to include the CLI version.

Cc @DanielJoyce in case I am incorrect on that assumption.

louislang: The version was added a while back to help debug when a customer was using a _super old_ version. I don't think it's as much of a worry now, but it did end up being useful at one point. If it's in the UA string, I think that's good too.

PR #671, opened 2022-09-08
Allow global install by specifying the `bin_dir`

### LINKED ISSUE #670: Support global installation in our installer
The install script currently only supports installing to the user's home directory (specifically, `~/.local/bin`). Some environments would benefit from being able to install to a globally-accessible directory (probably `/usr/local/bin`).

## Additional context
The need for this has come up as a part of phylum-dev/phylum-ci#118. The docker image created by the phylum-ci project installs the CLI into the home directory of the `root` user, which prevents the docker image from being run as a non-privileged user.

kylewillmon:
Fix #670 by allowing a user to specify `--bin-dir` and `--no-data-files` when running `install.sh`.

For a global install, the user would run something like this:

```
./install.sh --bin-dir /usr/local/bin --no-data-files
```

I've still got a few concerns with this...

1. `phylum update` gets weird...
2. With these options, the install script is essentially reduced to this one-liner:
   ```
   install -m 0755 phylum /usr/local/bin/phylum
   ```

Opened as a draft PR to elicit feedback

### DIFF
--- a/cli/src/install.sh
+++ b/cli/src/install.sh
@@ -23,6 +23,19 @@ banner() {
     printf "\n    %bphylum-cli%b installer\n\n" "${GREEN}" "${NC}"
 }
 
+usage() {
+    cat 1>&2 <<EOF
+install.sh [options]
+
+Install the phylum CLI.
+
+Options
+    -b <BINDIR>, --bin-dir <BINDIR>     Specify the install directory
+    --no-data-files                     Do not install data files (e.g., completion files)
+    -h, --help                          Show this help message
+EOF
+}
+
 # Get the platform name.
 get_platform() {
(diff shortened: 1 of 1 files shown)


cd-work:
Seems somewhat strange to me to be honest. Especially the combination of `--bin-dir` and `--no-data-files` which introduce two new flags that are basically just there to fix an issue with a single specific usecase.

This all just seems like a very hacky workaround trying to reinvent a distribution platform for a target that in theory already has a functional non-problematic distribution process built-in.

kylewillmon:
Agreed. Some day we will create a `.deb` package for the CLI. Until then, installing for all users is as simple as:

```
install -m 0755 phylum /usr/local/bin/phylum
```

PR #673, opened 2022-09-09
Add sandbox to extensions API

### LINKED ISSUE #672: Allow extensions to run external commands in a sandbox
When an extension runs an external command (e.g., `npm`), the extension may want to limit the capabilities of that command (e.g., disallowing network access, etc) for additional security.

## Acceptance Criteria
- [ ] A CLI extension can restrict network access of an external command that it runs
- [ ] A CLI extension can restrict disk access for any external commands that are run
- [ ] (Optional) macOS support

## Out of Scope

 - Windows support
 - Sandboxing outside of Filesystem/Network (system calls, IPC, etc.)

cd-work:
This adds a new `runSandboxed` function to Phylum's extension API. This allows running subprocesses with restricted filesystem or network access.

This also wraps the extension itself in another sandboxing layer around Deno, ensuring that the globally advertised permissions are always accurate.

The extension's sandboxing is derived from the `PhylumExt.toml`'s manifest permissions, while `runSandboxed` accepts the sandboxing restrictions as parameters.

Closes #672.

## TODO

- [x] Update extensions other than `yarn`
- [x] Tests
- [x] Documentation (?)

### DIFF
--- a/cli/src/commands/extensions/permissions.rs
+++ b/cli/src/commands/extensions/permissions.rs
@@ -195,6 +195,8 @@ pub fn default_sandbox() -> SandboxResult<Birdcage> {
     add_exception(&mut birdcage, Exception::ExecuteAndRead("/lib32".into()))?;
     add_exception(&mut birdcage, Exception::ExecuteAndRead("/libx32".into()))?;
     add_exception(&mut birdcage, Exception::ExecuteAndRead("/lib64".into()))?;
+    add_exception(&mut birdcage, Exception::ExecuteAndRead("/opt/homebrew".into()))?;
+    add_exception(&mut birdcage, Exception::ExecuteAndRead("/usr/local/lib".into()))?;
 
     // Allow `env` exec to resolve binary paths.
     add_exception(&mut birdcage, Exception::ExecuteAndRead("/usr/bin/env".into()))?;
--- a/cli/src/commands/extensions/permissions.rs
+++ b/cli/src/commands/extensions/permissions.rs
@@ -209,7 +209,7 @@ pub fn default_sandbox() -> SandboxResult<Birdcage> {
     //
     // This is required since many package manager's build scripts will use various
     // executables in their build scripts.
-    for path in env::var("PATH").iter().map(|path| path.split(':')).flatten() {
+    for path in env::var("PATH").iter().flat_map(|path| path.split(':')) {
         add_exception(&mut birdcage, Exception::ExecuteAndRead(path.into()))?;
     }
 
@@ -227,7 +227,7 @@ pub fn add_exception(birdcage: &mut Birdcage, exception: Exception) -> SandboxRe
         Ok(_) => Ok(()),
         // Ignore invalid path errors.
         Err(SandboxError::InvalidPath(_)) => Ok(()),
-        Err(err) => Err(err.into()),
+        Err(err) => Err(err),
     }
 }
 
--- a/cli/src/commands/extensions/permissions.rs
+++ b/cli/src/commands/extensions/permissions.rs
@@ -1,14 +1,21 @@
+use std::borrow::Cow;
 use std::path::PathBuf;
+use std::{env, fs};
 
 use anyhow::{anyhow, Result};
+use birdcage::error::Error as SandboxError;
+use birdcage::{Birdcage, Exception, Sandbox};
 use deno_runtime::permissions::PermissionsOptions;
 use serde::de::Error as _;
 use serde::{Deserialize, Deserializer, Serialize};
+
+use crate::dirs;
 
 /// Resource permissions.
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
 #[serde(untagged)]
 pub enum Permission {
+    #[serde(deserialize_with = "deserialize_permission_paths")]
     List(Vec<String>),
     Boolean(bool),
 }
@@ -33,6 +40,15 @@ impl Permission {
         }
     }
 
+    /// Get Birdcage sandbox exception resource paths.
+    pub fn sandbox_paths(&self) -> Cow<'_, Vec<String>> {
+        match self {
+            Permission::List(paths) => Cow::Borrowed(paths),
+            Permission::Boolean(true) => Cow::Owned(vec!["/".into()]),
+            Permission::Boolean(false) => Cow::Owned(Vec::new()),
+        }
+    }
+
     /// Check if access to resource is permitted.
     pub fn validate(&self, resource: &String, resource_type: &str) -> Result<()> {
         if self.get().map_or(false, |allowed| allowed.contains(resource)) {
@@ -43,7 +59,26 @@ impl Permission {
     }
 }
 
-#[derive(Clone, Default, Deserialize, Debug, Serialize)]
+/// Deserializer for automatically resolving `~/` path prefix.
+pub fn deserialize_permission_paths<'de, D>(deserializer: D) -> Result<Vec<String>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    // Ensure field is a valid string.
+    let mut paths = Vec::<String>::deserialize(deserializer)?;
+
+    // Resolve `~/` home prefix.
+    let home = dirs::home_dir().map_err(D::Error::custom)?;
+    for path in &mut paths {
+        if let Some(suffix) = path.strip_prefix("~/") {
+            *path = home.join(suffix).display().to_string();
+        }
+    }
+
+    Ok(paths)
+}
+
+#[derive(Serialize, Deserialize, Clone, Default, Debug)]
 pub struct Permissions {
     #[serde(default)]
     pub read: Permission,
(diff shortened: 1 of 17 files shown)


cd-work:
@kylewillmon / @andreaphylum I'd like some feedback on the design/approach if possible before I update the remaining extensions and start writing tests and docs. I think the core of the functionality should already be done and I've hopefully documented the API sufficiently.

extensions/yarn/main.ts
```diff
@@ -85,9 +92,57 @@ if (Deno.args.length >= 1
...
+        },
+    });
+
+    // Failure here could indicate vulnerabilities; report to the user.
+    if (!output.success) {
+        console.log(`[${red("phylum")}] Sandboxed build failed.`);
+        console.log(`[${red("phylum")}]`);
+
+        // Check for SIGSYS, since this confirms sandbox violation.
+        if (output.signal === 31) {
```
kylewillmon, line 129: Checking for a specific signal at this high level feels a bit weird. Maybe it would make more sense to have a boolean field that `runSandbox` sets when a sanbox violation is detected (something like `output.violation`)?

Otherwise we may have problems when we abstract this to different platforms, since (for example) macos may use a different signal.

kylewillmon (changes requested):

cli/src/commands/extensions/permissions.rs
```diff
@@ -117,6 +174,89 @@ impl From<&Permissions> for PermissionsOptions {
...
+    add_exception(&mut birdcage, Exception::Read("/etc/resolv.conf".into()))?;
+
+    // Allow reading SSL certificates.
+    add_exception(&mut birdcage, Exception::Read("/etc/ca-certificates".into()))?;
+    add_exception(&mut birdcage, Exception::Read("/etc/ssl".into()))?;
+
+    // TODO: I really don't like this
+    if let Ok(config_dir) = dirs::config_dir() {
+        add_exception(&mut birdcage, Exception::Write(config_dir.join("phylum/settings.yaml")))?;
+    }
```
cd-work, line 212: Maybe we can resolve the config writing issues before this PR gets merged?

extensions/yarn/main.ts
```diff
@@ -85,9 +92,57 @@ if (Deno.args.length >= 1
...
+            console.log(`[${red("phylum")}] Sandbox violation confirmed during install.`);
+            console.log(`[${red("phylum")}] Do NOT retry installation.`);
+            console.log(`[${red("phylum")}]`);
+            console.log(`[${red("phylum")}] Please submit your lockfile to Phylum.`);
+        } else {
+            console.log(`[${red("phylum")}] This could mean one of your packages attempted to access a restricted resource.`);
+            console.log(`[${red("phylum")}] Do not retry installation without Phylum's extension.`);
+            console.log(`[${red("phylum")}]`);
+            console.log(`[${red("phylum")}] Please submit your lockfile to Phylum should this error persist.`);
+        }
```
cd-work, line 139: Feedback on error messages would be appreciated, we'll probably want to share this wording across our extensions and I'm not sure this is ideal already.

extensions/yarn/main.ts
cd-work: I don't know what macOS does at all, that's kinda why I'm exposing just the process level API here.

As far as I know, there's no way to reliably detect a sandbox violation so I don't think it would be a good idea to provide an API to read this. Since that API is going to be unreliable and thus misleading.

Reading the `SIGSYS` signal is a heuristic at best.

cd-work:
Sharing the Yarn extension output here, for people that don't want to run it themselves:

tmp

extensions/yarn/main.ts
kylewillmon: It may be just a heuristic, but it's important enough that we are printing different messages for this extension (and likely many others). I'm just worried that these heuristic checks might get out of hand. But I guess we can make that determination after we figure out macos.

cd-work: My concern is just that when we put an `output.violation` field, people are going to assume there's been a violation when it's `true` and there hasn't been one when it's `false`.

But we can't say that for sure in either direction.

That's why we aren't actually doing anything different. We're just adapting our logging to provide the best possible feedback.

kylewillmon: I share the same concern. But it seems like a naming problem more than anything else...

We could name it `output.violation_likely`... Or we could make it something other than a bool. Like a string that gets set to either `"likely"` or `"¯\_(ツ)_/¯"` depending on the heuristic.

cd-work: I like the shrug. :smile:

extensions/yarn/main.ts
kylewillmon: The error message is pretty good as is. I've wondered if we should say anything about the possible error message that `yarn` may have output above this message, but I'm really not sure what to say there.

We could probably include a link with the "Please submit" wording. Something like this:

```
Please consider reporting this to Phylum using the following link:
https://github.com/phylum-dev/cli/issues/new?template=sandbox_violation.md
```

We would need to add `.github/ISSUE_TEMPLATE/sandbox_violation.md` for that link to work, but we could add prompts in that template for details such as full command output and the lockfile.

kylewillmon: Cc @furi0us333 since he may have opinions on the wording.

extensions/find-permissions/main.ts
```diff
@@ -0,0 +1,207 @@
...
+{
+    console.log('find-permissions');
+    console.log('CLI extension to help find required sandboxing permissions.');
+    console.log();
+    console.log('USAGE:');
+    console.log('    phylum find-permissions [OPTIONS] --bin <PATH>');
+    console.log();
+    console.log('OPTIONS:');
+    console.log('    --read               Check if tested paths need to be readadle or executable');
+    console.log('    --write              Check if tested paths need to be writeable');
```
andreaphylum, line 17:
```suggestion
    console.log('    --read               Check if tested paths need to be readable or executable');
    console.log('    --write              Check if tested paths need to be writable');
```

extensions/npm/main.ts
```diff
@@ -2,59 +2,125 @@ import { red, green, yellow } from 'https://deno.land/std@0.150.0/fmt/colors.ts'
 import { PhylumApi } from 'phylum';
 
 class FileBackup {
-  readonly fileName: string
-  readonly fileContent: string | null
+  readonly fileName: string;
```
andreaphylum, line 5: No need to change this for now, but for the future we should consider applying consistent automated styling to extensions. Generally I recommend https://standardjs.com/ which _doesn't_ have semicolons.

andreaphylum (changes requested):

extensions/npm/main.ts
cd-work: Not having semicolons in JavaScript seems extremely unconventional to me.

andreaphylum: It's been pretty commonplace for quite a few years now. `npm` itself doesn't use them, I have not been using them for the largest part of my career when writing JS, most JS developers I know also do not use them. Then again the point is not whether they're good or bad, but having a consistent and easy to enforce system, and `standard` has been pretty great at doing that. But we can keep this for another PR.

docs/extensions/extension_sandboxing.md
```diff
@@ -0,0 +1,119 @@
...
+---
+
+Phylum's CLI extensions allow developers to impose additional restrictions when
+running third party applications. This can protect the system from damage when
+these applications contain vulnerabilities or execute untrusted code.
+
+## Example
+
+The following code provides an example on how you could sandbox `cat` to only
+allow access to files in the current working directory or below it:
```
andreaphylum, line 14:
```suggestion
allow access to files in the current working directory or its descendants:
```
(optional)

docs/extensions/extension_sandboxing.md
```diff
@@ -0,0 +1,119 @@
...
+        net: false,
+    },
+});
+
+// Propagate `cat`'s exit code.
+Deno.exit(output.code);
+```
+
+When running this against a file in your local directory, it will print its
+content, otherwise , you'll see `cat` printing the following error:
```
andreaphylum, line 44:
```suggestion
content, otherwise, you'll see `cat` printing the following error:
```

docs/extensions/extension_sandboxing.md
```diff
@@ -0,0 +1,119 @@
...
+allow your application to run without any errors. To simplify this a bit you can
+use the [`find-permissions`] extension.
+
+[`find-permissions`]: https://github.com/phylum-dev/cli/tree/main/extensions/find-permissions
+
+This extension will run a script against each path in your filesystem
+recursively to validate what the most granular necessary exceptions are. Once
+completed, it will output all necessary paths.
+
+Most invokations of this extension will probably look something like this:
```
andreaphylum, line 100:
```suggestion
Most invocations of this extension will probably look something like this:
```

andreaphylum (changes requested):

docs/extensions/extension_sandboxing.md
cd-work: I think "descendants" might be more difficult to understand for some users than "below it". Especially because it can be interpreted as some kind of derived directory rather than a hierarchical structure.

I'm open to other suggestions, but I don't think descendants is the way to go.

andreaphylum (dismissed):

kylewillmon (approved):

PR #676, opened 2022-09-13
Re-enable compression

### LINKED ISSUE #614: Reenable release zip compression
Release 3.7.2 introduced a regression which prevents it from updating to a compressed zip archive. To work around this zip archive compression was temporarily disabled in https://github.com/phylum-dev/cli/pull/613.

Once we're confident that most users will have migrated away from 3.7.2, we should reenable compression.

kylewillmon:
Delete a few characters to save the trees and close #614


maxrake (approved):
Was just about to ask when we would do this.

cd-work (approved):

PR #677, opened 2022-09-13
Remove XDG migration code

kylewillmon:
The migration to XDG directories was months ago, so it should now be safe to remove the migration code.


maxrake (approved):
A search of code and docs for additional instances turned up nothing new. This is good timing as the `phylum-ci` project is also removing support for pre-XDG-compliant versions of the CLI. LGTM.

PR #678, opened 2022-09-14
Include and install default extensions

### LINKED ISSUE #641: Include default extensions with each release
The extensions in the `./extensions` folder should be shipped with the CLI.

## Acceptance Criteria
- [ ] When a user installs the CLI default extensions should automatically be installed
- [ ] When the CLI is updated the extensions should also be updated

## Notes
Because these extensions are maintained by Phylum and automatically installed, we do not need to prompt the user to accept permissions when doing these installs. Also, we can simply overwrite the extensions during upgrade as we do not expect users to modify these extensions.

We should consider adding the ability to skip this in the installer for some environments. Perhaps a command-line argument or environment variable to disable the install of extensions. Ideally, this choice would persist across upgrades.

kylewillmon:
Install the `npm`, `yarn`, and `poetry` extensions by default. Closes #641

### DIFF
--- a/cli/src/install.sh
+++ b/cli/src/install.sh
@@ -9,7 +9,7 @@ set -eu
 
 data_dir="${XDG_DATA_HOME:-${HOME}/.local/share}/phylum"
 completions_dir="${data_dir}/completions"
-extensions_dir="${data_dir}/extensions"
+extensions_dir="${data_dir}/official-extensions"
 bin_dir="${HOME}/.local/bin"
 
 error() {
@@ -146,7 +146,7 @@ copy_files() {
         # Install the extension
         cp -Rf "extensions/${ext_name}" "${dest}"
     done
-    success "Copied extensions to ${data_dir}/extensions"
+    success "Copied extensions to ${extensions_dir}"
 }
 
 cd "$(dirname "$0")"
--- a/cli/src/install.sh
+++ b/cli/src/install.sh
@@ -9,6 +9,7 @@ set -eu
 
 data_dir="${XDG_DATA_HOME:-${HOME}/.local/share}/phylum"
 completions_dir="${data_dir}/completions"
+extensions_dir="${data_dir}/extensions"
 bin_dir="${HOME}/.local/bin"
 
 error() {
@@ -131,6 +132,21 @@ copy_files() {
     (umask 077; mkdir -p "${data_dir}")
     cp -a "completions" "${data_dir}/"
     success "Copied completions to ${completions_dir}"
+
+    # Copy extensions over
+    (umask 077; mkdir -p "${extensions_dir}")
+    for ext in extensions/*/PhylumExt.toml
+    do
+        ext_name=$(basename "$(dirname "${ext}")")
+        dest="${extensions_dir}/${ext_name}"
+
+        # Delete the old version if it exists
+        [ -e "${dest}" ] && rm -rf "${dest}"
+
+        # Install the extension
+        cp -Rf "extensions/${ext_name}" "${dest}"
+    done
+    success "Copied extensions to ${data_dir}/extensions"
 }
 
 cd "$(dirname "$0")"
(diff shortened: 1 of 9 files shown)


cli/src/install.sh
```diff
@@ -131,6 +132,21 @@ copy_files() {
...
+
+    # Copy extensions over
+    (umask 077; mkdir -p "${extensions_dir}")
+    for ext in extensions/*/PhylumExt.toml
+    do
+        ext_name=$(basename "$(dirname "${ext}")")
+        dest="${extensions_dir}/${ext_name}"
+
+        # Delete the old version if it exists
+        [ -e "${dest}" ] && rm -rf "${dest}"
```
cd-work, line 144: This is the reason why I mentioned in the original issue that we might want to hold until we have an idea of what our `upgrade` command should look like.

Do we just want it to delete && install? IIRC one of the main reasons why @andreaphylum didn't want to have install implicitly overwrite the existing version because users might have made local changes. This would apply here too. We'd just nuke anything the user might have changed.

cd-work (changes requested):

maxrake (changes requested):
A review of the code turned up no issues.

The changes still need to be tested with a local install of the release artifacts from this branch. A [manual run was created](https://github.com/phylum-dev/cli/actions/runs/3055326038) to do just that. This comment will be updated to state whether the extensions function as advertised.

The existing documentation should be updated to make it more clear that these official extensions are included during install and available for use right away. The [Extension Overview page](https://github.com/phylum-dev/cli/blob/main/docs/extensions/extension_overview.md) and the [Official Extensions section](https://github.com/phylum-dev/cli/blob/main/docs/extensions/extension_overview.md?plain=1#L33-L38) in particular. There may be other pages (or new ones) that would make the existence and use of these extensions easier for users.

cli/src/install.sh
kylewillmon: Users shouldn't be making changes directly in these folders. But it is definitely true that someone may have installed a modified version of one of our extensions or created their own extension of the same name.

Maybe we should install our extensions into a separate folder?

maxrake:
A [new release build is in process](https://github.com/phylum-dev/cli/actions/runs/3056129180) to try out the recent changes. However, a few observations are already available:

* The READMEs in each "default" extension need to be updated
  * They currently specify the old way of installing...which is no longer necessary
* There are test files copied
  * These may or may not be meant for installation
  * At the very least, the `authors` section of `poetry/fixtures/pyproject.toml` should likely be changed to `Phylum, Inc. <engineering@phylum.io>`

kylewillmon:
> There are test files copied

This is not new. And without a true packaging system for extensions, this is difficult to avoid.

cli/src/install.sh
cd-work: I mean then users just can't make modifications to our extensions at all anymore. That seems very counter-productive.

Extensions are all about hacking some stuff together to fit your own usecase. People should be encouraged to modify the stock extensions, not blocked from doing so.

cd-work: You could argue that someone modifying the extensions should have a local for of our git repository available, though at that point I don't see why `install` wouldn't overwrite too. Either way I'd just be consistent?

kylewillmon: I believe the current code strikes a decent balance here. Users can install an extension with the same name as an official extension if they want to use a modified version.

Any extensions in the regular extensions directory are used first. And the official extension is only used if no extension exists with the same name.

maxrake (approved):
Things appear to work as expected/advertised. There was one observation that might require further discussion. It does not appear to be possible to uninstall an official extension:

image

Is this expected behavior?

cli/src/install.sh
cd-work: I **really** don't like the idea of creating separate directories for official and unofficial extensions. That just sounds like a cheap hack avoiding a proper solution to me.

And imo it shows in the code, it just introduces so much unnecessary complexity. It doesn't really go hand in hand with the spirit of extensions imo and seems like it's just a worse way to ship built-in subcommands at that point.

kylewillmon:
This is now on hold and the final PR will look very different. I may revive this PR later or (more likely) just re-use some of the code in a new PR.

PR #679, opened 2022-09-14
separate parsers into new crate

### LINKED ISSUE #680: Extract lock file parsing so it can be used outside cli
# Overview
Other Phylum code needs the ability to parse lock files and extract dependencies. It's currently possible by using the cli crate as a library, but because of the way the cli crate is structured you end up bringing in the entire cli and extension system as well.

# Acceptance Criteria
- [ ] The lock file parsing logic from this repository can be reused from other crates.
- [ ] Parsers can be resolved by name so parser names are consistent across Phylum software.

matt-phylum:
This PR breaks the parsers out of CLI into their own crate so they can be used outside CLI.

The new parser crate contains the same parsers, but with some added introspection capabilities about what formats are supported and what formats are being used.

The parser test fixtures have been moved up to the top level because they are used by both the cli crate and the parse crate.

Closes #680

## Checklist


Cargo.toml
```diff
@@ -2,5 +2,6 @@
 resolver = "2"
 members = [
   "cli",
+  "parse",
```
cd-work, line 5: `parse` is not exactly a descriptive name. I appreciate the brevity but I feel like it might be misplaced for a crate name? Especially if we ever want to publish any of these to crate.io.

I'd at the very least include *what* we're parsing in the name? In fact I'd say that's the more important part here, that we're parsing is mostly implementation detail really.

cli/Cargo.toml
```diff
@@ -34,8 +34,8 @@ lazy_static = "1.4.0"
 log = "^0.4.6"
 maplit = "1.0.2"
 minisign-verify = "0.2.1"
-nom = "7.1.1"
 open = "3.0.1"
+phylum-parse = { path = "../parse" }
 phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "development" }
```
cd-work, line 39: Inconsistent naming here, I strongly suggest using underscore everywhere.

cli/src/commands/extensions/api.rs
```diff
@@ -223,10 +224,10 @@ async fn parse_lockfile(
...
     // Attempt to parse as requested lockfile type.
 
-    let parser = LOCKFILE_PARSERS
-        .iter()
-        .find_map(|(name, parser)| (*name == lockfile_type).then(|| *parser))
-        .ok_or_else(|| anyhow!("Unrecognized lockfile type: `{lockfile_type}`"))?;
+    let parser = lockfile_type
+        .parse::<LockFileFormat>()
+        .with_context(|| format!("Unrecognized lockfile type: `{lockfile_type}`"))?
+        .parser();
```
cd-work, line 230: I'd assume this changes the CLI's error message? Could you provide the before/after?

parse/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "phylum-parse"
+description = "Parsers used by Phylum"
+version = "0.1.0"
+authors = ["Phylum, Inc. <engineering@phylum.io>"]
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
cd-work, line 8: This comment is unnecessary.

We should probably also specify MSRV considering we do so in cli too.

parse/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod lockfiles;
```
cd-work, line 1: Why did you go for this crate structure? It seems rather strange to me to have everything inside a single module with that being the only thing directly exposed in the crate root.

parse/src/lockfiles/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+mod python;
+mod ruby;
+
+/// A file format that can be parsed.
+#[derive(
+    Clone, Copy, Debug, Deserialize, EnumIter, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize,
+)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub enum LockFileFormat {
```
cd-work, line 28: Lockfile is generally used as a single word in our CLI repository afaik. So I wouldn't capitalize the F here.

parse/src/lockfiles/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+pub type ParseResult = anyhow::Result<Vec<PackageDescriptor>>;
+
+pub trait Parse {
+    /// Parse from a string
+    fn parse(&self, data: &str) -> ParseResult;
+
+    /// Indicate the type of file parsed by this parser
+    fn format(&self) -> LockFileFormat;
```
cd-work, line 84: Why did you add this function? As far as I can tell, it's only ever used in tests?

parse/src/lockfiles/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+mod csharp;
+mod java;
+mod javascript;
+mod parsers;
+mod python;
+mod ruby;
+
+/// A file format that can be parsed.
+#[derive(
+    Clone, Copy, Debug, Deserialize, EnumIter, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize,
```
cd-work, line 24: I'm personally not a big fan of pulling in dependencies just to implement iteration on enums.

cd-work (changes requested):
Preliminary review:

This PR kind of does a lot of stuff at the same time. I personally would have preferred it a lot of it didn't move the stuff to a new crate, refactor the entire thing **and** fix some issues you noticed along the way.

That's likely just going to drag out the review process unnecessarily.

parse/src/lib.rs
matt-phylum: The parser crate was expected to have other parsers besides lock files.

parse/src/lockfiles/mod.rs
matt-phylum: This function is required for knowing what format was used when the format is autodetected based on the file name unless you want to try casting the `&dyn Parse` back to every known implementation.

matt-phylum:
> This PR kind of does a lot of stuff at the same time. I personally would have preferred it a lot of it didn't move the stuff to a new crate, refactor the entire thing **and** fix some issues you noticed along the way.

This is actually a merge back from another repository where the changes to the parse crate were previously reviewed.

cli/src/commands/extensions/api.rs
matt-phylum: It apparently does not change anything. Previously, there was no inner because find_map returns Option. The new code preserves the error from FromStr, but that information apparently doesn't survive the translation into Deno.

Before
```
Extension error: Uncaught (in promise) Error: Unrecognized lockfile type: `poetry3`
    at async Function.parseLockfile (deno:phylum:243:16)
    at async file:///Users/matt.donoughe/Documents/GitHub/cli/cli/tests/fixtures/extensions/api/main.ts:2:18
```

After
```
Extension error: Uncaught (in promise) Error: Unrecognized lockfile type: `poetry3`
    at async Function.parseLockfile (deno:phylum:243:16)
    at async file:///Users/matt.donoughe/Documents/GitHub/cli/cli/tests/fixtures/extensions/api/main.ts:2:18
```

Cargo.toml
matt-phylum: The thinking was that this crate would be for any common parsers because people didn't want to have a lot of repositories. Maybe it doesn't matter anymore since this code has rejoined its original repository.

parse/src/lockfiles/mod.rs
matt-phylum: In this repository, lock file is written as two words 18 times, mostly in user-facing locations. It is written as one word 177 times.

parse/src/lib.rs
cd-work: I'm not sure it's a good idea to have a "jack of all trades" parsing crate. I personally prefer specialized crates that focus on doing one thing well.

matt-phylum: My original crate was phylum-lockfile. I can bring that back.

cd-work: Yeah I'd prefer `phylum_lockfile` myself over `parse`.

parse/src/lockfiles/mod.rs
cd-work: Thanks for getting the numbers on that one, I had a hunch but always good to confirm.

parse/src/lockfiles/mod.rs
cd-work: So the goal is mostly for non-CLI usage? Do you have a usecase for this already lined up or did you just add it to be safe?

cli/src/commands/parse.rs
```diff
@@ -61,11 +43,11 @@ pub fn try_get_packages(path: &Path) -> Result<(Vec<PackageDescriptor>, PackageT
...
-    for (name, parser) in LOCKFILE_PARSERS.iter() {
-        if let Ok(pkgs) = parser.parse(data.as_str()) {
+    for format in LockfileFormat::iter() {
+        if let Ok(pkgs) = format.parser().parse(data.as_str()) {
             if !pkgs.is_empty() {
-                print_user_success!("Identified lockfile type: {}", name);
-                return Ok((pkgs, parser.package_type()));
+                print_user_success!("Identified lockfile type: {}", format);
+                return Ok((pkgs, format.parser().package_type()));
             }
```
cd-work, line 51: The `parser()` function is called twice here. I think it would be cleaner to extract that into a `parser` variable and reuse it.

cli/src/commands/parse.rs
```diff
@@ -76,63 +58,37 @@ pub fn try_get_packages(path: &Path) -> Result<(Vec<PackageDescriptor>, PackageT
...
-        "poetry.lock" => parse(Poetry, path)?,
-        "effective-pom.xml" => parse(Pom, path)?,
-        "gradle.lockfile" => parse(GradleLock, path)?,
-        ".csproj" => parse(CSProj, path)?,
-        _ => try_get_packages(path)?,
+    let res = if let Some(parser) = get_path_parser(path) {
+        let data = read_to_string(path)?;
+        (parser.parse(&data)?, parser.package_type())
+    } else {
+        try_get_packages(path)?
```
cd-work, line 65: I'd say with option it's more consistent internally and generally within the Rust ecosystem to `match` rather than using `if let/else`.

lockfile/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "phylum_lockfile"
+description = "Package lock file parsers used by Phylum"
+version = "0.1.0"
+authors = ["Phylum, Inc. <engineering@phylum.io>"]
+edition = "2021"
+rust-version = "1.60"
```
cd-work, line 7: Just for future reference: We should probably add a build with MSRV to our CI. Right now there is zero guarantee that this crate will actually build on 1.60.

lockfile/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+[dependencies]
+anyhow = "1.0.44"
+log = "^0.4.6"
+nom = "7.1.1"
+phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "development" }
+serde = { version = "^1.0", features = ["derive"] }
+serde_json = "^1.0"
+serde_yaml = "0.9.2"
+serde-xml-rs = "0.5.1"
+toml = "0.5.8"
```
cd-work, line 18: You might have just copied that over, but I would **always** specify the dependencies with major, minor AND patch version. Specifically for serde/serde_json here.

This makes sure that our semver compatability is actually correct. And rather than just blindly putting `.0` in the last position you should either test that `.0` actually works or use the latest version if you're just lazy (both are fine by me). This way when people see `.0` they **know** that we will actually build with `.0`.

lockfile/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    /// Indicate the type of file parsed by this parser
+    fn format(&self) -> LockfileFormat;
+
+    /// Indicate the type of packages parsed by this parser
+    fn package_type(&self) -> PackageType;
+
+    /// Test if a file name could be a lock file supported by this parser.
+    ///
+    /// The file does not need to exist.
```
cd-work, line 138:
```suggestion
    /// Will return `false` if the file does not exist.
```

I think this is a neat tiny modification that provides additional useful information while barely increasing the length of the comment.

However it's probably important to point out that this is a guarantee that must be ensured by the implementation. It's difficult on this level to make guarantees about the behavior of each implementation of this function.

lockfile/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+
+pub trait Parse {
+    /// Parse from a string
+    fn parse(&self, data: &str) -> ParseResult;
+
+    /// Indicate the type of file parsed by this parser
+    fn format(&self) -> LockfileFormat;
+
+    /// Indicate the type of packages parsed by this parser
+    fn package_type(&self) -> PackageType;
```
cd-work, line 134:
```suggestion
    /// Parse from a string.
    fn parse(&self, data: &str) -> ParseResult;

    /// Indicate the type of file parsed by this parser.
    fn format(&self) -> LockfileFormat;

    /// Indicate the type of packages parsed by this parser.
    fn package_type(&self) -> PackageType;
```

Just a little extra consistency.

lockfile/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    /// Test if a file name could be a lock file supported by this parser.
+    ///
+    /// The file does not need to exist.
+    fn is_path_lockfile(&self, path: &Path) -> bool;
+}
+
+/// Get the parser of a potential lock file.
```
cd-work, line 142:
```suggestion
/// Get the parser for a potential lock file.
```

Your english is most likely much better than mine. But I think this reads a little nicer?

lockfile/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+///
+/// If the file name does not look like a lock file supported by this crate,
+/// `None` is returned.
+///
+/// The file does not need to exist.
```
cd-work, line 147: These two statements could probably be combined, which again would allow us to actually tell the user what will be returned when the file doesn't exist.

cd-work (changes requested):

lockfile/src/lib.rs
matt-phylum: Actually this suggestion is not the correct behavior. The file does not need to exist because the function only cares about the path value, not the file at that location.

lockfile/Cargo.toml
matt-phylum: These lines with other version formats are copied verbatim from cli/cargo.toml. Should the cli crate dependencies also be updated?

lockfile/src/lib.rs
matt-phylum: I thought it was weird that these comments were like this, but I didn't change it.

lockfile/src/lib.rs
matt-phylum: `for` makes more sense if it's getting a parser. Maybe at some point it was getting the expected format instead.

lockfile/src/lib.rs
cd-work: Oh right, you're ignoring if the file exists or not since the comparison is just done based on the filename.

lockfile/Cargo.toml
cd-work: I'd appreciate it tbh. But I don't mind either way. I just wouldn't want it in **new** code. Feel free to update though if you know which versions are guaranteed to build.

parse/src/lockfiles/mod.rs
matt-phylum: This is related to roadmap#140. However, it could probably be useful for API as well. There's nothing very important using this `.format()` at the moment, but it can be useful for things like logging `Parsed file requirements-all.txt as pip` consistently regardless of whether the parser was selected based on the name of the file or by trying formats until finding a parser that works.

Maybe it'd be better to just replace `get_path_parser` with `get_path_format` since you can always go from a format to a parser (and that's actually how `get_path_parser` is implemented).

lockfile/Cargo.toml
matt-phylum: I'll just put in the versions from the package lock file since that's what CLI has been using.

lockfile/Cargo.toml
matt-phylum: I also noticed clippy is not being run against test code.

parse/src/lockfiles/mod.rs
matt-phylum: Only implementing LockfileFormat->Parser and then having `get_path_format` seems better.

cd-work:
@matt-phylum Please ensure you do not revert https://github.com/phylum-dev/cli/pull/681 when updating your PR on top of master.

cd-work (approved):
I gave it another rough read-over and things seem fine now. However since my time is somewhat limited I'm not able to do a thorough review, if you still have doubts about anything I recommend pinging kyle/andrea.

PR #684, opened 2022-09-16
Fix extension name regex

cd-work:
The extension name error message pointed out `_` as a legal character, however the regex itself did not allow for this character.

To ensure extension naming is consistent, the `_` is removed from the error and only `-` is allowed.

### DIFF
--- a/cli/src/commands/extensions/extension.rs
+++ b/cli/src/commands/extensions/extension.rs
@@ -207,11 +207,7 @@ pub fn validate_name(name: &str) -> Result<(), anyhow::Error> {
     if EXTENSION_NAME_RE.is_match(name) {
         Ok(())
     } else {
-        Err(anyhow!(
-            "{}: invalid extension name, must be lowercase alphanumeric, dash (-) or underscore \
-             (_)",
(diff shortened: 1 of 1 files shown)


cli/src/commands/extensions/extension.rs
```diff
@@ -207,11 +207,7 @@ pub fn validate_name(name: &str) -> Result<(), anyhow::Error> {
     if EXTENSION_NAME_RE.is_match(name) {
         Ok(())
     } else {
-        Err(anyhow!(
-            "{}: invalid extension name, must be lowercase alphanumeric, dash (-) or underscore \
-             (_)",
-            name
-        ))
+        Err(anyhow!("{}: invalid extension name, must be lowercase alphanumeric, dash (-)", name))
```
andreaphylum, line 210: I'd slightly reword, including the "starts alphabetic":
```suggestion
        Err(anyhow!("{}: invalid extension name, must start with a letter and can contain only lowercase alphanumeric character or dashes (-)", name))
```

andreaphylum (approved):

PR #693, opened 2022-09-21
Allow upgrade in phylum extension install

### LINKED ISSUE #400: Allow extension upgrades
We may want to implement the `phylum extension upgrade` subcommand on top of `install`/`uninstall`. In that case, we may allow overwriting the extension installation directory, which we disallow at the moment.

kylewillmon:
This patch adds code to allow extension upgrades with the `phylum extension install` command. 

Fix #400


kylewillmon:
The code to compare directories was a bit more complicated than I expected. Opening a draft PR for early feedback on that code.

Cc: @cd-work

kylewillmon:
A few notes about the current implementation:

* The check for an installed extension happens **after** the permissions prompt
  ```
  ❯ phylum extension install ./yarn
  The `yarn` extension requires the following permissions:
  
    Read the following paths:
      './package.json'
      'yarn.lock'
  
    Write the following paths:
      './package.json'
      'yarn.lock'
  
    Run the following commands:
      'yarn'
  
  
  Do you accept? yes
  Installing extension yarn...
  ❗ Error: identical extension already installed, skipping
  ```
* The `--yes` flag only applies to the permissions prompt. Users will still be prompted about overwriting a previous version.
  ```
  ❯ phylum extension install --yes ./yarn2
  Installing extension yarn...
  Another version of the 'yarn' extension is already installed. Overwrite? yes
  Removing old extension yarn...
  Extension yarn installed successfully
  ```

cli/src/commands/extensions/extension.rs
```diff
@@ -104,12 +93,47 @@ impl Extension {
...
+            if dir_compare(&self.path, &target_prefix)? {
+                return Err(anyhow!("identical extension already installed, skipping"));
+            } else {
+                let mut prompt = Confirm::new();
+                prompt
+                    .with_prompt(format!(
+                        "Another version of the '{}' extension is already installed. Overwrite?",
+                        self.name()
+                    ))
+                    .default(true);
```
andreaphylum, line 124: I would cut down on complexity, remove this check and reword to something like `The '{}' extension is already installed. Overwrite?`.

IIUC, not having it would only mean that we overwrite files that are already identical, which is a very modest overhead unless we anticipate cli extensions to be in the hundreds of MB of size -- and even then, it is much more a networking issue rather than disk writes issue. Having it has a cost in term of complexity.

Do we need the `dir_compare` check, or to report to the user that the installed extension is identical? Are there other reasons/places it could be useful? If so, I'd also add a few tests with fixtures for it.

andreaphylum (changes requested):

cli/src/commands/extensions/extension.rs
kylewillmon: > Do we need the `dir_compare` check, or to report to the user that the installed extension is identical? Are there other reasons/places it could be useful?

The `dir_compare` check is used to avoid excessive prompting during the following scenario: The `phylum extension install` command will be used during `phylum update` to install or upgrade official extensions. Since extensions may not change between every release, we want to avoid prompting for every official extension during every `phylum update`.

kylewillmon: > If so, I'd also add a few tests with fixtures for it.

I'll have to think a bit more on this. Testing the interactive prompt is difficult and we avoid it in other cases (e.g., the permissions prompt). But I guess I could add a test case to confirm that install is aborted when attempting to install a modified version of an extension in a non-interactive setting.

kylewillmon:
> The `--yes` flag only applies to the permissions prompt. Users will still be prompted about overwriting a previous version.

I'm currently struggling with this. And I'm beginning to think that we will need separate flags for `--accept-permissions` and `--overwrite` because they are different decisions. For example, during an automated install of official extensions it is reasonable to automatically accept extension permissions, but not reasonable to automatically overwrite whatever is already there.

cli/src/commands/extensions/extension.rs
andreaphylum: > The `dir_compare` check is used to avoid excessive prompting during the following scenario: The `phylum extension install` command will be used during `phylum update` to install or upgrade official extensions. Since extensions may not change between every release, we want to avoid prompting for every official extension during every `phylum update`.

In that case, yeah, the check is a good thing.

> I'll have to think a bit more on this. Testing the interactive prompt is difficult and we avoid it in other cases (e.g., the permissions prompt). But I guess I could add a test case to confirm that install is aborted when attempting to install a modified version of an extension in a non-interactive setting.

I'd say it wouldn't be necessary to go that far, though an e2e test of this feature is welcome.

What I'd like to mainly have more coverage on is just `dir_compare` and `file_compare`, and we could do that with no prompting at all, by just having tests that create, populate and compare some temporary directories and files (respectively).

kylewillmon: Good point. I was thinking e2e test, but unit testing is much simpler. I will add unit tests.

andreaphylum:
> > The `--yes` flag only applies to the permissions prompt. Users will still be prompted about overwriting a previous version.
> 
> I'm currently struggling with this. And I'm beginning to think that we will need separate flags for `--accept-permissions` and `--overwrite` because they are different decisions. For example, during an automated install of official extensions it is reasonable to automatically accept extension permissions, but not reasonable to automatically overwrite whatever is already there.

The use case for `--yes` and similar flags is unsupervised execution, so we need to make sure the presence of any of those doesn't end up in a prompt, and not specifying a flag that is required should result in exiting with an error.

Potential approaches: 
- Have separate flags for separate decisions. Pro: granular control. Con: verbosity
- Have one single `--yolo` flag for everything. Pro: easy and probably what's intended by the user. Con: all of its consequences may not be understood by the user.
- A mix of the two, where one flag can specify all of them but also individual ones are available. Pro: flexibility for the user. Con: more code to maintain and we need to design and document it well to avoid it being confusing.

kylewillmon:
> * A mix of the two, where one flag can specify all of them but also individual ones are available

This is what I'm leaning towards. Specifically:

```
  --overwrite           Overwrite prior installed extensions
  --accept-permissions  Automatically accept requested permissions
  --yes                 Accept permissions and overwrite existing (equivalent to --overwrite --accept-permissions)
```

andreaphylum (approved):

maxrake (changes requested):
Code and behavior looks good. The only thing found was a missing update to the `docs/command_line_tool/phylum_extension_install.md` file to match the new behavior/options.

maxrake (approved):

Issue #54, opened 2021-06-21
`phylum` does not gracefully display error conditions when `analyze` submission is not correctly associated with a project


peterjmorgan:
**Is your feature request related to a problem? Please describe.**
Yes. When a user submits an `analyze` request but has not created a project for the request, the rust error message is returned to `stderr`. 

**Describe the solution you'd like**
This feature request centers on removing the `stderr` output and cleanly reporting the error in the design of the CLI tool

PR #705, opened 2022-09-30
Replace `Deno.run` instances with `PhylumApi.runSandboxed`

### LINKED ISSUE #697: Replace `Deno.run` executions in extensions with `Phylum.runSandboxed`
Currently, a few `Deno.run` calls in the `npm` extension have side effects, e.g. `preinstall` scripts are run even with the `--package-lock-only` flag, and those aren't gated by the sandbox. We should update the relevant code once #673 is merged in and alongside #692.

andreaphylum:
Closes #697.

The permissions placed on each invocation, laxer than specifying directly each file, are provisional but should be safe.

We do need to investigate the fact that with the current setup on my machine (`Linux phylum-ubuntu 5.15.0-48-generic #54-Ubuntu SMP Fri Aug 26 13:26:29 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`) specifying exceptions directly regarding a file (e.g. `./package-lock.json` or even `./node_modules`) in `runSandboxed` does not grant access as expected, whereas specifying the same exception on the parent directory (`./`) does.


kylewillmon:
The commits on this branch messed up because it's based on the pre-merged #673 branch. You can fix that with a rebase like this one:

```
git rebase --onto origin/main 5de375f
```

cli/Cargo.toml
```diff
@@ -61,7 +61,7 @@ once_cell = "1.12.0"
 deno_runtime = { version = "0.71.0" }
 deno_core = { version = "0.145.0" }
 deno_ast = { version = "0.17.0", features = ["transpiling"] }
-birdcage = { git = "https://github.com/phylum-dev/birdcage" }
+birdcage = { git = "https://github.com/phylum-dev/birdcage", branch = "andrea/abi-v2" }
```
andreaphylum, line 64: We may want to hold out and merge that branch in `birdcage` before merging this.

extensions/poetry/PhylumExt.toml
```diff
@@ -4,5 +4,5 @@ entry_point = "main.ts"
 
 [permissions]
 run = ["poetry"]
-read = ["./poetry.lock", "./pyproject.toml"]
-write = ["./poetry.lock", "./pyproject.toml"]
+read = ["~/.cache/pypoetry", "~/.local/lib", "./"]
```
andreaphylum, line 7: Added `~/.local/lib` as a common path for unprivileged installations of Python. Other places should be covered by our whitelist. This doesn't take into account other exotic installation locations but I think it should be fine.

extensions/npm/main.ts
```diff
@@ -56,7 +56,16 @@ if (Deno.args.length >= 1
...
 
     // Install packages without executing build scripts.
-    let status = await Deno.run({ cmd: ['npm', ...Deno.args, '--ignore-scripts']}).status();
+    let status = PhylumApi.runSandboxed({
+        cmd: 'npm',
+        args: [...Deno.args, '--ignore-scripts'],
+        exceptions: {
+            write: ['~/.npm', './'],
+            run: ['npm'],
+            read: ['~/.npm', './'],
```
cd-work, line 65: I'd always cluster read/write together and just move run above/below the two. Especially because they're likely to share some content. Will look much cleaner.

extensions/npm/main.ts
```diff
@@ -94,19 +103,33 @@ if (Deno.args.length >= 1
...
+    let status = PhylumApi.runSandboxed({
+        cmd: 'npm', 
+        args: Deno.args,
+        exceptions: {
+            write: ['~/.npm', './'],
+            read: ['~/.npm', './'],
+            run: ['npm'],
+            net: true,
+        }
+    });
```
cd-work, line 115: I'd honestly be surprised if these were all the permissions required to run *any* `npm` command. But I guess we'll find out?

extensions/npm/main.ts
cd-work, line 65: I think these read permissions won't work with filesystem dependencies, like `npm i /tmp/test`. Pretty sure you need `read: true` for that.

extensions/poetry/PhylumExt.toml
cd-work: Who commonly installs Python to `~/.local/lib`? Like which application.

Also I'm not opposed to adding `~/.local/lib` to our general exceptions. Very unlikely there's going to be sensitive data there.

cd-work, line 7: Also want to mention filesystem dependencies here, doubt this suffices for them.

extensions/yarn/PhylumExt.toml
```diff
@@ -3,6 +3,6 @@ description = "yarn package manager hooks"
 entry_point = "main.ts"
 
 [permissions]
-write = ["./package.json", "yarn.lock"]
-read = ["./package.json", "yarn.lock"]
+write = ["~/.cache", "~/.yarn", "./"]
+read = ["~/.cache", "~/.yarn", "./"]
```
cd-work, line 7: I'd restrict this to `~/.cache/node`, no reason to allow it to write to my Firefox cache for example.

extensions/yarn/main.ts
```diff
@@ -55,7 +55,16 @@ if (Deno.args.length >= 1
...
+    let status = PhylumApi.runSandboxed({
+        cmd: 'yarn',
+        args: [...Deno.args, '--mode=skip-build'],
+        exceptions: {
+            write: ['~/.cache',  '~/.yarn', './'],
+            read: ['~/.cache',  '~/.yarn', './'],
+            run: false,
+            net: true,
+        }
+    })
```
cd-work, line 67: Won't work for local dependencies: `yarn add /tmp/test`.

cd-work (changes requested):

extensions/poetry/PhylumExt.toml
andreaphylum: I'd probably think about adding the entirety of `~/.local`, honestly, as the chance that there's anything there that could pose more of a risk than anything in `/usr` does is slim.

What do you mean by "which application"? I do that myself because I usually stay on top of newer releases of Python, which aren't in package managers yet, plus I try to avoid privileged installations in `/usr/local`, and in those cases installation usually happens in `~/.local`.

cli/src/commands/extensions/api.rs
```diff
@@ -340,7 +340,11 @@ fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
     // Apply sandbox to subprocess after fork.
     unsafe {
         command.pre_exec(move || {
-            let into_ioerr = |err| io::Error::new(io::ErrorKind::Other, err);
+            fn into_ioerr<E: Into<Box<dyn std::error::Error + Send + Sync>>>(err: E) -> io::Error {
+                io::Error::new(io::ErrorKind::Other, err)
+            }
```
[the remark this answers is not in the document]
andreaphylum: Yeah, type inference was having troubles.

extensions/poetry/PhylumExt.toml
cd-work: I definitely would **not** add all of `~/.local`. You can easily have sensitive data in there because it is used as a storage directory for applications. Say your text editor stores undo history there for example. So there's a **very high** chance *something* in there might be sensitive.

> What do you mean by "which application"? I do that myself because I usually stay on top of newer releases of Python, which aren't in package managers yet, plus I try to avoid privileged installations in /usr/local, and in those cases installation usually happens in ~/.local.

How did you decide on `~/.local/lib`? Not that I don't trust your judgement, but adding an exception just because one developer decided to put his stuff there isn't really a good reason. Was this documented somewhere as the recommended location?

cli/src/commands/extensions/api.rs
```diff
@@ -349,11 +353,13 @@ fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
...
 
             for path in process.exceptions.read.sandbox_paths().iter() {
-                permissions::add_exception(&mut birdcage, Exception::Read(PathBuf::from(path)))
+                let path = permissions::expand_home_path(path, &home_dir);
+                permissions::add_exception(&mut birdcage, Exception::Read(path))
                     .map_err(into_ioerr)?;
             }
             for path in process.exceptions.write.sandbox_paths().iter() {
-                permissions::add_exception(&mut birdcage, Exception::Write(PathBuf::from(path)))
+                let path = permissions::expand_home_path(path, &home_dir);
```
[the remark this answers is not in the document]
andreaphylum: I think this is the only place we use it in at the moment -- I placed it next to the `resolve_bin_path` method that was related to it. Semantically, `dirs` probably makes more sense, though.

extensions/poetry/PhylumExt.toml
andreaphylum: > I definitely would **not** add all of `~/.local`. You can easily have sensitive data in there because it is used as a storage directory for applications. Say your text editor stores undo history there for example. So there's a **very high** chance _something_ in there might be sensitive.

We could then add multiple exceptions, such as `~/.local/bin`, `etc`, `include` and so on. In the specific case of Python, no other paths are read.

> How did you decide on `~/.local/lib`? Not that I don't trust your judgement, but adding an exception just because one developer decided to put his stuff there isn't really a good reason. Was this documented somewhere as the recommended location?

I have been using `~/.local` as install prefix for as long as I remember the XDG spec being a thing. `lib` is just because that's where the Python installation is stored as a consequence. But I just checked and can't find an explicit reference to this, so we may want to let go of this assumption. Of course this creates a problem for users having installations in uncommon paths, but I'm not sure we want to support that just yet.

cli/src/commands/extensions/api.rs
```diff
@@ -364,12 +370,15 @@ fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
...
             }
+
             birdcage.lock().map_err(into_ioerr)?;
             Ok(())
         });
     }
 
-    let output = command.output()?;
+    let output = command.output().with_context(|| {
+        format!("Executing sandboxed process `{}` `{}`", process.cmd, process.args.join(" "))
```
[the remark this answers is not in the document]
andreaphylum: I put backticks there to provide some separation. Though, with empty args that would look like

```
Error: Executing sandbox process `touch` ``
```
with the current solution, and
```
Error: Executing sandbox process `touch `
```
by removing the ones in the middle, so suboptimal in both cases. Nothing that can't be easily improved, though.

cli/src/commands/extensions/permissions.rs
```diff
@@ -229,12 +225,11 @@ pub fn default_sandbox() -> SandboxResult<Birdcage> {
...
-        Ok(_) => Ok(()),
+    birdcage.add_exception(exception).map(|_| ()).or_else(|e| match e {
         // Ignore invalid path errors.
-        Err(SandboxError::InvalidPath(_)) => Ok(()),
-        Err(err) => Err(err),
-    }
+        SandboxError::InvalidPath(_) => Ok(()),
+        err => Err(err),
+    })
 }
```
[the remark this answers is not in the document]
andreaphylum: I had to rewrite that part a few times due to other changes, and by the time I needed to revert to the previous logic I wrote it in this slightly different way. Can revert the changes altogether if needed.

extensions/npm/main.ts
andreaphylum: Agreed -- this is probably an oversight left over from to testing different stuff.

extensions/npm/main.ts
andreaphylum: I'd be surprised too, but as a developer I also would not want `npm` to go and modify files in unpredictable places. I'd rather find out and evaluate more exceptions as we go :D

extensions/npm/main.ts
andreaphylum: That's true, but having `read: true` and `net: true` would expose us to exfiltration (not through preinstall scripts at least, in this case). I still think it's ok.

cli/src/commands/extensions/api.rs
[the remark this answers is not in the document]
andreaphylum: Oh, yeah, that's a good point.

extensions/npm/main.ts
cd-work: Usually we should allow read+net only without executing build scripts, then when we execute the build scripts we don't need `net` anymore. So that way we should be safe?

andreaphylum: Yeah, exactly.

cli/Cargo.toml
```diff
@@ -61,7 +61,7 @@ once_cell = "1.12.0"
 deno_runtime = { version = "0.79.0" }
 deno_core = { version = "0.153.0" }
 deno_ast = { version = "0.19.0", features = ["transpiling"] }
-birdcage = { git = "https://github.com/phylum-dev/birdcage" }
+birdcage = { git = "https://github.com/phylum-dev/birdcage", branch = "andrea/abi-v2" }
```
cd-work, line 64: Would like to get this to main before merging.

cli/src/commands/extensions/api.rs
```diff
@@ -349,27 +353,34 @@ fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
...
             }
             for path in process.exceptions.write.sandbox_paths().iter() {
-                permissions::add_exception(&mut birdcage, Exception::Write(PathBuf::from(path)))
+                let path = dirs::expand_home_path(path, &home_dir);
+                permissions::add_exception(&mut birdcage, Exception::Write(path))
                     .map_err(into_ioerr)?;
             }
             for path in process.exceptions.run.sandbox_paths().iter() {
-                let absolute_path = permissions::resolve_bin_path(path);
+                let absolute_path = permissions::resolve_bin_path(path).map_err(into_ioerr)?;
```
cd-work, line 366: I would have preferred resolving home dirs before bin_path, for consistency and to keep resolve_bin_path focused. I don't think we need this behavior anywhere else.

But I'm fine with both.

cli/src/commands/extensions/permissions.rs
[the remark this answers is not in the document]
cd-work: I'd still like to revert this if possible.

The current code is just more convoluted and (in a way that likely gets optimized out) less efficient.

cli/src/commands/extensions/permissions.rs
```diff
@@ -230,36 +226,39 @@ pub fn default_sandbox() -> SandboxResult<Birdcage> {
...
 
 /// Resolve non-absolute bin paths from `$PATH`.
-pub fn resolve_bin_path(bin: &str) -> PathBuf {
+pub fn resolve_bin_path(bin: &str) -> Result<PathBuf> {
```
cd-work, line 237: Having to change this to a `Result` here is another good argument for not resolving home paths inside this function.

Could even make it accept something like `AsRef<Path>` or something to make it absolutely clear that this function only takes an already valid path as argument.

extensions/npm/main.ts
```diff
@@ -94,19 +103,33 @@ if (Deno.args.length >= 1
...
+    let status = PhylumApi.runSandboxed({
+        cmd: 'npm',
+        args: [subcommand, '--package-lock-only', ...args],
+        exceptions: {
+            write: ['~/.npm', './'],
+            read: ['~/.npm', './'],
+            run: ['npm'],
+            net: true,
+        }
+    })
```
cd-work, line 132: Have you tested this with local packages? It should also need `read: true`, right?

extensions/yarn/main.ts
```diff
@@ -93,17 +102,33 @@ if (Deno.args.length >= 1
...
+    let status = PhylumApi.runSandboxed({
+        cmd: 'yarn',
+        args: [...Deno.args],
+        exceptions: {
+            read: true,
+            write: ['~/.cache/node', '~/.cache/yarn', '~/.yarn', './'],
+            run: false,
+            net: true,
+        }
+    })
```
cd-work, line 114: This has a `read` + `net` as true, which is a bit unfortunate for sandboxing. I don't think we have a better solution here but it might be worth it at some point to go through the `yarn` subcommands and see if there's anything else we want to protect against here.

Otherwise this is basically running unsandboxed, which is totally fine and will certainly be required by some extensions.

cd-work (changes requested):
Just a couple more minor things.

cli/src/commands/extensions/permissions.rs
andreaphylum: Yeah, I'd honestly prefer this to _not_ return `Result<_>`.

cli/src/commands/extensions/api.rs
andreaphylum: Agreed, and reimplemented this way.

extensions/npm/main.ts
andreaphylum: I haven't but you're right, `read: true` should be good since we're not executing scripts anyway.

cli/src/commands/extensions/api.rs
```diff
@@ -349,27 +353,35 @@ fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
...
             for path in process.exceptions.write.sandbox_paths().iter() {
-                permissions::add_exception(&mut birdcage, Exception::Write(PathBuf::from(path)))
+                let path = dirs::expand_home_path(path, &home_dir);
+                permissions::add_exception(&mut birdcage, Exception::Write(path))
                     .map_err(into_ioerr)?;
             }
             for path in process.exceptions.run.sandbox_paths().iter() {
-                let absolute_path = permissions::resolve_bin_path(path);
+                let absolute_path =
+                    permissions::resolve_bin_path(dirs::expand_home_path(path, &home_dir));
```
cd-work, line 367: We've got a linebreak here anyway, I think it would be nice for consistency to extract the home path resolution into an intermediate value. That way it's much more consistent and thus it's going to be easy to spot where changes need to be made if we need to do so in the future.

cd-work (changes requested):
Just a super tiny stylistic suggestion we can apply when the birdcage changes are upstreamed. Other than that this should be good to go.

cli/src/commands/extensions/api.rs
andreaphylum: Agreed

cli/Cargo.toml
andreaphylum: Just merged `birdcage` and pushed a commit with the branch spec removed.

extensions/poetry/PhylumExt.toml
```diff
@@ -4,5 +4,5 @@ entry_point = "main.ts"
 
 [permissions]
 run = ["poetry"]
-read = ["./poetry.lock", "./pyproject.toml"]
+read = true,
```
cd-work, line 7: Why is this still modified? Also I'm not sure the comma is even valid toml? I'd assume this was a bug during rebase?

cd-work (changes requested):

extensions/poetry/PhylumExt.toml
andreaphylum: Reverted

cd-work (approved):

cd-work (approved):

PR #709, opened 2022-10-03
Create project extensions API

### LINKED ISSUE #698: Expose a method for creating projects in the Extensions API

andreaphylum:
Closes #698.

## Checklist

### DIFF
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -18,16 +18,18 @@ use deno_runtime::deno_core::{op, OpDecl, OpState};
 use deno_runtime::permissions::Permissions;
 use phylum_lockfile::LockfileFormat;
 use phylum_types::types::auth::{AccessToken, RefreshToken};
-use phylum_types::types::common::JobId;
+use phylum_types::types::common::{JobId, ProjectId};
 use phylum_types::types::group::ListUserGroupsResponse;
 use phylum_types::types::job::JobStatusResponse;
 use phylum_types::types::package::{
     Package, PackageDescriptor, PackageStatusExtended, PackageType,
 };
 use phylum_types::types::project::ProjectSummaryResponse;
+use reqwest::StatusCode;
 use serde::{Deserialize, Serialize};
 use tokio::fs;
 
+use crate::api::{PhylumApiError, ResponseError};
 use crate::auth::UserInfo;
 #[cfg(unix)]
 use crate::commands::extensions::permissions;
@@ -252,6 +254,24 @@ async fn get_projects(
     let api = state.api().await?;
 
     api.get_projects(group.as_deref()).await.map_err(Error::from)
+}
+
+/// Create a project.
+#[op]
+async fn create_project(
+    op_state: Rc<RefCell<OpState>>,
+    name: &str,
+    group: Option<&str>,
+) -> Result<ProjectId> {
+    let state = ExtensionState::from(op_state);
+    let api = state.api().await?;
+
+    api.create_project(name, group).await.map_err(|e| match e {
+        PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. }) => {
+            anyhow!("Project '{}' already exists", name)
+        },
+        err => err.into(),
+    })
 }
 
 /// Analyze a single package.
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -260,13 +260,13 @@ async fn get_projects(
 #[op]
 async fn create_project(
     op_state: Rc<RefCell<OpState>>,
-    name: &str,
-    group: Option<&str>,
+    name: String,
+    group: Option<String>,
 ) -> Result<ProjectId> {
     let state = ExtensionState::from(op_state);
     let api = state.api().await?;
 
-    api.create_project(name, group).await.map_err(|e| match e {
+    api.create_project(&name, group.as_deref()).await.map_err(|e| match e {
         PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. }) => {
             anyhow!("Project '{}' already exists", name)
         },
@@ -432,6 +432,7 @@ pub(crate) fn api_decls() -> Vec<OpDecl> {
         get_current_project::decl(),
         get_groups::decl(),
         get_projects::decl(),
+        create_project::decl(),
         get_package_details::decl(),
         parse_lockfile::decl(),
         run_sandboxed::decl(),
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -265,7 +265,7 @@ struct CreatedProject {
 #[derive(Serialize)]
 enum CreatedProjectStatus {
     Created,
-    Existing,
+    Exists,
 }
 
 /// Create a project.
@@ -284,7 +284,7 @@ async fn create_project(
         Err(PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. })) => api
             .get_project_id(&name, group.as_deref())
             .await
-            .map(|id| CreatedProject { id, status: CreatedProjectStatus::Existing })
+            .map(|id| CreatedProject { id, status: CreatedProjectStatus::Exists })
             .map_err(|e| e.into()),
         Err(e) => Err(e.into()),
         Ok(id) => Ok(CreatedProject { id, status: CreatedProjectStatus::Created }),
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -266,12 +266,28 @@ async fn create_project(
     let state = ExtensionState::from(op_state);
     let api = state.api().await?;
 
-    api.create_project(&name, group.as_deref()).await.map_err(|e| match e {
-        PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. }) => {
-            anyhow!("Project '{}' already exists", name)
+    // Retrieve the id if the project already exists, otherwise return the id or the error.
+    match api.create_project(&name, group.as_deref()).await {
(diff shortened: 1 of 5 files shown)


cli/src/commands/extensions/api.rs
```diff
@@ -254,6 +256,24 @@ async fn get_projects(
...
+    name: String,
+    group: Option<String>,
+) -> Result<ProjectId> {
+    let state = ExtensionState::from(op_state);
+    let api = state.api().await?;
+
+    api.create_project(&name, group.as_deref()).await.map_err(|e| match e {
+        PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. }) => {
+            anyhow!("Project '{}' already exists", name)
+        },
```
kylewillmon, line 272: The API needs to handle this "already exists" condition in a well-defined manner. Many extensions will want to treat this as success and not an error. For an example of that, see [the `build-gradle` extension](https://github.com/phylum-dev/community-extensions/blob/8cb3b6bef8868dd96fb96142fe7ab861360cd601/extensions/build-gradle/main.ts#L265-L269) (which is where the requirement for this function began).

With this code returning an `anyhow::Error`, I guess that code would need a `try...catch` and a check against the error message. But that seems kinda ugly. Is there a cleaner for us to define this API?

kylewillmon (changes requested):
Is there a way to add a test for this function in our e2e tests?

Also, should we add `delete_project()` as well?

cli/src/commands/extensions/api.rs
andreaphylum: My original idea was to use the `try...catch` as most if not all other API endpoints return a `Result<T>`. I agree that the "already exists" case is actually another success state, though.

In that case we could either go for silent success (i.e. return the already existing project's id, the easier way) or with a supplementary field stating whether the creation was a success, or whether we are returning an already existing project's id.

andreaphylum:
Agreed -- we should add `delete_project` (and, gradually, probably also most of the other API endpoints). I forgot we had those e2e tests for each of the extensions APIs, but yeah, should definitely add a test there for this.

cli/src/commands/extensions/api.rs
cd-work: We can maybe avoid using `anyhow` and create a custom error struct for our API errors that includes things like an error code? That would make it possible to easily match for downstream.

andreaphylum: Moving away from `anyhow` throughout the module is too big a diff to be handled in this PR. I created #719 to track this, though, because I think it is a good idea.

kylewillmon (approved):

cli/src/extension_api.ts
```diff
@@ -181,6 +181,28 @@ export class PhylumApi {
...
+    /// The id of the newly created project:
+    ///
+    /// ```
+    /// "5d6eaa97-dff8-dead-a619-bcafffefeef0"
+    /// ```
+    static async createProject(name: string, group?: string): string {
+        return Deno.core.opAsync('create_project', name, group);
+    }
+
+    /// Deletes a project.
```
cd-work, line 197:
```suggestion
    /// Delete a project.
```

Just for consistency, should either be creates/deletes or create/delete.

cli/tests/end_to_end/extension.rs
```diff
@@ -75,6 +75,33 @@ pub async fn get_groups() {
...
+    test_cli
+        .extension(
+            r#"
+            try {
+                await PhylumApi.deleteProject("create_and_delete_test_project")
+            } catch (e) {
+            }
+
+            let projectId = await PhylumApi.createProject("create_and_delete")
+            let duplicatedProjectId = await PhylumApi.createProject("create_and_delete")
```
cd-work, line 91: I don't understand, why are we deleting `create_and_delete_test_project` here, when `create_and_delete` is the project name we use for the test? I assumed this was just a setup to ensure a clean state, but this wouldn't actually delete the project.

cd-work, line 91: I'm not sure we want to just silently ignore duplicates? In a follow-up I'd want to introduce error handling, but that doesn't mean it shouldn't fail at all right now? I wouldn't want to change an API from non-failing to failing, that seems kinda inconsistent?

An endpoint to get project info might help with that, just to see if it needs to be created or not ahead of time. I don't think we have one of those yet but it probably would make sense? We have `getCurrentProject` already.

cd-work (changes requested):

cli/tests/end_to_end/extension.rs
andreaphylum: It was indeed a typo.

andreaphylum: I agree that a breaking change in an API from non-failing to failing would be bad. But, as per the previous discussion, I think the way forward would make the "created" and "already exists" both _successful_ states; if anything, those two states would optionally be different in the response format (e.g. carry an "existing" flag or something similar alongside the id), but I don't even think the user would benefit from that.

I think erroring out on a project existing is bad from the user's perspective, as it forces them to do error management when all they care about is either creating or retrieving a project. The use case for checking whether a project exists would be covered by a "get project by name" API -- which we already have in the CLI, just not exposed to extensions. Once we have that, the user has no benefit in manually checking for the existence of a project, unless they need to for purposes different than deciding whether to create a project or not.

Also, "duplicated" here is a misnomer: a project that was already existing is returned with its id, rather than copied into a new project.

Given the above, I think the current solution is appropriate, but I'm open to other feedback.

An argument in favor of simplifying this API would be to more faithfully mimick our server side API, but I don't think that is necessarily an added value over providing a more ergonomic API to the users.

andreaphylum: That makes sense; the only point I question there is whether the user ever cares about actually knowing the project's state. I think not: all of the operations that can be done on a project (submitting an analysis to it; querying its information; deleting it) do not need to know anything about its state beforehand, and will error out if the project doesn't exist, as intended. Whatever we decide to name it, I think the semantics of "create project or return project handle" is the one that is going to be more in line with what a user instinctively does: 
- if someone cares about only creating a project if it doesn't exist, they can check manually with other APIs
- if someone _does not_ care about the state of the project, they don't have extra state to manage, and an error will have to be handled as an error
- if someone is not aware of this mechanism, they won't notice, and nothing bad is going to happen; this is going to be true also in the augmented response case where we might have something like `{ "id": "abcd1234", "status": "created" }` vs `{ "id": "abcd1234", "status": "existing" }`

On the other hand, returning the project to a clean slate could actually be what surprises a user, who might have their analysis canceled.

Considering this, what would you suggest a good way forward is? Pinging also @kylewillmon for feedback as he raised the issue originally.

kylewillmon: First, we should take the return value as a separate discussion. The Project ID is basically unused in our extension API because `analyze` take a Project Name rather than an ID. This function may not need a return value at all.

> it's basically just an `ensureProjectExists` method at that point rather than a `createProject` endpoint.

Renaming this function might solve all our problems. Ensuring that a project exists is the [primary use case](https://github.com/phylum-dev/community-extensions/blob/c26f325afa88607e2f9f3df7af7ba18d19188655/build-gradle/main.ts#L245-L272) for this feature.

Creating a clean API for `createProject` would probably require that we resolve #719 first. That problem does not exist with `ensureProjectExists`.

cd-work: It's just that `ensureProjectExists` is a little less RESTy/CRUDy, at least it feels a little bit weird to me.

I don't want to block this PR on just this question, but I'd definitely say there's only two good choices:
 - We never want this to return an error, even after #719, then we can just keep it as-is
 - We want it to return a custom error after #719, then I think we should already return an error

I'd at least like a clear decision for one of the two, my understanding right now is that it's basically a mix of the two which I don't really like.

kylewillmon: Yeah... We might need to go back to the issue (#698) and discuss the details of the API before continuing to go back and forth on this PR...

cd-work (changes requested):
After some discussion with @kylewillmon, I think our best solution for this PR to not keep it blocked forever is to return an object that contains both the project ID and a flag indicating if the project was newly created or not.

This will also mean that we will **not** change this API in the future to return an error, but keep it as-is.

I'd probably wait for the +1 by @kylewillmon before implementation.

andreaphylum:
> My only concern with this is that I am left wondering if we should be returning the entire project "object" (i.e, what is currently returned by `getProjects()`) and not just the ID.
> 
> But since that change might require backend support (or extra round trips) and is probably a part of a bigger discussion about adding type declarations to our TypeScript API, I am okay with putting that concern aside for now in the interest of keeping this PR moving.

If we are to now return an object like 
```json
{ "id": "abcd1234", "status": "created" }
```
nothing is stopping us from, later on, adding the other fields in the [`ProjectConfig`](https://github.com/phylum-dev/cli/blob/andrea/extensions-create-project/cli/src/config.rs#L66-L72) struct which is already returned by `get_current_project`.

I feel like augmenting the object with more fields at a later date definitely won't break any existing code, and we can decide whether to shoulder additional round trips or request more support from the backend without worrying too much.

cd-work:
I agree with @andreaphylum, I think this solution should allow us to easily adjust in the future while also providing a simple solution that works **now**.

cli/src/commands/extensions/api.rs
```diff
@@ -256,24 +256,38 @@ async fn get_projects(
...
+#[derive(Serialize)]
+struct CreatedProject {
+    id: ProjectId,
+    status: CreatedProjectStatus,
+}
+
+#[derive(Serialize)]
+enum CreatedProjectStatus {
+    Created,
+    Existing,
```
cd-work, line 268: I'm not sure "Existing" is the best choice here, but I don't have a better idea. Feels kinda weird when put next to "Created".

Maybe "Exists" would be better?

cd-work (changes requested):

cd-work (approved):

cd-work (approved):

PR #714, opened 2022-10-04
Removed ansi_term

### LINKED ISSUE #696: Replace `ansi_term` with `termcolor` or `console`
We are currently using [`ansi_term`](https://crates.io/crates/ansi_term) for color terminal output, which is now abandoned.

Thanks to our use of [`dialoguer`](https://crates.io/crates/dialoguer), we are already pulling in [`console`](https://crates.io/crates/console), which can also handle terminal color printing.

So in order to shorten our list of dependencies and avoid relying on abandonware, we should switch to `console` for all of our color terminal output.

AntoniosBarotsis:
Closes #696.

Changed `console` from `0.15.1` to `0.15.2` as the [`cargo.lock`](https://github.com/AntoniosBarotsis/cli/blob/ff090bb40e0b77bc36b04ecff6d5d3d36df7402a/Cargo.lock#L566) file used the latter anyways.

Tests are passing locally.

### DIFF
--- a/lockfile/src/javascript.rs
+++ b/lockfile/src/javascript.rs
@@ -273,7 +273,7 @@ mod tests {
             name: "@yarnpkg/lockfile".to_string(),
             version: "1.1.0".to_string(),
             package_type: PackageType::Npm,
-        }]);
+        }])
     }
 
     #[test]
--- a/lockfile/src/javascript.rs
+++ b/lockfile/src/javascript.rs
@@ -273,7 +273,7 @@ mod tests {
             name: "@yarnpkg/lockfile".to_string(),
             version: "1.1.0".to_string(),
             package_type: PackageType::Npm,
-        }])
+        }]);
     }
 
     #[test]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,15 +106,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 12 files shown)


cd-work (approved):
Looks good, thanks for the PR.

Generally I'd suggest not importing external functions directly and use `console::style` rather than importing it. But considering how verbose it already is to do `style("text").color()` I think it's probably best to keep it the way it is right now.

cd-work:
Please do fix the remaining CI formatting issues and rebase on top of master so CI is all green and this PR can be merged.

lockfile/src/javascript.rs
```diff
@@ -273,7 +273,7 @@ mod tests {
             name: "@yarnpkg/lockfile".to_string(),
             version: "1.1.0".to_string(),
             package_type: PackageType::Npm,
-        }]);
```
cd-work, line 276: Did you remove this semicolon? I don't think `rustfmt` should do this.

AntoniosBarotsis: I was fixing them manually but both `cargo +nightly fmt --all -- --check` and `cargo build` seem to be fine with it, I'll add it back just in case.

AntoniosBarotsis:
Why is the pipeline just an `exit 1` 0_0

cd-work:
> Why is the pipeline just an `exit 1` 0_0

Currently the pipeline relies on GitHub secrets, so it will not succeed when running on PRs sent from forks. This has nothing to do with your code so I'll just have to fixup CI and this should be good to go.

cd-work (approved):

PR #716, opened 2022-10-04
Skip end-to-end test for external PRs

kylewillmon:
Github Actions don't allow access to secrets when triggered by a `pull_request` event from a fork. This PR adds a condition to skip this test so that it does not fail or hang (like we have seen in #714)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,7 +65,9 @@ jobs:
           command: test
           args: --locked
 
-      - name: All Features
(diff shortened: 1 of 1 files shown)


kylewillmon:
Looks like it works. Here are the test runs:

External PR (test is skipped): https://github.com/phylum-dev/cli/actions/runs/3185063428/jobs/5194190647

Internal PR (test is run): https://github.com/phylum-dev/cli/actions/runs/3185036821/jobs/5194132295

PR #717, opened 2022-10-04
[EXTERNAL] Skip end-to-end test for external PRs

kylewillmon:
An external PR to test #716

This PR should be ignored in favor of #716


kylewillmon:
This test workflow already runs on every push to `main` and on a weekly schedule. Plus, we can manually run this workflow if we are ever concerned about anything specific.

PR #720, opened 2022-10-05
golang support

### LINKED ISSUE #718: Add support for parsing go.sum files
# Overview
Add support for parsing Go `go.sum` files.

The `go.sum` lists the checksum of direct and indirect dependency required along with the version. 

# Acceptance Criteria
- [ ] Update lockfile parsers to extract `Go` package names and versions
- [ ] Update command line hints and documentation

## Associated User Story
https://github.com/phylum-dev/roadmap/issues/122

ein-tier:
This implements support for golang and parsing go.sum files in the cli. I've attempted to find all the places where things need to be updated, but I'm very new to the codebase so let me know if I've missed anything!

NOTE: For testing purposes this is using an unmerged phylum-types branch – it needs to be switched back before merging

## Checklist

- [x] Switch back to main phylum-types branch

## Issue
closes #718


Cargo.lock
```diff
@@ -2912,9 +2912,10 @@ dependencies = [
 [[package]]
 name = "phylum_types"
 version = "0.1.0"
-source = "git+https://github.com/phylum-dev/phylum-types?branch=development#bb39c21c3d6f42f83219ff24bbe3ce3bdf4b25a1"
+source = "git+https://github.com/phylum-dev/phylum-types?branch=golang-support#0e6d0baa9e432e9f71710d9c047a1a74c6fa7cd1"
```
kylewillmon, line 2915: As mentioned in the PR notes, this needs to be fixed before merging. I'm just marking it here so that it doesn't slip through the cracks.

kylewillmon (changes requested):

lockfile/src/lib.rs
```diff
@@ -185,6 +193,7 @@ mod tests {
             ("gradle", LockfileFormat::Gradle),
             ("nuget", LockfileFormat::Msbuild),
             ("msbuild", LockfileFormat::Msbuild),
+            ("gosum", LockfileFormat::GoSum),
```
cd-work, line 196: Do Go developers actually refer to this as "gosum" or is the "sum" here basically just like the ".lock" suffix in other ecosystems?

If it's just the standard lockfile format for go, we might want to consider just naming it "go" so users can more intuitively pick their desired format.

lockfile/src/golang.rs
```diff
@@ -0,0 +1,54 @@
+use std::ffi::OsStr;
+use std::path::Path;
+
+use anyhow::{anyhow, Context};
+use nom::error::convert_error;
+use nom::Finish;
+use phylum_types::types::package::PackageType;
+
+use super::parsers::go_sum;
```
cd-work, line 9:
```suggestion
use crate::parsers::go_sum;
```

I'd assume you've just copied this from another parser, but there's really no need for this confusing indirection when it's literally just in the crate root anyway.

lockfile/src/parsers/go_sum.rs
```diff
@@ -0,0 +1,51 @@
+use phylum_types::types::package::{PackageDescriptor, PackageType};
+
+use super::*;
```
cd-work, line 3: I really don't like this habit in our parsers and it has cost me some completely unnecessary debugging time in the past. So I'd recommend to import everything directly in this file for the nom stuff we need since it's a new file anyway.

It's not like we need to bundle imports in one location to reduce file size, this is just unnecessary indirection that makes troubleshooting and adding additional imports more complicated.

lockfile/src/parsers/go_sum.rs
```diff
@@ -0,0 +1,51 @@
...
+fn package(input: &str) -> Option<PackageDescriptor> {
+    let (input, name) = package_name(input).ok()?;
+    let (input, version) = package_version(input).ok()?;
+    let (_, _hash) = package_hash(input).ok()?;
+
+    // If the package version ends in "/go.mod" then this entry
+    // just records the hash for a go.mod file. The package this
+    // go.mod file belongs to will also be listed, and that's what
+    // we're interested in, so we simply discard this entry.
+    if version.ends_with(r"/go.mod") {
```
cd-work, line 19: Could use `.then(|| ...)` if you want to be fancy.

cd-work (changes requested):

lockfile/src/lib.rs
ein-tier: I went back and forth on this myself, I'm open to change. I picked `GoSum` because:

- Go has had a number of different dependency management systems over the years. I think it is likely this latest one will remain the defacto standard, but there has been enough change that it seems prudent to be specific with the name (and who knows, maybe we will end up supporting one of the older systems also)
- Similar to other languages/ecosystems, there's a package file `go.mod` and lock file `go.sum`. Some of the other languages here parse things that are more like a package file than a lock file, and I when we were discussing go there was some debate about also parsing `go.mod` files, so I think the name should leave that option open.

ein-tier: (and fwiw I think that go developers do say "go sum file", but I'm not deeply into the community so I don't really know)

lockfile/src/parsers/go_sum.rs
ein-tier: Makes sense, done

lockfile/src/parsers/go_sum.rs
ein-tier: love to be fancy

lockfile/src/lib.rs
cd-work: Fair enough, having had other systems might be reason enough to be more explicit.

Though the difference between `.mod` and `.sum` isn't really one that matters considering the same could be said about `packge-lock.json`, about yarn's config, about gradle, ... . Pretty much every package manager that has a lockfile also has an "unlocked" manifest. We're more pointing out the package management system here than pointing out the lockfile specifically.

cd-work (approved):
> Does anyone know if there is more documentation that needs to be updated? I updated the help strings in `cli/src/app.rs` and found a couple things to update in the docs directory, but did I miss anything?

Based on my memory alone, I think you should be good. But if you want to be super safe you could probably also take a look at other somewhat recently added ecosystems like poetry. Should be simple enough to find in the git log and include doc updates.

lockfile/src/lib.rs
```diff
@@ -38,6 +40,8 @@ pub enum LockfileFormat {
     #[serde(rename = "nuget")]
     #[serde(alias = "msbuild")]
     Msbuild,
+    #[serde(rename = "gosum")]
+    GoSum,
```
kylewillmon, line 44: 🤔 Maybe it would be better to just change the `rename_all` on this enum to use `lowercase` instead of `camelCase`?

All of the other variants are one "word" (or at least they only have one capital letter), so this is the first variant where the distinction between `camelCase` and `lowercase` actually matters. If we want this one to be all lowercase, we might as well make that the decision going forward.

lockfile/src/parsers/go_sum.rs
```diff
@@ -0,0 +1,53 @@
...
+use nom::character::complete::space0;
+use nom::combinator::recognize;
+use phylum_types::types::package::{PackageDescriptor, PackageType};
+
+use super::Result;
+
+pub fn parse(input: &str) -> Result<&str, Vec<PackageDescriptor>> {
+    let pkgs = input.lines().filter_map(package).collect::<Vec<_>>();
+    Ok((input, pkgs))
+}
```
kylewillmon, line 13: This function never returns an error...

We've had problems in the past with other parsers being too permissive and handling files that they shouldn't. I definitely think that will be a problem here.

lockfile/src/parsers/go_sum.rs
```diff
@@ -0,0 +1,53 @@
...
+    version.ends_with(r"/go.mod").not().then(|| PackageDescriptor {
+        name: name.to_string(),
+        version: version.to_string(),
+        package_type: PackageType::Golang,
+    })
+}
+
+fn package_name(input: &str) -> Result<&str, &str> {
+    // take away any leading whitespace
+    let (input, _) = recognize(space0)(input)?;
```
kylewillmon, line 34: I don't think the `recognize` call here does anything

```suggestion
    let (input, _) = space0(input)?;
```

kylewillmon (changes requested):

lockfile/src/parsers/go_sum.rs
cd-work: We need to be especially careful that no other lockfile accidentally produces go packages, since our `auto` parser would then get confused.

lockfile/src/lib.rs
ein-tier: good call, done

lockfile/src/lib.rs
kylewillmon: > Though the difference between .mod and .sum isn't really one that matters considering the same could be said about packge-lock.json, about yarn's config, about gradle,

I agree with this. And the name should probably just be `go` for now.

This may get complicated in the future if we allow submission of either `go.mod` or `go.sum` files (See phylum-dev/roadmap#173). But we're going to have to deal with that across the board, so it's not a reason to deviate from normal now.

cd-work: Are lockfiles not standard for Go the way they are for Rust for example? Generally I'd avoid accepting non-lock files wherever possible. Nobody should **ever** submit a `Cargo.toml` for analysis.

lockfile/src/parsers/go_sum.rs
ein-tier: I reworked this to error when encountering unexpected input – now if it fails to find lines with three strings separated by spaces then it will fail. Hopefully that will give us more actionable insight vs before when it would have just returned no descriptors

lockfile/src/lib.rs
ein-tier: Fair enough, I changed to `go`!

ein-tier:
> Doesn't look like things will get merged today, but if @kylewillmon stays on as reviewer then I'm happy to take over and carry it over the finish line. No need to have it sit around for weeks and decay.

🙏 Thank you!!!

lockfile/src/parsers/go_sum.rs
```diff
@@ -0,0 +1,57 @@
...
+use phylum_types::types::package::{PackageDescriptor, PackageType};
+
+use super::{take_till_line_end, Result};
+
+pub fn parse(input: &str) -> Result<&str, Vec<PackageDescriptor>> {
+    let (input, pkg_options) = many1(package)(input)?;
+    let pkgs = pkg_options.iter().flatten().cloned().collect::<Vec<_>>();
+
+    Ok((input, pkgs))
+}
```
kylewillmon, line 16: This is still a very permissive parser. It accepts any file with at least 3 words in the first line. In fact, it successfully parses every file in `tests/fixtures` (the `pip` and `gradle` parsers also accomplish this... But that's a separate issue).

A few ways that it could be made more strict:

1. Check for EOF at the end of the parser
2. Check the format of the hash (i.e., `preceded(tag("h1"), hex_digit1)`)
3. Check for extra data at the end of each line

kylewillmon (approved):

PR #721, opened 2022-10-05
Update from clap 3 to 4

cd-work:


cd-work:
Worth noting that there's been a big change to clap's output again, removing the colorful output in favor of a more reduced palette:

tmp

Having to implement a custom parser for our lockfile types also means error messages do not match the rest of clap:

tmp

I'd assume it's probably possible to go back to colored if we want to reduce the fatigue on users with these constant clap changes, but I'm not sure we want that since obviously the colors have been removed for a reason.

cli/src/app.rs
```diff
@@ -210,7 +211,7 @@ pub fn app<'a>() -> clap::Command<'a> {
                     .long("lockfile-type")
                     .value_name("type")
                     .help("The type of the lock file (default: auto)")
-                    .possible_values(parse::lockfile_types()),
+                    .value_parser(LockfileValueParser),
```
kylewillmon, line 214: Is there a reason to implement our own `ValueParser` rather than simply using [`PossibleValuesParser`](https://docs.rs/clap/4.0.9/clap/builder/struct.PossibleValuesParser.html)?

cd-work: Oh wow, so if you pass a `Vec` to `value_parser` it doesn't work, but if you wrap it in `PossibleValuesParser` it does. So `value_parser` only accepts owned slices.

cli/src/commands/parse.rs
```diff
@@ -8,19 +8,20 @@ use anyhow::{anyhow, Result};
...
 pub fn handle_parse(matches: &clap::ArgMatches) -> CommandResult {
-    let lockfile_type = matches.value_of("lockfile-type").unwrap_or("auto");
+    let auto_type = String::from("auto");
+    let lockfile_type = matches.get_one::<String>("lockfile-type").unwrap_or(&auto_type);
     // LOCKFILE is a required parameter, so .unwrap() should be safe.
-    let lockfile = matches.value_of("LOCKFILE").unwrap();
+    let lockfile = matches.get_one::<String>("LOCKFILE").unwrap();
 
-    let pkgs = if lockfile_type == "auto" {
+    let pkgs = if lockfile_type == &auto_type {
```
andreaphylum, line 24: I think the `let auto_type = ...` binding is a bit confusing.
Another possible solution could be to have `.unwrap_or_else(|| "auto".to_string())` and, below, leave the `lockfile_type == "auto"` comparison. No strong opinion on this, the current solution is fine too.

andreaphylum (dismissed):

kylewillmon (dismissed):

cli/src/commands/jobs.rs
```diff
@@ -122,25 +123,27 @@ pub async fn handle_submission(api: &mut PhylumApi, matches: &clap::ArgMatches)
...
 
-        label = matches.value_of("label");
-        verbose = matches.is_present("verbose");
-        pretty_print = !matches.is_present("json");
-        display_filter = matches.value_of("filter").and_then(|v| Filter::from_str(v).ok());
-        is_user = !matches.is_present("force");
+        label = matches.get_one::<String>("label");
+        verbose = matches.get_flag("verbose");
+        pretty_print = !matches.get_flag("json");
+        display_filter = matches.get_one::<String>("filter").and_then(|v| Filter::from_str(v).ok());
```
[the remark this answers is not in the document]
cd-work: I'd rather just keep it simple tbh.

cli/src/commands/project.rs
```diff
@@ -59,35 +59,35 @@ pub async fn handle_project(api: &mut PhylumApi, matches: &clap::ArgMatches) ->
...
         print_user_success!("Successfully created new project, {}", name);
     } else if let Some(matches) = matches.subcommand_matches("delete") {
-        let project_name = matches.value_of("name").unwrap();
-        let group_name = matches.value_of("group");
+        let project_name = matches.get_one::<String>("name").unwrap();
+        let group_name = matches.get_one::<String>("group");
 
         let proj_uuid = api
-            .get_project_id(project_name, group_name)
+            .get_project_id(project_name, group_name.map(|g| &**g))
```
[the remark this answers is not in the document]
cd-work: A little bit longer, but I guess it's also a bit easier to understand for readers.

kylewillmon (approved):

PR #724, opened 2022-10-06
Sign archives with openssl

kylewillmon:
Putting this up as a draft PR to start discussion around the idea.


signing-key.pub
```diff
@@ -0,0 +1,9 @@
+-----BEGIN PUBLIC KEY-----
+MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAyGgvuy6CWSgJuhKY8oVz
+42udH1F2yIlaBoxAdQFuY2zxPSSpK9zv34B7m0JekuC5WCYfW0gS2Z8Ryu2RVdQh
+7DXvQb7qwzZT0H11K9Pw8hIHBvZPM+d61GWgWDc3k/rFwMmqd+kytVZy0mVxNdv4
+P2qvy6BNaiUI7yoB1ahR/6klfkPit0X7pkK9sTHwW+/WcYitTQKnEnRzA3q8EmA7
+rbU/sFEypzBA3C3qNJZyKSwy47kWXhC4xXUS2NXvew4FoVU6ybMoeDApwsx1AgTu
+CPPnPlCwuCIyUPezCP5XYczuHfaWeuwArlwdJFSUpMTc+SqO6REKgL9yvpqsO5Ia
+sQIDAQAB
+-----END PUBLIC KEY-----
```
kylewillmon, line 9: Really wondering if we should just host this somewhere instead?

Cc: @brent-phylum in case he's got a good idea of where we could store this so that people can download it.

.github/workflows/release.yml
```diff
@@ -151,13 +151,16 @@ jobs:
       - name: Sign archives
         run: |
           echo -e $MINISIGN_KEY > minisign.key
+          echo "$OPENSSL_KEY" > key.pem
           for archive in phylum-*.zip;
           do
             echo $MINISIGN_PASSWORD | minisign -Sm ${archive} -s minisign.key -t 'Phylum - The Software Supply Chain Company'
```
kylewillmon, line 157: This will be phased out over time. If this PR is accepted, I will create an issue to track the eventual removal of `minisign` signatures from our release assets.

README.md
```diff
@@ -21,17 +21,16 @@ The command line interface (CLI) allows users to submit their project package de
...
-   | aarch64-apple-darwin | [phylum-aarch64-apple-darwin.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-apple-darwin.zip) <br /> [phylum-aarch64-apple-darwin.zip.minisig](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-apple-darwin.zip.minisig) |
+   | x86_64-unknown-linux-musl | [phylum-x86_64-unknown-linux-musl.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-x86_64-unknown-linux-musl.zip) <br /> [phylum-x86_64-unknown-linux-musl.zip.signature](https://github.com/phylum-dev/cli/releases/latest/download/phylum-x86_64-unknown-linux-musl.zip.signature) |
+   | aarch64-unknown-linux-musl | [phylum-aarch64-unknown-linux-musl.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-unknown-linux-musl.zip) <br /> [phylum-aarch64-unknown-linux-musl.zip.signature](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-unknown-linux-musl.zip.signature) |
```
andreaphylum, line 25: Why the `-musl` suffix changes? Shouldn't we be phasing those out?

docs/command_line_tool/quickstart.md
```diff
@@ -23,17 +23,16 @@ The command line interface (CLI) allows users to submit their project package de
...
-   | aarch64-apple-darwin | [phylum-aarch64-apple-darwin.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-apple-darwin.zip) <br /> [phylum-aarch64-apple-darwin.zip.minisig](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-apple-darwin.zip.minisig) |
+   | x86_64-unknown-linux-musl | [phylum-x86_64-unknown-linux-musl.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-x86_64-unknown-linux-musl.zip) <br /> [phylum-x86_64-unknown-linux-musl.zip.signature](https://github.com/phylum-dev/cli/releases/latest/download/phylum-x86_64-unknown-linux-musl.zip.signature) |
+   | aarch64-unknown-linux-musl | [phylum-aarch64-unknown-linux-musl.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-unknown-linux-musl.zip) <br /> [phylum-aarch64-unknown-linux-musl.zip.signature](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-unknown-linux-musl.zip.signature) |
```
andreaphylum, line 27: In the same way as the comment above, shouldn't we change these `-musl` suffixes to `-gnu`?

signing-key.pub
andreaphylum: I don't like the idea of adding files to the top level, but I also don't think there is a better spot at the moment. I'd be cool to have it hosted somewhere on phylum.io but I think it's also worth having it in VCS as well, just for redundancy. So I'd say let's go ahead and commit this.

andreaphylum (changes requested):
Other than the very minor thing about `musl`, looks good to me.

signing-key.pub
kylewillmon: We do have the [`assets`](https://github.com/phylum-dev/cli/tree/main/assets) folder, which I considered. Or we could put it in [`scripts`](https://github.com/phylum-dev/cli/tree/main/scripts) next to the script which uses it. I'm not sure I have a strong preference.

andreaphylum: `assets` feels a bit unrelated, given the context; `scripts` would be semantically a bit weird but it would also make sense for the file to be next to the script that's using it. I don't have a strong preference either, if I had to pick I'd probably go with `scripts`.

andreaphylum (approved):

maxrake (commented):
Changes reviewed and issues in the CLI repo related to `minisign` and artifact signatures were updated. A new issue was created in `phylum-dev/phylum-ci` to account for the new signature verification method in the `phylum-init` script entry point: phylum-dev/phylum-ci#141
(see referenced issues above ☝️ )

PR #731, opened 2022-10-07
Add CLI flags for log level control

### LINKED ISSUE #722: Verbose flag
We currently manage logging with the `RUST_LOG` environment variable, but a `-v` flag for verbosity might be more natural for most users.

~~We can consider using the [`clap-verbosity-flag`](https://crates.io/crates/clap-verbosity-flag) crate as a simple way to add this option.~~

cd-work:
This adds the new `-v` and `-q` CLI flags which allow controlling the log level for the `env_logger`.

The default log level is `warn`, logging both errors and warnings. The `-q` flag can be specified up to two times to disable logging completely while the `-v` flag can be specified up to three times to enable trace-level logging.

Closes #722.


cd-work:
I deviated from the original plan suggested in #722, since the default log level was at `warn` and not `error`. So it starts with `-qq` instead of just `-q`. I think this makes sense and has worked well for me in other applications.

cli/src/bin/phylum.rs
```diff
@@ -63,6 +63,19 @@ async fn handle_commands() -> CommandResult {
...
+        (_, 2..) => "off",
+        (_, 1) => "error",
+        (0, 0) => "warn",
+        (1, _) => "info",
+        (2, _) => "debug",
+        (3.., _) => "trace",
+    };
+
+    // Initialize logger.
+    env_logger::Builder::from_env(Env::default().default_filter_or(log_level)).init();
```
kylewillmon, line 77: I'm wondering about this (admittedly obscure) case:

```
export RUST_LOG=trace
phylum -qq ping
```

I think a user would expect logging to be turned off for that call. But the code here would use the environment variable instead of the flags.

cli/src/app.rs
```diff
@@ -47,6 +47,15 @@ pub fn app() -> Command {
...
                 .help("Don't validate the server certificate when performing api requests"),
+            Arg::new("verbose")
+                .short('v')
+                .help("Increase the level of verbosity (the maximum is -vvv)")
+                .action(ArgAction::Count),
+            Arg::new("quiet")
+                .short('q')
+                .help("Reduce the level of verbosity (the maximum is -qq)")
+                .action(ArgAction::Count)
+                .conflicts_with("verbose"),
```
kylewillmon, line 58: If there is any `Arg` that deserves to be marked as [`global`](https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.global), this is probably it.

Otherwise, I expect users will end up in this loop:
```
# Run a command
phylum analyze foo.txt

# Some weird error happened?!? Add the verbose flag and try again.
phylum analyze -vv foo.txt

# Whoops... Can't put it there
phylum -vv analyze foo.txt
```

kylewillmon (changes requested):

cli/src/bin/phylum.rs
cd-work: Hm I could make the CLI take precedence, sure.

kylewillmon (approved):

maxrake (approved):
Manually tested the changes. LGTM.

PR #734, opened 2022-10-10
Optimize test workflows without caching

kylewillmon:
Attempting to make the cache smaller in #733 was a failure. Maybe we can speed up the tests in other ways and remove caching altogether.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,10 +38,6 @@ jobs:
           profile: minimal
           components: rustfmt
 
-      - uses: Swatinem/rust-cache@v1
-        with:
-          key: v3
-
       - name: Format check
         if: github.event_name != 'schedule'
         run: cargo +nightly fmt --all -- --check
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -17,7 +17,14 @@ jobs:
...
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        platform: [ubuntu-latest, macos-latest, windows-latest]
+        include:
+          - platform: ubuntu-latest
+            os: ubuntu-latest
+          - platform: windows-latest
+            os: windows-latest
+          - platform: macos-latest
+            os: [self-hosted, macos, arm64]
```
cd-work, line 27: Just out of curiosity, do you know if it would be possible to just specify `os: [ubuntu-latest, windows-latest, [self-hosted, macos, arm64]]`?

cd-work (dismissed):

maxrake (approved):
Approving now...but not sure these changes improve the workflow times. Perhaps this is an "open" PR...in the sense that it will be used to trial new techniques and only be closed when/if one is found that does improve the test workflow times.

cd-work:
> not sure these changes improve the workflow times

They certainly improve reliability and remove a dependency on a third party action.

.github/workflows/test.yml
kylewillmon: Turns out [it is possible](https://github.com/phylum-dev/cli/actions/runs/3229981438)

cd-work (approved):

maxrake (approved):

PR #740, opened 2022-10-11
Improve extension subcommand conflict resolution

### LINKED ISSUE #735: Improve extension name conflict resolution
# Overview

Currently when an extension conflicts with a built-in command, a warning will be printed for commands like `extension list` that the extension's name is in conflict.

However, there's no warnings or errors when the extension is created, creating a new `ping` extension for example works without any issues at all.

Additionally, the extension also isn't filtered out of the `extension list` output, despite stating that it would be.

We should try harder for this scenario to be prevented in the first place, instead of just accepting it and then printing a warning later. I propose the following new behavior:
 - Error out when creating a new extension that conflicts with a built-in name
 - Error out when trying to install an extension that conflicts with a built-in name
 - Keep warning when a "manually" installed extension conflicts with a built-in name
 - Actually filter out extensions from the `extension list` output when they cannot be used

cd-work:
This patch attempts to be more proactive at catching potential conflicts between new extensions and existing subcommands, erroring out during creation and installation instead of waiting until usage.

Closes #735.


cli/src/app.rs
```diff
@@ -59,7 +59,31 @@ pub fn app() -> Command {
...
+        app = app.subcommand(
+            Command::new("uninstall").about("Uninstall the Phylum CLI").arg(
+                Arg::new("purge")
+                    .action(ArgAction::SetTrue)
+                    .short('p')
+                    .long("purge")
+                    .help("Remove all files, including configuration files (default: false)"),
+            ),
+        );
+    }
```
kylewillmon, line 79: Does this mean an extension named `uninstall` will be allowed?

cd-work: Hm I guess so. Should probably move this to the `add_subcommands` function.

maxrake (approved):
The changes appear good and were verified with a checkout and manual testing.

kylewillmon (approved):

PR #741, opened 2022-10-11
Fix shell completion file generation

kylewillmon:
The release workflow has been [failing](https://github.com/phylum-dev/cli/actions/runs/3228666298/jobs/5285052846) because of shell completion file generation errors.

This patch adds shell completion file generation to the test workflow. And also fixes the error.


cli/src/app.rs
```diff
@@ -184,7 +184,6 @@ pub fn app() -> Command {
             Command::new("package").about("Retrieve the details of a specific package").args(&[
                 Arg::new("name").value_name("name").help("The name of the package.").required(true),
                 Arg::new("version")
-                    .action(ArgAction::SetTrue)
```
cd-work, line 187: This is why manual repetitive tasks are always terrible ideas as a migration strategy, probably thought this was the `-v` flag.

cd-work (approved):

maxrake (approved):

PR #743, opened 2022-10-12
Add support for parsing cargo lockfiles

### LINKED ISSUE #736: Add support for parsing Cargo.lock files
# Overview
Update lockfile parsing to support Rust `Cargo.lock` files.

The `Cargo.lock` files lists the direct and indirect dependencies. 

# Acceptance Criteria
- [ ] Update lockfile parsers to extract `Cargo` package names and versions
- [ ] Update command line hints and documentation

JosephPhylum:
## Checklist
This implements support for rust and parsing cargo.lock files in the cli.

closes #736 


docs/command_line_tool/phylum_parse.md
```diff
@@ -10,7 +10,7 @@ phylum parse [OPTIONS] <LOCKFILE>
 
 ### Options
 `-t`, `--lockfile-type`
-&emsp; The type of the lockfile (default: `auto`): `yarn`, `npm`, `gem`, `pip`, `pipenv`, `poetry`, `mvn`, `gradle`, `nuget`, `auto`
+&emsp; The type of the lockfile (default: `auto`): `yarn`, `npm`, `gem`, `pip`, `pipenv`, `poetry`, `mvn`, `gradle`, `nuget`, `rust`, `auto`
```
andreaphylum: Is `rust` intended, or should it be `cargo`?

docs/knowledge_base/analyzing-dependencies.md
```diff
@@ -20,6 +20,8 @@ The Phylum CLI natively supports processing the lock/requirements files for seve
 * Maven
     * `effective-pom.xml`
     * `gradle.lockfile`
+* Cargo
+    * `cargo.lock`
```
andreaphylum:
```suggestion
    * `Cargo.lock`
```

lockfile/src/cargo.rs
```diff
@@ -0,0 +1,164 @@
...
+            },
+        ];
+
+        for expected_pkg in expected_pkgs {
+            assert!(pkgs.contains(&expected_pkg));
+        }
+    }
+
+    #[test]
+    fn parse_cargo_lockv1() {
```
andreaphylum, line 102: This is just a very minor nitpick, but I'd separate the `v1` here with an underscore (and below, in the fixture names, with a dash):
```suggestion
    fn parse_cargo_lock_v1() {
```

andreaphylum (changes requested):
Looks good overall, a couple minor details.

docs/knowledge_base/analyzing-dependencies.md
JosephPhylum: updated

lockfile/src/cargo.rs
JosephPhylum: Thank you, it looks much cleaner with the underscore.

tests/fixtures/Cargo.lock
```diff
@@ -0,0 +1,5585 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+version = 3
```
cd-work, line 3: Why do we have this plus a version 3 lockfile? Shouldn't one of the two be sufficient?

lockfile/src/cargo.rs
```diff
@@ -0,0 +1,164 @@
...
+    #[serde(rename = "package")]
+    packages: Vec<Package>,
+}
+
+#[derive(Deserialize, Debug, Clone)]
+struct Package {
+    name: String,
+    version: String,
+    source: Option<String>,
+}
```
cd-work, line 22: Could we rely on Cargo's internal structs rather than making our own? Like https://docs.rs/cargo/0.65.0/cargo/core/resolver/struct.EncodableResolve.html?

lockfile/src/cargo.rs
```diff
@@ -0,0 +1,164 @@
...
+
+impl Parse for Cargo {
+    /// Parses `cargo.lock` files into a vec of packages
```
cd-work, line 25:
```suggestion
    /// Parses `Cargo.lock` files into a vec of packages.
```

lockfile/src/cargo.rs
```diff
@@ -0,0 +1,164 @@
...
+            .packages
+            .drain(..)
+            .filter_map(|package| {
+                package.source.as_ref().map(|_| PackageDescriptor::from(package.clone()))
+            })
+            .collect())
+    }
+
+    fn package_type(&self) -> PackageType {
+        println!("declared package type");
```
cd-work, line 38: I'd assume this is a leftover from debugging?

lockfile/src/cargo.rs
```diff
@@ -0,0 +1,164 @@
...
+
+impl Parse for Cargo {
+    /// Parses `cargo.lock` files into a vec of packages
+    fn parse(&self, data: &str) -> ParseResult {
+        let mut lock: CargoLock = toml::from_str(data)?;
+        Ok(lock
+            .packages
+            .drain(..)
+            .filter_map(|package| {
+                package.source.as_ref().map(|_| PackageDescriptor::from(package.clone()))
```
cd-work, line 32: This looks rather wonky, especially with the `clone`. Seems like we just want `try_from` maybe?

lockfile/src/cargo.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn is_path_lockfile(&self, path: &Path) -> bool {
+        path.file_name() == Some(OsStr::new("Cargo.lock"))
+    }
+}
+
+impl From<Package> for PackageDescriptor {
+    fn from(package: Package) -> Self {
+        let source = package.source.unwrap_or_default();
+        let version = if source.starts_with("git+") {
+            source.trim_start_matches("git+").into()
```
cd-work, line 51: I don't think this is what we want. It would trim `git+git+git+git+` for example?

Looks like what you want is `if let Some(git_version) = source.strip_prefix("git+")`?

cd-work (changes requested):

tests/fixtures/Cargo.lock
JosephPhylum: we needed an example of using the URL as the version, but I can move the longer lockfile into the v3 spot.

lockfile/src/cargo.rs
JosephPhylum: Fixed.

lockfile/src/cargo.rs
JosephPhylum: Removed.

lockfile/src/cargo.rs
JosephPhylum: Fixed.

lockfile/src/cargo.rs
JosephPhylum: Fixed.

lockfile/src/cargo.rs
JosephPhylum: I don't see how to use EncodableResolve, the fields are there but they're all private. If you have suggestions I'm up for implementing them.

tests/fixtures/Cargo_v2.lock
```diff
@@ -0,0 +1,220 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
```
kylewillmon, line 2: Do we know which rust versions correspond to the different `Cargo.lock` file versions?

I'd like to get an idea of how important it is to support the old versions.

JosephPhylum: I don't see which cargo versions associate with which lockfile version, but the change from version 2 to version 3 was February of 2021.

kylewillmon: So that's rust 1.50 ... Probably old enough that we don't really have to worry about it.

No change required here, just wanted to know. Thanks

andreaphylum (dismissed):

lockfile/src/cargo.rs
cd-work: You could deserialize and convert it to a `Resolve`. But it might be easier to have our own version. I was just curious if you've looked into this at all.

cd-work (changes requested):
Super minor style thing, other than that things are looking good.

lockfile/src/cargo.rs
JosephPhylum: We looked into it, and it seems like more trouble than it would be worth. We would need to update the parse() function to include a path to the cargo.toml file that we want to parse, whereas we're currently passing in the contents of the lockfile as a string. We would also need to link against the Cargo crate, which we don't do yet.

cd-work (approved):
LGTM

PR #744, opened 2022-10-12
Use RSA signature in phylum update

kylewillmon:


maxrake (approved):
Looks good...but did not verify. I suppose a release candidate can be made to test the update feature with these changes...might be worthwhile before the next CLI release.

PR #745, opened 2022-10-13
Add CI compile-test for MSRV

### LINKED ISSUE #683: Add test for MSRV
Currently our `Cargo.toml` specifies a MSRV, but we have no tests making sure this is actually correct and working. Since it's very easy for dependencies to break this guarantee, we should add CI to ensure our MSRV builds correctly.

cd-work:
Closes #683.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,7 +69,7 @@ jobs:
           args: --locked
 
       - name: Oldstable
-        if: matrix.os != windows-latest
+        if: matrix.os != 'windows-latest'
         run: |
           oldstable=$(cat "./cli/Cargo.toml" | grep "rust-version" | sed 's/.*"\(.*\)".*/\1/')
           rustup toolchain install --profile minimal "$oldstable"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,7 +69,7 @@ jobs:
           args: --locked
 
       - name: Oldstable
-        if: matrix.os == ubuntu-latest
+        if: matrix.os != windows-latest
         run: |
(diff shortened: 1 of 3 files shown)


andreaphylum (approved):

lockfile/Cargo.toml
```diff
@@ -4,7 +4,7 @@ description = "Package lock file parsers used by Phylum"
 version = "0.1.0"
 authors = ["Phylum, Inc. <engineering@phylum.io>"]
 edition = "2021"
-rust-version = "1.60"
+rust-version = "1.62.1"
```
kylewillmon, line 7: I'm not sure how we should handle this... I think the `lockfile` crate would build just fine with 1.61 (or maybe even 1.60).

Since this is a library, we are restricting our users by putting a MSRV that is higher than necessary here. Maybe this crate shouldn't have an MSRV?

cd-work: I don't think not providing a MSRV guarantee at all is even worse tbh.

Initially I thought about separating the two MSRVs with two separate CI tests, I can go back to that if you'd prefer it? I just thought it would be unnecessarily complicated to do so.

kylewillmon: I agree that two separate tests is unnecessarily complicated.

I guess keeping it the same as the `phylum-cli` crate is fine for now

.github/workflows/test.yml
```diff
@@ -61,6 +61,13 @@ jobs:
           command: test
           args: --locked
 
+      - name: Oldstable
+        if: matrix.os != windows-latest
```
kylewillmon, line 65:
```suggestion
        if: matrix.os != 'windows-latest'
```

kylewillmon (approved):

PR #746, opened 2022-10-13
Layered sandbox permissions

### LINKED ISSUE #692: Sandbox layering and improvements
# Overview
Currently, there are two independent sandboxes: one enforced by Deno which pertains to its calls, and one enforced by the OS.
The latter is intended for sandboxing invoked commands. We need to integrate the two sandboxes, making sure the OS sandbox doesn't have more privileges than the Deno sandbox.

# Acceptance Criteria
- [ ] `Deno.run` should not be accessible from an extension; the user should use `PhylumApi.runSandboxed` instead (note: check deno.rs:40).
- [ ] If permissions specified in `runSandboxed` are not a subset of the ones in `PhylumExt.toml`, an error should be emitted and the operation should not be allowed.

andreaphylum:
Closes #692.

## Checklist

### DIFF
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -91,6 +91,18 @@ struct ProcessException {
     strict: bool,
 }
 
+impl From<ProcessException> for permissions::Permissions {
+    fn from(process_exception: ProcessException) -> Self {
+        Self {
+            read: process_exception.read,
+            write: process_exception.write,
+            run: process_exception.run,
+            env: Permission::Boolean(true),
+            net: Permission::Boolean(process_exception.net),
+        }
+    }
+}
+
 /// Standard I/O behavior.
 #[derive(Serialize, Deserialize, Debug)]
 enum ProcessStdio {
@@ -380,13 +392,21 @@ async fn parse_lockfile(
 /// possible even after the command has been spawned.
 #[op]
 #[cfg(unix)]
-fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
+fn run_sandboxed(op_state: Rc<RefCell<OpState>>, process: Process) -> Result<ProcessOutput> {
+    let Process { cmd, args, stdin, stdout, stderr, exceptions } = process;
+    let exceptions_strict = exceptions.strict;
+    let real_permissions = {
+        let mut state = op_state.borrow_mut();
+        let permissions = state.borrow_mut::<permissions::Permissions>();
+        permissions::Permissions::from(exceptions).subset_of(permissions)
+    }?;
+
     // Setup process to be run.
-    let mut command = Command::new(&process.cmd);
-    command.args(&process.args);
-    command.stdin(process.stdin);
-    command.stdout(process.stdout);
-    command.stderr(process.stderr);
+    let mut command = Command::new(&cmd);
+    command.args(&args);
+    command.stdin(stdin);
+    command.stdout(stdout);
+    command.stderr(stderr);
 
     // Apply sandbox to subprocess after fork.
     unsafe {
@@ -397,29 +417,29 @@ fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
 
             let home_dir = dirs::home_dir().map_err(into_ioerr)?;
 
-            let mut birdcage = if process.exceptions.strict {
+            let mut birdcage = if exceptions_strict {
                 Birdcage::new().map_err(into_ioerr)?
             } else {
                 permissions::default_sandbox().map_err(into_ioerr)?
             };
 
-            for path in process.exceptions.read.sandbox_paths().iter() {
+            for path in real_permissions.read.sandbox_paths().iter() {
                 let path = dirs::expand_home_path(path, &home_dir);
                 permissions::add_exception(&mut birdcage, Exception::Read(path))
                     .map_err(into_ioerr)?;
             }
-            for path in process.exceptions.write.sandbox_paths().iter() {
+            for path in real_permissions.write.sandbox_paths().iter() {
                 let path = dirs::expand_home_path(path, &home_dir);
                 permissions::add_exception(&mut birdcage, Exception::Write(path))
                     .map_err(into_ioerr)?;
             }
-            for path in process.exceptions.run.sandbox_paths().iter() {
+            for path in real_permissions.run.sandbox_paths().iter() {
                 let path = dirs::expand_home_path(path, &home_dir);
                 let absolute_path = permissions::resolve_bin_path(path);
                 permissions::add_exception(&mut birdcage, Exception::ExecuteAndRead(absolute_path))
                     .map_err(into_ioerr)?;
             }
-            if process.exceptions.net {
+            if let permissions::Permission::Boolean(true) = real_permissions.net {
                 birdcage.add_exception(Exception::Networking).map_err(into_ioerr)?;
             }
 
@@ -429,8 +449,7 @@ fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
     }
 
     let output = command.output().with_context(|| {
-        let cmd = process.cmd;
-        let args = process.args.iter().map(|arg| format!("`{arg}`")).collect::<Vec<_>>().join(" ");
+        let args = args.iter().map(|arg| format!("`{arg}`")).collect::<Vec<_>>().join(" ");
         format!("Executing sandboxed process failed: `{cmd}` {args}",)
     })?;
 
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -32,8 +32,7 @@ use tokio::fs;
 use crate::api::{PhylumApiError, ResponseError};
 use crate::auth::UserInfo;
 #[cfg(unix)]
-use crate::commands::extensions::permissions;
-use crate::commands::extensions::permissions::Permission;
+use crate::commands::extensions::permissions::{self, Permission};
 use crate::commands::extensions::state::ExtensionState;
 use crate::commands::parse;
 use crate::config::{self, ProjectConfig};
@@ -91,6 +90,7 @@ struct ProcessException {
     strict: bool,
 }
 
+#[cfg(unix)]
 impl From<ProcessException> for permissions::Permissions {
     fn from(process_exception: ProcessException) -> Self {
         Self {
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -32,7 +32,8 @@ use tokio::fs;
 use crate::api::{PhylumApiError, ResponseError};
 use crate::auth::UserInfo;
 #[cfg(unix)]
-use crate::commands::extensions::permissions::{self, Permission};
+use crate::commands::extensions::permissions;
+use crate::commands::extensions::permissions::Permission;
 use crate::commands::extensions::state::ExtensionState;
 use crate::commands::parse;
 use crate::config::{self, ProjectConfig};
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -98,7 +98,7 @@ impl From<ProcessException> for permissions::Permissions {
             read: process_exception.read,
             write: process_exception.write,
             run: process_exception.run,
-            env: Permission::Boolean(false), // TODO: replace with inherit
+            env: Permission::Boolean(true),
             net: Permission::Boolean(process_exception.net),
         }
     }
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -443,6 +443,7 @@ fn run_sandboxed(op_state: Rc<RefCell<OpState>>, process: Process) -> Result<Pro
             if let permissions::Permission::Boolean(true) = resolved_permissions.net {
                 birdcage.add_exception(Exception::Networking).map_err(into_ioerr)?;
             }
+            birdcage.add_exception(Exception::FullEnvironment).map_err(into_ioerr)?;
 
             birdcage.lock().map_err(into_ioerr)?;
             Ok(())
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -98,7 +98,7 @@ impl From<ProcessException> for permissions::Permissions {
             read: process_exception.read,
             write: process_exception.write,
             run: process_exception.run,
-            env: Permission::Boolean(true),
+            env: Permission::Boolean(false), // TODO: replace with inherit
             net: Permission::Boolean(process_exception.net),
         }
     }
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -98,7 +98,7 @@ impl From<ProcessException> for permissions::Permissions {
             read: process_exception.read,
             write: process_exception.write,
             run: process_exception.run,
-            env: Permission::Boolean(true),
+            env: Permission::Boolean(false),
             net: Permission::Boolean(process_exception.net),
         }
     }
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -395,7 +395,7 @@ async fn parse_lockfile(
 fn run_sandboxed(op_state: Rc<RefCell<OpState>>, process: Process) -> Result<ProcessOutput> {
     let Process { cmd, args, stdin, stdout, stderr, exceptions } = process;
     let exceptions_strict = exceptions.strict;
-    let real_permissions = {
+    let resolved_permissions = {
         let mut state = op_state.borrow_mut();
         let permissions = state.borrow_mut::<permissions::Permissions>();
         permissions::Permissions::from(exceptions).subset_of(permissions)
@@ -423,23 +423,23 @@ fn run_sandboxed(op_state: Rc<RefCell<OpState>>, process: Process) -> Result<Pro
                 permissions::default_sandbox().map_err(into_ioerr)?
             };
 
-            for path in real_permissions.read.sandbox_paths().iter() {
+            for path in resolved_permissions.read.sandbox_paths().iter() {
                 let path = dirs::expand_home_path(path, &home_dir);
                 permissions::add_exception(&mut birdcage, Exception::Read(path))
                     .map_err(into_ioerr)?;
             }
-            for path in real_permissions.write.sandbox_paths().iter() {
+            for path in resolved_permissions.write.sandbox_paths().iter() {
                 let path = dirs::expand_home_path(path, &home_dir);
                 permissions::add_exception(&mut birdcage, Exception::Write(path))
                     .map_err(into_ioerr)?;
             }
-            for path in real_permissions.run.sandbox_paths().iter() {
+            for path in resolved_permissions.run.sandbox_paths().iter() {
                 let path = dirs::expand_home_path(path, &home_dir);
                 let absolute_path = permissions::resolve_bin_path(path);
                 permissions::add_exception(&mut birdcage, Exception::ExecuteAndRead(absolute_path))
                     .map_err(into_ioerr)?;
             }
-            if let permissions::Permission::Boolean(true) = real_permissions.net {
+            if let permissions::Permission::Boolean(true) = resolved_permissions.net {
                 birdcage.add_exception(Exception::Networking).map_err(into_ioerr)?;
             }
 
--- a/cli/src/commands/extensions/permissions.rs
+++ b/cli/src/commands/extensions/permissions.rs
@@ -69,9 +69,7 @@ impl Permission {
             // Child deny-all always succeeds, returning deny-all.
             (_, &Boolean(false)) => Ok(Boolean(false)),
             // Parent deny-all fails with everything else.
-            (&Boolean(false), child) => {
-                Err(anyhow!("Requested permissions incompatible with parent"))
-            },
+            (&Boolean(false), _) => Err(anyhow!("Requested permissions incompatible with parent")),
             // Child allow-all inherits the parent's permissions.
             (&List(ref parent), &Boolean(true)) => Ok(List(parent.clone())),
             // Parent allow-all returns the child's permissions.
--- a/cli/src/commands/extensions/permissions.rs
+++ b/cli/src/commands/extensions/permissions.rs
@@ -69,13 +69,13 @@ impl Permission {
             (_, &Permission::Boolean(false)) => Ok(Permission::Boolean(false)),
             // Parent deny-all fails with all child permissions but deny-all.
             (&Permission::Boolean(false), _) => {
-                Err(anyhow!("Requested permissions incompatible with the manifest"))
+                Err(anyhow!("All permissions are denied by the manifest"))
             },
             // Parent allow-all always succeeds, returning the child's permissions.
             (&Permission::Boolean(true), child) => Ok(child.clone()),
             // Child allow-all fails with more restrictive parent permissions.
             (_, &Permission::Boolean(true)) => {
-                Err(anyhow!("Requested permissions incompatible with the manifest"))
+                Err(anyhow!("The requested permissions are denied by the manifest"))
             },
             // Parent set vs child set have to be validated.
             // This will error if child is not subset of parent, and return the child set otherwise.
@@ -84,7 +84,7 @@ impl Permission {
                     .map(|_| Permission::List(child.clone()))
                     .map_err(|mismatches| {
                         anyhow!(
-                            "Requested permission paths incompatible with the manifest: {}",
+                            "The following paths are denied by the manifest: {}",
                             mismatches.join(", ")
                         )
                     })
@@ -212,7 +212,7 @@ impl Permissions {
     }
 
     pub fn subset_of(&self, other: &Permissions) -> Result<Permissions> {
-        let err_ctx = |name: &'static str| move |e| anyhow!("{name}: {}", e);
+        let err_ctx = |name: &'static str| move |e| anyhow!("Invalid {name} permissions: {}", e);
 
         Ok(Permissions {
             read: self.read.subset_of(&other.read).map_err(err_ctx("read"))?,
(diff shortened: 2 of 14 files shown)


cli/src/commands/extensions/api.rs
```diff
@@ -380,13 +392,21 @@ async fn parse_lockfile(
...
 #[cfg(unix)]
-fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
+fn run_sandboxed(op_state: Rc<RefCell<OpState>>, process: Process) -> Result<ProcessOutput> {
+    let Process { cmd, args, stdin, stdout, stderr, exceptions } = process;
+    let exceptions_strict = exceptions.strict;
+    let resolved_permissions = {
+        let mut state = op_state.borrow_mut();
+        let permissions = state.borrow_mut::<permissions::Permissions>();
+        permissions::Permissions::from(exceptions).subset_of(permissions)
+    }?;
```
kylewillmon, line 402: I think we decided that `run_sanboxed` should always error if `exceptions` is not a strict subset of the extension permissions. So instead of resolving a common subset, this should just be a strict permission check.

That would look something like this:

```suggestion
    let mut state = op_state.borrow_mut();
    let ext_perms = state.borrow_mut::<permissions::Permissions>();
    let child_perms = permissions::Permissions::from(exceptions);
    if !child_perms.subset_of(ext_perms) {
        return Err(anyhow!("Permissions denied by manifest file!"));
    }
```

Or you could have the error message generated by the subset function, which might be able to provide more detail:


```suggestion
    let mut state = op_state.borrow_mut();
    let ext_perms = state.borrow_mut::<permissions::Permissions>();
    let child_perms = permissions::Permissions::from(exceptions);
    child_perms.enforce_subset_of(ext_perms)?;
```

andreaphylum: When A is subset of B, the common subset is, by definition, A itself. In fact, that's exactly what `subset_of` is returning, rather than a resolution; see [this](https://github.com/phylum-dev/cli/pull/746/files#diff-061d5eeda494ae877735811ed5f387c830d778a7f49a345622fc17ba8f17f0f6R75). In all other cases, there exists one a in A such that for each b in B, b is not prefix of a, which means that A is not subset of B.

The [`paths_subset`](https://github.com/phylum-dev/cli/pull/746/files#diff-061d5eeda494ae877735811ed5f387c830d778a7f49a345622fc17ba8f17f0f6R80) method does indeed return either success (`Ok(())`) or a list of forbidden paths, and that's probably the functionality you are referring to; `subset_of` only takes care of the [additional cases](https://github.com/phylum-dev/cli/pull/746/files#diff-061d5eeda494ae877735811ed5f387c830d778a7f49a345622fc17ba8f17f0f6R69-R73) where at least one of parent and child are boolean permissions, and returns the appropriate permissions set, which could be either the parent's, the child's, or a boolean permission deriving from a combination of those.

I think this is mostly an issue of unclear naming and/or documentation, but the semantics is what we intend.

andreaphylum: 1. You're right, that should probably be split into a separate match with an explicit error.
2. I think it is important that we address this immediately, both in code and in documentation, as it is a public facing API and altering this behavior at a later point would result in a breaking change. I'd strongly suggest supporting the "inherit" use case as it feels like it follows naturally from the distinction we already make between list and boolean permissions.
3. This is a bona fide mistake on my end, and it should definitely result in an error too.

All in all, I think the current overall design makes sense in its intention, that is, centralizing the resolution of permission subsetting in a single testable spot rather than providing a boolean value at the outermost call site.

kylewillmon: I am happy with this design. We'll just need to settle on the correct way to handle the inherit use case.

> altering this behavior at a later point would result in a breaking change

This is only half true. Changing this case from an error to not an error does not break existing code. That is why I'd rather it be an error for now.

> I'd strongly suggest supporting the "inherit" use case as it feels like it follows naturally from the distinction we already make between list and boolean permissions.

My concern is that there are many different ways implement "inherit" and I'm not certain that `true` is the best option. For example:

* `write: true,`
* `write: PhylumApi.permissions().write,` or more likely `...PhylumApi.permissions()`
* `write: "inherit",`

kylewillmon: Pinging @cd-work since he may have an opinion on this

cd-work: We discussed in the last sync that `write: true` should not be allowed for now unless `write: true` is also set in the manifest, right? Specifically because it's trivial to change in the future and it might be more confusing than beneficial?

andreaphylum: My bad then -- must've slipped my mind. In that case, I suggest we ask the user to be explicit with a list for the time being, and we create an issue to add support for a `PhylumApi.permission()` method.

cli/tests/fixtures/extensions/permissions/correct-run-perms/main.ts
```diff
@@ -1,10 +1,16 @@
...
+  cmd: '/bin/echo',
+  args: ['hello'],
+  stdout: 'piped',
+  stderr: 'piped',
+  exceptions: {
+    read: false,
+    write: false,
+    net: false,
+    run: false,
+  }
```
cd-work, line 13: They all default to `false` anyway, right? Are these mandatory? Would be nice to just omit them.

docs/extensions/extension_sandboxing.md
```diff
@@ -56,6 +56,15 @@ permission is a superset of `read` that allows for execution. While `read`,
 `write`, and `run` accept either a path to be allowed or a boolean, `net` only
 allows for a boolean value.
 
+## Limitations
+
+The `PhylumApi.runSandboxed` method is the only allowed means of spawning
+child processes from an extension. In order to prevent extension to escape the
```
cd-work, line 62:
```suggestion
child processes from an extension. In order to prevent extensions from escaping the
```

docs/extensions/extension_sandboxing.md
```diff
@@ -56,6 +56,15 @@ permission is a superset of `read` that allows for execution. While `read`,
...
+sandbox, methods such as `Deno.run` are explicitly disabled.
```
cd-work, line 63: "Methods such as" is incredibly unclear here. Makes it sound like we deny a bunch of other stuff but don't actually document it.

Should just be "the `Deno.run` is explicitly disabled." or something like that. Sounds a little nicer if you turn it around and start with `The Deno.run method is explicitly disabled, to prevent bla bla bla`.

docs/extensions/extension_sandboxing.md
```diff
@@ -56,6 +56,15 @@ permission is a superset of `read` that allows for execution. While `read`,
...
+
+The method is not allowed to specify permissions that are less restrictive than
+the ones specified in the manifest.
```
cd-work, line 66: This is super duper annoying semantics, but it might be easier to understand "more permissive" than "less restrictive"? Almost a double negative kinda deal, some people might struggle a bit more to wrap their head around it.

cd-work (changes requested):

cli/src/commands/extensions/api.rs
```diff
@@ -380,13 +393,21 @@ async fn parse_lockfile(
 /// possible even after the command has been spawned.
 #[op]
 #[cfg(unix)]
-fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
+fn run_sandboxed(op_state: Rc<RefCell<OpState>>, process: Process) -> Result<ProcessOutput> {
+    let Process { cmd, args, stdin, stdout, stderr, exceptions } = process;
+    let exceptions_strict = exceptions.strict;
```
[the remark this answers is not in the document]
andreaphylum: It's because `Permissions::from` would partially move out of `Process`. If need be, we could reimplement `Permissions::from` to take a reference.

cli/src/commands/extensions/permissions.rs
```diff
@@ -61,6 +62,74 @@ impl Permission {
             Err(anyhow!("Requires {resource_type} access to {resource:?}"))
         }
     }
+
+    pub fn subset_of(&self, parent: &Permission) -> Result<Permission> {
+        use Permission::*;
```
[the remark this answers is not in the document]
andreaphylum: I did this in order to avoid peppering the `match` statement with `Permission::`. I'd argue it is one of those use cases where _not_ having imports at the top helps legibility a great deal. Besides, it would not make sense to have this import at the top as the intent is just importing the _variants_ in scope.

docs/extensions/extension_sandboxing.md
andreaphylum: Wanted to discuss this -- in newer releases of Deno, I've noticed there are other methods of spawning processes besides `run` such as [`spawnChild`](https://doc.deno.land/deno/unstable/~/Deno.spawnChild). We don't necessarily want to bump Deno releases anytime soon, and I don't think those methods exist for the runtime version we are using; but if we ever do, we should be aware of those, otherwise escaping the sandbox could be a possibility.

I'm cool with simplifying the wording for the time being.

cli/tests/fixtures/extensions/permissions/correct-run-perms/main.ts
andreaphylum: Those _should_ default to `false` but I need to check again -- and if they don't already, I'd say defaulting to false is the correct expected behavior.

cli/tests/fixtures/extensions/permissions/correct-run-perms/PhylumExt.toml
```diff
@@ -3,4 +3,4 @@ description = "An extension with correctly specified permissions"
 entry_point = "main.ts"
 
 [permissions]
-run = ["cargo"]
+run = ["/bin/echo"]
```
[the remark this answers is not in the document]
andreaphylum: Good point.

cli/src/commands/extensions/permissions.rs
```diff
@@ -61,6 +62,74 @@ impl Permission {
...
+        let without_parent: Vec<_> = child_paths
+            .iter()
+            .filter_map(|child| {
+                // Using PathBuf::starts_with rather than String::starts_with in order to get
+                // the correct semantics.
+                if !parent_paths.iter().any(|p| child.starts_with(p)) {
+                    Some(child.to_string_lossy().to_string())
+                } else {
+                    None
+                }
```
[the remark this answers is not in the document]
andreaphylum: It feels kinda weird to put the `None` before the `Some` branch, but I don't like the negation either. Wish we had a `none` method that's just `!any`, it would be perfect.

cli/src/commands/extensions/permissions.rs
```diff
@@ -61,6 +62,74 @@ impl Permission {
...
+        // The above definition is tested in `tests::paths_subset_algorithm` below.
+        //
+        // In this method, A is `child` and B is `parent`.
+
+        let parent_paths = parent.iter().map(PathBuf::from).collect::<Vec<_>>();
+        let child_paths = child.iter().map(PathBuf::from).collect::<Vec<_>>();
+
+        // List all child paths for which no parent path exists such that child <
+        // parent. O(n*m) where n = child_paths,len() and m =
+        // parent_paths.len().
```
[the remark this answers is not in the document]
andreaphylum: Went back and forth myself here -- I'm not sure what is the best way of expressing 1) a potentially impactful O(nm) complexity and 2) the fact that this computes the partial ordering operation. It is unfortunate that that's tied to the definition above but I don't really see a better alternative. Again, the intent here was more "literate programming" than "explain what the code does" which I feel is appropriate -- given the context, of course, not everywhere.

cli/src/commands/extensions/permissions.rs
```diff
@@ -61,6 +62,74 @@ impl Permission {
...
         }
     }
+
+    pub fn subset_of(&self, parent: &Permission) -> Result<Permission> {
+        use Permission::*;
+        match (parent, self) {
+            // Child deny-all always succeeds, returning deny-all.
+            (_, &Boolean(false)) => Ok(Boolean(false)),
+            // Parent deny-all fails with all child permissions but deny-all.
+            (&Boolean(false), _) => Err(anyhow!("Requested permissions incompatible with parent")),
```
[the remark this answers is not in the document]
andreaphylum: My point here was exactly that -- we don't necessarily want this to be strongly tied to `runSandboxed` but also finding a general enough wording for the error is problematic. `manifest` is probably a good enough compromise.

cli/src/commands/extensions/api.rs
```diff
@@ -397,29 +418,29 @@ fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
...
             }
-            for path in process.exceptions.run.sandbox_paths().iter() {
+            for path in resolved_permissions.run.sandbox_paths().iter() {
                 let path = dirs::expand_home_path(path, &home_dir);
                 let absolute_path = permissions::resolve_bin_path(path);
                 permissions::add_exception(&mut birdcage, Exception::ExecuteAndRead(absolute_path))
                     .map_err(into_ioerr)?;
             }
-            if process.exceptions.net {
+            if let permissions::Permission::Boolean(true) = resolved_permissions.net {
```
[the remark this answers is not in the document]
andreaphylum: That would be a _lot_ of work -- basically moving all modifications of this PR into `ProcessExceptions` which IIUC is little more than a DTO, thus splintering important functionality away from the place where it makes sense. I agree that this isn't the best looking code ever, but the solution would be much worse than the problem.

cli/src/commands/extensions/permissions.rs
```diff
@@ -61,6 +62,74 @@ impl Permission {
...
+            // Child allow-all fails with more restrictive parent permissions.
+            (_, &Boolean(true)) => Err(anyhow!("Requested permissions incompatible with parent")),
+            // Parent set vs child set have to be validated.
+            // This will error if child is not subset of parent, and return the child set otherwise.
+            (&List(ref parent), &List(ref child)) => Permission::paths_subset(parent, child)
+                .map(|()| Permission::List(child.clone()))
+                .map_err(|mismatches| {
+                    anyhow!(
+                        "Requested permission paths incompatible with parent: {}",
+                        mismatches.join(", ")
```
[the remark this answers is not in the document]
andreaphylum:
```
Requested permission paths incompatible with the manifest: /tmp/foo, /dev
```

cli/src/commands/extensions/permissions.rs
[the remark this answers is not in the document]
cd-work: I **strongly** disagree with this improving legibility, it also very much goes against what is generally done in Rust.

The whole idea that the goal is "just importing the *variants* in scope" already shows why this is a terrible, terrible idea: You're basically intentionally creating an import that can be bound differently in a different scope, making things unnecessarily confusing. If an import is useful, it should be globally consistent in the module and enum variants generally should not be imported directly in Rust.

cli/src/commands/extensions/permissions.rs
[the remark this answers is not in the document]
cd-work: Yeah I think manifest is the bare minimum I'm willing to accept here.

I understand that you might not want an overly specific error, but if the error is useless then we have to handle it at a higher level instead. At the point where it is presented to the user, it must be clear in what is going on and our internal code structure should not have a negative impact on its clarity.

cli/src/commands/extensions/permissions.rs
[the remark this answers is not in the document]
cd-work: That's probably going to be a bit difficult to connect back with `runSanboxed`, I'd love to be able to tell the extension author that's where the error came from.

I also think that printing this error without mentioning the specific permission is problematic. Imagine a user accidentally swapping read/write, it would likely be much faster to troubleshoot when we print the specific permission in the output here.

cli/src/commands/extensions/permissions.rs
```diff
@@ -61,6 +62,74 @@ impl Permission {
...
+        //
+        // Let "<<" be the partial order relation over the power set of P, such that "A
+        // << B" reads "A is paths-subset of B" where A and B are subsets of P.
+        //
+        // A << B if and only if, for each a in A, there exist at least one b in B such
+        // that a < b.
+        //
+        // The above definition is tested in `tests::paths_subset_algorithm` below.
+        //
+        // In this method, A is `child` and B is `parent`.
```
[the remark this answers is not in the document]
cd-work: What benefit does this comment bring when it only confuses and doesn't help anyone? I haven't bothered reading it.

I mean if you want to write this down because it helps you with the implementation, that's fine of course. But what benefit does it bring to our codebase over a **descriptive** comment detailing the expected behavior.

cli/src/commands/extensions/permissions.rs
```diff
@@ -61,6 +62,74 @@ impl Permission {
...
+                .map_err(|mismatches| {
+                    anyhow!(
+                        "Requested permission paths incompatible with parent: {}",
+                        mismatches.join(", ")
+                    )
+                }),
+        }
+    }
+
+    fn paths_subset(parent: &[String], child: &[String]) -> StdResult<(), Vec<String>> {
```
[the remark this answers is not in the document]
cd-work: I'd suggest something like `validate/check_path_includes_child`. My focus is mainly on correctly clarifying in the name that this function will return an error to indicate a validation failure.

cli/src/commands/extensions/permissions.rs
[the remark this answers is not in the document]
cd-work: I disagree strongly that this is appropriate, considering it puts a heavy focus on writing the code when in reality the focus should be on making it as readable as possible.

A comment like this should be easier to understand without requiring any prior knowledge:
```
// Collect all child paths which are not a subset of any parent.
```

The entirety of the trailing end of this comment is completely unnecessary imo and does not provide information that is helpful to a reader. It only serves to confuse and extend the time it takes to read and understand the comment.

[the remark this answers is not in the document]
andreaphylum: I had already replaced the above with a similar comment.

cli/src/commands/extensions/permissions.rs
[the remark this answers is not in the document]
andreaphylum: We deal with that [here](https://github.com/phylum-dev/cli/pull/746/files#diff-061d5eeda494ae877735811ed5f387c830d778a7f49a345622fc17ba8f17f0f6R228-R235).

The overall output should be something like
```
read: Requested permissions incompatible with the manifest: /here/there, /tmp/something
```
Not married to the style, so I'm open to suggestions here.

cli/src/commands/extensions/permissions.rs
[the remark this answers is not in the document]
andreaphylum: We deal with that [here](https://github.com/phylum-dev/cli/pull/746/files#diff-061d5eeda494ae877735811ed5f387c830d778a7f49a345622fc17ba8f17f0f6R228-R235).

The overall output should be something like
```
read: Requested permissions incompatible with the manifest: /here/there, /tmp/something
```
Not married to the style, so I'm open to suggestions here.

cli/src/commands/extensions/permissions.rs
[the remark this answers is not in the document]
andreaphylum: My reply below should address this as well.

cli/src/commands/extensions/permissions.rs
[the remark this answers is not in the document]
andreaphylum: >it will keep providing this purpose if we ever need to get back to this code for whatever reason.

I honestly feel like debating these matters any further is terribly fruitless though so I will just get rid of the comment,

cli/src/commands/extensions/permissions.rs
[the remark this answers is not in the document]
andreaphylum: I like that a lot better. I also updated the other error messages to state "permissions are denied by the manifest".

andreaphylum:
On a Mac M1, tests pass and CLI compiles successfully.

Manual tests of our `npm`, `yarn`, `poetry` extensions were successful on OS X, and resulted in an issue for `npm` that is most likely related to `birdcage` on Ubuntu Linux.

cli/src/commands/extensions/permissions.rs
```diff
@@ -61,6 +62,65 @@ impl Permission {
...
+                        anyhow!(
+                            "The following paths are denied by the manifest: {}",
+                            mismatches.join(", ")
+                        )
+                    })
+            },
+        }
+    }
+
+    fn check_paths_include_childs(
```
cd-work, line 95:
```suggestion
    fn check_paths_include_children(
```

cd-work (changes requested):

cli/src/commands/extensions/api.rs
```diff
@@ -91,6 +91,19 @@ struct ProcessException {
...
 }
 
+#[cfg(unix)]
+impl From<ProcessException> for permissions::Permissions {
+    fn from(process_exception: ProcessException) -> Self {
+        Self {
+            read: process_exception.read,
+            write: process_exception.write,
+            run: process_exception.run,
+            env: Permission::Boolean(false), // TODO: replace with inherit
```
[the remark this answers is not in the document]
andreaphylum: If that's ok, since we went with no inherit, I'd switch this to `true` so it mimicks the current behavior (i.e. allow all) and we can tackle this in the other PR.

andreaphylum:
We should additionally document the instances where using `Deno.run` is appropriate, but I'd leave this for a future PR as there's a number of considerations to be made -- i.e. how deep to go into explaining how the sandbox works and why we need to have `Deno.run` in the first place -- and it would probably be more appropriate to address this alongside the addition of the `unsandboxed-run` permission request.

cli/src/commands/extensions/permissions.rs
```diff
@@ -61,6 +62,65 @@ impl Permission {
...
+        let parent_paths = parent.iter().map(PathBuf::from).collect::<Vec<_>>();
+        let child_paths = child.iter().map(PathBuf::from).collect::<Vec<_>>();
+
+        // Find all paths in `child` that don't have a prefix in `parent`.
+        let without_parent: Vec<_> = child_paths
+            .iter()
+            .filter_map(|child| {
+                // Using PathBuf::starts_with rather than String::starts_with in order to get
+                // the correct semantics.
+                if parent_paths.iter().any(|p| child.starts_with(p)) {
```
kylewillmon, line 108: The `starts_with` function is in `Path` and not `PathBuf`. Which means that you could probably refactor this function to avoid extra allocations and copies.

For a simple option, if `c` and `p` are both `&String`, then you can call `Path::starts_with` like this: `Path::new(c).starts_with(p)`

cli/tests/extensions/mod.rs
```diff
@@ -7,6 +7,7 @@ use std::sync::Mutex;
 
 use lazy_static::lazy_static;
 use phylum_cli::commands::extensions::extension::Extension;
+use phylum_cli::commands::extensions::permissions::{Permission, Permissions};
```
kylewillmon, line 10: This warns on Windows.

```suggestion
#[cfg(unix)]
use phylum_cli::commands::extensions::permissions::{Permission, Permissions};
```

kylewillmon (changes requested):

cd-work (approved):

cli/src/commands/extensions/permissions.rs
andreaphylum: Usually these sorts of optimizations are automatically tackled by LLVM and generate the same code; out of curiosity, I benchmarked these two functions and both consistently take ~170ns per iteration:

```rust
pub fn with_path() -> bool {
    let parent = "/tmp".to_string();
    let child = "/tmp/child".to_string();

    let parent_path = Path::new(&parent);
    let child_path = Path::new(&child);

    child_path.starts_with(parent_path)
}

pub fn with_pathbuf() -> bool {
    let parent = "/tmp".to_string();
    let child = "/tmp/child".to_string();

    let parent_path = PathBuf::from(parent);
    let child_path = PathBuf::from(child);

    child_path.starts_with(parent_path)
}
```

The change is minuscule, literally switching `PathBuf::from` for `Path::new` at the top of the method, so I think it's worth it anyway just in case.

kylewillmon: I also should have been more clear: Using `Path` means you can completely avoid building the `parent_paths` and `child_paths` vectors. I'll add a suggestion on the new code to show this better.

cli/src/commands/extensions/permissions.rs
```diff
@@ -61,6 +62,65 @@ impl Permission {
...
+            .filter_map(|child| {
+                // Using Path::starts_with rather than String::starts_with in order to get
+                // the correct semantics.
+                if parent_paths.iter().any(|p| child.starts_with(p)) {
+                    None
+                } else {
+                    Some(child.to_string_lossy().to_string())
+                }
+            })
+            .collect::<Vec<_>>();
```
kylewillmon, line 114:
```suggestion
        // Find all paths in `child` that don't have a prefix in `parent`.
        let without_parent: Vec<_> = child
            .iter()
            .filter_map(|child| {
                // Using Path::starts_with rather than String::starts_with in order to get
                // the correct semantics.
                if parent.iter().any(|p| Path::new(child).starts_with(p)) {
                    None
                } else {
                    Some(child.clone())
                }
            })
            .collect::<Vec<_>>();
```

kylewillmon (approved):

cli/src/commands/extensions/permissions.rs
andreaphylum: Oh, yeah, this is a great point indeed. I like it.

cli/src/commands/extensions/permissions.rs
andreaphylum: Here's the same thing with `String` in lieu of `&str`: https://godbolt.org/z/Kzse56ohe

Interestingly enough, `with_path` calls `drop_in_place<PathBuf>`.

kylewillmon: > Interestingly enough, `with_path` calls `drop_in_place<PathBuf>`.

For what it's worth, I think that's actually `drop_in_place<String>` (which gets collapsed into `drop_in_place<PathBuf>` during optimization). And it's only called in the unwind section (i.e., during a panic)

kylewillmon (approved):

kylewillmon (approved):

PR #747, opened 2022-10-13
Fix CLI certificate override modifying config

### LINKED ISSUE #664: Command line argument `--no-check-certificate` is saved to config
# Overview
Setting `--no-check-certificate` on the command line gets saved to the config so that all future runs no longer check certificates. And since we do not have a `--check-certificate` option (see #483), the only way to get phylum to start checking certificates again is to manually modify the config file. This is very similar to #582

# How To Reproduce
```
❯ grep 'ignore_certs' ~/.config/phylum/settings.yaml 
ignore_certs: false

❯ phylum --no-check-certificate ping
[2022-09-02T15:12:10Z WARN  phylum] Ignoring TLS server certificate verification per user request.
✅ alive!

❯ grep 'ignore_certs' ~/.config/phylum/settings.yaml
ignore_certs: true
```

# Expected Behavior
Config from command line arguments or environment variables should never be written out to the config file.

cd-work:
This resolves an issue with the `--no-check-certificate` CLI option which would cause the `ignore_certs` flag to be permanently changed to `true` in the user's Phylum configuration file.

Closes #664.

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -149,12 +149,12 @@ impl PhylumApi {
         // Do we have a refresh token?
         let tokens: TokenResponse = match &config.auth_info.offline_access() {
             Some(refresh_token) => {
-                handle_refresh_tokens(refresh_token, config.ignore_certs, &config.connection.uri)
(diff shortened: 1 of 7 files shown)


maxrake (approved):
The code changes look good. They were tested manually with a local build.

PR #749, opened 2022-10-13
Fix verbosity errors

cd-work:
This fixes a regression introduced in 8748152 where the CLI would panic when running the jobs/analyze subcommands since these had CLI arguments conflicting with the new global `verbose` flag.

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -88,11 +88,6 @@ pub fn add_subcommands(command: Command) -> Command {
                 Arg::new("JOB_ID")
                     .value_name("JOB_ID")
                     .help("The job id to query (or `current` for the most recent job)"),
-                Arg::new("verbose")
-                    .action(ArgAction::SetTrue)
-                    .short('v')
-                    .long("verbose")
-                    .help("Increase verbosity of api response."),
                 Arg::new("filter").long("filter").value_name("filter").help(FILTER_ABOUT),
                 Arg::new("json")
                     .action(ArgAction::SetTrue)
@@ -259,11 +254,6 @@ pub fn add_subcommands(command: Command) -> Command {
                          system)",
                     ),
                     Arg::new("label").short('l').value_name("label"),
-                    Arg::new("verbose")
-                        .action(ArgAction::SetTrue)
-                        .short('v')
-                        .long("verbose")
-                        .help("Increase verbosity of api response."),
                     Arg::new("filter").long("filter").value_name("filter").help(FILTER_ABOUT),
                     Arg::new("json")
                         .action(ArgAction::SetTrue)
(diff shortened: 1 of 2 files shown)


cli/src/app.rs
```diff
@@ -88,11 +88,6 @@ pub fn add_subcommands(command: Command) -> Command {
                 Arg::new("JOB_ID")
                     .value_name("JOB_ID")
                     .help("The job id to query (or `current` for the most recent job)"),
-                Arg::new("verbose")
-                    .action(ArgAction::SetTrue)
-                    .short('v')
-                    .long("verbose")
```
kylewillmon, line 94: The global argument only has the short form and not the long form. So calls like this will now fail:

```
phylum analyze --verbose --json
```

And I think that is exactly how `phylum-ci` runs that command... Pinging @maxrake to confirm.

cli/src/app.rs
```diff
@@ -259,11 +254,6 @@ pub fn add_subcommands(command: Command) -> Command {
                          system)",
                     ),
                     Arg::new("label").short('l').value_name("label"),
-                    Arg::new("verbose")
-                        .action(ArgAction::SetTrue)
-                        .short('v')
-                        .long("verbose")
```
maxrake, line 265: Removing the long form of this option for the analyze subcommand is a breaking change...at least for `phylum-ci`, which makes calls with this long form. Can the global verbosity option be updated to provide a long form? If not, then a release of the CLI with this breaking change will need to be coordinated with an update to `phylum-ci` to use the short form instead.

cli/src/app.rs
maxrake: I already have a comment on a draft review in place saying exactly this. You beat me to it!

maxrake (changes requested):
Documentation that would need to be updated if the long form of `-v` (`--verbose`) is not added back in:

* `docs/knowledge_base/analyzing-dependencies.md`
* `docs/command_line_tool/phylum_analyze.md`
* `docs/command_line_tool/phylum_history.md`

cli/src/app.rs
cd-work: We can just add the long form tot he global command, no issue there.

maxrake (commented):
Yay...`--verbose` is back!

However, when testing out the branch, this was observed:

image

It was not clear if it is related to the changes in this PR, so here is the output from the current released version:

image

...and current `main`:

image

So...since this appears to be an issue in `main`...albeit with a different error output, a new issue will be created.

cd-work:
Yet another clap regression, yay. @maxrake can you open an issue?

maxrake:
> Yet another clap regression, yay. @maxrake can you open an issue?

Done - #751

kylewillmon (approved):

maxrake (approved):

PR #759, opened 2022-10-17
Recommend WSL on Windows

kylewillmon:
Our support for native binaries on Windows is currently minimal, but our docs continue to claim "binaries can easily be built".

This patch updates the wording to recommend WSL for those interested in using our CLI on Windows.

Additionally, this patch adds `--locked` to our `cargo install` command so that the `Cargo.lock` file is not ignored. Due to our usage of git dependencies, this is critical and the command will often fail without it.


furi0us333 (approved):

maxrake (approved):

docs/command_line_tool/alternate_install.md
```diff
@@ -23,7 +23,7 @@ Phylum is written in Rust, so you'll need a recent Rust installation to build it
 3. You can use the executable directly as `./target/debug/phylum` or install it like so:
 
    ```sh
-   cargo install --path cli
+   cargo install --locked --path cli
```
cd-work, line 26: Do we really want this? IIRC `--locked` outputs an error if the lockfile is not up to date, right? Do we expect users, potentially not particularly experienced with Rust, to run into this error?

If we remove this flag, the lockfile will be updated automatically and should probably fix the issue for the user automagically.

kylewillmon: There are only two cases where `--locked` should cause an error:

1. A crate in `Cargo.lock` is yanked
2. The packages in `Cargo.lock` don't satisfy the requirements in `Cargo.toml`

Case 1 is rare and case 2 should never happen on the `main` branch

On the other hand, without `--locked`, the build will fail any time the code is incompatible with the latest updates to `phylum_types` or `birdcage` (which happens quite often)

You can see this for yourself by running the command on `main` right now, as it fails without `--locked` and passes with it.

cd-work: > You can see this for yourself by running the command on `main` right now, as it fails without `--locked` and passes with it.

Oh right, I completely forgot about https://github.com/rust-lang/cargo/issues/7169. 

Honestly I just think it's a good idea to always specify the git hash you're building against when running against git repositories. In fact I generally build against tagged commits since those can't be easily dropped by force-pushing.

But I suppose this is not something worth holding this PR up for.

cd-work (approved):

PR #764, opened 2022-10-19
Bump to v3.11.0

andreaphylum:
Release-Version: v3.11.0
Release-Summary: Extension API: Sandbox processes, create and delete projects


maxrake:
I thought it was going to be 4.0

andreaphylum:
> I thought it was going to be 4.0

The 4.0 breaking change would be gating `Deno.run` behind a new permission, but we don't have that yet, so `main` doesn't have breaking changes so far. Once that feature is complete, we'll release 4.0.

maxrake (approved):

PR #765, opened 2022-10-19
Re-execute phylum for sandboxing extensions

### LINKED ISSUE #763: Avoid pre_exec/fork for `run_sandboxed`
Currently `run_sandboxed` uses `pre_exec` to setup the sandboxing rules, switching to `fork` once https://github.com/phylum-dev/cli/pull/760 is merged. However both of these have issues with signal safety due to making allocations and modifying the environment when setting up our sandbox.

To avoid any potential issues during sandbox setup, we should instead move the creation of the sandboxing profile and its application to a separate program.

The easiest solution would probably be to add a hidden option to the CLI to execute any command in a sandboxed environment, that way we could rely on `argv[0]` to spawn a new process.

cd-work:
This fixes any potential issues with setting up our sandboxing
environment in `pre_exec` or after a `fork` by adding a new `phylum
sandbox` subcommand. Using this subcommand we can simply re-execute
`phylum` with the appropriate arguments to setup a sandboxing
environment.

Closes #763.

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -350,6 +350,57 @@ pub fn add_subcommands(command: Command) -> Command {
         )
         .subcommand(extensions::command());
 
+    #[cfg(unix)]
+    {
+        app = app.subcommand(
+            Command::new("sandbox").hide(true).about("Run an application in a sandbox").args(&[
+                Arg::new("allow-read")
+                    .help("Add filesystem read sandbox exception")
+                    .long("allow-read")
+                    .value_name("PATH")
+                    .value_hint(ValueHint::FilePath)
+                    .action(ArgAction::Append),
+                Arg::new("allow-write")
+                    .help("Add filesystem write sandbox exception")
+                    .long("allow-write")
+                    .value_name("PATH")
+                    .value_hint(ValueHint::FilePath)
+                    .action(ArgAction::Append),
+                Arg::new("allow-run")
+                    .help("Add filesystem execute sandbox exception")
+                    .long("allow-run")
+                    .value_name("PATH")
+                    .value_hint(ValueHint::FilePath)
+                    .action(ArgAction::Append),
+                Arg::new("allow-env")
+                    .help("Add environment variable access sandbox exception")
+                    .long("allow-env")
+                    .value_name("ENV_VAR"),
+                // TODO: Can we combine this with allow-env?
+                Arg::new("allow-all-env")
+                    .help("Add full environment variable access sandbox exception")
+                    .long("allow-all-env")
+                    .conflicts_with("allow-env")
+                    .action(ArgAction::SetTrue),
+                Arg::new("allow-net")
+                    .help("Add network access sandbox exception")
+                    .long("allow-net")
+                    .action(ArgAction::SetTrue),
+                Arg::new("strict")
+                    .help("Do not add any default sandbox exceptions")
+                    .long("strict")
+                    .action(ArgAction::SetTrue),
+                Arg::new("cmd").help("Command to be executed").value_name("CMD").required(true),
+                Arg::new("args")
+                    .help("Command arguments")
+                    .value_name("ARG")
+                    .trailing_var_arg(true)
+                    .allow_hyphen_values(true)
+                    .action(ArgAction::Append),
+            ]),
+        )
+    }
+
     #[cfg(feature = "selfmanage")]
     {
         app = app.subcommand(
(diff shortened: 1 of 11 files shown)


cli/src/app.rs
```diff
@@ -347,6 +347,54 @@ pub fn add_subcommands(command: Command) -> Command {
...
+                    .help("Add filesystem execute sandbox exception")
+                    .long("allow-run")
+                    .value_name("PATH")
+                    .value_hint(ValueHint::FilePath)
+                    .action(ArgAction::Append),
+                Arg::new("allow-env")
+                    .help("Add environment variable access sandbox exception")
+                    .long("allow-env")
+                    .value_name("ENV_VAR")
+                    .num_args(0..=1)
```
kylewillmon, line 376: This seems like it might cause some parsing issues since we accept positional arguments.

```
phylum sandbox --allow-env cat ./dangerous_script.sh
```

Due to the nature of this command, there will always be arguments after `--allow-env`. So when will clap use the 0 arg version? Only when the following argument starts with `--`?

cli/src/commands/extensions/api.rs
```diff
@@ -465,123 +405,83 @@ fn run_sandboxed(op_state: Rc<RefCell<OpState>>, process: Process) -> Result<Pro
...
     }
-}
 
-/// Lock down the current process.
-#[cfg(unix)]
-fn lock_process(exceptions: permissions::Permissions, strict: bool) -> Result<()> {
-    let home_dir = dirs::home_dir()?;
+    // Execute sandboxed command.
+    let output = Command::new(env::current_exe()?)
+        .args(sandbox_args.iter_mut().map(|arg| arg.to_mut()))
```
kylewillmon, line 424: The mutable references here feel wrong. There's got to be some way to make it work with something like this:

```suggestion
        .args(sandbox_args.into_iter().map(|arg| arg.as_ref()))
```

Although maybe it needs some hint to know that `.as_ref()` should return `&str` in that case.

cli/src/commands/extensions/api.rs
```diff
@@ -465,123 +405,83 @@ fn run_sandboxed(op_state: Rc<RefCell<OpState>>, process: Process) -> Result<Pro
...
 
-    // Apply filesystem exceptions.
-    for path in exceptions.read.sandbox_paths().iter() {
+    // Add filesystem exception arguments.
+    let home_dir = dirs::home_dir()?;
+    for path in permissions.read.sandbox_paths().iter() {
         let path = dirs::expand_home_path(path, &home_dir);
-        permissions::add_exception(&mut birdcage, Exception::Read(path))?;
+        sandbox_args.push("--allow-read".into());
+        sandbox_args.push(path.to_string_lossy().into_owned().into());
```
kylewillmon, line 454: `Cow -> String -> Cow`? Do the lifetimes not match? Should both be `'a` based on the function signature.

cli/src/commands/extensions/api.rs
```diff
@@ -465,123 +405,83 @@ fn run_sandboxed(op_state: Rc<RefCell<OpState>>, process: Process) -> Result<Pro
...
-        permissions::add_exception(&mut birdcage, Exception::ExecuteAndRead(absolute_path))?;
+        sandbox_args.push("--allow-run".into());
+        sandbox_args.push(path.to_string_lossy().into_owned().into());
     }
 
-    // Apply network exceptions.
-    if let permissions::Permission::Boolean(true) = exceptions.net {
-        birdcage.add_exception(Exception::Networking)?;
+    // Add network exception argument.
+    if let permissions::Permission::Boolean(true) = permissions.net {
```
kylewillmon, line 468: Should this be `if matches!(...)` instead of `if let`?

kylewillmon (changes requested):

cli/src/commands/extensions/api.rs
cd-work: I'd avoid `matches!` unless necessary tbh.

cd-work: More specifically by "necessary" I mean if you need to evaluate a `let` thing to true/false **and** chain something after it.

cli/src/commands/extensions/api.rs
cd-work: Can't use `into_iter` because then the reference doesn't last long enough.

Can't use `as_ref` for similar reasons, `.borrow()` is the only way we could convert here afaik but that requires massive type annotations so it's not worth it. Afaik `to_mut` is the most concise way to transform a `Cow<'_, T>` into `&T` in this scenario (I've tried a couple things actually).

kylewillmon (approved):
Looks good (once the Windows issues are fixed)

kylewillmon (approved):

PR #770, opened 2022-10-21
`prettier` format checks

### LINKED ISSUE #704: Extensions: setup linting/formatting rules
We aren't currently enforcing any linting and formatting rules on our official extensions.
We should setup consistent linting and checks across all extensions.
I suggest relying on [standard](https://standardjs.com/) for simplicity.

andreaphylum:
Closes #704.

This PR adds a job for checking extensions' formatting with [prettier](https://prettier.io/).
`prettier` is opinionated and requires no configuration to be run.

In this PR, the only manual modifications are in the `test.yml` workflow; the remainder of the diffs derive from applying the formatting.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+*
+!(cli/src|extensions)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,13 +96,15 @@ cargo clippy --all-features
 #### Extensions
 
 First-party extensions are written in TypeScript. Extensions code must be
-formatted with [prettier](https://prettier.io/). Install it via `npm`:
+formatted with [Prettier](https://prettier.io/). 
+
+Prettier should be installed globally with a method of your choice. For example, via `npm`:
 
 ```sh
 npm install -g prettier
 ```
 
-`prettier` should be run from the top-level directory of the repository:
+Prettier should be run from the top-level directory of the repository:
 
 ```sh
 prettier --write .
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,8 @@ documentation available in the `docs/` directory.
 
 ### Style
 
+#### Rust
+
 General code format is maintained using `rustfmt`:
 
 ```sh
@@ -90,6 +92,23 @@ This should also pass for all pre-release features:
 ```sh
 cargo clippy --all-features
 ```
+
+#### Extensions
+
+First-party extensions are written in TypeScript. Extensions code must be
+formatted with [prettier](https://prettier.io/). Install it via `npm`:
+
+```sh
+npm install -g prettier
+```
+
+`prettier` should be run from the top-level directory of the repository:
+
+```sh
+prettier --write .
+```
+
+#### Shell scripts
 
 Additionally, there are a couple of script files that are part of the Phylum CLI
 project. These are linted by using [shellcheck]. The following command performs
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,7 +95,7 @@ jobs:
       - run: npm i -g prettier
       - name: Extensions Style Check
         if: github.event_name != 'schedule'
-        run: prettier --check extensions cli/src/extension_api.ts
+        run: prettier --check .
 
   shellcheck:
     # TODO: Update to `ubuntu-latest` once `ubuntu-22.04` support is stabilized.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,6 +84,19 @@ jobs:
           args: --all-features
         env:
           PHYLUM_API_KEY: ${{ secrets.PHYLUM_TOKEN_STAGING }}
+
+  prettier:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actionsx/setup-node@v2
+        with:
+          node-version: 18
+      - run: npm i -g prettier
+      - name: Extensions Style Check
+        if: github.event_name != 'schedule'
+        run: prettier --check extensions
+
   shellcheck:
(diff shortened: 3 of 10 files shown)


maxrake (changes requested):
This looks like a good idea and the implementation appears straight forward. The `CONTRIBUTING.md` file should be updated (in the "Style" section) to let folks know this tool is in use, how it should/could be configured locally, and which files it is meant to cover. To that end, are there additional configuration files for this tool to be more explicit about which files/directories it is meant to cover? That is, if folks have this tool installed locally in their IDE, will it start formatting other files unexpectedly...possibly leading to "flapping" formatting for files that are not included in the test workflow checks?

andreaphylum:
> This looks like a good idea and the implementation appears straight forward. The `CONTRIBUTING.md` file should be updated (in the "Style" section) to let folks know this tool is in use, how it should/could be configured locally, and which files it is meant to cover.

Good point. @kylewillmon was suggesting making that part of a different PR where we'd also tackle other aspects of this (for example, making an `xtask` out of this which would prompt the user to install the linter if that was not found).

> To that end, are there additional configuration files for this tool to be more explicit about which files/directories it is meant to cover? That is, if folks have this tool installed locally in their IDE, will it start formatting other files unexpectedly...possibly leading to "flapping" formatting for files that are not included in the test workflow checks?

I need to look into this, though I suppose we could short-circuit that by just stating `prettier --check .`... I don't believe we have other js/ts files outside of `extensions` and `extension_api.ts`, and even if we did (test fixtures maybe?) I don't think it's a bad idea to have those checked/formatted too.

kylewillmon:
> > This looks like a good idea and the implementation appears straight forward. The `CONTRIBUTING.md` file should be updated (in the "Style" section) to let folks know this tool is in use, how it should/could be configured locally, and which files it is meant to cover.
> 
> Good point. @kylewillmon was suggesting making that part of a different PR where we'd also tackle other aspects of this (for example, making an `xtask` out of this which would prompt the user to install the linter if that was not found).

You are both correct. I'd like to improve the flow for contributors and will open an issue for that.

However, `CONTRIBUTING.md` is the current place where we have these things documented and we should add a mention of `prettier` for js/ts files.

maxrake (dismissed):

.prettierignore
```diff
@@ -0,0 +1,2 @@
+*
+!(cli/src|extensions)
```
cd-work, line 2: I suppose this means ignore everything by default, but then whitelist `./cli/src` and `./extensions`?

Which additional files would get formatted without this ignore file?

CONTRIBUTING.md
```diff
@@ -91,6 +93,23 @@ This should also pass for all pre-release features:
...
 ```
 
+#### Extensions
+
+First-party extensions are written in TypeScript. Extensions code must be
+formatted with [prettier](https://prettier.io/). Install it via `npm`:
+
+```sh
+npm install -g prettier
+```
```
cd-work, line 103: Installation isn't documented for the Rust side either and I don't think it's necessary. Some people might use `npm`, others use a different mechanism. All possible options should be way better documented by prettier so we don't need to duplicate it here. That way things are also more future-proof.

cd-work (changes requested):

.prettierignore
andreaphylum: `prettier` doesn't work _only_ on js/ts files but also on a host of other formats, among which markdown. Having this whitelist means we can safely run `prettier --write .` without risking a billion lines long diff going forward as the only files we actually _want_ to be formatted are in one of those two directories. This intentionally keeps out stuff such as test fixtures and documentation.

CONTRIBUTING.md
andreaphylum: The [install docs](https://prettier.io/docs/en/install.html) only cover the use case of a pure JS codebase and a non-global install, which is not our case. We could either remove the install part and let the users figure it out for themselves, or just nudge them via rewording (i.e. _here's a way you could possibly get this done_) but linking to the original documentation is not only not future proof but also confusing _right now_.

cd-work: I would just let the user figure it out. It really shouldn't be that difficult. I'd install it from the official Arch repos for example, so the docs wouldn't even apply to all our current uses.

It's just that excessive verbosity in documentation is evil and I don't see much benefit to the statement right now, since nobody should be stumped by the task of installing prettier.

.prettierignore
cd-work: Have you actually tested this though? Does it actually format markdown by default?

andreaphylum: I have, it does. Also, as @maxrake mentioned above, we want to explicitly avoid being able to have someone set their IDE up in any way that could lead them to inadvertently format files we don't want to have formatted.

CONTRIBUTING.md
andreaphylum: >avoid any unnecessary details that are documented in other places anyway (like installing prettier for example

As I said above, due to how our repo is structured, prettier _doesn't_ provide that specific piece of documentation and thus a user would be left wondering how to fit prettier's recommended installation workflow (i.e. add it locally as a dev-dependency) into our repo's structure.

>Especially considering that a software developer who can't figure out how to install prettier probably won't be able to send a patch that requires doing so anyway.

That's a flawed assumption, though one that I would have agreed with in the past. Most users who are faced with a list of instructions with a missing step are just going to either bail unless they are motivated to see something through, which is a disservice to many potential contributors, or open an issue to ask for clarifications, which is a waste of everybody's time.

This doesn't automatically mean that those users are not smart enough, though, just that they are lacking proper context.

I agree that verbosity is mostly undesirable, but having a _tiny bit_ more information often goes a long way into ensuring some legitimate doubts, however small, never even crop up.

andreaphylum: Anyway, I reworded the docs ever so slightly to make the concept more "optional" without implying that `npm` is the only intended way.

cd-work (approved):

PR #771, opened 2022-10-21
Improve strictness of Gradle parser

### LINKED ISSUE #754: `gradle` parser is seeing ghosts
# Overview
The `gradle` parser is incredibly permissive and often detects packages in files that are not actually gradle lockfiles.

# How To Reproduce
```
phylum parse -t gradle tests/fixtures/yarn.lock
```

# Expected Behavior
The parser should return an empty list or error when given a file that looks nothing like a gradle lockfile.

cd-work:
Previously the Gradle parser was **very** permissive, so much so that
some of our non-Gradle test lockfiles were accepted as valid Gradle
lockfiles.

This patch rewrites the Gradle parser while attempting to be as strict
as possible. Running against our lockfiles in tests/fixtures now shows
that all parsers can only parse the lockfiles in their specific format.

Closes #754.

### DIFF
--- a/lockfile/src/lib.rs
+++ b/lockfile/src/lib.rs
@@ -160,6 +160,8 @@ pub fn get_path_format<P: AsRef<Path>>(path: P) -> Option<LockfileFormat> {
 
 #[cfg(test)]
 mod tests {
+    use std::fs;
+
     use super::*;
 
     #[test]
@@ -225,6 +227,7 @@ mod tests {
             ("gradle", LockfileFormat::Gradle),
             ("nuget", LockfileFormat::Msbuild),
             ("go", LockfileFormat::Go),
+            ("cargo", LockfileFormat::Cargo),
         ] {
             let actual_name = format.to_string();
             assert_eq!(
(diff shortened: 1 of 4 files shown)


lockfile/src/lib.rs
```diff
@@ -248,4 +251,42 @@ mod tests {
...
+    /// Ensure no new lockfiles are accidentally picked up by an unrelated
+    /// parser.
+    #[test]
+    fn parsers_only_parse_their_lockfiles() {
+        for (format, lockfile_count) in [
+            (LockfileFormat::Yarn, 4),
+            (LockfileFormat::Npm, 2),
+            (LockfileFormat::Gem, 1),
+            (LockfileFormat::Pipenv, 2),
+            (LockfileFormat::Poetry, 1),
```
maxrake, line 264: It looks like an entry was left off for `pip`:

`(LockfileFormat::Pip, 2),`

Was this on purpose, because the pip parser also seems to be pretty permissive? It appears to parse every test fixture, albeit only three of them result in a non-zero list of dependencies:

```
❯ for i in tests/fixtures/*
do
  echo $i $(target/debug/phylum parse -t pip $i >/dev/null 2>&1 && echo YES || echo NO) $(target/debug/phylum parse -t pip $i | jq length)
done

tests/fixtures/Calculator.csproj YES 0
tests/fixtures/Cargo_v1.lock YES 0
tests/fixtures/Cargo_v2.lock YES 0
tests/fixtures/Cargo_v3.lock YES 0
tests/fixtures/Gemfile.lock YES 0
tests/fixtures/Pipfile YES 0
tests/fixtures/Pipfile.lock YES 0
tests/fixtures/complex-requirements.txt YES 8
tests/fixtures/effective-pom.xml YES 0
tests/fixtures/go.sum YES 0
tests/fixtures/gradle.lockfile YES 0
tests/fixtures/package-lock-v6.json YES 0
tests/fixtures/package-lock.json YES 0
tests/fixtures/poetry.lock YES 0
tests/fixtures/requirements.txt YES 131
tests/fixtures/sample.csproj YES 0
tests/fixtures/workspace-effective-pom.xml YES 0
tests/fixtures/yarn-v1.lock YES 0
tests/fixtures/yarn-v1.lock.bad YES 0
tests/fixtures/yarn-v1.simple.lock YES 0
tests/fixtures/yarn-v1.trailing_newlines.lock YES 0
tests/fixtures/yarn.lock YES 1
```

kylewillmon (approved):

maxrake (approved):
This appears to do the job for gradle parsing...as confirmed with local testing. There is still a question about `pip` being left off the new test.

lockfile/src/lib.rs
cd-work: Yeah, I decided to give up on the pip parser since the requirements.txt format is somewhat insane and not actually just a lockfile.

I'm not saying it would be impossible to make it more permissive, but it would probably take a complete rewrite and some potentially dangerous assumptions.

PR #773, opened 2022-10-21
Avoid stdout when run with `--json`

### LINKED ISSUE #769: `--json` flag does not produce JSON output
# Overview
Starting with the release of CLI v3.11.0, the `--json` flag on commands is no longer producing _strictly_ JSON output on stdout. A number of commands offer the `--json` flag, but the `analyze` command is the one that is directly affecting `phylum-ci` since it gets used like an API call there and fails when the output is not parsable JSON format.

# How To Reproduce

image

# Expected Behavior
All commands with the `--json` flag specified produce JSON output. This is what it looks like for the previously released version of the CLI, v3.10.1:

image

# Additional Context
The way the JSON output is produced is up for discussion. That is, it does not necessarily _have to_ be the same way as CLI v3.10.1 and prior. For instance, maybe an output file is also specified with the `--json` flag.

kylewillmon:
There were two places where we might print to stdout when running `phylum analyze --json`. One of them is printing the Job ID. This patch avoids printing in that case. The other is identifying a previously unknown lockfile. This patch sends that output to stderr.

Closes #769

### DIFF
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -420,10 +420,10 @@ async fn parse_lockfile(
     let lockfile_type = match lockfile_type {
         Some(lockfile_type) => lockfile_type,
         None => {
-            let (packages, package_type) = parse::get_packages_from_lockfile(Path::new(&lockfile))?;
+            let parsed = parse::get_packages_from_lockfile(Path::new(&lockfile))?;
             return Ok(PackageLock {
-                package_type,
-                packages: packages.into_iter().map(PackageSpecifier::from).collect(),
+                package_type: parsed.package_type,
+                packages: parsed.packages.into_iter().map(PackageSpecifier::from).collect(),
             });
         },
     };
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -9,7 +9,12 @@ use phylum_lockfile::{get_path_format, LockfileFormat};
 use phylum_types::types::package::{PackageDescriptor, PackageType};
 
 use crate::commands::{CommandResult, ExitCode};
-use crate::{print_user_success, print_user_warning};
+
+pub struct ParsedLockfile {
+    pub format: LockfileFormat,
+    pub packages: Vec<PackageDescriptor>,
+    pub package_type: PackageType,
+}
 
 pub fn lockfile_types() -> Vec<&'static str> {
     LockfileFormat::iter().map(|format| format.name()).chain(["auto"]).collect()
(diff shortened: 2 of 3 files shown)


maxrake (approved):
This was tested locally and appears to work. LGTM.

cli/src/commands/parse.rs
```diff
@@ -48,7 +48,7 @@ pub fn try_get_packages(path: &Path) -> Result<(Vec<PackageDescriptor>, PackageT
         let parser = format.parser();
         if let Ok(pkgs) = parser.parse(data.as_str()) {
             if !pkgs.is_empty() {
-                print_user_success!("Identified lockfile type: {}", format);
+                print_user_warning!("Identified lockfile type: {}", format);
```
cd-work, line 51: This feels very wrong? We shouldn't use `warning` only because it prints to stderr instead of stdout. That isn't really a solution, that's just a hacky workaround.

cd-work (changes requested):

cli/src/commands/parse.rs
maxrake: Maybe if the `parse` command had a `--json` flag like the other commands that can produce JSON output? Would that help...like to use the `pretty_print` predicate as is done elsewhere?

kylewillmon: The output in this function definitely feels a bit weird... Perhaps these should just be log messages rather than `print` or `eprint`?

cd-work: Yeah logging was my first thought too, but we **definitely** want to print which lockfile was identified and only error and warn are printed by default.

kylewillmon: Alright. I've updated the code to use a `pretty_print` condition and continue to use stdout.

cli/src/commands/parse.rs
```diff
@@ -36,19 +36,19 @@ pub fn handle_parse(matches: &clap::ArgMatches) -> CommandResult {
     Ok(ExitCode::Ok.into())
 }
 /// Attempt to get packages from an unknown lockfile type
-pub fn try_get_packages(path: &Path) -> Result<(Vec<PackageDescriptor>, PackageType)> {
-    print_user_warning!(
-        "Attempting to obtain packages from unrecognized lockfile type: {}",
-        path.to_string_lossy()
-    );
-
```
kylewillmon, line 44: This warning was misplaced here because this code gets run in the `--lockfile-type auto`. In that case, we don't even attempt to check if it's a recognized lockfile and we are using behavior explicitly requested by the user, so a warning is unnecessary.

cli/src/commands/extensions/api.rs
```diff
@@ -420,7 +420,8 @@ async fn parse_lockfile(
     let lockfile_type = match lockfile_type {
         Some(lockfile_type) => lockfile_type,
         None => {
-            let (packages, package_type) = parse::get_packages_from_lockfile(Path::new(&lockfile))?;
+            let (packages, package_type) =
+                parse::get_packages_from_lockfile(Path::new(&lockfile), false)?;
```
kylewillmon, line 424: I chose to err on the side of caution here because extensions should probably retain control over what gets printed to stdout. But maybe we should expose this as an option in the extension API?

maxrake (approved):

cd-work (commented):
Have you checked if it would be possible to move the actual output to stderr to a higher level, ideally the `handle_whatever` part of the code?

I think generally that might be a smart way to approach things, considering the "deeper" functions are usually likely to be shared and uniform stdout/stderr output might not be sensible. At that level logging might be the better choice.

cd-work (approved):

PR #774, opened 2022-10-21
Fix --package-type option

kylewillmon:
Fix the `--package-type` option by avoiding the call to `get_flag` (since this option is not a flag).

### DIFF
--- a/cli/src/commands/packages.rs
+++ b/cli/src/commands/packages.rs
@@ -21,7 +21,7 @@ fn parse_package(matches: &ArgMatches, request_type: PackageType) -> Result<Pack
     // to the global setting
     let package_type = match matches.get_one::<String>("package-type") {
         Some(pt) => {
-            PackageType::from_str(pt).map_err(|()| anyhow!("invalid package type: {}", pt))?
+            PackageType::from_str(pt).map_err(|_| anyhow!("invalid package type: {}", pt))?
         },
         None => request_type,
     };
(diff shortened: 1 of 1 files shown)


maxrake (dismissed):
Confirmed with a local build and testing. LGTM.

cli/src/commands/packages.rs
```diff
@@ -11,28 +12,28 @@ use crate::filter::{Filter, FilterIssues};
...
     // If a package type was provided on the command line, prefer that
-    //  to the global setting
-    if options.get_flag("package-type") {
-        package_type = PackageType::from_str(options.get_one::<String>("package-type").unwrap())
-            .unwrap_or(package_type);
-    }
+    // to the global setting
+    let package_type = match matches.get_one::<String>("package-type") {
+        Some(pt) => {
+            PackageType::from_str(pt).map_err(|()| anyhow!("invalid package type: {}", pt))?
```
cd-work, line 24:
```suggestion
            PackageType::from_str(pt).map_err(|_| anyhow!("invalid package type: {}", pt))?
```

cd-work (changes requested):

cli/src/commands/packages.rs
kylewillmon: My goal in using `()` was to highlight the fact that the error type is `()`. If it were any other reasonable error type, I wouldn't have to use `map_err`. And if we change the error type in the future (and we should), then we should reconsider the use of `map_err` here.

cd-work: I'm not sure anyone reading this in the future will get that from just looking at the `()`? Like I understand where you're coming from with the intention, but at the end of the day all this really does is reduce code consistency.

cd-work (approved):

maxrake (approved):

PR #777, opened 2022-10-24
Add `unsandboxed_run` manifest permission

cd-work:
This adds a new permission specifically for `Deno.run`, which makes it
clear that executing commands with `Deno.run` will escape the sandbox
and not have any safety restrictions applied.


cd-work:
One thing I noticed while writing this is that currently our `run` permission behaves somewhat strangely. Since we add everything from `$PATH` by default for example, you will not need a `run` permission at all for anything you specify with a relative path.

This is why I have also just copy/pasted the documentation for now. I think we should probably rethink our standard `run` permission somewhat. Ultimately I don't think this is an issue since `run` fundamentally isn't dangerous as long as it's inside the sandbox. But it's certainly confusing.

cli/src/commands/extensions/api.rs
```diff
@@ -100,6 +100,7 @@ impl From<ProcessException> for permissions::Permissions {
             run: process_exception.run,
             env: process_exception.env,
             net: Permission::Boolean(process_exception.net),
+            unsandboxed_run: Permission::default(),
```
andreaphylum, line 103: I assume this would default to `false`; why is that? Wouldn't we rather want to augment `ProcessException` here?

cli/src/commands/extensions/permissions.rs
```diff
@@ -228,6 +231,7 @@ impl Permissions {
             env: self.env.subset_of(&other.env).map_err(err_ctx("env"))?,
             run: self.run.subset_of(&other.run).map_err(err_ctx("run"))?,
             net: self.net.subset_of(&other.net).map_err(err_ctx("net"))?,
+            unsandboxed_run: Permission::default(),
```
andreaphylum, line 234: Can we put a comment here clarifying the reason of this being different from the above? I _assume_ that it's because it has nothing to do with `birdcage` and doesn't carry over in the sandbox layering, but I may be missing something and it took a bit of time for me to get to that conclusion.

cli/src/commands/extensions/permissions.rs
```diff
@@ -242,7 +246,7 @@ impl From<&Permissions> for PermissionsOptions {
 
         let allow_env = value.env.get().cloned();
         let allow_net = value.net.get().cloned();
-        let allow_run = value.run.get().cloned();
+        let allow_run = value.unsandboxed_run.get().cloned();
```
andreaphylum, line 249: Does this replace `run` with `unsandboxed_run` because those `run` permissions are actually managed by Deno?

andreaphylum (changes requested):
Looks good overall, there are a couple places where I'd add clarifying comments as I find some choices and the interaction with the previos `run` permissions difficult to understand.

Besides fixing the struct initializations in the tests, I'd also add a few more test cases to demo what `unsandboxed_run` can and cannot do.

cli/src/commands/extensions/api.rs
cd-work: I'm not sure what you mean? `ProcessException` only applies to `runSandboxed` and as such `unsandboxed_run` should have no impact on it?

cli/src/commands/extensions/permissions.rs
cd-work: Yeah `unsandboxed_run` is **only** for `Deno.run`, so obviously we don't want it anywhere near sandboxed run.

cli/src/commands/extensions/permissions.rs
cd-work: It just renames the `run` permissions used by deno in our manifest to `unsandboxed_run` to make it a little more clear these do not use any sandboxing.

cli/src/commands/extensions/mod.rs
```diff
@@ -276,6 +276,7 @@ fn ask_permissions(extension: &Extension) -> Result<()> {
     print_permissions_list("Read", "path", permissions.read.get());
     print_permissions_list("Write", "path", permissions.write.get());
     print_permissions_list("Run", "command", permissions.run.get());
+    print_permissions_list("Run", "unsandboxed command", permissions.unsandboxed_run.get());
```
kylewillmon, line 279: Should we do something to make this stand out? Like make it print in red or something?

kylewillmon (approved):

cli/tests/fixtures/extensions/permissions/incorrect-run-perms/PhylumExt.toml
```diff
@@ -0,0 +1,3 @@
+name = "incorrect-run-perms"
+description = "An extension with correctly specified permissions"
```
kylewillmon, line 2: As far as I can tell, this extension isn't actually called from any tests.

Additionally, the description and name are in conflict (`correctly specified permissions` vs. `incorrect-run-perms`)

kylewillmon (changes requested):

andreaphylum (dismissed):

cli/src/commands/extensions/mod.rs
cd-work: All our permission requests are already somewhat highlighted in color. I feel like all of them should be read carefully anyway so there's no immediate **necessity** to highlight it even more. Though it is the most dangerous permission.

cli/tests/fixtures/extensions/permissions/incorrect-run-perms/PhylumExt.toml
cd-work: I recently added a new test and contemplated if it was worth to add a fixture and decided against it myself.

Imho it's just additional overhead at this point to check the extension fixture directory rather than having everything relevant defined right in the test.

kylewillmon: > Imho it's just additional overhead at this point to check the extension fixture directory rather than having everything relevant defined right in the test.

I agree. So this `incorrect-run-perms` extension is going away?

kylewillmon (dismissed):

kylewillmon (approved):

PR #784, opened 2022-10-27
Remove $PATH exception for `run` permission

cd-work:
This removes the default permission to run anything in the `$PATH` when calling `runSandboxed`. Executables now have to be specified manually instead.

The existing extensions have been tested by trying to add one remote and one local dependency to the project. Additional sandboxing exceptions have been added where necessary.


cd-work:
This has honestly gotten a little more complicated than I originally expected. Primarily with updating all our extensions. I'm fairly confident with the `yarn` and `npm` stuff since I wrote that myself, but the `poetry` extensions seemed somewhat janky since it didn't work for me before this patch.

Would probably be good to have some additional extension testing, specifically on macOS.

andreaphylum (changes requested):
`npm` and `yarn` look good to me.

Wrt `poetry`:
- It wouldn't work for me until I explicitly added `python3` to the execution permissions. Most likely, `poetry` internally uses some discovery method that both on my latest ubuntu and on osx resolved to `python3`.
- On OS X, the cache path is not XDG but rather `~/Library/Caches`.
- One directory was present in the `runSandboxed` write permissions but not in the manifest.
- `poetry install` fails at the updating lockfile stage because the `--lock` flag doesn't exist (anymore?).

I added suggestions to deal with all of these but the last one. Both `poetry add` and `poetry update` work as intended. For the last one, I'd open a separate ticket.

andreaphylum (approved):

PR #786, opened 2022-10-27
`phylum auth set-token`

kylewillmon:
This patch adds a command to set the token in the config file. This provides an alternative to `phylum auth login` provided you can get the token from some other place. The other place may be `phylum auth token` from another machine or, in the future, we may support copying the token directly from the web UI.

### DIFF
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -88,28 +88,26 @@ pub async fn handle_auth(
     app_helper: &mut Command,
     timeout: Option<u64>,
 ) -> CommandResult {
-    if matches.subcommand_matches("register").is_some() {
-        match handle_auth_register(config, config_path).await {
+    match matches.subcommand() {
+        Some(("register", _)) => match handle_auth_register(config, config_path).await {
             Ok(_) => {
                 print_user_success!("{}", "User successfuly regsistered");
                 Ok(ExitCode::Ok.into())
             },
             Err(error) => Err(error).context("User registration failed"),
-        }
-    } else if matches.subcommand_matches("login").is_some() {
-        match handle_auth_login(config, config_path).await {
+        },
+        Some(("login", _)) => match handle_auth_login(config, config_path).await {
             Ok(_) => {
                 print_user_success!("{}", "User login successful");
                 Ok(ExitCode::Ok.into())
             },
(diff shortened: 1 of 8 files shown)


furi0us333 (changes requested):
We should update the docs for this new subcommand

cli/src/commands/auth.rs
```diff
@@ -88,28 +109,27 @@ pub async fn handle_auth(
     app_helper: &mut Command,
     timeout: Option<u64>,
 ) -> CommandResult {
-    if matches.subcommand_matches("register").is_some() {
-        match handle_auth_register(config, config_path).await {
+    match matches.subcommand() {
```
andreaphylum, line 112: I always like it when `if` chains become `match` blocks. 💯

andreaphylum (approved):

maxrake (approved):
The behavior was confirmed with a local build and tested with both a token provided as an option and one provided from stdin. Other than adding documentation, LGTM.

maxrake (approved):

furi0us333 (approved):
LGTM

docs/command_line_tool/phylum_auth.md
```diff
@@ -9,5 +9,6 @@ Manage authentication, registration, and API keys
 ### Commands
 * [phylum auth login](https://docs.phylum.io/docs/phylum_auth_login)
 * [phylum auth register](https://docs.phylum.io/docs/phylum_auth_register)
+* [phylum auth set-token](https://docs.phylum.io/docs/phylum_auth_set-token)
```
maxrake, line 12:
```suggestion
* [phylum auth set-token](https://docs.phylum.io/docs/phylum_auth_set_token)
```
The other possibility is to change the file name to `docs/command_line_tool/phylum_auth_set-token.md`...which would match the other command/page with a hyphenated name: `docs/command_line_tool/phylum_project_set-thresholds.md`

The working assumption here is the file name is what creates the page URL/slug.

maxrake (changes requested):
The new link or the new doc filename needs to change.

maxrake (approved):

cd-work (dismissed):

cd-work (approved):

PR #787, opened 2022-10-27
fix: avoid stdout when run with --json

### LINKED ISSUE #769: `--json` flag does not produce JSON output

maxrake:
This is a continuation of #773, to resolve #769

## Screenshots

Behavior before this change:

image

---

Behavior after this change:

image

### DIFF
--- a/cli/src/commands/jobs.rs
+++ b/cli/src/commands/jobs.rs
@@ -136,7 +136,7 @@ pub async fn handle_submission(api: &mut PhylumApi, matches: &clap::ArgMatches) 
         let res = get_packages_from_lockfile(Path::new(lockfile))
             .context("Unable to locate any valid package in package lockfile")?;
 
-        if !pretty_print {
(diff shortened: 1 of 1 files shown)


kylewillmon (approved):
🤦‍♂️

PR #791, opened 2022-10-31
Add ignore scripts when updating package-lock.json

louislang:
Updates the read permissions to limit access to disk from the npm extension. Also adds the `--ignore-scripts` to the lock file generation.


kylewillmon (changes requested):
This should be split into 2 PRs. The addition of `--ignore-scripts` is simple and we can merge that quickly. Changing the read permissions when executing scripts is significantly more complicated and will take more time and testing (and must be reconciled with #784)

kylewillmon (approved):

PR #792, opened 2022-11-01
Sandbox outer error management implementation with IPC file

### LINKED ISSUE #781: Throw exception for failure to create a process in `PhylumApi.runSandboxed()`
Calling `PhylumApi.runSandboxed()` does not distinguish between failure to create process and process exit failure. For example, this code:

```
let output = PhylumApi.runSandboxed({cmd: "blargle", args: [], exceptions: {}})
console.log(`output: success=${output.success} code=${output.code}`)
```

Gives the following output:

```
❗ Error: No such file or directory (os error 2)
output: success=false code=1
```

The exit code of 1 here is misleading because the `blargle` process was never actually created. (Instead, this refers to the exit code of `phylum sandbox`).

When using `Deno.run`, failure to create a process results in an exception. We should probably also throw an exception in this case.

_This was discovered during analysis of a separate issue in [this ticket](https://github.com/phylum-dev/cli/issues/775#issuecomment-1293807879)_

andreaphylum:
A proposed solution to #781.

`run_sandboxed` creates a temporary directory, and the `phylum sandbox` command writes to a file in it to signal failures that aren't coming from its child process. The file has to be created and opened before the sandbox is locked, but can be accessed afterwards as long as its file descriptor is not closed.

Drawback: I/O based communication which introduces errors that can't be solved by more I/O based communication. This means that there are some top-level failure cases which won't be distinguishable from child process failures (i.e. couldn't create the IPC file). This could be acceptable as I doubt it's ever going to be an attack vector (the consequences of a miscommunication here aren't security sensitive) and the conditions leading to a failure (missing write permission to a temporary path, full disk, other write errors) are most likely going to affect other parts of this flow as well.

Opening as draft PR in order to gather ideas and feedback before moving forward.

Closes #781.

## Checklist

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -354,6 +354,12 @@ pub fn add_subcommands(command: Command) -> Command {
     {
         app = app.subcommand(
             Command::new("sandbox").hide(true).about("Run an application in a sandbox").args(&[
+                Arg::new("ipc-path")
+                    .help("Set directory to store IPC messages in")
+                    .long("ipc-path")
+                    .value_name("PATH")
+                    .value_hint(ValueHint::FilePath)
+                    .action(ArgAction::Set),
                 Arg::new("allow-read")
                     .help("Add filesystem read sandbox exception")
                     .long("allow-read")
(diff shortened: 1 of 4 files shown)


kylewillmon:
As [noted in the issue](https://github.com/phylum-dev/cli/issues/781#issuecomment-1307637398), we have decided to go a different direction to resolve #781

PR #797, opened 2022-11-02
Require "selfmanage" feature flag for `phylum update`

### LINKED ISSUE #782: Allow disabling self-update
Many package managers, such as Homebrew, [don't like tools that upgrade themselves](https://docs.brew.sh/Acceptable-Formulae#we-dont-like-tools-that-upgrade-themselves).

We hope to support Homebrew soon, so we need a way to disable `phylum update` (and `phylum uninstall`) when installed by Homebrew.

I think the simplest way to accomplish this is with a feature flag, but I am open to other suggestions.

kylewillmon:
Fixes #782 by putting `phylum update` behind the "selfmanage" feature flag.

### DIFF
--- a/cli/src/bin/phylum.rs
+++ b/cli/src/bin/phylum.rs
@@ -1,5 +1,5 @@
 use std::fmt::Display;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::time::UNIX_EPOCH;
 
 use anyhow::{anyhow, Context, Result};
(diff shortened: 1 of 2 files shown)


cli/src/bin/phylum.rs
```diff
@@ -103,29 +130,9 @@ async fn handle_commands() -> CommandResult {
...
 
-    // Check for updates, if we haven't explicitly invoked `update`.
-    //
-
-    if matches.subcommand_matches("update").is_none() {
-        let now = UNIX_EPOCH.elapsed().expect("Time went backwards").as_secs() as usize;
-
-        let check_for_updates = config.last_update.map_or(true, |last_update| {
-            const SECS_IN_DAY: usize = 24 * 60 * 60;
-            now - last_update > SECS_IN_DAY
```
kylewillmon, line 114: Since `last_update` comes from the config, there was a chance this could underflow if somebody threw a big number in there (e.g., maybe just trying to disable the update check). I switched this to a `.saturating_sub()` so that it evaluates to 0 if `last_update > now`.

cd-work (approved):

PR #801, opened 2022-11-08
Add `phylum init` subcommand

### LINKED ISSUE #795: `phylum init` command
The `phylum init` command will be an interactive prompt that assists users in setting up a new project. The command should automatically determine the project name from the basename of the current working directory, but should prompt to confirm and allow the user to enter another name if desired.

Something like this:
```
kyle@computer:/users/kyle/code/fooproj$ phylum init
Creating a new Phylum project!
Enter a project name (fooproj): 

✅ Successfully created project "fooproj"
```

cd-work:
This adds a new subcommand for setting up a new Phylum project.

Currently this command will only negotiate the desired project name and corresponding group, but further information like lockfile can be detected in the future.

Project and group can either be specified directly in the CLI, or specified interactively.

The user is not prompted for a group name when passing the project name on the CLI, to enable non-interactive usage without passing extra CLI flags. So in this scenario a group must be specified with `--group`.

Closes #795.

### DIFF
--- a/cli/src/commands/init.rs
+++ b/cli/src/commands/init.rs
@@ -4,7 +4,7 @@ use std::io;
 
 use clap::ArgMatches;
 use dialoguer::theme::ColorfulTheme;
-use dialoguer::{Input, Confirm};
+use dialoguer::{Confirm, Input};
 
 use crate::api::PhylumApi;
 use crate::commands::{project, CommandResult};
@@ -36,10 +36,8 @@ fn prompt_project() -> io::Result<String> {
 
 // Ask for the desired group.
 fn prompt_group() -> io::Result<Option<String>> {
-    let should_prompt = Confirm::new()
-        .with_prompt("Use a project group?")
-        .default(false)
-        .interact()?;
+    let should_prompt =
+        Confirm::new().with_prompt("Use a project group?").default(false).interact()?;
 
     if !should_prompt {
         return Ok(None);
--- a/cli/src/commands/init.rs
+++ b/cli/src/commands/init.rs
@@ -29,17 +29,7 @@ pub async fn handle_init(api: &mut PhylumApi, matches: &ArgMatches) -> CommandRe
     let cli_group = matches.get_one::<String>("group");
 
     // Interactively prompt for missing information.
-    let (project, group) = match cli_project {
-        Some(project) => (project.clone(), cli_group.cloned()),
-        None => {
-            let project = prompt_project()?;
-            let group = match cli_group {
-                Some(group) => Some(group.clone()),
-                None => prompt_group()?,
-            };
-            (project, group)
-        },
-    };
+    let (project, group) = prompt(cli_project, cli_group)?;
 
     // Attempt to create the project.
     let response = project::create_project(api, &project, group.clone()).await?;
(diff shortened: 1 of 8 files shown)


cli/src/commands/init.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::api::PhylumApi;
+use crate::commands::{project, CommandResult};
+
+/// Handle `phylum init` subcommand.
+pub async fn handle_init(api: &mut PhylumApi, matches: &ArgMatches) -> CommandResult {
+    let cli_project = matches.get_one::<String>("project");
+    let cli_group = matches.get_one::<String>("group");
+
+    let (project, group) = match cli_project {
+        Some(project) => (project.clone(), cli_group.cloned()),
```
andreaphylum, line 18: Why is the `clone()` on `project` necessary? I see only a reference to it is used afterwards and it doesn't look like it's getting moved elsewhere, but I may be missing something or it may be related to `project`'s type.

cli/src/commands/init.rs
```diff
@@ -0,0 +1,51 @@
...
+        None => {
+            let project = prompt_project()?;
+            let group = match cli_group {
+                Some(group) => Some(group.clone()),
+                None => prompt_group()?,
+            };
```
andreaphylum, line 24: I feel like this part could also be made a bit clearer by chaining combinators, or switching to `if let`.

andreaphylum (changes requested):
Looks good, just a couple non-mandatory things I'd change for improving clarity.

cli/src/commands/init.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use crate::api::PhylumApi;
+use crate::commands::{project, CommandResult};
+
+/// Handle `phylum init` subcommand.
+pub async fn handle_init(api: &mut PhylumApi, matches: &ArgMatches) -> CommandResult {
+    let cli_project = matches.get_one::<String>("project");
+    let cli_group = matches.get_one::<String>("group");
+
+    let (project, group) = match cli_project {
```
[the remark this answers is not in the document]
cd-work: I'd say `if let ... else` is basically always the wrong choice. Generally I'd always recommend `match` over `let` if there's an `else` branch for the sake of consistency.

I think extracting it into a function works nicely though.

cli/src/commands/init.rs
cd-work: Just because I can't get a `&String` back from prompt_project, so branch consistency basically.

I could wrap it in a cow if we really wanted to, but that's just pointless complexity really for a command that executes interactively in a couple of seconds once.

andreaphylum: I honestly wouldn't mind `clone`s even in hot paths as optimizations _should_ take care of that anyway more often than not, so that's fine.

andreaphylum (approved):

cli/src/commands/init.rs
cd-work: Getting `clone`s optimized away is extremely rare, so I wouldn't rely on that *at all*. But yeah in the CLI we don't really have any hot paths anyway.

PR #802, opened 2022-11-08
Clarify connection between read and run permissions

### LINKED ISSUE #785: Clarify run permission read access
Currently the `run` sandbox permissions always automatically give read access too. However the permissions do not make this clear to the user. This is especially impactful when the `run: true` permission is specified.

At the very least this should be communicated to the user appropriately, rather than just pointing out the run permissions requested.

kylewillmon:
Update the docs and install code to clarify the connection between read and run permissions.

Fixes #785

### DIFF
--- a/cli/src/commands/extensions/mod.rs
+++ b/cli/src/commands/extensions/mod.rs
@@ -276,11 +276,12 @@ fn ask_permissions(extension: &Extension) -> Result<()> {
 
     // Calculate effective read permissions
     let read = match (&permissions.read, &permissions.run) {
-        (Permission::Boolean(true), _) => Permission::Boolean(true),
-        (_, Permission::Boolean(true)) => Permission::Boolean(true),
+        (Permission::Boolean(true), _) | (_, Permission::Boolean(true)) => {
+            Permission::Boolean(true)
+        },
         (Permission::Boolean(false), run) => run.clone(),
         (read, Permission::Boolean(false)) => read.clone(),
-        (Permission::List(ref read), Permission::List(ref run)) => {
+        (Permission::List(read), Permission::List(run)) => {
             let mut read = read.clone();
             for path in run {
                 if !read.contains(path) {
--- a/cli/src/commands/extensions/mod.rs
+++ b/cli/src/commands/extensions/mod.rs
@@ -15,6 +15,7 @@ use log::error;
 
 use crate::api::PhylumApi;
 use crate::commands::extensions::extension::{Extension, ExtensionManifest};
+use crate::commands::extensions::permissions::Permission;
 use crate::commands::{CommandResult, CommandValue, ExitCode};
 use crate::print::print_sc_help;
 use crate::{app, print_user_success, print_user_warning};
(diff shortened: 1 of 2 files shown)


cli/src/commands/extensions/mod.rs
```diff
@@ -273,9 +274,16 @@ fn ask_permissions(extension: &Extension) -> Result<()> {
         }
     }
 
-    print_permissions_list("Read", "path", permissions.read.get());
+    // Special case `run = true` to always say "Read any path"
+    if let Permission::Boolean(true) = &permissions.run {
```
cd-work, line 278: I don't understand this special case? What's the goal here? Just to reduce the duplication between run and read? Seems weird if we don't have that consistently for everything.

cli/src/commands/extensions/mod.rs
```diff
@@ -273,7 +274,18 @@ fn ask_permissions(extension: &Extension) -> Result<()> {
         }
     }
 
-    print_permissions_list("Read", "path", permissions.read.get());
+    // Calculate effective read permissions
+    let read = match (&permissions.read, &permissions.run) {
+        (Permission::Boolean(true), _) => Permission::Boolean(true),
+        (_, Permission::Boolean(true)) => Permission::Boolean(true),
```
kylewillmon, line 280: Is this case worthy of a warning to let the extension author know that `run = true` implies `read = true` (so they might as well set `read = true`)?

We're in the extension install code here, which might not be the best place for that warning. But I can't think of any other place to put it...

cd-work: I have similar concerns here as I had for the other special case. Why handle `true` differently than anything else? Should we warn if one path is in `run` but not in `read`? Should either go all or nothing imo.

cd-work: I mean that's like saying there's no legitimate reason for `run = ['./test']; read = ['./test']` though?

An author that uses booleans for `run` and `read` will probably just add it to one or the other in order to avoid duplication.

kylewillmon: That's a fair point. I will not go forward with any warnings.

cli/src/commands/extensions/mod.rs
```diff
@@ -273,7 +274,24 @@ fn ask_permissions(extension: &Extension) -> Result<()> {
...
```
cd-work, line 280:
```suggestion
        (Permission::Boolean(true), _) | (_, Permission::Boolean(true)) => Permission::Boolean(true),
```

I don't really mind it either way, but this would technically reduce the duplicate `Permission::Boolean(true)` and thus make it easier to change or add new branches in the future.

cli/src/commands/extensions/mod.rs
```diff
@@ -273,7 +274,24 @@ fn ask_permissions(extension: &Extension) -> Result<()> {
...
+        (Permission::Boolean(false), run) => run.clone(),
+        (read, Permission::Boolean(false)) => read.clone(),
+        (Permission::List(ref read), Permission::List(ref run)) => {
```
cd-work, line 283: You don't need `ref` here, these are already borrowed due to the `&permissions.read`.

cd-work (changes requested):

cli/src/commands/extensions/mod.rs
kylewillmon: `rustfmt` split the line because it's too long after making this change. But it's still probably better this way.

I suppose we could do the same thing for the `false` case. That is:

```
        (Permission::Boolean(false), other) | (other, Permission::Boolean(false)) => other.clone(),
```

But binding on both sides of an `|` might be confusing to read.

maxrake (approved):

PR #805, opened 2022-11-10
update phylum-types

JosephPhylum:
update phylum-types to support "phylum package -t cargo" previously it was expecting -t rust, which is inconsistent with the rest of the options, and inconsistent with the help description. 

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3182,7 +3182,7 @@ dependencies = [
 [[package]]
 name = "phylum_types"
 version = "0.1.0"
-source = "git+https://github.com/phylum-dev/phylum-types?branch=development#60bd7fbd04a6303764e6225f8e16e6169ea1ebd1"
+source = "git+https://github.com/phylum-dev/phylum-types?branch=development#138ab11a4db60313adea6cc9cbd57a3731c7eb4a"
 dependencies = [
  "chrono",
  "schemars",


kylewillmon:
> update phylum-types to support "phylum package -t cargo" previously it was expecting -t rust, which is inconsistent with the rest of the options, and inconsistent with the help description.

`phylum package -t cargo` is also inconsistent with the rest of the options... Here are the current options (from the help output)

```
  -t, --package-type <type>  The type of the package ("npm", "rubygems", "pypi", "maven", "nuget", "golang", "cargo")
```

As can be seen, the existing options refer to the registry on which the package resides:

* `npm` -> `npmjs.org`
* `rubygems` -> `rubygems.org`
* `pypi` -> `pypi.org`
* `maven` -> `maven.org`
* `nuget` -> `nuget.org`

It makes sense to refer to the registry because that is where a given package name and version can be resolved. Outside of a registry (or even inside of a poorly designed registry...), there may be multiple packages with the same name and version.

The new options of `golang` and `cargo` do not follow this trend. To follow the trend, these should probably be `go` (for `go.dev`) and `crates-io` (for `crates.io`) or something to that effect.

kylewillmon (changes requested):
The new behavior matches the help output, but is still inconsistent with other package types.

ejortega:
> The new behavior matches the help output, but is still inconsistent with other package types.

I believe @JosephPhylum was just following https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst which `rubygems` doesn't follow.

Do we want to update the ecosystem name as well to reflect the registry?

@louislang @furi0us333

furi0us333:
My understanding is that we were going to try and follow purl, will leave that to @louislang for the final decision

louislang:
I totally see the logic in using `go` and `cargo-io`, but lets standardize on purl across the platform.

kylewillmon:
That makes sense and it is fortunate that we already use most of the correct names for PURL.

The only exception is `rubygems` should be `gem`. Is there a ticket somewhere to track that switch? If not, I will create one.

kylewillmon (approved):

louislang:
I don't think we've got a ticket tracking that. We should definitely make one @kylewillmon!

kylewillmon:
Done. phylum-dev/phylum-types#61

Issue #793, opened 2022-11-01
Lockfile path in `.phylum_project` file


kylewillmon:
The `phylum analyze` command currently pulls project and group information from the `.phylum_project` file (if not provided on the command line). In contrast, the lockfile path must be specified on the command line. This should be fixed.

## Acceptance criteria
  - [ ] Running `phylum analyze` with no arguments submits the lockfile specified in `.phylum_project`

## Notes
We can currently only submit one lockfile at a time, but that is likely to change. We should prepare for this to be a list of lockfiles in the future.

PR #806, opened 2022-11-11
Add lockfile path and type to .phylum_project

cd-work:
This patch extends the existing .phylum_project with two new optional
fields: `lockfile` and `lockfile_type`.

When the `lockfile` type is present in the .phylum_project file and no
lockfile override is specified on the CLI, then the project file from
the .phylum_project file will be used with the lockfile type specified
in the same file.

If the lockfile type is specified on the CLI without the lockfile being
specified, it will be ignored.

The lockfile name and type can be set when running `phylum init`, either
by passing `--lockfile` and (optionally) `--lockfile-type`, or by
specifying them interactively.

Closes https://github.com/phylum-dev/cli/issues/793.

### DIFF
--- a/cli/src/commands/init.rs
+++ b/cli/src/commands/init.rs
@@ -12,7 +12,7 @@ use reqwest::StatusCode;
 use crate::api::{PhylumApi, PhylumApiError, ResponseError};
 use crate::commands::{project, CommandResult, ExitCode};
 use crate::config::PROJ_CONF_FILE;
-use crate::{config, print_user_failure, print_user_success, print_user_warning};
+use crate::{config, print_user_success, print_user_warning};
 
 /// Handle `phylum init` subcommand.
 pub async fn handle_init(api: &mut PhylumApi, matches: &ArgMatches) -> CommandResult {
@@ -43,8 +43,11 @@ pub async fn handle_init(api: &mut PhylumApi, matches: &ArgMatches) -> CommandRe
     // If project already exists, just link to it.
     let mut project_config = match result {
         Err(PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. })) => {
+            let project_config = project::link_project(api, &project, group)
+                .await
+                .context("Unable to link project")?;
             print_user_success!("Successfully linked to project {project:?}");
-            project::link_project(api, &project, group).await.context("Unable to link project")?
+            project_config
         },
         project_config => {
             if project_config.is_ok() {
@@ -64,9 +67,8 @@ pub async fn handle_init(api: &mut PhylumApi, matches: &ArgMatches) -> CommandRe
     project_config.lockfile = Some(lockfile);
 
     // Save project config.
-    config::save_config(Path::new(PROJ_CONF_FILE), &project_config).unwrap_or_else(|err| {
-        print_user_failure!("Failed to save project file: {}", err);
-    });
+    config::save_config(Path::new(PROJ_CONF_FILE), &project_config)
+        .context("Failed to save project file")?;
 
     Ok(ExitCode::Ok.into())
 }
@@ -128,12 +130,16 @@ fn prompt_lockfile(
     cli_lockfile: Option<&String>,
     cli_lockfile_type: Option<&String>,
 ) -> io::Result<(String, Option<String>)> {
-    if let Some(lockfile) = cli_lockfile {
-        return Ok((lockfile.clone(), cli_lockfile_type.cloned()));
-    }
+    let cli_lockfile_type = cli_lockfile_type.cloned();
 
-    // Prompt for lockfile name.
-    let lockfile = prompt_lockfile_name()?;
+    let lockfile = match (cli_lockfile.cloned(), &cli_lockfile_type) {
+        // Do not prompt if name and type were specified on CLI.
+        (Some(lockfile), Some(_)) => return Ok((lockfile, cli_lockfile_type)),
+        // Prompt for type if only lockfile was passed.
+        (Some(lockfile), _) => lockfile,
+        // Prompt for lockfile if it wasn't specified.
+        (None, _) => prompt_lockfile_name()?,
+    };
 
     // Try to determine lockfile name from known lockfiles.
     for format in LockfileFormat::iter() {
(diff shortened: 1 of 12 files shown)


andreaphylum (approved):

cli/src/commands/init.rs
```diff
@@ -23,28 +27,52 @@ pub async fn handle_init(api: &mut PhylumApi, matches: &ArgMatches) -> CommandRe
...
+    let result = project::create_project(api, &project, group.clone()).await;
 
     // If project already exists, just link to it.
-    match response {
-        CommandValue::Code(ExitCode::AlreadyExists) => {
-            project::link_project(api, &project, group).await
+    let mut project_config = match result {
+        Err(PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. })) => {
+            print_user_success!("Successfully linked to project {project:?}");
+            project::link_project(api, &project, group).await.context("Unable to link project")?
```
kylewillmon, line 47:
```suggestion
            project::link_project(api, &project, group).await.context("Unable to link project")?
            print_user_success!("Successfully linked to project {project:?}");
```

It's probably a good idea to swap the order of these lines to avoid the possibility of:
```
✅ Successfully linked to project foo
❗ Unable to link to project ...
```

cli/src/commands/init.rs
```diff
@@ -23,28 +27,52 @@ pub async fn handle_init(api: &mut PhylumApi, matches: &ArgMatches) -> CommandRe
...
+    // Add lockfile and its type to the project configuration.
+    println!();
+    let (lockfile, lockfile_type) = prompt_lockfile(cli_lockfile, cli_lockfile_type)?;
+    project_config.lockfile_type = lockfile_type;
+    project_config.lockfile = Some(lockfile);
+
+    // Save project config.
+    config::save_config(Path::new(PROJ_CONF_FILE), &project_config).unwrap_or_else(|err| {
+        print_user_failure!("Failed to save project file: {}", err);
+    });
```
kylewillmon, line 69: It definitely seems like this should cause the entire command to fail rather than just printing the failure

```suggestion
    // Save project config.
    config::save_config(Path::new(PROJ_CONF_FILE), &project_config).context("Failed to save project file")?;
```

cli/src/commands/init.rs
```diff
@@ -90,9 +117,79 @@ fn prompt_group() -> io::Result<Option<String>> {
...
+            .interact()?;
+
+        // Return selected lockfile unless `other` was chosen.
+        if index + 1 != lockfiles.len() {
+            return Ok(lockfiles.remove(index));
+        }
+    }
+
+    // Prompt for lockfile name if none was selected.
+    Input::new().with_prompt("Project lockfile name").interact_text()
```
kylewillmon, line 182: This might be clearer if it asked for the `path` rather than the `name` so that the user doesn't think this is just some arbitrary name.

```suggestion
    Input::new().with_prompt("Project lockfile path").interact_text()
```

cli/src/commands/init.rs
```diff
@@ -90,9 +117,79 @@ fn prompt_group() -> io::Result<Option<String>> {
...
 }
+
+/// Interactively ask for missing lockfile information.
+fn prompt_lockfile(
+    cli_lockfile: Option<&String>,
+    cli_lockfile_type: Option<&String>,
+) -> io::Result<(String, Option<String>)> {
+    if let Some(lockfile) = cli_lockfile {
+        return Ok((lockfile.clone(), cli_lockfile_type.cloned()));
+    }
```
kylewillmon, line 133: It seems like we should handle the case where `cli_lockfile_type` here is `None`. For example, if the user does this:

```
phylum init --lockfile yarn.lock foo
```

We should probably set the lockfile type to `yarn` because we can detect it by the filename.

It's the same behavior on any future `phylum analyze` because the command will look up the parser by the filename of the lockfile. But this way we can pre-calculate the parser and maybe even show a warning if it's not known (to discourage the use of the `auto` parser)

cli/src/commands/init.rs
```diff
@@ -90,9 +117,79 @@ fn prompt_group() -> io::Result<Option<String>> {
...
+fn prompt_lockfile_type() -> io::Result<String> {
+    let lockfile_types: Vec<_> = LockfileFormat::iter().map(|format| format.name()).collect();
+
+    let index = FuzzySelect::new()
+        .with_prompt("Select lockfile's type")
+        .items(&lockfile_types)
+        .interact()?;
+
+    Ok(lockfile_types[index].to_owned())
+}
```
kylewillmon, line 195: With command line arguments, it's possible to set the lockfile type to `None`. But that is not an option in the prompt. This list doesn't even include `auto` as an option.

I think we should stay consistent. If we are okay with `None` (or `auto`) as an option, we should add it to this list. If we are not okay with it, then it should be an error when command line arguments would result in a lockfile type of `None`.

kylewillmon (changes requested):

cli/src/commands/init.rs
cd-work: I think the difference makes a lot of sense. You might not care/want to specify it when running an analyze because you trust Phylum to detect it automatically. But when configuring your project there is one guaranteed lockfile type that is going to be correct and that shouldn't be left to be figured out at a later point.

We're already trying to detect it automatically based on file name, but if that isn't enough I think we should definitely force the user to specify which format it actually is.

And I also think that our existing analyze/parse CLI should stay with an auto option.

kylewillmon: > But when configuring your project there is one guaranteed lockfile type that is going to be correct and that shouldn't be left to be figured out at a later point.

I agree with that. But by that logic, `phylum init` should never set the lockfile type to `None`. So passing `--lockfile` without `--lockfile-type` should error if we can't detect the lockfile type.

cli/src/commands/init.rs
```diff
@@ -90,9 +119,86 @@ fn prompt_group() -> io::Result<Option<String>> {
...
+        Input::new().with_prompt("Project Group (default: none)").interact_text()?;
 
     Ok(Some(group))
 }
+
+/// Interactively ask for missing lockfile information.
+fn prompt_lockfile(
+    cli_lockfile: Option<&String>,
+    cli_lockfile_type: Option<&String>,
+) -> io::Result<(String, Option<String>)> {
```
kylewillmon, line 132: The returned option is always `Some(_)`. So it might make sense to refactor the `Option` away.

```suggestion
fn prompt_lockfile(
    cli_lockfile: Option<&String>,
    cli_lockfile_type: Option<&String>,
) -> io::Result<(String, String)> {
```

Not really a big deal either way though, so I'll approve anyways.

kylewillmon (dismissed):

cli/src/commands/init.rs
cd-work: Maintainable code is always a big deal, good catch.

kylewillmon (approved):

PR #809, opened 2022-11-14
Add group member management subcommands

### LINKED ISSUE #810: Add/remove group members with CLI
Adding and removing members from a group is currently only accessible in the UI or via direct usage of the REST API. For easier scripting, the CLI should have commands to add and remove members from groups.

## Acceptance Criteria
- [ ] A CLI command exists to add a user to a group
- [ ] A CLI command exists to remove a user from a group
- [ ] Informative messages are returned for errors (Community edition user, user does not exist, etc)
- [ ] All documentation is updated to reflect the new commands

cd-work:
This patch adds the new `phylum group members` subcommand which has the additional `list`/`add`/`remove` subcommands. These allow managing and inspecting the users which are part of a group.

The `members`/`list` commands behave similarly to all our other commands, doing the same thing and just listing all members that are part of the specified group.

The group parameter is a required parameter for the `members` subcommand, which allows clap to provide helpful messages guiding the user to specify it in the correct location, rather than having to handle all possible locations internally.

Closes #810.

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -3,7 +3,9 @@ use std::time::Duration;
 use anyhow::{anyhow, Context};
 use phylum_types::types::auth::TokenResponse;
 use phylum_types::types::common::{JobId, ProjectId};
-use phylum_types::types::group::{CreateGroupRequest, CreateGroupResponse, ListUserGroupsResponse};
+use phylum_types::types::group::{
+    CreateGroupRequest, CreateGroupResponse, ListGroupMembersResponse, ListUserGroupsResponse,
+};
 use phylum_types::types::job::{
     AllJobsStatusResponse, JobStatusResponse, SubmitPackageRequest, SubmitPackageResponse,
 };
@@ -114,6 +116,16 @@ impl PhylumApi {
         path: U,
         body: Option<B>,
     ) -> Result<T> {
+        let body = self.send_request_raw(method, path, body).await?;
+        serde_json::from_str::<T>(&body).map_err(|e| PhylumApiError::Other(e.into()))
+    }
+
+    async fn send_request_raw<B: Serialize, U: IntoUrl>(
+        &self,
+        method: Method,
+        path: U,
+        body: Option<B>,
+    ) -> Result<String> {
         let mut request = self.client.request(method, path);
         if let Some(body) = body {
             request = request.json(&body);
@@ -135,7 +147,7 @@ impl PhylumApi {
             return Err(err.into());
         }
 
-        serde_json::from_str::<T>(&body).map_err(|e| PhylumApiError::Other(e.into()))
+        Ok(body)
     }
 }
 
@@ -353,6 +365,26 @@ impl PhylumApi {
     pub async fn create_group(&self, group_name: &str) -> Result<CreateGroupResponse> {
         let group = CreateGroupRequest { group_name: group_name.into() };
         self.post(endpoints::group_create(&self.config.connection.uri)?, group).await
+    }
+
+    /// Get members of a group.
+    pub async fn group_members(&self, group_name: &str) -> Result<ListGroupMembersResponse> {
+        let url = endpoints::group_members(&self.config.connection.uri, group_name)?;
+        self.get(url).await
+    }
+
+    /// Add user to a group.
+    pub async fn group_add(&self, group_name: &str, user_email: &str) -> Result<()> {
+        let url = endpoints::group_usermod(&self.config.connection.uri, group_name, user_email)?;
+        self.send_request_raw(Method::POST, url, None::<()>).await?;
+        Ok(())
+    }
+
+    /// Remove user from a group.
+    pub async fn group_remove(&self, group_name: &str, user_email: &str) -> Result<()> {
+        let url = endpoints::group_usermod(&self.config.connection.uri, group_name, user_email)?;
+        self.send_request_raw(Method::DELETE, url, None::<()>).await?;
+        Ok(())
     }
 
     pub fn config(&self) -> &Config {
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -347,7 +347,49 @@ pub fn add_subcommands(command: Command) -> Command {
                             .help("Name for the new group")
                             .required(true),
                     ),
-                ),
+                )
+                .subcommand(
+                    Command::new("members").about("List group members").args(&[
+                        Arg::new("group")
+                            .short('g')
+                            .long("group")
+                            .value_name("GROUP")
+                            .help("Group to list the members for")
+                            .required(true),
+                        Arg::new("json")
+                            .action(ArgAction::SetTrue)
+                            .short('j')
+                            .long("json")
+                            .help("Produce group list in json format (default: false)"),
+                        ])
+                    .subcommand(
+                        Command::new("list").about("List all existing projects").args(&[
+                            Arg::new("json")
+                                .action(ArgAction::SetTrue)
+                                .short('j')
+                                .long("json")
+                                .help("Produce output in json format (default: false)"),
+                        ]),
+                    )
+                    .subcommand(
+                        Command::new("add").about("Add user to group").args(&[
+                            Arg::new("user")
+                                .value_name("USER")
+                                .help("User to be added")
+                                .action(ArgAction::Append)
+                                .required(true),
+                        ]),
+                    )
+                    .subcommand(
+                        Command::new("remove").about("Remove user from group").args(&[
+                            Arg::new("user")
+                                .value_name("USER")
+                                .help("User(s) to be removed")
+                                .action(ArgAction::Append)
+                                .required(true),
+                        ]),
+                    )
+                )
         )
         .subcommand(
             Command::new("init")
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -360,7 +360,7 @@ pub fn add_subcommands(command: Command) -> Command {
                             .action(ArgAction::SetTrue)
                             .short('j')
                             .long("json")
-                            .help("Produce group list in json format (default: false)"),
+                            .help("Produce member list in json format (default: false)"),
                         ])
                     .subcommand(
                         Command::new("list").about("List group members").args(&[
@@ -368,7 +368,7 @@ pub fn add_subcommands(command: Command) -> Command {
                                 .action(ArgAction::SetTrue)
                                 .short('j')
                                 .long("json")
-                                .help("Produce output in json format (default: false)"),
+                                .help("Produce member list in json format (default: false)"),
                         ]),
                     )
                     .subcommand(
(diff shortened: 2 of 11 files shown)


cli/src/api/mod.rs
```diff
@@ -114,6 +116,16 @@ impl PhylumApi {
...
+        let body = self.send_request_raw(method, path, body).await?;
+        serde_json::from_str::<T>(&body).map_err(|e| PhylumApiError::Other(e.into()))
+    }
+
+    async fn send_request_raw<B: Serialize, U: IntoUrl>(
+        &self,
+        method: Method,
+        path: U,
+        body: Option<B>,
+    ) -> Result<String> {
```
kylewillmon, line 128: If we're going to introduce this function, we should use it for `delete_project()` as well (which is currently using [`IgnoredAny`](https://docs.rs/serde/1.0.147/serde/de/struct.IgnoredAny.html) to throw away the output)

cli/src/app.rs
```diff
@@ -347,7 +347,49 @@ pub fn add_subcommands(command: Command) -> Command {
                             .help("Name for the new group")
                             .required(true),
                     ),
-                ),
+                )
+                .subcommand(
+                    Command::new("members").about("List group members").args(&[
```
kylewillmon, line 352: 1. This command can do more than simply `List group members` and the about message should reflect that
2. For consistency with our other commands, `members` should not be plural (this will require changing more than just this line)

```suggestion
                    Command::new("member").about("Manage group members").args(&[
```

cli/src/app.rs
```diff
@@ -347,7 +347,49 @@ pub fn add_subcommands(command: Command) -> Command {
...
+                            .help("Group to list the members for")
+                            .required(true),
+                        Arg::new("json")
+                            .action(ArgAction::SetTrue)
+                            .short('j')
+                            .long("json")
+                            .help("Produce group list in json format (default: false)"),
+                        ])
+                    .subcommand(
+                        Command::new("list").about("List all existing projects").args(&[
```
kylewillmon, line 366: Incorrect about message.

```suggestion
                        Command::new("list").about("List group members").args(&[
```

cli/src/app.rs
```diff
@@ -347,7 +347,49 @@ pub fn add_subcommands(command: Command) -> Command {
...
+                        Command::new("add").about("Add user to group").args(&[
+                            Arg::new("user")
+                                .value_name("USER")
+                                .help("User to be added")
+                                .action(ArgAction::Append)
+                                .required(true),
+                        ]),
+                    )
+                    .subcommand(
+                        Command::new("remove").about("Remove user from group").args(&[
```
kylewillmon, line 384: Should we add an `rm` alias?

cli/src/app.rs
```diff
@@ -347,7 +347,49 @@ pub fn add_subcommands(command: Command) -> Command {
...
+                            .short('g')
+                            .long("group")
+                            .value_name("GROUP")
+                            .help("Group to list the members for")
+                            .required(true),
+                        Arg::new("json")
+                            .action(ArgAction::SetTrue)
+                            .short('j')
+                            .long("json")
+                            .help("Produce group list in json format (default: false)"),
```
kylewillmon, line 363: This does not work as expected:

* `phylum group members --group x --json` outpus JSON ✅
* `phylum group members --group x list --json` outputs JSON ✅
* `phylum group members --group x --json list` outputs a table ❌
* `phylum group --json members --group x list` outputs a table ❌
* `phylum --json group members --group x list` is an error 🤷 (included for completeness)

cli/src/app.rs
```diff
@@ -347,7 +347,49 @@ pub fn add_subcommands(command: Command) -> Command {
...
-                ),
+                )
+                .subcommand(
+                    Command::new("members").about("List group members").args(&[
+                        Arg::new("group")
+                            .short('g')
+                            .long("group")
+                            .value_name("GROUP")
+                            .help("Group to list the members for")
+                            .required(true),
```
kylewillmon, line 358: This might be personal preference, but I think required arguments are more natural when they are positional. So I much prefer this:

```
phylum group members list <GROUP_NAME>
phylum group members add <GROUP_NAME> <EMAIL>
phylum group members remove <GROUP_NAME> <EMAIL>
```

Over this:

```
phylum group members -g <GROUP_NAME> list
phylum group members -g <GROUP_NAME> add <EMAIL>
phylum group members -g <GROUP_NAME> remove <EMAIL>
```

cli/src/app.rs
```diff
@@ -347,7 +347,49 @@ pub fn add_subcommands(command: Command) -> Command {
...
+                                .short('j')
+                                .long("json")
+                                .help("Produce output in json format (default: false)"),
+                        ]),
+                    )
+                    .subcommand(
+                        Command::new("add").about("Add user to group").args(&[
+                            Arg::new("user")
+                                .value_name("USER")
+                                .help("User to be added")
```
kylewillmon, line 378: This should match the help for `remove`

```suggestion
                                .help("User(s) to be added")
```

kylewillmon (changes requested):

cli/src/api/mod.rs
cd-work: Fyi `IgnoredAny` works if we want to ignore the output, but will not work if the output is empty. That's why I handled it this way. Ideally I'd rather not have to do it, but serde cannot handle deserializing an empty string to anything.

cli/src/app.rs
cd-work: > For consistency with our other commands, members should not be plural (this will require changing more than just this line)

Yeah I found myself mixing up members/member while testing myself. Sole reason for choosing members was the API naming.

kylewillmon: Right. Plural is standard in the API (for some reason that I don't understand). But we should keep singular as the standard in CLI.

I guess we could add aliases to allow both (we do that for `phylum project` for historical reasons). But I'd rather not add that complexity everywhere.

cli/src/app.rs
kylewillmon: > I'm pretty sure this should be consistent with our other commands?

I pointed it out because I thought it was inconsistent. I was thinking of `phylum project list` which has these cases:

* `phylum project list --json` outputs JSON
* `phylum project --json list` outputs JSON

But I hadn't checked `phylum group list` which has these cases.

* `phylum group list --json` outputs JSON
* `phylum group --json list` outputs a table

So we are currently inconsistent on how we handle this... That makes this a separate issue and I will open an issue so that we can discuss a resolution.

> I personally think that our command/command list duplication is kind of a hot mess anyway.

I agree. We should also discuss plans to clean up that mess in that separate issue.

cli/src/app.rs
kylewillmon: > With `--group` it's a little easier since all positional arguments can be swapped since they're all just user emails

That's fair. You have convinced me.

> The way it's currently done the `--group` location is clear and enforced by clap, leaving no room for screwups.

I'm not sure this is a good thing. I think it makes sense, for example, to allow `phylum group members add --group my-group kyle@phylum.io` in addition to `phylum group members --group my-group add kyle@phylum.io`. And clap has an easy way to do exactly that with [`Arg::global`](https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.global).

cd-work: We've run into consistency issues with `Arg::global` before. I **really** don't think it's as good as a solution as you make it out to be.

All `group member` options require a specific group, so enforcing it at **ONE** location makes it absolutely unambiguous which group is operated upon.

You say that it makes sense to allow both `member add --group` and `member --group add`, but you give no reason for why you think this is beneficial. If you ask me, this just adds unnecessary ambiguity. The user needs to specify it either way and with the way it's done right now, that's absolutely clear. Having it both way will just introduce potential for confusion.

kylewillmon: > We've run into consistency issues with `Arg::global` before. I really don't think it's as good as a solution as you make it out to be.

This is referring to #749? The issue was simply that we had an argument with the same name at two levels, causing overlap once we marked the higher level flag as global. That issue was simple to fix and does not arise in this case.

> You say that it makes sense to allow both `member add --group` and `member --group add`, but you give no reason for why you think this is beneficial. If you ask me, this just adds unnecessary ambiguity. The user needs to specify it either way and with the way it's done right now, that's absolutely clear. Having it both way will just introduce potential for confusion.

`phylum group member add --group my-group kyle@phylum.io` is not ambiguous. It is clear what the command will do. I don't see how that could confuse someone. On the contrary, if they remember that there is a `phylum group member add` command, it's probably more natural that they would type it this way.

cd-work: > This is referring to https://github.com/phylum-dev/cli/pull/749? The issue was simply that we had an argument with the same name at two levels, causing overlap once we marked the higher level flag as global. That issue was simple to fix and does not arise in this case.

I don't remember the exact issues, but wasn't there a problem where global flags could still be specified in multiple locations but then only one of them would actually work? I recall it not being particularly intuitive.

> phylum group member add --group my-group kyle@phylum.io is not ambiguous. It is clear what the command will do. I don't see how that could confuse someone.

The ambiguity here is not what the command will do, but what the "correct" way to do things are. Having three different ways to do the same thing with no good reason for it isn't an advantage for a CLI, it's a massive disadvantage. It confuses users and makes correct usage unclear and will seed doubt in the CLI actually doing what it's supposed to.

If you give a user **one** correct way to do things and that way is always how they're forced to do things then things are clear and make sense, thus giving people trust in the stuff actually working like it's supposed to. It's one of the big complaints people have about `git`.

kylewillmon: > I recall it not being particularly intuitive.

That is true of most things in `clap`, so you are probably correct here.

> If you give a user **one** correct way to do things and that way is always how they're forced to do things then things are clear and make sense, thus giving people trust in the stuff actually working like it's supposed to.

This is generally true. But it can be taken too far. For example, we allow all of these similar commands:

```
phylum group member --group x list
phylum group member --group=x list
phylum group member -g x list
phylum group member -g=x list
```

Which one is "correct"? I'm not sure. If you ask `phylum group member list --help`, it says the first one:

```
❯ phylum group member list --help
List group members

Usage: phylum group member --group <GROUP> list [OPTIONS]
[ -- SNIP -- ]
```

But it definitely be too restrictive to disallow the others just so that we can pick one correct way. How is this any different?

cd-work: > This is generally true. But it can be taken too far. For example, we allow all of these similar commands:

The examples listed here have nothing to do with our CLI though. They're just commonly known aliases that terminal users should generally be aware of. No matter how our CLI behaves, users should know that `--group=x` and `--group x` would be the same even without using our CLI.

kylewillmon: After further discussion, we discovered that [`Arg::global`](https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.global) cannot be used in conjunction with [`Arg::required`](https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.required). Other methods of making this work would almost certainly result in either unmaintainable code or inconsistent and/or confusing error messages.

That makes my suggestion impractical and means that we should continue with the code as written.

kylewillmon (approved):

kylewillmon (approved):

PR #811, opened 2022-11-15
Add code of conduct and security policy

kylewillmon:
Add code of conduct and security policy in order to meet the community standards for a public repository.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,132 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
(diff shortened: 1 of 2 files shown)


andreaphylum (dismissed):

maxrake (approved):
Looks like the same files from other Phylum repos. One possible addition would be to include a "Supported Versions" section to the security policy. There is one for [phylum-ci](https://github.com/phylum-dev/phylum-ci/security/policy) and might help to let folks know which major versions will be supported with security fixes.

kylewillmon:
> Looks like the same files/language from other Phylum repos.

Yep. That's where I got them 😄

> One possible addition would be to include a "Supported Versions" section to the security policy. There is one for [phylum-ci](https://github.com/phylum-dev/phylum-ci/security/policy) and might help to let folks know which major versions will be supported with security fixes.

I removed that section as we don't currently have a policy for this. And at this point, I don't think we need to commit one way or the other on long-term release branches. We can make decision case-by-case as security issues arise.

maxrake (approved):

PR #813, opened 2022-11-16
Fix `phylum batch` command

kylewillmon:
The `phylum batch` command was not correctly handling the `--type` parameter due to a regression in #721. This patch fixes that issue


ejortega:
lgtm, I was able to submit my list of packages using your branch.

andreaphylum (approved):

PR #815, opened 2022-11-17
Remove minisign artifacts

### LINKED ISSUE #738: Remove minisign artifacts
Once enough time has passed to allow users to migrate to the new releases which are signed with openssl, we can remove the minisign signatures from any future releases.

_Originally posted by @kylewillmon in https://github.com/phylum-dev/cli/pull/724#discussion_r991669531_

kylewillmon:
Remove `.minisig` files from future releases and close #738 

After a `grep -r minisig .`, I realized that I forgot to update the alternate install docs during the 2517c58 patch. So I have fixed that here as well.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -129,12 +129,6 @@ jobs:
     if: startsWith(github.ref, 'refs/tags/')
     runs-on: ubuntu-latest
     steps:
-      - name: Install host dependencies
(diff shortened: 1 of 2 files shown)


maxrake (approved):
`MINISIGN_KEY` and `MINISIGN_PASSWORD` can also be removed as secrets from the repo settings.

PR #828, opened 2022-11-23
Fix package-lock parsing with 3rd-party registries

cd-work:
This changes the v7+ package-lock.json parser to accept third-party registries and submit their packages pretending they are hosted on npmjs.org.

Since the v6 parser has the same behavior, it should allow for a more consistent user experience while still providing as good of an analysis as we can get with the information we have.


maxrake (approved):
Approving now...but also taking a note so we come back to this with a more holistic solution.

PR #830, opened 2022-11-28
Add automatic CLI doc generation

### LINKED ISSUE #803: Automatically generate CLI documentation
# Overview

Right now the CLI documentation on Phylum's website is created by hand. Not only does this take a lot of effort, it is also extremely error-prone.

Considering that the CLI output is very consistent and we're already using `clap` to generate the completions automatically, it should not be much more difficult to automatically generate our CLI documentation from this.

cd-work:
This patch adds a new `clap_markdown` crate which automatically converts our internal `clap::Command` constructed for the CLI into a markdown page compatible with reame.io's markdown renderer.

Generation is done using `cargo run --bin xtask -- gendocs` or with the `cargo xtask gendocs` alias.

A test was added which validates in CI that the currently generated documentation is always up to date with the latest CLI, reminding developers to update the documentation when making any changes to Phylum's CLI.

Closes #803.


clap_markdown/LICENSE
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
cd-work, line 1: I just copied this over, but I can nuke it if we think the license in the root of the directory is sufficient.

kylewillmon (changes requested):

maxrake (changes requested):
The test is working...there are currently failures 😅. It looks like the `phylum.md` page is the issue...showing the extensions in one and not the other.

Some questions:
* Where, when, and how does `gendocs` get called?
* Why are the number of `doc_templates` files not the same as the number of `docs/command_line_tool` files?
* How much of `gendocs` is automated and what, if anything, needs to be done manually by a developer?
* What is the process for adding a new doc?
  * What does it mean to add a new template?
  * Are templates always added or are there exceptions?
  * What placeholders are available for a template and are any required?

Instead of answering the questions in this PR (some answers are already given in the PR summary), it would be better to document this new feature for future developers in the source itself. `CONTRIBUTING.md` already has a `Documentation` section which will need to be updated as part of this PR. It might also make sense to place `README` file(s) in the `doc_templates` and/or `docs/command_line_tool` directories...even if they are simple pointers to more robust directions...since that is likely where devs will look first when adding new content.

Cargo.lock
```diff
@@ -3148,7 +3156,7 @@ dependencies = [
  "maplit",
  "once_cell",
  "open",
- "phylum_lockfile",
+ "phylum_lockfile 0.1.0",
```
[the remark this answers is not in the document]
chrisduerr: Because there were multiple phylum_lockfile versions. This has been reverted now so it should be good again.

xtask/src/gendocs.rs
```diff
@@ -0,0 +1,94 @@
...
+const TEMPLATE_DIR: &str = "./doc_templates";
+
+/// Default fallback template.
+const DEFAULT_TEMPLATE: &str = "{CLAP-MARKDOWN}";
+
+/// File header inserted at the top of each page.
+const HEADER: &str = "---
+title: {MARKDOWN-TITLE}
+category: 6255e67693d5200013b1fa3e
+hidden: false
```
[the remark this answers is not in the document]
chrisduerr: That's a great question. You can add a template but even that would get the header added before it so it wouldn't work.

We might need to change the templates to include the header.

docs/command_line_tool/phylum_update.md
```diff
@@ -3,7 +3,22 @@ title: phylum update
...
 ```sh
-phylum update
+Usage: phylum update [OPTIONS]
 ```
+
+### Options
+
+
+
+-v, --verbose...
```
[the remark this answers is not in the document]
chrisduerr: Fixed now, good catch.

doc_templates/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+These templates provide extra information for the automatically generated CLI
+documentation.
+
+## Adding new templates
+
+By default, no template is necessary. If you want to modify the automatically
+generated CLI documentation, you can lookup the filename for the specific
+command in `../docs/command_line_tool/` and add a file with identical name to
+this repository to use as a template.
```
maxrake, line 11:
```suggestion
this directory to use as a template.
```

doc_templates/README.md
```diff
@@ -0,0 +1,28 @@
...
+automatically with their values during documentation generation.
+
+These placeholers are currently supported:
+
+| Name             | Description                                                                            |
+| ---------------- | -------------------------------------------------------------------------------------- |
+| `{PH-HEADER}`    | ReadMe.com metadata header; omit this when manually overriding the header              |
+| `{PH-TITLE}`     | Command title (i.e. `phylum init`); use this for the ReadMe.com metadata header title  |
+| `{PH-MARKDOWN}
+`  | Automatically generated command documentation; this should be present in all templates |
```
maxrake, line 28:
```suggestion
| `{PH-MARKDOWN}`  | Automatically generated command documentation; this should be present in all templates |
```

maxrake (changes requested):
Looking good. Just two small changes requested...plus the `phylum.md` doc differences causing CI to fail.

clap_markdown/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    // Don't show hidden arguments.
+    if arg.is_hide_set() {
+        return None;
+    }
+
+    let mut markdown = String::new();
+
+    // Add short option.
+    if let Some(short) = arg.get_short() {
+        let _ = write!(markdown, "-{short}");
```
maxrake, line 124:
```suggestion
        let _ = write!(markdown, "`-{short}`");
```
To be more like the original (current) docs.

clap_markdown/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    // Add long option.
+    if let Some(long) = arg.get_long() {
+        if !markdown.is_empty() {
+            markdown += ", ";
+        }
+        let _ = write!(markdown, "--{long}");
```
maxrake, line 132:
```suggestion
        let _ = write!(markdown, "`--{long}`");
```
To be more like the original (current) docs.

clap_markdown/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+        for (i, value_name) in value_names.iter().enumerate() {
+            // Add separator between parameters.
+            if i != 0 {
+                markdown.push(delimiter);
+            }
+
+            if i >= min_required || all_optional {
+                let _ = write!(markdown, "[{value_name}]");
+            } else {
+                let _ = write!(markdown, "<{value_name}>");
```
maxrake, line 154:
```suggestion
                let _ = write!(markdown, "`[{value_name}]`");
            } else {
                let _ = write!(markdown, "`<{value_name}>`");
```
To make the positional arguments stand out in the docs.

clap_markdown/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+            }
+        }
+    }
+
+    // Add repetition indicator.
+    if matches!(arg.get_action(), ArgAction::Count) {
+        let _ = write!(markdown, "...");
```
maxrake, line 161: Adding backticks here would break the markdown rendering. The repetition indicator does not necessarily need to be in backticks, though, and this line could stay the same.

docs/command_line_tool/phylum.md
```diff
@@ -4,37 +4,51 @@ category: 6255e67693d5200013b1fa3e
...
 ### Options
-  `-c`, `--config <FILE>`
-  &emsp; Use a specific configuration file
-  
-  `-h`, `--help`
-  &emsp; Print help information
-  
-  `--no-check-certificate`
-  &emsp; Don't validate the server certificate when performing API requests
-  &emsp; DANGER: this should only be used when SSL termination breaks the Phylum certificate pinning
```
maxrake, line 34: This line got dropped but appears important enough that is should continue to be included somehow.

docs/command_line_tool/phylum_extension_install.md
```diff
@@ -4,29 +4,32 @@ category: 6255e67693d5200013b1fa3e
...
-`--accept-permissions`
-&emsp; Automatically accept requested permissions
 
-`--overwrite`
-&emsp; Overwrite existing extension
+-y, --yes
+&emsp; Accept permissions and overwrite existing extensions (same as --overwrite --accept-permissions)
 
-`-y`, `--yes`
-&emsp; Accept permissions and overwrite existing (same as `--overwrite --accept-permissions`)
```
maxrake, line 21: The backticks got dropped here. It might be worth adding them back in at the command definition.

docs/command_line_tool/phylum_extension_install.md
```diff
@@ -4,29 +4,32 @@ category: 6255e67693d5200013b1fa3e
...
+--accept-permissions
+&emsp; Automatically accept requested permissions
 
-### Details
+--overwrite
+&emsp; Overwrite existing extension
 
-The extension will be installed under `$XDG_DATA_HOME/phylum/extensions/<EXT_NAME>`.
-If `$XDG_DATA_HOME` is not set, it will default to `$HOME/.local/share/phylum/extensions/<EXT_NAME>`.
```
maxrake, line 26: These details were dropped. Was that on purpose? It might be worth adding them back in at the command definition so they show up in the CLI help and docs.

docs/command_line_tool/phylum_extension_new.md
```diff
@@ -7,13 +7,20 @@ hidden: false
...
-phylum extension new <PATH>
+Usage: phylum extension new [OPTIONS] <PATH>
 ```
 
-### Details
+### Arguments
 
-This command will create a "Hello, World!" extension skeleton that can be
-modified for a specific use. See the [Extension Quickstart] guide for more
-detail.
```
maxrake, line 17: These details were dropped. Was that on purpose? It might be worth adding them in a template file.

docs/command_line_tool/phylum_extension_run.md
```diff
@@ -7,20 +7,26 @@ hidden: false
...
 &emsp; Automatically accept requested permissions
 
-### Details
-
-The extension will be run without prior installation.
+-v, --verbose...
+&emsp; Increase the level of verbosity (the maximum is -vvv)
 
-The first set of options are for the `run` command. The second set of options
-are for the extension.
```
maxrake, line 23: More details that might be nice to include in the template file.

docs/command_line_tool/phylum_project_set-thresholds.md
```diff
@@ -3,17 +3,34 @@ title: phylum project set-thresholds
...
 hidden: false
 ---
+
 Interactively set risk domain thresholds for a project
+
 ```sh
-phylum project set-thresholds [OPTIONS] <name>
+Usage: phylum project set-thresholds [OPTIONS] <name>
 ```
-Resulting scores at or below the defined threshold will fail.
```
maxrake, line 10: Another lost detail.

docs/command_line_tool/phylum_parse.md
```diff
@@ -4,16 +4,31 @@ category: 6255e67693d5200013b1fa3e
 hidden: false
 ---
 
-Parse a lockfile and output the packages as JSON
```
maxrake, line 7: Another lost detail that might be worth included in the command definition.

maxrake (changes requested):
Did another pass...for doc formatting and content related changes.

clap_markdown/src/lib.rs
kylewillmon: @cd-work mentioned elsewhere that he intentionally removed the backticks:

> I've actually intentionally removed the backticks because they were not working well in the existing format (imo).

And for what it's worth, I like the way the new docs look without the backticks.

doc_templates/README.md
kylewillmon: Prepositional phrases suck... But it might be clearer if we re-order them?

> add a file to this directory with an identical name for use as a template

clap_markdown/src/lib.rs
maxrake: Roger that. Well...I may be outnumbered then 😢

cd-work: Yeah, just to clarify when you have flags with arguments, we previously had this:

`-s`, `--short <THING>`

However this does not correctly show that `-s` also needs an argument, an alternative would be this:

`-s`, `--short` `<THING>`

But that also looks kinda odd.

So since the CLI itself just prints everything as plain text anyway, I thought we might as well specify it in a way the user should already be familiar with, just with a proportional font:

-s, --short &lt;THING&gt;

docs/command_line_tool/phylum.md
cd-work: Considering it's not in the CLI either, I'm not sure it's worth it tbh.

Not checking certificates should be pretty obviously dangerous to any user, so I'm not sure if this comment provides actual value?

docs/command_line_tool/phylum_extension_install.md
cd-work: Good catch, I only looked for example sections, I wasn't aware of this one.

docs/command_line_tool/phylum_project_set-thresholds.md
cd-work: Expanded it a bit and moved it to a separate details section. Hopefully should be a bit more clear and not collide with the new template system.

docs/command_line_tool/phylum_parse.md
cd-work: Decided to expand the about section of the Command in general.

doc_templates/README.md
cd-work: It might be because I'm not a native speaker, but I honestly like @maxrake's suggestion much better (though maybe that's just because it's what I intended to write?).

docs/command_line_tool/phylum.md
maxrake: I like having this kind of information in the CLI help output. It makes that output more verbose, but that is the place I (and I assume most users) will look first for "documentation." My guess is this language was initially added b/c someone was (ab)using this feature the wrong way or otherwise had a question about the security implications.

doc_templates/README.md
kylewillmon: I am a native speaker and I have no clue which is actually better...

Using @maxrake's suggestion is fine.

xtask/src/gendocs.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/// Get a vec with all pages and their respective paths.
+fn pages() -> Result<Vec<(PathBuf, String)>> {
+    let template_dir = Path::new(TEMPLATE_DIR);
+    let target_dir = Path::new(OUTPUT_DIR);
+
+    // Set `XDG_DATA_HOME` to a bogus directory so regardless of installed
+    // extensions, none of them are ever documented.
+    env::set_var("XDG_DATA_HOME", "/i/n/v/a/l/i/d");
```
cd-work, line 43: Now when you first look at this, you might think it's a bit of a hack. And you would be right.

But I think this is the best solution we have to just ensure no extensions are taken into account for documentation generation without actually having to modify the CLI's interface. We could add an `extensions: bool` flag to `phylum_cli::app::app`, but I don't really like the idea.

This should work just fine, even if the bogus directory exists, just as long as it's not actually where users are installing their extensions. It generates an error log message, but since we never initialize any logger nobody gets to see that anyway.

maxrake (approved):

xtask/src/gendocs.rs
maxrake: Oh...sorry...missed this comment before I submitted my review.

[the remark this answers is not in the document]
cd-work: I don't like using a directory that actually might exist because if our CLI starts messing with it things might go **horribly** wrong.

/dev/null isn't a directory for example, worst-case scenario someone goes ahead and creates a directory if it didn't explicitly find a directory. Now normally that would just be a bad time nuking a potentially existing file, but if you nuke /dev/null to replace it with a /dev/null directory things are going to be **bad**.

I've considered /dev/null myself, but ultimately there's no **advantage** to using it, we don't actually need a null sink or use a null sink. And the hazards are just unnecessary.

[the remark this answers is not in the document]
maxrake: Roger that...withdrawn...I should have tested it first anyway...b/c it doesn't work:

```sh
❯ XDG_DATA_HOME=/dev/null poetry run phylum-init
 [*] Getting https://github.com/phylum-dev/cli/releases/download/v3.12.1/phylum-aarch64-apple-darwin.zip file ...Done
 [*] Saving https://github.com/phylum-dev/cli/releases/download/v3.12.1/phylum-aarch64-apple-darwin.zip file to /var/folders/gh/wnf14j7n4q34y2t36hq2jz800000gn/T/tmpdu9b4y5c/phylum-aarch64-apple-darwin.zip ...Done
 [*] Getting https://github.com/phylum-dev/cli/releases/download/v3.12.1/phylum-aarch64-apple-darwin.zip.signature file ...Done
 [*] Saving https://github.com/phylum-dev/cli/releases/download/v3.12.1/phylum-aarch64-apple-darwin.zip.signature file to /var/folders/gh/wnf14j7n4q34y2t36hq2jz800000gn/T/tmpdu9b4y5c/phylum-aarch64-apple-darwin.zip.signature ...Done
 [*] Verifying /var/folders/gh/wnf14j7n4q34y2t36hq2jz800000gn/T/tmpdu9b4y5c/phylum-aarch64-apple-darwin.zip with signature from /var/folders/gh/wnf14j7n4q34y2t36hq2jz800000gn/T/tmpdu9b4y5c/phylum-aarch64-apple-darwin.zip.signature ...SUCCESS

    phylum-cli installer

mkdir: /dev/null: Not a directory
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/maxrake/dev/phylum/phylum-ci/src/phylum/init/cli.py", line 404, in main
    subprocess.run(cmd, check=True, cwd=extracted_dir)
  File "/Users/maxrake/.pyenv/versions/3.11.0/lib/python3.11/subprocess.py", line 569, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['sh', 'install.sh']' returned non-zero exit status 1.
```

kylewillmon (approved):

PR #831, opened 2022-11-28
Revert to using a PAT for cargo update

kylewillmon:
In #816 (and #817), I tried to change the `cargo-update.yml` workflow to use `GITHUB_TOKEN` instead of a PAT. That was a mistake. The workflow_dispatch trigger does not properly connect to the opened PR.

This patch reverts the 8f4f7e5 and 82fb0a3 changes to that workflow. Other changes made in those commits (namely, to the `release.yml` workflow) are not reverted.


maxrake (approved):
Bummer...the `GH_RELEASE_PAT` can't really be a `release` environment secret now.

PR #833, opened 2022-11-30
Add `phylum group transfer` subcommand

### LINKED ISSUE #832: Group transfer command
The API now allows [group transfer](https://api.staging.phylum.io/api/v0/swagger/index.html#/Groups/groups_change_group_owner), but this is not yet available in the CLI.

The command should look something like this:

```
> phylum group transfer --group earth overlords@insects.com

This will transfer ownership of `earth` to `overlords@insects.com`. You will no longer own the group.
Are you sure [y/N]? y

✅ Successfully transferred group ownership!
```

## Acceptance criteria

* A command is available to transfer group ownership
* The effect of the command is well documented
* The command requires confirmation via prompt or a `--yes` or `--force` argument

cd-work:
This patch adds support for the new API endpoint to transfer group ownership between users.

To ensure that groups are not accidentally transferred to incorrect users and since the operation cannot be reverted easily, extra precautions are taken in the form of a confirmation prompt.

Closes #832.

### DIFF
--- a/cli/src/api/endpoints.rs
+++ b/cli/src/api/endpoints.rs
@@ -112,6 +112,13 @@ pub fn group_members(api_uri: &str, group: &str) -> Result<Url, BaseUriError> {
     Ok(url)
 }
 
+/// PUT /groups/<groupName>/owner/<userEmail>
+pub fn set_owner(api_uri: &str, group: &str, owner: &str) -> Result<Url, BaseUriError> {
+    let mut url = get_api_path(api_uri)?;
+    url.path_segments_mut().unwrap().pop_if_empty().extend(["groups", group, "owner", owner]);
+    Ok(url)
+}
+
 /// GET /.well-known/openid-configuration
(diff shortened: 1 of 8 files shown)


cd-work:
While writing this PR, I've encountered some issues with the responses made by the API:

Apparently in some scenarios missing groups will cause a 403 and in others it will cause a 404. I've tested with an "interj" group since it's similar to the "interject" group I do own and get a 404, however when using "nada6134495" I get an access forbidden error. Since the "nada6134495" certainly shouldn't exist, I don't see how these API error responses make any sense?

Considering that the only error that uniquely identifies one event by code is the 404, I'm not sure if it's even worth it to do any special error handling at all? We could also just go with the plain response error messages for every reply but that has been deemed insufficient for some common errors in the past.

cd-work:
After looking through several confusing API errors with matching codes together with @kylewillmon, we've decided to opt against trying to provide custom error message and instead stick with the response error bodies instead for now.

kylewillmon (approved):

kylewillmon (approved):

PR #838, opened 2022-12-01
Add `parentDoc` metadata to grouped CLI docs

### LINKED ISSUE #837: Specify `parentDoc` for subpages in CLI doc templates
Currently the groupings of pages in our documentation is manually defined. To improve reliability the [parentDoc](https://docs.readme.com/main/reference/createdoc) field should be used to define a fixed grouping of all nested subcommands.

cd-work:
This patch adds the `parentDoc` readme.com metadata field to all CLI documentation pages which should be grouped beneath another page.

Closes #837.


maxrake (changes requested):
These changes all appear to be correct. The request is for _additional_ changes...to update the `doc_templates/README.md` file. Before, the category was the same for all the CLI doc pages and so it was kind of obvious what it should be for new pages. It is less obvious what the `parentDoc` value should be, if it is needed at all, and how to obtain it for new commands/pages.

If it helps for wording, the `phylum-ci` repo has [a similar README](https://github.com/phylum-dev/phylum-ci/blob/main/docs/sync/README.txt) that might provide some inspiration. NOTE: that file is `.txt` instead of `.md` on purpose...so that it will NOT be included when providing a directory to the `rdme` CLI.

maxrake:
> Honestly I'm not sure how to document that considering I had to crawl through the inspector to find them. I feel like it's the kind of thing where you wouldn't find it anyway if I have to tell you where I got it from, especially since that might change in the future?

That is a bummer. The worry is that future contributors (and our future selves!) will have these exact questions and will also spend time trying to figure it out. If there is no obvious way to do it, then explicit directions can be omitted. Bread crumbs of any sort are still desired over nothing at all. Maybe even just mentioning what `parentDoc` is, if it is required or not, and references to the ReadMe docs about it:

* https://docs.readme.com/main/docs/reordering-pages#subpages
* https://docs.readme.com/main/reference/createdoc
  * > For a subpage, specify the parent doc ID, which you can get through https://docs.readme.com/main/reference/docs#getdoc

cd-work:
Added some documentation linking to ReadMe.com's docs.

maxrake (approved):

PR #840, opened 2022-12-01
Ignore parent directory projects for `phylum init`

### LINKED ISSUE #839: `phylum init` incorrectly warns about overwrite
When `phylum init` is run from a subdirectory of a phylum project, it incorrectly warns that it is going to overwrite an existing project config.

# How To Reproduce
From a directory `~/tmp/foo`:

```
❯ find ..
..
../.phylum_project
../foo

❯ phylum init
⚠️  Workspace is already linked to a Phylum project
Overwrite existing project configuration? yes

Project Name: foo
Use a project group? no
✅ Successfully created project "foo"

Project lockfile path: package-lock.json

❯ find ..
..
../.phylum_project
../foo
../foo/.phylum_project
```

# Expected Behavior
We should remove the warning in this case.

cd-work:
Previously, `phylum init` would query if the project should be overwritten when a parent project contained a project file already.

Since we do not actually overwrite any project files outside of the current directory, and since having nested Phylum projects is perfectly valid, a warning is now only emitted when there's already a Phylum project file in the current directory.

Closes #839.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -197,27 +197,27 @@ pub fn read_configuration(path: &Path) -> Result<Config> {
     Ok(config)
 }
 
-pub fn find_project_conf(starting_directory: &Path) -> Option<PathBuf> {
-    let mut path: PathBuf = starting_directory.into();
-    let mut attempts = 0;
-    const MAX_DEPTH: u8 = 32;
+pub fn find_project_conf(
+    starting_directory: impl AsRef<Path>,
+    recurse_upwards: bool,
+) -> Option<PathBuf> {
+    let max_depth = if recurse_upwards { 32 } else { 1 };
+    let mut path = starting_directory.as_ref();
 
-    loop {
-        let search_path = path.join(PROJ_CONF_FILE);
-        if search_path.is_file() {
-            return Some(search_path);
+    for _ in 0..max_depth {
+        let conf_path = path.join(PROJ_CONF_FILE);
+        if conf_path.is_file() {
+            return Some(conf_path);
         }
 
-        if attempts > MAX_DEPTH {
-            return None;
-        }
-        path.push("..");
-        attempts += 1;
+        path = path.parent()?;
     }
+
+    None
 }
 
 pub fn get_current_project() -> Option<ProjectConfig> {
-    find_project_conf(Path::new(".")).and_then(|config_path| {
+    find_project_conf(".", true).and_then(|config_path| {
         log::info!("Found project configuration file at {config_path:?}");
         let mut config: ProjectConfig = parse_config(&config_path).ok()?;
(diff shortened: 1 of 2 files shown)


cli/src/config.rs
```diff
@@ -170,27 +170,27 @@ pub fn read_configuration(path: &Path) -> Result<Config> {
...
-    let mut attempts = 0;
-    const MAX_DEPTH: u8 = 32;
-
-    loop {
-        let search_path = path.join(PROJ_CONF_FILE);
-        if search_path.is_file() {
-            return Some(search_path);
+pub fn find_project_conf(
+    starting_directory: impl AsRef<Path>,
+    recurse_upwards: bool,
```
cd-work, line 175: While we could have checked if the `find_project_conf` returned Path's parent is equal to the current directory, that would have meant traversing the entire directory tree just to discard it. So I've decided to go for the simpler solution and just added an option to skip the traversal.

cli/src/config.rs
```diff
@@ -170,27 +170,27 @@ pub fn read_configuration(path: &Path) -> Result<Config> {
...
+) -> Option<PathBuf> {
+    let max_depth = if recurse_upwards { 32 } else { 1 };
+    let mut path = starting_directory.as_ref();
+
+    for _ in 0..max_depth {
```
cd-work, line 180: I've decided to just rewrite the loop here completely. It's shorter and using `parent` should be easier to understand and at the same time aborts early instead of pointlessly looping when we're already at the root.

andreaphylum (approved):

cd-work:
Don't merge this, I think I want to actually add the project file's root directory to our project conf struct.

cd-work:
Will rewrite this once https://github.com/phylum-dev/cli/pull/845 is merged.

cd-work:
I decided to keep things pretty much identical since reading the `root` field on the config still isn't really easier than not recursing. We'd have to do expensive recursive lookup and parse the entire config just to at the end still have to compare paths that we might have to canonicalize first. So not really worth it, much easier to just skip the directory traversal and see if we find something at all.

andreaphylum (approved):

andreaphylum (dismissed):

cli/src/config.rs
```diff
@@ -197,27 +197,27 @@ pub fn read_configuration(path: &Path) -> Result<Config> {
...
 
-        if attempts > MAX_DEPTH {
-            return None;
-        }
-        path.push("..");
-        attempts += 1;
+        path = path.parent()?;
     }
+
+    None
```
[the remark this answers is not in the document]
cd-work: Definitely a good idea, yeah.

kylewillmon (approved):

PR #845, opened 2022-12-01
Fix subdir analysis without lockfile parameter

### LINKED ISSUE #843: Analysis cannot be submitted without lockfile from subdirectories
When running `phylum analyze` in a subdirectory without adding a lockfile path, relying on the linked project's lockfile, then the path is taken as relative to the current directory and as such will not correctly perform an analysis.

When in the subdirectory of a phylum project, the lockfile path should always be resolved relative to the project file's location. We don't want to resolve the path absolutely since this would prevent users from moving directories around.

cd-work:
Since the lockfile was specified as a relative path in the project lockfile, an analysis would fail when the analysis was started from a child directory of the Phylum project directory.

To ensure that analysis will work in any of the child directories, a root path of the project is now stored in the config on load and the lockfile location is determined based on that root directory whenever it is required.

While an absolute path to the lockfile might work temporarily, it was intentionally avoided as a solution since it would break when the project is moved to a different location or shared between different systems.

Closes #843.

### DIFF
--- a/cli/src/commands/init.rs
+++ b/cli/src/commands/init.rs
@@ -63,8 +63,7 @@ pub async fn handle_init(api: &mut PhylumApi, matches: &ArgMatches) -> CommandRe
     // Add lockfile and its type to the project configuration.
     println!();
     let (lockfile, lockfile_type) = prompt_lockfile(cli_lockfile, cli_lockfile_type)?;
-    project_config.lockfile_type = Some(lockfile_type);
-    project_config.lockfile = Some(lockfile);
+    project_config.set_lockfile(lockfile_type, lockfile);
 
     // Save project config.
     config::save_config(Path::new(PROJ_CONF_FILE), &project_config)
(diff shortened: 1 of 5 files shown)


maxrake (approved):
The expected behavior was confirmed with local tests. LGTM.

One thing observed while testing that does not need to hold up this review:
* `phylum init -t` help output does not provide the available options like `parse` and `analyze` do
  * Invalid values (guesses) provided there already get accepted as-is (e.g., try `-t auto`)
  * Is it possible to provide the options in the help output, specify the default when one isn't provided (e.g., `auto`), and restrict input to those options?
    * It might require special handling of the default/`auto` entry

cd-work:
> phylum init -t help output does not provide the available options like parse and analyze do 

This is a good point, but completely unrelated to this PR. Could you open an issue for this? Should be a simple fix, just getting all the lockfile types and filtering out `auto`.

maxrake:
> This is a good point, but completely unrelated to this PR. Could you open an issue for this? Should be a simple fix, just getting all the lockfile types and filtering out `auto`.

Done...in #847

maxrake (approved):

PR #848, opened 2022-12-02
Reorder project initialization

### LINKED ISSUE #841: Improve `phylum init` project creation process
Currently when running `phylum init` and entering a project, immediately after the project name and group are selected, the project is looked up and a message is printed whether the linking was successful.

This has two main issues: The project isn't linked yet at this step and it prevents the user from defining the lockfile until the request completes.

As a solution, I would recommend gathering all the information like lockfile name/type and project/group first and then performing all mutable operations and lookups without additional user interactions at the end.

The `commands::project::link_project` method should also likely be renamed in the process, since it doesn't actually perform the linking.

cd-work:
This reorders the steps in the project initialization for the `phylum init` command, to improve user experience with the command.

Instead of trying to create or link to the project before prompting for the lockfile path and type, all API operations are now moved to the end after the user has finished his inputs. This way looking can go on in the background without users waiting for long-running API calls to complete.

Closes #841.

### DIFF
--- a/cli/src/commands/extensions/permissions.rs
+++ b/cli/src/commands/extensions/permissions.rs
@@ -79,7 +79,7 @@ impl Permission {
             },
             // Parent set vs child set have to be validated.
             // This will error if child is not subset of parent, and return the child set otherwise.
-            (&Permission::List(ref parent), &Permission::List(ref child)) => {
+            (Permission::List(parent), Permission::List(child)) => {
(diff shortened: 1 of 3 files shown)


kylewillmon (approved):

maxrake (approved):
Code was reviewed and mostly understood. This was tested locally and confirmed to work. LGTM.

PR #849, opened 2022-12-02
Add possible values to `phylum init -t`

### LINKED ISSUE #847: Lockfile type for `init` command is undocumented and unrestricted
# Overview

Help output for `phylum init` does not provide the available options like `parse` and `analyze` do for the `--lockfile-type` option. Invalid values (users guessing) provided for that option get accepted as-is (e.g., try `-t auto`) and added to the Phylum project file.

# How To Reproduce

1. Run `phylum analyze -h` or `phylum parse -h`
2. See that there is an entry of the form:
```
-t, --lockfile-type <type>    The type of the lock file (default: auto) [possible values: yarn, npm, gem, pip, pipenv, poetry, mvn, gradle, nuget, go, cargo, auto]
```
3. Run `phylum init -h`
4. See that there is an entry of the form:
```
-t, --lockfile-type <LOCKFILE_TYPE>  Project lockfile type
```
5. In a Rust project directory with no existing Phylum project file, run `phylum init --lockfile Cargo.lock --lockfile-type rust delme_type_guess`
6. See that the command succeeds and the `.phylum_project` file created includes an entry for `lockfile_type: rust`, which was a guess and not the correct/valid `cargo` type

# Expected Behavior

Provide the possible options in the help output, specify the default when one isn't provided (e.g., `auto`), and restrict input to those options.

# Additional Context

Special handling of the default/`auto` entry may be needed.

The `value_name` used in the `init` command for the `lockfile-type` argument is not the same as the ones used for the `analyze` and `parse` commands. FWIW...the one used for `init` appears to be the better choice since it is more descriptive and follows the ALL_CAPS convention.

cd-work:
This adds restrictions on the allowed parameters that can be passed to the lockfile types parameter on `phylum init` to clap, allowing for proper documentation and feedback when an invalid value is specified.

Closes #847.

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -233,7 +233,7 @@ pub fn add_subcommands(command: Command) -> Command {
                     .long("lockfile-type")
                     .value_name("type")
                     .help("The type of the lock file (default: auto)")
-                    .value_parser(PossibleValuesParser::new(parse::lockfile_types())),
+                    .value_parser(PossibleValuesParser::new(parse::lockfile_types(true))),
             ]),
         )
         .subcommand(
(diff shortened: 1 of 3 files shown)


maxrake (approved):
Verification was performed with local tests.

PR #854, opened 2022-12-05
Remove actions-rs GitHub actions

cd-work:
This removes the unnecessary usage of actions-rs/toolchain and actions-rs/cargo in favor of using rustup and cargo directly.

The patch should also ensure that when running with a specific toolchain version (like `stable`), that this version is always up to date.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,16 +21,10 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
+        run: rustup toolchain install stable --profile minimal
 
       - name: Generate shell completion files
-        uses: actions-rs/cargo@v1
-        with:
-          command: xtask
-          args: gencomp
+        run: cargo +stable xtask gencomp
 
       - name: Upload shell completions
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -21,16 +21,10 @@ jobs:
...
-          toolchain: stable
-          profile: minimal
+        run: rustup toolchain install stable --profile minimal
 
       - name: Generate shell completion files
-        uses: actions-rs/cargo@v1
-        with:
-          command: xtask
-          args: gencomp
+        run: cargo +stable xtask gencomp
```
kylewillmon, line 27: It makes me a bit nervous that we have to remember to add `+stable` (or `+whatever`) to all of our `cargo` commands. I can see the benefit, but I also know that it would be easy to forget.

There are a few options for setting the default when we do the `rustup toolchain install`.
1. Set the global default
   ```
   rustup default stable
   ```
1. Add an override
   ```
   rustup override set stable
   ```
1. Set an environment variable
   ```
   echo "RUSTUP_TOOLCHAIN=stable" >> $GITHUB_ENV
   ```

I'm not sure I have a strong preference between these options. But any of them would allow us to omit the `+whatever` when calling `cargo`.

kylewillmon (dismissed):

.github/workflows/release.yml
cd-work: I've told you before that I'd personally just slap `rustup default` on it, but that said it's not like "forgetting" the `+stable` or whatever would be a catastrophic failure. In most cases it would likely work out just fine.

So either way works fine for me tbh.

kylewillmon (approved):

PR #865, opened 2022-12-07
Recommend brew and curl installs in the README

### LINKED ISSUE #824: Document curl and Homebrew in Quickstart
The quickstart and top-level README should be updated to recommend the `curl` install for Linux users and the Homebrew install for macos users.

The current install instructions in the quickstart can be removed.

kylewillmon:
This patch updates the quickstart to recommend curl and brew as the primary methods of installing phylum. Closes #824


README.md
```diff
@@ -11,39 +11,30 @@
 [![Test Status](https://github.com/phylum-dev/cli/actions/workflows/test.yml/badge.svg?branch=master)](https://github.com/phylum-dev/cli/actions/workflows/test.yml)
 [![README](https://img.shields.io/badge/docs-README-blue)](https://docs.phylum.io/docs/welcome)
 
-The command line interface (CLI) allows users to submit their project package dependencies to [Phylum's](https://phylum.io) API for analysis. Currently [pre-built binaries](https://github.com/phylum-dev/cli/releases) for Linux and macOS are available. On Windows, we recommend using the Linux binaries under [WSL](https://learn.microsoft.com/en-us/windows/wsl/). For more options, see [Alternate Installation Methods](https://docs.phylum.io/docs/alternate_install).
+The command line interface (CLI) allows users to submit their project package dependencies to [Phylum's](https://phylum.io) API for analysis. Currently [pre-built binaries](https://github.com/phylum-dev/cli/releases) for Linux and macos are available. On Windows, we recommend using the Linux binaries under [WSL](https://learn.microsoft.com/en-us/windows/wsl/). For more options, see [Alternate Installation Methods](https://docs.phylum.io/docs/alternate_install).
```
maxrake, line 14: Why change to `macos`? It appears that [Apple](https://www.apple.com/macos/ventura/) and [Wikipedia](https://en.wikipedia.org/wiki/MacOS) use the `macOS` form. The change appears deliberate since it was made in a few spots.

README.md
```diff
@@ -11,39 +11,30 @@
...
 
 [![asciicast](https://asciinema.org/a/431262.svg)](https://asciinema.org/a/431262)
 
-## Quickstart for Linux or macOS
+## Install phylum
 
-1. Download the latest release package and signature file for your target:
```
maxrake, line 20: Does it make sense to keep the manual download/verify steps...but just move them to the "Alternate Installation Methods" page?

README.md
```diff
@@ -11,39 +11,30 @@
...
-   | x86_64-apple-darwin | [phylum-x86_64-apple-darwin.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-x86_64-apple-darwin.zip) <br /> [phylum-x86_64-apple-darwin.zip.signature](https://github.com/phylum-dev/cli/releases/latest/download/phylum-x86_64-apple-darwin.zip.signature) |
-   | aarch64-apple-darwin | [phylum-aarch64-apple-darwin.zip](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-apple-darwin.zip) <br /> [phylum-aarch64-apple-darwin.zip.signature](https://github.com/phylum-dev/cli/releases/latest/download/phylum-aarch64-apple-darwin.zip.signature) |
+Install on Linux with the following command:
 
-1. Confirm the signature of the archive with `openssl` and the [public key for Phylum](https://raw.githubusercontent.com/phylum-dev/cli/main/scripts/signing-key.pub):
+```sh
+curl https://sh.phylum.io/ | sh -
```
maxrake, line 25: Now that this is the first/default recommendation for install, the `scripts/phylum-init.sh` script should really be automatically sync'ed to the location that enables it to be hosted at https://sh.phylum.io/ ...unless it already is?

maxrake (approved):
There are a few questions...but looks good enough to approve now.

README.md
kylewillmon: Cloudfront is pulling the script directly from the main branch on GitHub. There is no manual process for updating it anymore

README.md
kylewillmon: The change was deliberate, but mistaken. Not sure why it was `macos` in my head....

README.md
kylewillmon: I don't think it makes sense to document it at the same level of detail because the alternate install page has a fairly different audience and is not kept up-to-date as carefully.

But a basic link to the releases and example of how to verify an archive might be useful. I've added that now.

maxrake (approved):
Oh yeah...the quickstart lives in _two_ places...nice catch. LGTM.

maxrake (approved):

PR #867, opened 2022-12-08
Use JSDoc comments in extension API

kylewillmon:
This patch switches the doc comments in our extension API definition file to JSDoc style comments. This change allows our docs to be consumed by normal deno tools for generating docs.

To see an example of these docs rendered, view them here on [doc.deno.land](https://doc.deno.land/https://raw.githubusercontent.com/phylum-dev/cli/jsdoc/cli/src/extension_api.ts/~/PhylumApi).

### DIFF
--- a/cli/src/extension_api.ts
+++ b/cli/src/extension_api.ts
@@ -24,13 +24,13 @@ export class PhylumApi {
    *
    * Analyze Job ID, which can later be queried with `getJobStatus`.
    */
-  static async analyze(
+  static analyze(
     package_type: string,
-    packages: [object],
+    packages: object[],
     project?: string,
     group?: string
-  ): string {
-    return await Deno.core.opAsync(
+  ): Promise<string> {
+    return Deno.core.opAsync(
       "analyze",
       package_type,
       packages,
@@ -58,18 +58,18 @@ export class PhylumApi {
    * }
    * ```
    */
-  static async getUserInfo(): object {
-    return await Deno.core.opAsync("get_user_info");
+  static getUserInfo(): Promise<object> {
+    return Deno.core.opAsync("get_user_info");
   }
 
   /** Get the current short-lived API access token. */
-  static async getAccessToken(): string {
-    return await Deno.core.opAsync("get_access_token");
+  static getAccessToken(): Promise<string> {
+    return Deno.core.opAsync("get_access_token");
   }
 
   /** Get the long-lived user refresh token. */
-  static async getRefreshToken(): string {
-    return await Deno.core.opAsync("get_refresh_token");
+  static getRefreshToken(): Promise<string> {
+    return Deno.core.opAsync("get_refresh_token");
   }
 
   /**
@@ -123,8 +123,8 @@ export class PhylumApi {
    * }
    * ```
    */
-  static async getJobStatus(jobId: string): object {
-    return await Deno.core.opAsync("get_job_status", jobId);
+  static getJobStatus(jobId: string): Promise<object> {
+    return Deno.core.opAsync("get_job_status", jobId);
   }
 
   /**
@@ -171,7 +171,7 @@ export class PhylumApi {
    * }
    * ```
    */
-  static async getGroups(): object {
+  static getGroups(): Promise<object> {
     return Deno.core.opAsync("get_groups");
   }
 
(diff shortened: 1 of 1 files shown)


cli/src/extension_api.ts
```diff
@@ -1,34 +1,36 @@
...
+   *   { name: "accepts", version: "1.3.8" },
+   *   { name: "ms", version: "2.0.0" },
+   *   { name: "negotiator", version: "0.6.3" },
+   *   { name: "ms", version: "2.1.3" }
+   * ]
+   * ```
+   *
+   * # Returns
+   *
+   * Analyze Job ID, which can later be queried with `getJobStatus`.
```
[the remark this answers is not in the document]
cd-work: I think having it all converted in one PR makes more sense than having it in some weird invalid middle ground between the two PRs.

cli/src/extension_api.ts
```diff
@@ -37,358 +39,380 @@ export class PhylumApi {
...
+  /** Get the current short-lived API access token. */
+  static getAccessToken(): Promise<string> {
+    return Deno.core.opAsync("get_access_token");
   }
 
-  /// Get the long-lived user refresh token.
-  static async getRefreshToken(): string {
-    return await Deno.core.opAsync("get_refresh_token");
+  /** Get the long-lived user refresh token. */
+  static getRefreshToken(): Promise<string> {
```
cd-work, line 71: What's the reason of having these not be async? Did you just change it because that's what Deno does?

I'd imagine it's more "compatible" since it doesn't require the "new" async keyword, but on the other hand I also feel like using `async` is probably the "recommended" way with the latest version of JavaScript?

cd-work (changes requested):

cli/src/extension_api.ts
kylewillmon: There are many technically valid ways to write this function:

1. Without `async` / `await`:
   ```
   static getRefreshToken(): Promise<string> {
     return Deno.core.opAsync("get_refresh_token");
   }
   ```
1. With `async`, but not `await`:
   ```
   static async getRefreshToken(): Promise<string> {
     return Deno.core.opAsync("get_refresh_token");
   }
   ```
1. With `async` / `await`:
   ```
   static async getRefreshToken(): Promise<string> {
     return await Deno.core.opAsync("get_refresh_token");
   }
   ```

You might be surprised (as I was) at a few details here:

1. They all use the `Promise<_>` return type. Type-check fails without it.
   ```
   TS1064 [ERROR]: The return type of an async function or method must be the global Promise<T> type. Did you mean to write 'Promise<string>'?
     static async getRefreshToken(): string {
   ```
1. It's valid to return an unawaited `Promise<_>` in an `async` function... This is a feature of JavaScript. An `async` function automatically wraps the return in a `Promise` if and only if it isn't already a `Promise`.

Since we can just return the same promise returned by `opAsync`, we don't need `await`. And if we don't need `await`, we don't need to mark the function as `async`.

cd-work: I was aware that you don't need to await it, but I wasn't aware that you need to declare the return type to be a Promise. I thought TypeScript did that automatically just like Rust. Sounds good then.

cli/src/extension_api.ts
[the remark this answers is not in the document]
kylewillmon: Done.

cli/src/extension_api.ts
```diff
@@ -1,395 +1,402 @@
...
+
+function opSync<T>(op: string, ...args: unknown[]): T {
+  // deno-lint-ignore no-explicit-any
+  return (Deno as any).core.opSync(op, ...args) as T;
+}
+
+function opAsync<T>(op: string, ...args: unknown[]): Promise<T> {
+  // deno-lint-ignore no-explicit-any
+  return (Deno as any).core.opAsync(op, ...args) as Promise<T>;
+}
```
cd-work, line 402: I'd assume it was complaining that `Deno.core` didn't exist and this was the easiest solution to solve it?

cd-work (changes requested):

cli/src/extension_api.ts
kylewillmon: Yep. That's exactly what happened.

cd-work (approved):

PR #868, opened 2022-12-08
Add extension helpers for direct API requests

### LINKED ISSUE #862: Document example of direct usage of API in an extension
The extension API includes a `.getAccessToken()` function so that extension authors can make use of the full API and not just the endpoints already used by the CLI.

We should provide an example of this in our extension docs to aid extension authors.

### LINKED ISSUE #863: Include API base URL in extension API
Consider this extension which makes direct usage of the API:

```
import { PhylumApi } from "phylum";

const token = await PhylumApi.getAccessToken();
const headers = new Headers({
  Accept: "application/json",
  Authorization: `Bearer ${token}`,
});

const res = await fetch(
  "https://api.staging.phylum.io/api/v0/preferences/current_user",
  { headers },
);
console.log(await res.json());
```

In order to write this extension, the author must hard code the base URL of the API (or, potentially, find and read the phylum config file). This is not ideal.

We should provide a way to retrieve the base URL via the `PhylumApi` object to avoid this complication.

cd-work:
This patch adds some new abstractions to the extension API which make it easier for extension authors to make requests directly to Phylum's REST API.

To ensure people can figure out how they're supposed to use these new functions, the functionality has also been documented in our extension documentation.

Closes #863.
Closes #862.

---

Outstanding questions before undraft:
 - [x] Should probably write some tests
 - [x] Do we like the `fetchPhylum` name? Suggestions welcome
 - [x] Should we add a general net exception for our API URIs, or require adding them manually?

### DIFF
--- a/cli/src/extension_api.ts
+++ b/cli/src/extension_api.ts
@@ -4,6 +4,62 @@
 
 export class PhylumApi {
   /**
+   * Get the Phylum REST API base URL.
+   *
+   * This will usually return `https://api.phylum.io/api`.
+   */
+  static async apiBaseUrl(): string {
+    return Deno.core.opAsync("api_base_url");
+  }
+
+  /**
+   * Send a request to the Phylum REST API.
+   *
+   * See https://api.staging.phylum.io/api/v0/swagger/index.html for available API endpoints.
+   *
+   * The `options` parameter matches the `options` parameter of the `fetch` function:
+   * https://developer.mozilla.org/en-US/docs/Web/API/fetch
+   */
+  static async fetchPhylum(
+    apiVersion: ApiVersion,
+    endpoint: string,
+    options?: object
+  ): object {
+    // Ensure header object is initialized.
+    const fetchOptions = options ?? {};
+    fetchOptions.headers = fetchOptions.headers ?? {};
+
+    // Set Authorization/Content-Type headers if they are missing.
+    if (fetchOptions.headers instanceof Headers) {
+      if (!fetchOptions.headers.has("Authorization")) {
+        const token = await PhylumApi.getAccessToken();
+        fetchOptions.headers.set("Authorization", `Bearer ${token}`);
+      }
+
+      if (fetchOptions.headers.has("Content-Type")) {
+        fetchOptions.headers.set("Content-Type", "application/json");
+      }
+    } else {
+      if (!fetchOptions.headers["Authorization"]) {
+        const token = await PhylumApi.getAccessToken();
+        fetchOptions.headers["Authorization"] = `Bearer ${token}`;
+      }
+
+      if (!fetchOptions.headers["Content-Type"]) {
+        fetchOptions.headers["Content-Type"] = "application/json";
+      }
+    }
+
+    // Set Content-Type header if it is missing.
+
+    // Get API base URI without version.
+    const baseUrl = await PhylumApi.apiBaseUrl();
+
+    // Send fetch request.
+    return fetch(`${baseUrl}/${apiVersion}${endpoint}`, fetchOptions);
+  }
+
+  /**
    * Analyze dependencies in a lockfile.
    *
    * Packages are expected in the following format:
@@ -28,7 +84,7 @@ export class PhylumApi {
     package_type: string,
     packages: object[],
     project?: string,
-    group?: string,
+    group?: string
   ): Promise<string> {
     return Deno.core.opAsync(
       "analyze",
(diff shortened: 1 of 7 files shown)


kylewillmon:
For the name of the function, we could just use `fetch`. There should never be a name conflict since this in the `PhylumApi` class.

A few other options:
1. `request()`
1. `rawRequest()`
1. `bareRequest()`

kylewillmon:
I do think we need a net exception for our API to make this work. Making extension authors hard code the URL isn't reasonable. And the security implications of that exception should be minimal.

cli/src/extension_api.ts
```diff
@@ -26,15 +82,9 @@ export class PhylumApi {
     package_type: string,
     packages: object[],
     project?: string,
-    group?: string,
+    group?: string
```
kylewillmon, line 85: Is this from `prettier`?

I was using `deno fmt` when I made my changes. I guess we'll need to make a choice there.

cli/src/extension_api.ts
```diff
@@ -400,3 +445,8 @@ function opAsync<T>(op: string, ...args: unknown[]): Promise<T> {
   // deno-lint-ignore no-explicit-any
   return (Deno as any).core.opAsync(op, ...args) as Promise<T>;
 }
+
+/** Available Phylum REST API versions. **/
```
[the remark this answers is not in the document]
cd-work: Hm I'd hope that the chances of that are **very** unlikely, considering we should provide support for new API versions once they're released. But at the same time I guess we can do this.

One of the concerns might be that a user wouldn't know what **exactly** to put here. Is it `0`, `v0`, `/v0`, etc... But I guess providing the option is better than not allowing it at all.

cli/src/extension_api.ts
```diff
@@ -1,6 +1,62 @@
...
+        const token = await PhylumApi.getAccessToken();
+        fetchOptions.headers["Authorization"] = `Bearer ${token}`;
+      }
+
+      if (!fetchOptions.headers["Content-Type"]) {
+        fetchOptions.headers["Content-Type"] = "application/json";
+      }
+    }
+
+    // Set Content-Type header if it is missing.
```
kylewillmon, line 42: This fails `deno check` because it doesn't handle this annoying but valid way to specify headers:

```
{ headers: [["Accept", "*/*"], ["Content-Type", "application/json"]] }
```

You can greatly simplify the code if you just go ahead and create the `Headers` object. Then there is only one case to handle.

```suggestion
    const headers = new Headers(fetchOptions.headers);

    // Set Authorization header if it is missing.
    if (!headers.has("Authorization")) {
      const token = await PhylumApi.getAccessToken();
      headers.set("Authorization", `Bearer ${token}`);
    }

    // Set Content-Type header if it is missing.
    if (headers.has("Content-Type")) {
      headers.set("Content-Type", "application/json");
    }
```

docs/extensions/extension_rest_api.md
```diff
@@ -0,0 +1,63 @@
...
+);
+
+// Parse the reply as JSON.
+const projects = await reply.json();
+
+// Output all our projects.
+console.log(projects);
+```
+
+Allowed options match the `fetch` API and can be overwritten to send more
```
kylewillmon, line 39: We could consider adding a link to [fetch in the deno docs](https://deno.land/api?s=fetch) or something.

But there are many places where users can find these docs, so maybe that's overkill.

kylewillmon (changes requested):

kylewillmon (approved):

cli/src/extension_api.ts
```diff
@@ -1,6 +1,55 @@
...
+   * https://deno.land/api@latest?s=fetch
+   */
+  static async fetch(
+    apiVersion: ApiVersion | string,
+    endpoint: string,
+    init?: RequestInit,
+  ): Promise<object> {
+    // Ensure header object is initialized.
+    const fetchInit = init ?? {};
+    fetchInit.headers = fetchInit.headers ?? {};
```
cd-work, line 19:
```suggestion
```

cli/tests/end_to_end/extension.rs
```diff
@@ -185,3 +185,21 @@ pub fn async_state() {
...
 }
+
+#[tokio::test]
+pub async fn rest_api() {
+    let test_cli = TestCli::builder().with_config(None).build();
+
+    test_cli
+        .extension(
+            "
+            const reply = await PhylumApi.fetchPhylum(ApiVersion.V0, '/health');
```
kylewillmon, line 196: Looks like this was missed.

```suggestion
            const reply = await PhylumApi.fetch(ApiVersion.V0, '/health');
```

kylewillmon (changes requested):

kylewillmon (approved):

PR #873, opened 2022-12-09
Use lib.deno_core.d.ts types

kylewillmon:
Update extension_api.ts to use the types from lib.deno_core.d.ts and avoid any.

Thanks to this update, I discovered that `Deno.core.opSync` is no longer supported and has been [removed from deno_core]

[removed from deno_core]: https://github.com/denoland/deno/pull/16379

### DIFF
--- a/cli/src/extension_api.ts
+++ b/cli/src/extension_api.ts
@@ -1,5 +1,7 @@
 // deno-lint-ignore-file ban-types
 
+/// <reference types="https://raw.githubusercontent.com/denoland/deno/v1.28.3/core/lib.deno_core.d.ts" />
+
 export class PhylumApi {
   /**
    * Analyze dependencies in a lockfile.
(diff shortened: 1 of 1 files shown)


cli/src/extension_api.ts
```diff
@@ -1,5 +1,7 @@
 // deno-lint-ignore-file ban-types
 
+/// <reference types="https://raw.githubusercontent.com/denoland/deno/v1.28.3/core/lib.deno_core.d.ts" />
```
cd-work, line 3: Wait does this comment actually do something? Are the triple slashes intentional?

kylewillmon: Yes and yes

<https://deno.land/manual@v1.28.3/advanced/typescript/types#using-a-triple-slash-directive>

kylewillmon: I should clarify. It is only used during type checking. It will not download that file when run as a phylum extension (and our module loader wouldn't allow it if it tried).

Very weird that TypeScript separates the idea of type checking and compiling... Plenty of code will happily compile even if it doesn't type check correctly.

cd-work (approved):

andreaphylum (approved):

cd-work (approved):

PR #876, opened 2022-12-12
Use `deno fmt` and `deno lint`

kylewillmon:
This patch makes use of `deno` tooling for our extension code. This includes the use of `deno lint` for linting and a switch to `deno fmt` instead of `prettier`.

`deno lint` revealed many issues in our current extensions which have also been fixed in this PR.


extensions/find-permissions/main.ts
```diff
@@ -89,11 +89,11 @@ for (const path of requiredPaths) {
 console.log("]");
 
 // Recursively check the path for required sandboxing exceptions.
-async function checkPath(allowed: [string], path: string) {
+async function checkPath(allowed: string[], path: string) {
```
kylewillmon, line 92: I've seen this a few times now. So to clarify: `[string]` is an array with exactly one item (which is a `string`). This is the [tuple type](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) in TypeScript. Very different than `string[]`, which is an array of `string`s

extensions/npm/main.ts
```diff
@@ -27,21 +27,21 @@ class FileBackup {
...
         await Deno.remove(this.fileName);
       }
-    } catch (e) {}
+    } catch (_e) { /* Do nothing */ }
   }
 }
 
 // Find project root directory.
-async function findRoot(manifest: string): string | undefined {
+async function findRoot(manifest: string): Promise<string | undefined> {
```
andreaphylum, line 35: Weird that this wasn't caught already by `prettier` or at runtime by `deno` itself. I definitely remember `deno` complaining about an `async` function that did not return `Promise` in its signature.

extensions/find-permissions/main.ts
andreaphylum: Good catch, I wasn't aware of that.

andreaphylum (approved):
Looks good to me. I like `deno fmt` better than `prettier`, I feel like it improves the quality of the code by quite a bit.

cd-work (changes requested):
Can probably remove the .prettierignore, right?

extensions/npm/main.ts
kylewillmon: Type information is erased at runtime and `prettier` doesn't do type checking. Even in this branch, type checking is never performed. This was only caught because of my editor setup. I will bring that up in a separate issue and PR.

kylewillmon:
I will rebase this branch and clear up any conflicts after #868 is merged

cd-work (approved):

PR #878, opened 2022-12-14
Type check TypeScript files

kylewillmon:
Our files have a lot of type checking issues because most of our `PhylumApi` function return `object` instead of an actual type.

Resolving these issues might mean adding lots of new types to our extension API. It might be a while before I can do that, but I am putting up the PR to keep track of it.

### DIFF
--- a/cli/src/extension_api.ts
+++ b/cli/src/extension_api.ts
@@ -15,8 +15,8 @@ type ProcessOutput = {
   stdout: string;
   stderr: string;
   success: boolean;
-  signal?: number;
-  code?: number;
+  signal: number | null;
+  code: number | null;
 };
 
 export class PhylumApi {
--- a/cli/src/extension_api.ts
+++ b/cli/src/extension_api.ts
@@ -1,7 +1,23 @@
-// deno-lint-ignore-file ban-types
-
 /// <reference types="https://raw.githubusercontent.com/denoland/deno/v1.28.3/core/lib.deno_core.d.ts" />
 /// <reference lib="deno.window" />
+
+type Package = {
+  name: string;
+  version: string;
+};
+
+type Lockfile = {
+  package_type: string;
+  packages: Package[];
+};
+
+type ProcessOutput = {
+  stdout: string;
+  stderr: string;
+  success: boolean;
+  signal?: number;
+  code?: number;
+};
 
 export class PhylumApi {
   /**
@@ -16,7 +32,7 @@ export class PhylumApi {
     apiVersion: ApiVersion | string,
     endpoint: string,
     init?: RequestInit,
-  ): Promise<object> {
+  ): Promise<Response> {
     // Ensure header object is initialized.
     const fetchInit = init ?? {};
 
@@ -75,7 +91,7 @@ export class PhylumApi {
    */
   static analyze(
     package_type: string,
-    packages: object[],
+    packages: Package[],
     project?: string,
     group?: string,
   ): Promise<string> {
@@ -106,7 +122,7 @@ export class PhylumApi {
    * }
    * ```
    */
-  static getUserInfo(): Promise<object> {
+  static getUserInfo(): Promise<Record<string, unknown>> {
     return Deno.core.opAsync("get_user_info");
   }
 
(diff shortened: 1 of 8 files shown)


cd-work:
> Resolving these issues might mean adding lots of new types to our extension API.

This was intentionally not done because it's unrealistic with the amount of types we have to maintain them manually, right? That's why we have the issue to have them generated automatically.

That said, I don't see why `object` would stop us from type-checking. Yes we can't make statements about the contents of an object but it's still a completely valid TypeScript type.

kylewillmon:
> That said, I don't see why `object` would stop us from type-checking. Yes we can't make statements about the contents of an object but it's still a completely valid TypeScript type.

It's valid, but very difficult to use. In fact, it's flagged by the [`ban-types`](https://lint.deno.land/#ban-types) lint in `deno lint` (which we [intentionally disable](https://github.com/phylum-dev/cli/blob/0f2e75b7b89f4493a1939f853d416cb2d9aab0ae/cli/src/extension_api.ts#L1) so that `deno lint` succeeds).

Maybe that lint has the answer:

> Finally, Object and {} means "any non-nullish value" rather than "any object type". Record<string, unknown> is a good choice for a meaning of "any object type". On the other hand object type means "any object type", but using Record<string, unknown> is much more preferable in this case too.

Using `Record<string, unknown>` for these APIs would be a significant improvement and would probably make type checking a lot easier.

maxrake (approved):
The proof is right there in the [deno checks job](https://github.com/phylum-dev/cli/actions/runs/3809468959/jobs/6480850579#step:6:1)

cli/src/extension_api.ts
```diff
@@ -1,8 +1,24 @@
...
+  package_type: string;
+  packages: Package[];
+};
+
+type ProcessOutput = {
+  stdout: string;
+  stderr: string;
+  success: boolean;
+  signal: number | null;
+  code: number | null;
```
kylewillmon, line 19: Before writing up these types, I hadn't thought about the fact that these values can be `null` (because they come from a rust `Option<i32>`)...

extensions/yarn/main.ts
```diff
@@ -220,9 +220,9 @@ async function checkDryRun() {
 //
 // This assumes that execution was not successful and it will automatically
 // revert to the last stored package manager files.
-async function abort(code: number) {
+async function abort(code: number | null) {
   await restoreBackup();
-  Deno.exit(code);
+  Deno.exit(code ?? -1);
```
cd-work, line 225: Why was this changed? I don't think we ever abort without a code, right? And aborting with `-1` doesn't sound like a great idea.

kylewillmon: The code field returned by `runSandboxed` is `Option<i32>` in Rust, which becomes `number | null` in TypeScript. And we never check it against `null`. So I think we need this case.

I believe this happens when the process is killed by a signal. I am open to other values than `-1`

cd-work: This function performs the abort, it doesn't handle it. I'm slightly confused about your comment.

kylewillmon: Sorry. This basic code exists in many of our extensions:

```
const output = PhylumApi.runSandboxed(...);
if(!output.success) {
  await abort(output.code);
}
```

That is the code that I had in mind when making this change and writing the above comment.

But I suppose we can avoid changing this function if we handle that at the caller:

```
const output = PhylumApi.runSandboxed(...);
if(!output.success) {
  await abort(output.code ?? -1);
}
```

I will make that change

cd-work: I still wouldn't return -1, but a proper error code instead.

kylewillmon: Why is `-1` not a "proper error code"? Would you prefer I use `1` so that it is in the 0-255 range? Or do you think this case is important enough to have its own documented exit code?

maxrake (approved):

extensions/yarn/main.ts
cd-work: I don't think it's necessary to give it its own documented code, but I think we should pick one. I'd probably pick something a bit higher than 1 since that's what we do for most of our exit codes to avoid conflicts.

cd-work (approved):

PR #879, opened 2022-12-14
Add --reauth flag to phylum auth login

### LINKED ISSUE #858: Add flag to force re-authentication in `phylum auth login`
In staging, users are no longer required to re-authenticate when running `phylum auth login`. Instead, if the user is already logged in (e.g., via the UI), the auth provider forwards the user automatically.

The downside here is that nothing in this flow confirms that the user is logged in to the correct account. A user with multiple accounts may want to log in to a different account when running `phylum auth login` (e.g., when setting up a bot account for CI).

To fix this, the `phylum auth login` command should have a flag to force re-authentication when desired. For example:

```
phylum auth login --re-auth
```

Using this flag should set the `prompt=login` query parameter in the OIDC request, which should force the user to login. They may then login to the account of their choosing.

kylewillmon:
This patch adds support for `phylum auth login --reauth` to force reauthentication.

## Display issue

There seems to be an issue with the way Keycloak displays informational messages. This isn't something that can be fixed in the CLI so we'll have to take care of it separately. Maybe @janasheehan can take a look. Here is how to reproduce:

1. Login to [staging](https://app.staging.phylum.io/)
2. Run `phylum auth login -r` from this branch. Or just click [this link](https://login.staging.phylum.io/realms/phylum/protocol/openid-connect/auth?client_id=phylum_cli&code_challenge=AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA&code_challenge_method=S256&redirect_uri=http%3A%2F%2F127.0.0.1%3A12345%2F&response_type=code&response_mode=query&scope=openid+offline_access+profile+email&state=AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA&prompt=login), which hopefully should have the same effect.
3. See a banner that looks like this: 
   banner

Closes #858

### DIFF
--- a/cli/src/auth/oidc.rs
+++ b/cli/src/auth/oidc.rs
@@ -24,6 +24,7 @@ pub const OIDC_SCOPES: [&str; 4] = ["openid", "offline_access", "profile", "emai
 /// OIDC Client id used to identify this client to the oidc server
 pub const OIDC_CLIENT_ID: &str = "phylum_cli";
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub enum AuthAction {
     Login,
     Reauth,
@@ -100,13 +101,13 @@ pub struct OidcServerSettings {
 
 /// Using config information, build the url for the keycloak login page.
 pub fn build_auth_url(
-    action: &AuthAction,
+    action: AuthAction,
     oidc_settings: &OidcServerSettings,
     callback_url: &Url,
     code_challenge: &ChallengeCode,
     state: impl AsRef<str>,
 ) -> Result<Url> {
-    let mut auth_url = match *action {
+    let mut auth_url = match action {
         // Login uses the oidc defined /auth endpoint as is
         AuthAction::Login | AuthAction::Reauth => oidc_settings.authorization_endpoint.to_owned(),
         // Register uses the non-standard /registrations endpoint
@@ -133,7 +134,7 @@ pub fn build_auth_url(
         .append_pair("scope", &OIDC_SCOPES.join(" "))
         .append_pair("state", state.as_ref());
 
-    if matches!(*action, AuthAction::Reauth) {
+    if action == AuthAction::Reauth {
         auth_url.query_pairs_mut().append_pair("prompt", "login");
     }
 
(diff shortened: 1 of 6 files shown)


cli/src/auth/oidc.rs
```diff
@@ -132,6 +133,10 @@ pub fn build_auth_url(
         .append_pair("scope", &OIDC_SCOPES.join(" "))
         .append_pair("state", state.as_ref());
 
+    if matches!(*action, AuthAction::Reauth) {
```
cd-work, line 136: Would be cleaner if we just implemented PartialEq+Eq for `AuthAction` which is trivial to do in our case.

cli/src/commands/auth.rs
```diff
@@ -130,12 +130,14 @@ pub async fn handle_auth(
...
-            Err(error) => Err(error).context("User login failed"),
+        Some(("login", matches)) => {
+            match handle_auth_login(config, config_path, matches.get_flag("reauth")).await {
+                Ok(_) => {
+                    print_user_success!("{}", "User login successful");
+                    Ok(ExitCode::Ok.into())
+                },
+                Err(error) => Err(error).context("User login failed"),
+            }
         },
```
cd-work, line 141: I'm not sure I really care in this example, but by convention I feel like we tend to pass through the commands more often than not instead of getting the flag directly on the toplevel.

Since these toplevel structures tend to grow kinda unwieldy, it's usually nice to have all the stuff specific about each endpoint defined inside of it. It also means that future things pulled out of the CLI arguments don't require a change here anymore.

cd-work (changes requested):

cli/src/commands/auth.rs
andreaphylum: I agree with this, but if anything, I'd say it would be a good idea to do a refactoring pass and adapt all the other command handlers to have a similar behavior where command matches don't get passed thorugh. Not a huge ticket nor super urgent, but I'd suggest we think about it as a tech debt repayment at some point in the future.

andreaphylum (commented):
Other than what @cd-work suggested, looks good to me!

cli/src/commands/auth.rs
cd-work: Yeah I'm sure other areas are problematic too, but considering we should already be pretty good in most areas I'd refrain from introducing further tech debt if we can avoid it.

PR #882, opened 2022-12-19
Fix parser lockfile consistency

cd-work:
This patch is a rewrite of the lockfile parser with the goal to improve the consistency between the individual parsers.

Previously, some of the lockfile parsers would just ignore things like version ranges in favor of trying to extract as much valid data as possible. Instead any unknown syntax is now an immediate parsing failure and only strict dependency versions are allowed. This way we only ever analyze proper lockfiles.

Another inconsistency was the treatment of different dependency types. Some parsers were submitting git dependencies while others were just ignoring them. To help with parsing these different version types, there's a new `PackageVersion` enum in place of a version string which encodes where the version came from.

Using the new `PackageVersion` enum, it is now possible to filter packages in the CLI and determine after parsing which packages should be submitted. This also allows for printing warnings in the future to inform users about potentially unanalyzed dependencies.


lockfile/src/parsers/gem.rs
```diff
@@ -1,24 +1,196 @@
...
+
+            // Check for type of section head.
+            let section_type = if consumed.ends_with("GEM") {
+                SectionType::Gem
+            } else if consumed.ends_with("GIT") {
+                SectionType::Git
+            } else if consumed.ends_with("PATH") {
+                SectionType::Path
+            } else {
+                break;
```
kylewillmon, line 47: This should be unreachable code, right? If it ever becomes reachable, I don't want it to be silently ignored.

```suggestion
                unreachable!();
```

Returning an error would also be a decent option, but nom errors are weird to make...

lockfile/src/parsers/gem.rs
```diff
@@ -1,24 +1,196 @@
...
+            // Find next section head.
+            let header_result = recognize(many_till(
+                take_till_line_end,
+                alt((tag("GEM"), tag("GIT"), tag("PATH"))),
+            ))(input);
+
+            // Stop if no more headers can be found.
+            let (new_input, consumed) = match header_result {
+                Ok(header_result) => header_result,
+                Err(_) => break,
```
kylewillmon, line 36: This means we will just ignore any gibberish at the end of the file, right? It seems like that might make this parser accept things that it shouldn't.

lockfile/src/parsers/gem.rs
```diff
@@ -1,24 +1,196 @@
-use super::*;
+use nom::branch::alt;
+use nom::bytes::complete::{tag, take_until};
+use nom::character::complete::{line_ending, none_of, space0};
+use nom::combinator::recognize;
+use nom::error::{VerboseError, VerboseErrorKind};
+use nom::multi::{many1, many_till};
+use nom::sequence::{delimited, tuple};
+use nom::Err as NomErr;
```
kylewillmon, line 8: Is this rename actually necessary? Or do you just not like the look of `Err(Err::Failure(error))`?

lockfile/src/parsers/mod.rs
```diff
@@ -1,12 +1,11 @@
...
 use nom::bytes::complete::{tag, take, take_until};
-use nom::character::complete::{line_ending, multispace0, none_of, space0};
+use nom::character::complete::{line_ending, multispace0, space0};
 use nom::combinator::{eof, opt, recognize};
-use nom::error::{context, ParseError, VerboseError};
+use nom::error::{context, VerboseError};
 use nom::multi::{count, many1, many_till};
 use nom::sequence::{delimited, tuple};
 use nom::{AsChar, IResult};
-use phylum_types::types::package::{PackageDescriptor, PackageType};
```
kylewillmon, line 9: Does the lockfile crate still use `phylum_types` somewhere else or can we remove it as a dependency now?

lockfile/src/python.rs
```diff
@@ -182,58 +201,47 @@ mod tests {
...
+                version: PackageVersion::Git("git+https://github.com/matiascodesal/git-for-pip-example.git@v1.0.0".into()),
+            },
+            Package {
+                name: "tomli".into(),
+                version: PackageVersion::Internet("https://files.pythonhosted.org/packages/97/75/10a9ebee3fd790d20926a90a2547f0bf78f371b2f13aa822c759680ca7b9/tomli-2.0.1-py3-none-any.whl".into()),
+            },
+            Package {
+                name: "git+ssh://git@github.com/phylum-dev/phylum-ci.git@7d6d859ad368d1ab0a933f24679e3d3c08a40eac#egg=phylum".into(),
+                version: PackageVersion::Path(None),
+            },
```
kylewillmon, line 241: This test case seems wrong... Shouldn't this be `PackageVersion::Git`?

lockfile/src/python.rs
```diff
@@ -182,58 +201,47 @@ mod tests {
...
+                name: "attr".into(),
+                version: PackageVersion::Path(Some("file:///tmp/attr".into())),
+            },
+            Package {
+                name: "numpy".into(),
+                version: PackageVersion::Path(Some("file:///tmp/testing/numpy-1.23.5-pp38-pypy38_pp73-win_amd64.whl".into())),
+            },
+            Package {
+                name: "git-for-pip-example".into(),
+                version: PackageVersion::Git("git+https://github.com/matiascodesal/git-for-pip-example.git@v1.0.0".into()),
```
kylewillmon, line 232: This parser doesn't strip the `git+` like others do

kylewillmon (changes requested):
The `lockfile/werkzeug` file seems to have been added by mistake.

lockfile/src/parsers/gem.rs
```diff
@@ -128,12 +141,17 @@ impl<'a> Section<'a> {
...
+            .collect::<StdResult<_, _>>()?;
 
         // Bail if there isn't exactly one member in the `specs` section.
         if specs_packages.len() != 1 {
-            let kind =
-                VerboseErrorKind::Context("Invalid number of packages listed in path dependency");
+            let kind = VerboseErrorKind::Context(
+                "Invalid number of packages listed
+        in path dependency",
+            );
```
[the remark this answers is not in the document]
chrisduerr: It was not, great catch. Rustfmt formatted my commented-out code and that must have gotten mixed up in the process.

kylewillmon (approved):

PR #887, opened 2022-12-30
Add a `.phylum_project` file for GitHub App

kylewillmon:
This is a temporary solution to convince the GitHub App to ignore the many lockfiles in this repository that are used for testing. Those lockfiles contain many issues that would be concerning if they were actually in use.

### DIFF
--- a/.phylum_project
+++ b/.phylum_project
@@ -1,0 +1,6 @@
+# This file is a temporary fix to appease the Phylum GitHub App.
+# The format of this data is likely to change in the near future.
+
+lock_files:
+  - path: Cargo.lock
+    format: cargo


maxrake (approved):
Would adding these lines to a `.phylum_project` file that is produced by `phylum init` also work? Approving now since this is an admitted temporary fix.

cd-work:
I don't understand this fix, this isn't actually a valid `.phylum_project` file, is it? Is there a reason why you didn't go with the proper format?

kylewillmon:
Putting this on hold due to ongoing discussions about the file format and difference in functionality between the CLI and GitHub App.

maxrake:
Closing as OBE due to #928 ...

PR #888, opened 2022-12-30
Make project ID and name optional in .phylum_project file

kylewillmon:
Sometimes a `.phylum_project` file is useful for specifying a lockfile, but unnecessary for specifying an associated project. This change makes the project details optional in the `.phylum_project` file to allow for that possibility.

With this patch, a project could use a simple `.phylum_project` file like this one:

```yaml
lockfile_path: Cargo.lock
lockfile_type: cargo
```

And then submit their analysis jobs to different projects depending on the context:

* Local developer testing: `phylum analyze -p myproj`
* Manual CI: `phylum analyze --group company-devs --project some-proj`
* GitHub App: Submitted to appropriate project based on org settings
* `phylum-ci`: Submitted to appropriate project as configured in the workflow

### Additional questions

- [ ] Would it be better to put all project details in an optional sub-key? For example:
  ```yaml
  project:
    created_at: 2022-12-30T09:39:27.626246-06:00 # Is this even useful?
    id: 06babf58-62dd-4ffb-8f44-b65257b18b0c
    name: foo
    group_name: blah # This is optional
  lockfile_path: Cargo.lock
  lockfile_type: cargo
  ```
  If yes, how should users update to this new format? Do we need a command for that?
- [ ] Should `phylum init` allow skipping project details?
- [ ] Should a `phylum project unlink` command be added to remove project details?

### DIFF
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -157,13 +157,9 @@ async fn analyze(
 
     let (project, group) = match (project, group) {
         (Some(project), group) => (api.get_project_id(&project, group.as_deref()).await?, None),
-        (None, _) => {
-            if let Some(p) = config::get_current_project() {
-                (p.id, p.group_name)
-            } else {
-                return Err(anyhow!("Failed to find a valid project configuration"));
-            }
-        },
+        (None, _) => config::get_current_project()
+            .and_then(|c| c.project_group())
+            .ok_or_else(|| anyhow!("Failed to find a valid project configuration"))?,
     };
 
     let packages = packages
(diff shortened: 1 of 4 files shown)


cd-work:
None of the examples listed here actually require the key to be optional. With the lockfile specifically we're trying to go out of our way to make things simpler by enforcing its existence in the future, why are we going the exact opposite way with the project?

Is there actually a need to setup a project where **no default project** can be named at all?

kylewillmon:
> Is there actually a need to setup a project where **no default project** can be named at all?

Probably not. In the worst case, you could link to a project and then delete that project if you wanted to force people to supply their own project.

PR #892, opened 2023-01-04
Make config file write atomic

### LINKED ISSUE #851: Sporadic test failures
I've encountered this locally a couple of times and have seen CI fail for the same reason. It would probably be a good idea to look into this sporadic test failure:

```
thread 'extensions::create_extension' panicked at 'Unexpected failure.
code-1
stderr=```"❗ Error: Failed to read configuration at `/Users/phylum/.config/phylum/settings.yaml`: missing field `connection`\n"```
command=`"/Users/phylum/actions-runner/_work/cli/cli/target/debug/phylum" "extension" "new" "my-ext"`
code=1
stdout=""
stderr="❗ Error: Failed to read configuration at `/Users/phylum/.config/phylum/settings.yaml`: missing field `connection`\n"
', /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/ops/function.rs:248:5
```

If it was a network error sporadic failure should kind of be expected, but this seems to be something filesystem related which should be preventable. My best guess would be that we're modifying the config from one test while reading it from another?

cd-work:
This patch changes the way configuration files are written to first write their contents to a temporary file before then moving that file into place atomically. As a result the configuration file is always present and never contains partial data.

While I wasn't able to reproduce the issue locally, it should fix any test failures due to the configuration file not being accessible temporarily after writing to it.

Closes #851.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1,11 +1,10 @@
 use std::env::VarError;
 #[cfg(unix)]
 use std::fs::DirBuilder;
-#[cfg(unix)]
-use std::fs::Permissions;
+use std::fs::OpenOptions;
 use std::io::{self, Write};
 #[cfg(unix)]
-use std::os::unix::fs::{DirBuilderExt, PermissionsExt};
+use std::os::unix::fs::{DirBuilderExt, OpenOptionsExt};
 use std::path::{Path, PathBuf};
 use std::{env, fs};
 
@@ -15,7 +14,6 @@ use phylum_types::types::auth::RefreshToken;
 use phylum_types::types::common::ProjectId;
 use phylum_types::types::package::PackageType;
 use serde::{Deserialize, Deserializer, Serialize};
-use tempfile::NamedTempFile;
 
 use crate::{dirs, print_user_warning};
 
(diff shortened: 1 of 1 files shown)


maxrake (approved):
LGTM

andreaphylum (dismissed):

cli/src/config.rs
```diff
@@ -127,28 +130,7 @@ impl ProjectConfig {
...
-    // Use OpenOptions so that we can specify the permission bits
-    let mut opts = fs::OpenOptions::new();
-    opts.write(true).create(true).truncate(true);
-    #[cfg(unix)]
-    {
-        opts.mode(0o600);
-    }
-
-    // Windows file permissions are complicated and home folders aren't usually
-    // globally readable, so we can ignore Windows for now.
```
maxrake, line 142: The first half of this comment appears to be the reason the config tests are failing for Windows. It looks like ignoring it will no longer be possible.

maxrake (changes requested):
Other than the failing tests on Windows, looks good.

cli/src/config.rs
cd-work: I'm not entirely sure what this comment means by "globally readable", but I don't think that should be cause for this issue.

kylewillmon: I suspect the Windows tests are failing because  something (probably another test running in parallel) still has an open handle on the old config file when the `fs::rename()` happens.

kylewillmon (approved):

maxrake (approved):

kylewillmon (approved):

PR #901, opened 2023-01-06
chore: fix README test status badge

maxrake:
See: https://github.com/badges/shields/issues/8671

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 [![GitHub release (latest by date)](https://img.shields.io/github/v/release/phylum-dev/cli)](https://github.com/phylum-dev/cli/releases/latest/)
 [![License](https://img.shields.io/github/license/phylum-dev/cli)](https://github.com/phylum-dev/cli/blob/main/LICENSE)
-[![Test Status](https://github.com/phylum-dev/cli/actions/workflows/test.yml/badge.svg?branch=master)](https://github.com/phylum-dev/cli/actions/workflows/test.yml)
+[![Test Status](https://img.shields.io/github/actions/workflow/status/phylum-dev/cli/test.yml?branch=main&label=tests&logo=GitHub)](https://github.com/phylum-dev/cli/actions/workflows/test.yml)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,7 +8,7 @@
 
 [![GitHub release (latest by date)](https://img.shields.io/github/v/release/phylum-dev/cli)](https://github.com/phylum-dev/cli/releases/latest/)
 [![License](https://img.shields.io/github/license/phylum-dev/cli)](https://github.com/phylum-dev/cli/blob/main/LICENSE)
-[![Test Status](https://github.com/phylum-dev/cli/actions/workflows/test.yml/badge.svg?branch=master)](https://github.com/phylum-dev/cli/actions/workflows/test.yml)
+[![Test Status](https://img.shields.io/github/actions/workflow/status/phylum-dev/cli/test.yml?branch=main&label=tests&logo=GitHub)](https://github.com/phylum-dev/cli/actions/workflows/test.yml)
```
kylewillmon, line 11: Why not just stick with GitHub's badge? It's not broken.

kylewillmon (changes requested):
Why not just stick with GitHub's badge? It's not broken.

README.md
maxrake: There was a thought about doing that...but then it was noticed that the `branch=master` instead of `branch=main`...which was the last vestige of `master` in the repo. The shield still worked b/c it would fall back to the default branch. The only reason it was changed here was so that it would match the other instances in the `phylum-dev` org (currently only `phylum-ci` repo). AFAICT, the only advantage to using the shield.io form instead of the GitHub form of the badge is to be able to use the full range of shield styles...which helps for uniformity.

That said, the badge was working so it will be reverted (but with `main` instead of `master`) with no complaint if that is desired.

kylewillmon: I didn't notice the `master` reference... Good catch.

I would prefer to stick with the GitHub badge

kylewillmon (approved):

PR #902, opened 2023-01-06
Add multi-lockfile support to `.phylum_project`

### LINKED ISSUE #904: Update `phylum_project` format
# Overview
We've recently changed what our `phylum_project` file supports as a result of our recent Github App addition. Changes have been made to API for handling the project file, we need to make similar changes in the CLI directly. The expected structure is as follows:

```
lockfiles:
    - path: /absolute/path.json
      type: npm
    - path: /foo/bar/requirements.txt
      type: pypi
    ...
```

# Acceptance Criteria
- [ ] We have support for multiple paths to lockfiles
- [ ] Lockfile structure/format matches with ticket outlined [here](https://github.com/phylum-dev/api/issues/699#issuecomment-1378170100)

kylewillmon:
This patch allows specifying multiple lockfiles and their respective
lockfile type in the `.phylum_project`, which will then be used to
submit all packages from all lockfiles when `phylum parse` or
`phylum analyze` is run.

Closes #904

### DIFF
--- a/cli/src/commands/jobs.rs
+++ b/cli/src/commands/jobs.rs
@@ -11,7 +11,7 @@ use reqwest::StatusCode;
 
 use crate::api::{PhylumApi, PhylumApiError};
 use crate::commands::{parse, CommandResult, CommandValue, ExitCode};
-use crate::config::get_current_project;
+use crate::config::{get_current_project, LockfileConfig};
 use crate::filter::{Filter, FilterIssues};
 use crate::format::Format;
 use crate::{print_user_success, print_user_warning};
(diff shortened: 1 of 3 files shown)


cli/src/commands/jobs.rs
```diff
@@ -228,28 +231,15 @@ impl JobsProject {
     /// Assumes that the clap `matches` has a `project` and `group` arguments
     /// option.
     async fn new(api: &mut PhylumApi, matches: &clap::ArgMatches) -> Result<JobsProject> {
-        let cli_lockfile_type = matches.try_get_one::<String>("lockfile-type").ok().flatten();
-        let cli_lockfile = matches.try_get_one::<String>("LOCKFILE").ok().flatten();
```
kylewillmon, line 232: The new code uses `get_one` instead of `try_get_one`.

I believe the reason for `try_get_one` was to handle the `phylum batch` command (which doesn't have these parameters)

As a side note on `phylum batch`: The `phylum batch` command is rarely used... But I think there might still be some scripts out there that rely on it. We could consider rewriting `phylum batch` as an extension in `community-extensions` and removing it from the core CLI.

cd-work (approved):

cd-work (approved):

andreaphylum (approved):

PR #911, opened 2023-01-12
Inline format arguments when possible

cd-work:
The latest clippy nightly version further encourages inlining format arguments by warning for every instance where it is missing. This lint will likely make its way to stable in the near future and is a sensible suggestion for consistency and brevity.


andreaphylum (approved):
I like this a lot.

PR #915, opened 2023-01-17
Abort on unknown extension subcommands

### LINKED ISSUE #896: Yarn/Poetry extensions will break when subcommand is not first argument
Currently to determine if an install is taking place, the first argument is inspected and compared against the known arguments. This will however break when executing something like `yarn --cwd=/tmp/project install`. Unfortunately this will fail in an unsafe manner, not actually checking the dependencies, rather than just not running the command at all.

We should make sure every way to reach the `install` command for each respective ecosystem is covered by our argument parser, to ensure at the very least we can fail when something is not understood or print a warning.

My understanding from looking at `npm --help` is that the `npm` extension is unaffected by this, since it does not allow any arguments before the first subcommand.

cd-work:
This adds a workaround to deal with the possibility that flags are passed before the first subcommand with ecosystem extensions by validating that the first passed argument is a known subcommand.

Closes #896.


extensions/poetry/main.ts
```diff
@@ -50,6 +50,49 @@ async function findRoot(manifest: string): Promise<string | undefined> {
...
+  "shell",
+  "show",
+  "update",
+  "version",
+  "cache",
+  "debug",
+  "env",
+  "self",
+  "source",
+];
```
andreaphylum, line 80: Do we have a way of future-proofing this? I'm not saying to add any code to the extension, per se, but I feel like actively monitoring new releases for every tool we support in a manual way could be prone to error. We could add a few scripts _somewhere_ that, say, check the `--help` output or something similar for exhaustiveness.

Definitely a matter for a separate issue and low priority, though.

andreaphylum (approved):

extensions/poetry/main.ts
cd-work: I've contemplated how we could future-proof this but I genuinely have no idea.

I hope this will serve just as a somewhat temporary solution until we can find something more proper. But as these things usually go, temporary solutions are the most permanent things there are.

maxrake (approved):

PR #920, opened 2023-01-18
Switch to `phylum init` for CLI docs

### LINKED ISSUE #917: Update docs to prefer `phylum init`
Now that `phylum init` is the recommended way to bootstrap (and update) environments, the documentation should be updated to reflect it. The older (but still functional) method of using `phylum project` and `phylum group` commands can be replaced everywhere they are referenced, but especially in the quickstart guides. It might also be nice to explain how multiple lockfiles are supported and what to do if any fundamentals of the repository change such that the project file needs to be updated. Finally, this documentation change should be coordinated with a CLI release so that the changes do not remain in limbo for very long.

cd-work:
Since `phylum init` provides a better user experience than manually creating projects yourself, we should recommend it over the other alternatives.

Closes #917.

### DIFF
--- a/docs/knowledge_base/analyzing-dependencies.md
+++ b/docs/knowledge_base/analyzing-dependencies.md
@@ -28,13 +28,13 @@ The Phylum CLI natively supports processing the lock/requirements files for seve
 After setting up a Phylum [project](https://docs.phylum.io/docs/phylum_init), you can begin analysis by running:
 
 ```sh
-phylum analyze <package-lock-file.ext>
+phylum analyze
 ```
 
 The default response will provide you with a high-level overview of your packages, including the total project score, score distributions across all packages, whether or not this analysis was a pass or fail and the total number of packages still processing.
 
 ```
-$ phylum analyze package-lock.json
+$ phylum analyze
 ✅ Job ID: 3cd30a5b-eeee-4ba1-b8e1-276c61e6502c
 
 
@@ -63,13 +63,13 @@ $ phylum analyze package-lock.json
 You can get more detailed output from the analysis, to include specific issues and their severity, by using the `--verbose` flag:
 
 ```sh
-phylum analyze --verbose <package-lock-file.ext>
+phylum analyze --verbose
 ```
 
 If you prefer JSON formatted output, you can leverage the `--json` flag.
 
 ```sh
-phylum analyze --verbose --json <package-lock-file.ext> > output.json
+phylum analyze --verbose --json > output.json
 ```
 
 If the analysis failed to meet the project's thresholds, the command's exit code will be set to `100`.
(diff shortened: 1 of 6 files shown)


cd-work:
Working on a patch for https://github.com/phylum-dev/documentation separately, since these are obviously two different repos.

docs/knowledge_base/analyzing-dependencies.md
```diff
@@ -25,7 +25,7 @@ The Phylum CLI natively supports processing the lock/requirements files for seve
 * Cargo
     * `Cargo.lock`
 
-After setting up a Phylum [project](https://docs.phylum.io/docs/phylum_project) , you can begin analysis by running:
+After setting up a Phylum [project](https://docs.phylum.io/docs/phylum_init), you can begin analysis by running:
 
 ```sh
 phylum analyze <package-lock-file.ext>
```
maxrake, line 31: The other instances of `phylum analyze` in this review were changed to remove `<package-lock-file.ext>`. Why not here?

cd-work:
Docs PR: https://github.com/phylum-dev/documentation/pull/4

maxrake (changes requested):
Basic changes look good. There are two additional suggestions:

* Be consistent in the use of `phylum analyze` (see comment)
* Provide detail somewhere that `phylum init` can be used for more than just setting up a _new_ Phylum project
  * It can also be used to _update_ an existing project, to change the group, lockfiles, or simply to get the latest format of the `.phylum_project` file
  * `docs/knowledge_base/groups.md` seems like one good candidate
    * Unrelated...but if modifying anyway...`groups.md` has a typo on line 13: `anlaysis` should be `analysis`
    * It could also be updated to combine steps 1 and 2 or indicate that they can be performed with a single `phylum init` command
  * Maybe `docs/command_line_tool/phylum_init.md` would be another place for this detail

docs/knowledge_base/analyzing-dependencies.md
cd-work: Just an oversight.

cd-work:
> It can also be used to update an existing project, to change the group, lockfiles, or simply to get the latest format of the .phylum_project file

I'm not sure we want to explicitly document it as such at this point?

> It could also be updated to combine steps 1 and 2 or indicate that they can be performed with a single phylum init command

Groups are not created with `phylum init`. If we want this to be possible, we'd have to create a separate issue.

That said, I'm not sure I'd modify the documentation in this case at all. Currently we'd probably write a `.phylum_project` file once the project is created but it would be nice if the steps outlined here worked without touching `.phylum_project` at all. So a user can just setup a project/group from anywhere.

maxrake (approved):

maxrake:
> Groups are not created with phylum init. If we want this to be possible, we'd have to create a separate issue.

Oh yeah...my mistake. I suppose the times I've used this command, the group was already created. I don't know if it should be a feature but can suggest that at the next team leads meeting.

PR #925, opened 2023-01-25
Add multi-lockfile ecosystems to analysis summary

### LINKED ISSUE #924: Ecosystem on analysis summary should list all identified ecosystems
Currently the CLI only shows one ecosystem in the analysis summary when submitting multiple different ecosystem's lockfiles at the same time. Instead, we should provide a comma-separated list of ecosystems.

### Before

```
Ecosystem: Cargo
```

### After

```
Ecosystems: Cargo, PyPi
```

Having it as `Ecosystem(s)` to avoid additional code for pluralization might be an option to keep things simple, but I think we can do better than that.

cd-work:
Since multi-lockfile is capable of running an analysis on multiple ecosystems, the existing `Ecosystem` field was no longer accurate since it only contained a single ecosystem.

Instead when multiple ecosystems are detected, they're now specified in a comma-separated list.

Closes #924.

### DIFF
--- a/cli/src/format.rs
+++ b/cli/src/format.rs
@@ -317,7 +317,7 @@ where
         .ecosystems
         .iter()
         .flat_map(|ecosystem| {
-            Some(match PackageType::from_str(&ecosystem).ok()? {
+            Some(match PackageType::from_str(ecosystem).ok()? {
                 PackageType::Npm => "NPM",
                 PackageType::RubyGems => "RubyGems",
                 PackageType::PyPi => "PyPI",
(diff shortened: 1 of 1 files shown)


cli/src/format.rs
```diff
@@ -312,16 +312,31 @@ fn response_to_table<T>(resp: &JobStatusResponse<T>) -> Table
...
+                PackageType::Maven => "Maven",
+                PackageType::Nuget => "NuGet",
+                PackageType::Golang => "Golang",
+                PackageType::Cargo => "Cargo",
+            })
+        })
+        .collect::<Vec<_>>();
+    let mut ecosystems_value = ecosystems.join(", ");
+
+    // Add fallback if no ecosystem could be identified.
```
maxrake, line 333: This fallback path was encountered in a situation where it wasn't expected:

image

---

The same `analyze` command performed with CLI v4.4.0 shows the ecosystem as `Cargo` whereas with the code from this PR it shows as `Unknown`.

maxrake (changes requested):
Something is not quite right here...but I can't say what the fix is. There is at least one change needed to make Clippy pass in CI. There may be another one for getting a single ecosystem to show correctly. Manual local testing stopped after this finding/regression, before multiple ecosystems could be tested.

cli/src/format.rs
maxrake: This turned out to be due to testing on PROD instead of STG.

maxrake (approved):
The code looks good. Functionality was confirmed with manual local testing against the staging environment. It worked for both a single and multiple lockfile analysis submission. LGTM.

PR #926, opened 2023-01-25
Fix poetry extension

### LINKED ISSUE #804: Poetry extension is broken
# Overview
Using the Phylum `poetry` extension fails for each of the supported subcommands: `add`, `update`, and `install`. This is the case for the latest major release version of `poetry`, v1.2.x, and has not been confirmed for the previous release, v1.1.x

# How To Reproduce
1. Ensure [poetry v1.2+](https://python-poetry.org/docs/) is installed
2. Initialize a new project or go to an existing Poetry project (e.g., `phylum-dev/phylum-ci`)
3. Attempt to install the project with the phylum `poetry` extension
4. See the error
5. Attempt to add a new package or update the lockfile
6. See the error

image

---

It is not limited to the `install` command

image

# Expected Behavior
The Poetry commands `add`, `update`, and `install` work when passed through the Phylum `poetry` extension. If any of those commands attempt to add/update/install package(s) that fail the risk domain thresholds, the action is prevented and the user notified of the offending package(s) and reason for failure.

# Additional Context
The latest version of Poetry is v1.2.x but the extension was written and tested for the v1.1.x release. These releases are very different and might play some role in the failure(s). However, the extension attempts to run the commands with the `-n` and `--lock` options and, at least in the case of the `install` command, the `--lock` option is not present in _either_ the v1.1 or v1.2 releases.

cd-work:
This patch updates the poetry extension to make use of the `--lock` flag for dry-running the package installation without executing any code.

Installation after analysis is done outside of the sandbox, since it seems to allow for near arbitrary code execution.

A more blacklist-based approach to deny some access (like $HOME) could likely be introduced in the future.

Closes #804.

### DIFF
--- a/cli/src/commands/extensions/permissions.rs
+++ b/cli/src/commands/extensions/permissions.rs
@@ -283,9 +283,6 @@ pub fn default_sandbox() -> SandboxResult<Birdcage> {
     add_exception(&mut birdcage, Exception::ExecuteAndRead("/opt/homebrew".into()))?;
     add_exception(&mut birdcage, Exception::ExecuteAndRead("/usr/local".into()))?;
 
-    // Allow `env` exec to resolve binary paths.
-    add_exception(&mut birdcage, Exception::ExecuteAndRead("/usr/bin/env".into()))?;
-
     // Allow access to DNS list.
     //
     // While this is required to send DNS requests for network queries, this does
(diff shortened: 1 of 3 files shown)


andreaphylum (commented):
Looks good to me, just a doubt on virtual envs.

andreaphylum (approved):

PR #927, opened 2023-01-25
Add option to specify multiple lockfiles on CLI

### LINKED ISSUE #906: Add support for multiple lockfiles in parse/analyze/init CLI
Currently if a user wants to analyze multiple lockfiles at once, they **must** use the `.phylum_project` file. It should be possible to do the same using only command line flags.

The main issue with this is that while `--lockfile` could easily be specified multiple times, the associated `--lockfile-type` can no longer be stored separately since it requires some association to one of the lockfiles.

One option would be to switch the lockfile to `npm:package-lock.json` instead, where an optional ecosystem prefix is allowed in the `--lockfile` to force a given ecosystem.

cd-work:
This patch changes the `--lockfile` option in all places to allow for multiple occurrences, allowing proper multi-lockfile support for the entire CLI.

The `--lockfile-type` parameter itself is still only allowed once and applies to all other specified lockfiles. This should be fine considering it's only a last-ditch escape hatch since it should always be possible to just let the auto-parser figure things out.

Closes #906.

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -232,16 +232,17 @@ pub fn add_subcommands(command: Command) -> Command {
         .subcommand(Command::new("ping").about("Ping the remote system to verify it is available"))
         .subcommand(
             Command::new("parse").about("Parse a lockfile and output its packages as JSON").args(&[
-                Arg::new("LOCKFILE")
+                Arg::new("lockfile")
                     .value_name("LOCKFILE")
                     .value_hint(ValueHint::FilePath)
-                    .help("The package lock file to submit."),
+                    .help("The package lock file to submit.")
+                    .action(ArgAction::Append),
                 Arg::new("lockfile-type")
                     .short('t')
                     .long("lockfile-type")
                     .value_name("type")
                     .requires("LOCKFILE")
-                    .help("The type of the lock file (default: auto)")
+                    .help("The type of the lock files (default: auto)")
                     .value_parser(PossibleValuesParser::new(parse::lockfile_types(true))),
             ]),
         )
@@ -275,16 +276,17 @@ pub fn add_subcommands(command: Command) -> Command {
                         .value_name("group_name")
                         .help("Specify a group to use for analysis")
                         .requires("project"),
-                    Arg::new("LOCKFILE")
+                    Arg::new("lockfile")
                         .value_name("LOCKFILE")
                         .value_hint(ValueHint::FilePath)
-                        .help("The package lock file to submit."),
+                        .help("The package lock file to submit.")
+                        .action(ArgAction::Append),
                     Arg::new("lockfile-type")
                         .short('t')
                         .long("lockfile-type")
                         .value_name("type")
                         .requires("LOCKFILE")
-                        .help("The type of the lock file (default: auto)")
+                        .help("The type of the lock files (default: auto)")
                         .value_parser(PossibleValuesParser::new(parse::lockfile_types(true))),
                 ]),
         )
@@ -445,14 +447,15 @@ pub fn add_subcommands(command: Command) -> Command {
                         .short('l')
                         .long("lockfile")
                         .value_name("LOCKFILE")
-                        .help("Project lockfile name"),
+                        .help("Project lockfile name")
+                        .action(ArgAction::Append),
                     Arg::new("lockfile-type")
                         .short('t')
                         .long("lockfile-type")
-                        .value_name("LOCKFILE_TYPE")
-                        .requires("lockfile")
-                        .help("Project lockfile type")
-                        .value_parser(PossibleValuesParser::new(parse::lockfile_types(false))),
+                        .value_name("type")
+                        .requires("LOCKFILE")
+                        .help("The type of the lock files (default: auto)")
+                        .value_parser(PossibleValuesParser::new(parse::lockfile_types(true))),
                     Arg::new("force")
                         .short('f')
                         .long("force")
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -241,7 +241,7 @@ pub fn add_subcommands(command: Command) -> Command {
                     .short('t')
                     .long("lockfile-type")
                     .value_name("type")
-                    .requires("LOCKFILE")
+                    .requires("lockfile")
                     .help("The type of the lock files (default: auto)")
                     .value_parser(PossibleValuesParser::new(parse::lockfile_types(true))),
             ]),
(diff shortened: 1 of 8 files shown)


cli/src/app.rs
```diff
@@ -232,16 +232,17 @@ pub fn add_subcommands(command: Command) -> Command {
...
                     .value_name("LOCKFILE")
                     .value_hint(ValueHint::FilePath)
-                    .help("The package lock file to submit."),
+                    .help("The package lock file to submit.")
+                    .action(ArgAction::Append),
                 Arg::new("lockfile-type")
                     .short('t')
                     .long("lockfile-type")
                     .value_name("type")
                     .requires("LOCKFILE")
```
maxrake, line 244:
```suggestion
                    .requires("lockfile")
```
This change will fix the broken test.

cli/src/app.rs
```diff
@@ -275,16 +276,17 @@ pub fn add_subcommands(command: Command) -> Command {
...
                         .value_name("LOCKFILE")
                         .value_hint(ValueHint::FilePath)
-                        .help("The package lock file to submit."),
+                        .help("The package lock file to submit.")
+                        .action(ArgAction::Append),
                     Arg::new("lockfile-type")
                         .short('t')
                         .long("lockfile-type")
                         .value_name("type")
                         .requires("LOCKFILE")
```
maxrake, line 288:
```suggestion
                        .requires("lockfile")
```

cli/src/app.rs
```diff
@@ -445,14 +447,15 @@ pub fn add_subcommands(command: Command) -> Command {
...
+                        .action(ArgAction::Append),
                     Arg::new("lockfile-type")
                         .short('t')
                         .long("lockfile-type")
-                        .value_name("LOCKFILE_TYPE")
-                        .requires("lockfile")
-                        .help("Project lockfile type")
-                        .value_parser(PossibleValuesParser::new(parse::lockfile_types(false))),
+                        .value_name("type")
+                        .requires("LOCKFILE")
```
maxrake, line 456:
```suggestion
                        .requires("lockfile")
```

maxrake (changes requested):
The proposed changes are needed to avoid panics and failing tests. The changes have been made locally and will be evaluated with the assumption that they will be made here. Additional comments will be provided separately.

maxrake:
A details section in the doc templates for the affected commands would be a good place to explain to users how the `--lockfile-type` option is meant to be used (or rather, not used). Otherwise, they (and our future selves) will likely have questions when multiple lockfile types are used but only one `--lockfile-type` is allowed.

maxrake:
It is not obvious from the help output or the documentation that `--lockfile` can be applied more than once. If getting the `...` postfix on the argument in the `--help` output is not possible, then the text of the `--help` output and/or the command `.md` documentation should be updated to reflect the multiple use possibility.

maxrake:
Not an issue with the changes in this PR...but noticed while testing...

The output of the `phylum init` command, when the project or lockfile(s) are provided as arguments, will add extra blank lines. It appears there are `println!();` statements applied unilaterally in several places in `cli/src/commands/init.rs`. The command still functions, but the output looks slightly odd with the extra blank lines:

image

---

This comment can certainly be extracted into a separate issue...just adding it here in case changes are already going to be made around that code.

cd-work:
> A details section in the doc templates for the affected commands would be a good place to explain to users how the --lockfile-type option is meant to be used (or rather, not used). Otherwise, they (and our future selves) will likely have questions when multiple lockfile types are used but only one --lockfile-type is allowed.

What exactly would you document here that isn't already documented by the CLI? The documentation makes it very clear that multiple lockfiles are allowed but only one type. The behavior should be obvious from that, considering there's really only one way a single lockfile type could be used at that point.

> It is not obvious from the help output or the documentation that --lockfile can be applied more than once. If getting the ... postfix on the argument in the --help output is not possible, then the text of the --help output and/or the command .md documentation should be updated to reflect the multiple use possibility.

I genuinely have no clue anymore how clap is supposed to work. It introduces breaking changes so often, stuff just changes every other week. The only sensible solution I can see here is filing a bug upstream if that makes sense to you.

> This comment can certainly be extracted into a separate issue...just adding it here in case changes are already going to be made around that code.

I'd put that in a separate issue, yeah.

maxrake (changes requested):
The suggested code changes in this review are meant to make it clear, from the `--help` output text, that one or more lockfile(s) can be specified.

---

This is a situation that is envisioned for new users who don't have clear guidance or examples as to how the `--lockfile-type` argument is expected to be used:

image

---

The notes for the PR explain this:

> The --lockfile-type parameter itself is still only allowed once and applies to all other specified lockfiles. This should be fine considering it's only a last-ditch escape hatch since it should always be possible to just let the auto-parser figure things out.

...but that explanation is not found or obvious from looking at the output of `--help` or the online documentation.

The suggestion here is to provide more context for how this argument works...either in the `--help` output directly, in the `.md` documentation files, or both. The documentation files allow for this kind of explanation with the inclusion of a "details" section in the template and/or adding examples. Bottom line, it needs to be very clear that the current expectation is that `auto` should be used for the `--lockfile-type` whenever there are multiple lockfile types specified.

---

Filing a bug with `clap` to get the `...` postfix notation to show for the option is fine. Another possibility, since this appears to only show up for the `init` command is to remove `LOCKFILE` as a positional argument for that command (it already isn't required and has default behavior when not specified) and make it an option. Then, `--lockfile` can be defined as an option which allows multiple values (see [clap examples](https://docs.rs/clap/latest/clap/struct.Arg.html#method.num_args)):

```rust
.action(ArgAction::Set)
.num_args(0..)
```

---

The errors in CI now are related to the docs not matching. Since `cargo xtask gendocs` is needed anyway, it would be a great thing to update the templates to add details and examples for how these three commands are expected to be called with one or more lockfiles and how detecting the lockfile type is preferred with the `auto` default. More detailed documentation is almost always better IMO.

cd-work:
Created a clap issue upstream: https://github.com/clap-rs/clap/issues/4681

maxrake (changes requested):
The changes to the `--help` output do improve things a bit.

The errors in CI are still related to the docs not matching. Since `cargo xtask gendocs` is needed anyway, it would be a great thing to update the templates to add details and examples for how these three commands are expected to be called with one or more lockfiles and how detecting the lockfile type is preferred with the `auto` default.

maxrake (approved):
LGTM

PR #928, opened 2023-01-25
Add Phylum project file

cd-work:
This adds the `.phylum_project` file with our lockfile information to ensure the GitHub App does not run analysis on all our test fixture lockfiles.

### DIFF
--- a/.phylum_project
+++ b/.phylum_project
@@ -1,0 +1,7 @@
+id: 91324b39-08f7-4de5-8266-230404b43a06
+name: cli
+created_at: 2023-01-25T20:51:40.553394940+01:00
+group_name: null
+lockfiles:
+- path: Cargo.lock
+  type: cargo
--- a/.phylum_project
+++ b/.phylum_project
@@ -1,7 +1,7 @@
-id: 91324b39-08f7-4de5-8266-230404b43a06
+id: d84b0bb3-fb64-43cb-a473-dc96b54db662
 name: cli
-created_at: 2023-01-25T20:51:40.553394940+01:00
-group_name: null
+created_at: 2023-01-26T20:02:11.276987834+01:00
+group_name: integrations
 lockfiles:
 - path: Cargo.lock
   type: cargo


.phylum_project
```diff
@@ -0,0 +1,7 @@
+id: 91324b39-08f7-4de5-8266-230404b43a06
+name: cli
+created_at: 2023-01-25T20:51:40.553394940+01:00
+group_name: null
```
maxrake, line 4:
```suggestion
group_name: phylum-dev-github-2
```
This is the group name currently in use.

.phylum_project
```diff
@@ -0,0 +1,7 @@
+id: 91324b39-08f7-4de5-8266-230404b43a06
```
maxrake, line 1:
```suggestion
id: 5091cbde-1187-4d59-992c-a46e85dd9f26
```
This is the existing project that will be linked/used if the `group_name` below is changed to `phylum-dev-github-2`

maxrake (changes requested):
Discussion on the `group_name` to use is needed. There is already a project for this repo, used by the GitHub app, that uses the `phylum-dev-github-2` group. If that one is to (continue to be) used here, then the `id` will need to be updated as well. Suggestions for both have been provided.

cd-work:
@maxrake I don't think we want to use the GitHub App's project, because that will cause conflicts with the App's analysis every time a dev runs an analysis?

We could use the same group, but I'm not entirely sure since it's everything but descriptive.

maxrake:
> @maxrake I don't think we want to use the GitHub App's project, because that will cause conflicts with the App's analysis every time a dev runs an analysis?

Yeah...that is the case now, with the way the most recent analysis, by label, will show in the UI. So...I guess a different group is okay.

> We could use the same group, but I'm not entirely sure since it's everything but descriptive.

I think it can be a different group...it just might be confusing to anyone who doesn't know about the "shadow" app group. There should be a group specified here since without one, I believe only to project creator (you) will see the results in the UI. If we want to use an existing group, there is `phylum_bot` that has a fair/correct number of users already.

maxrake (approved):

PR #930, opened 2023-01-28
Fix install with poetry extension

cd-work:
Since the `poetry install` subcommand does not have a `--lock` argument, it would cause the subcommand to fail. Since `install` should generally be based on an updated lockfile already anyway, it should be fine to run an analysis without first updating the lockfile.

However there are options allowing `install` to modify the lockfile, which is why a proper solution would likely involve parsing the `--dry-run` output instead.

### DIFF
--- a/extensions/poetry/main.ts
+++ b/extensions/poetry/main.ts
@@ -188,6 +188,8 @@ async function poetryCheckDryRun(
       console.error(`[${red("phylum")}] Lockfile update failed.\n`);
       await abort(status.code ?? 255);
     }
+
+    console.log(`[${green("phylum")}] Lockfile updated successfully.\n`);
   }
 
   const lockfileData = await PhylumApi.parseLockfile("./poetry.lock", "poetry");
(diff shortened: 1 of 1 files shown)


cd-work:
Follow-up issue for a "proper", more complicated fix: https://github.com/phylum-dev/cli/issues/931

andreaphylum (approved):

cd-work:
> The extensions/poetry/README.md documentation should be updated to make it more clear that the install command is not currently supported...even if it is only temporary.

But it is supported? It's just *some* subset of the install command that might not run an analysis on the new stuff. And I'm sure that's not the only way to escape analysis.

maxrake:
> But it is supported? It's just _some_ subset of the install command that might not run an analysis on the new stuff. And I'm sure that's not the only way to escape analysis.

Oh...I see it now! 🤦 
The lockfile is still analyzed before the `install` command is allowed to proceed. I was not paying close enough attention to the code flow...sorry.

maxrake (changes requested):
The output of running this extension did cause some confusion since it prints `[phylum] Lockfile updated successfully.` for the `install` subcommand...which is misleading b/c that subcommand is excluded from the block of code that does update the lockfile.

It appears maybe this line just needs to be moved up, to go inside the larger block that excludes the `install` subcommand:

```
  console.log(`[${green("phylum")}] Lockfile updated successfully.\n`);
```

maxrake (approved):
LGTM

louislang (approved):
lgtm! 🚀

PR #939, opened 2023-02-07
Switch to crates.io version of birdcage

cd-work:


maxrake (approved):
That's funny...the 1 package that Phylum analysis shows as still processing is `birdcage@0.2.0`.
Approving now with the assumption that `birdcage` is good!

PR #944, opened 2023-02-13
Fix Go parser ignoring dependencies

### LINKED ISSUE #942: Golang parser missing dependencies
# Overview
Submitting a `go.sum` file does not always yield all expected dependencies. This appears to be a parsing issue.

# How To Reproduce
1. Acquire a sample [go.sum](https://github.com/sonatype-nexus-community/intentionally-vulnerable-golang-project/blob/master/go.sum) file
2. Run `phylum parse go.sum`
3. See limited results returned

# Expected Behavior
The parser identifies all package dependencies.

cd-work:
Previously the assumption was made that any line in the Go lockfile including `/go.mod` after its version was only included for the hash and would still have its version included without the `/go.mod` in a different line. This assumption was incorrect.

To ensure we don't ignore dependencies that are actually used, we now include all versions with and without the suffix and deduplicate that list after parsing.

This will include a lot more dependencies than what is actually used, but for security purposes over-reporting is much likely more desirable than under-reporting, considering we can't get an accurate list just by parsing the lockfile.

Closes #942.

### DIFF
--- a/lockfile/src/cargo.rs
+++ b/lockfile/src/cargo.rs
@@ -129,6 +129,7 @@ mod tests {
             assert!(pkgs.contains(&expected_pkg));
         }
     }
+
     #[test]
     fn parse_cargo_lock_v3() {
         let pkgs = Cargo.parse(include_str!("../../tests/fixtures/Cargo_v3.lock")).unwrap();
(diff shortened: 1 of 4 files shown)


ejortega (approved):
lgtm

PR #947, opened 2023-02-17
Fix project history endpoint

### LINKED ISSUE #378: `phylum history --project` always shows 0 jobs
After the latest API updates:

Even when a project has jobs, `phylum history --project <project-name>` shows zero jobs:

```
❯ phylum history --project test
Last 0 runs
```

cd-work:
This patch fixes a long-standing issue where querying for project-specific history using `phylum history --project <PROJECT>` would return an empty list. A new `--group` flag was also added to get history for a specific group's project.

The full history endpoint does not give any group information and as such cannot be used to retrieve this information. But projects and groups have a separate endpoint providing a less detailed list of historic jobs which can be used instead.

The new history for projects includes Job ID, Label, and Creation Time.

Closes #378.

### DIFF
--- a/cli/src/api/endpoints.rs
+++ b/cli/src/api/endpoints.rs
@@ -233,14 +233,6 @@ mod test {
     }
 
     #[test]
-    fn get_project_details_is_correct() {
-        assert_eq!(
-            get_project_history(API_URI, "acme/widgets").unwrap().as_str(),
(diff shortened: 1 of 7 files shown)


cd-work:
Here's what the new history looks like:

tmp

andreaphylum (approved):

PR #950, opened 2023-02-22
Add automatic lockfile detection

### LINKED ISSUE #946: Automatically detect multiple lockfiles for analysis
# Overview

When performing a `phylum analyze`, it would be nice if the CLI is able to automatically detect multiple lockfiles recursively.

* This would operate like the GitHub app
* `phylum-ci` could use this feature instead of rolling it's own
  * This would help to complete https://github.com/phylum-dev/phylum-ci/issues/147
* Need to be careful about looking into `node_modules` and/or any local packages that contain lockfiles
  * It might be possible to leverage a `.gitignore` file for the first pass of this
* It should be possible to "override" the automatic detection
  * The GitHub app does this by using the lockfile path(s) in a `.phylum_project` file, if one exists

# Acceptance Criteria
- [x] `phylum analyze` detects and scans multiple lockfiles without having to specify them on the command line or in a `.phylum_project` file
- [x] Lockfiles can be from different ecosystems
- [x] Lockfiles can be found in sub-directories (recursively)
- [x] Documentation is updated

cd-work:
This patch adds a fallback mechanism where the CLI will attempt to find lockfiles by walking the directory tree when neither CLI nor the .phylum_project file specified any lockfile.

To prevent unwanted files from being included, this will automatically respect ignore files like `.gitignore` or `.ignore`. However as a result some desirable files will be excluded like `Cargo.lock` for Rust libraries. However this is already how CI analysis operates anyway, so it should not cause major problems.

The automatic fallback is taken into account for both `phylum parse` and `phylum analyze`.

Closes #946.

---

I was honestly somewhat surprised just how simple this implementation is using BurnSushi's great `ignore` crate. Some testing is appreciated of course but it just worked™ for me.

@maxrake you suggested documentation updates in the original issue, but I'm not entirely sure where we would make them. I think for general workflow at this point we still want to suggest running `phylum init`, so user docs don't really need to be updated?


cli/src/commands/jobs.rs
```diff
@@ -127,9 +127,6 @@ pub async fn handle_submission(api: &mut PhylumApi, matches: &clap::ArgMatches)
         synch = true;
 
         jobs_project = JobsProject::new(api, matches).await?;
-        if jobs_project.lockfiles.is_empty() {
-            return Err(anyhow!("Missing lockfile parameter"));
-        }
```
cd-work, line 132: This code was removed since we already check for this in `config::lockfiles(...)` which is called from `JobsProject::new(...)` and the error is just bubbled up throughout. So it should be unreachable code anyway.

lockfile/src/lib.rs
```diff
@@ -22,6 +23,9 @@ mod parsers;
 mod python;
 mod ruby;
 
+/// Maximum directory depth to recurse for finding lockfiles.
+const MAX_LOCKFILE_DEPTH: usize = 5;
```
cd-work, line 27: I've picked 5 here because I like the number 5. a depth of 10 seems like it would already be kinda deep for a lockfile and 3 seemed somewhat shallow. It probably isn't too big of a deal to go even deeper, I don't think this should be too much of a performance problem, but I'm not sure it's necessary.

maxrake:
> @maxrake you suggested documentation updates in the original issue, but I'm not entirely sure where we would make them. I think for general workflow at this point we still want to suggest running phylum init, so user docs don't really need to be updated?

I think adding a `### Details` section to `doc_templates/phylum_analyze.md` and `doc_templates/phylum_parse.md` would be a good place to add documentation. It would also help to add to the existing `### Examples` section for each of those templates. Without any documentation, users have no way of knowing (short of reading the code):

* That lockfiles will be searched for automatically when not provided in a `.phylum_project` file or as a command option
* The order of resolution for lockfile determination
* That there is a maximum depth for searching recursively...and what that value is
* That there is an ability to ignore lockfile detection by using `.gitignore` or `.ignore` files
  * They may be surprised that a lockfile they expect to be detected is not...until they know this wrinkle in the implementation

cd-work:
@maxrake I added documentation detailing the fallback order we use for lockfiles and stated that ignore files are respected. However I did intentionally leave out the maximum search depth since that seemed more like an implementation detail that shouldn't affect users. I contemplated adding that if the limit is too low, people should contact us, but that seemed a little weird at this location and users are instructed to contact us with any issues in other places anyway.

maxrake (approved):
Code changes look okay to my eye. The behavior was tested locally and it works! One suggestion is to update the output for specifying the path of each detected lockfile:

image

So...a line like:

`Successfully parsed lockfile as type: cargo`

Would become:

`Successfully parsed ./Cargo.lock as type: cargo`

...where maybe the path is even provided as absolute:

`Successfully parsed /Users/maxrake/dev/phylum/localdev/cli/Cargo.lock as type: cargo`

This update would help to understand where lockfiles are found and make it easier to add them to ignore files.

cd-work:
Sounds good, I think paths relative to the working directory would be better, but that would likely be more complicated to implement if that's not what is reported by default.

cd-work:
Turns out relative paths is exactly what is reported by default, so we should be good.

maxrake (approved):
LGTM.

image

PR #954, opened 2023-02-24
ci: pin workflow actions

### LINKED ISSUE #953: Pin workflow actions
## Description

Update all actions within all workflows to pin to a hash instead of a tag and use Dependabot for weekly updates.

## Additional Details

* All GitHub actions should be pinned to an explicit SHA instead of a tag name
* Each pin should include a comment about the version of the action to which it corresponds
* Dependabot will update these comments at the same time that it updates the pin
  * [Reference](https://github.com/dependabot/dependabot-core/issues/4691#issuecomment-1296532496)

## Acceptance Criteria

- [x] All GitHub actions should be pinned to an explicit SHA instead of a tag name
- [x] Documentation is updated

maxrake:
All GitHub actions were pinned to an explicit SHA instead of a tag name/version. Each pin includes a comment about the version it represents, which will be updated automatically by Dependabot at the same time as the pin.

Closes #953

### DIFF
--- a/.github/workflows/cargo-update.yml
+++ b/.github/workflows/cargo-update.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout the repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
 
       - name: Cargo update
         run: cargo update
@@ -23,12 +23,12 @@ jobs:
         run: |
           git config user.name 'Phylum Bot'
           git config user.email 'phylum-bot@users.noreply.github.com'
-          git commit -a -m "build: Bump Cargo.lock dependencies"
+          git commit -a -m "build: Bump `Cargo.lock` dependencies"
           git push --force origin HEAD:auto-cargo-update
 
       - name: Create Pull Request
         if: ${{ steps.commit.outcome == 'success' }}
-        uses: actions/github-script@v6
+        uses: actions/github-script@98814c53be79b1d30f795b907e553d8679345975 # v6.4.0
         with:
           github-token: ${{ secrets.GH_RELEASE_PAT }}
           script: |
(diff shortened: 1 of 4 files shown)


.github/workflows/cargo-update.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout the repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
```
cd-work, line 15: Why comment the version behind the hash? Aren't versions ultimately meaningless when we don't use them anyway?

Will the next dependabot update not just break this, causing pointless additional work having to update the version every time?

Something like @v4 can get updated, but can a tag like @v4.1.0 get updated? Doesn't that always link to one specific version anyway? Or is it fully up to the action creator on which version they want to overwrite.

cd-work (changes requested):

.github/workflows/cargo-update.yml
maxrake: Dependabot will update these version comments at the same time that it updates the pin ([Reference](https://github.com/dependabot/dependabot-core/issues/4691#issuecomment-1296532496)). There is no manual update required for either the hash or the version comment. This is nice because it makes it easier to map the pin to the tagged version without having to look it up. The same thing is done [for `pre-commit` hooks in `phylum-ci`](https://github.com/phylum-dev/phylum-ci/blob/main/.pre-commit-config.yaml) and that has worked out well.

Yes...action creators are responsible for bumping tags. The convention most follow is to provide specific/fixed patch versions (e.g., vX.Y.Z) and floating major and minor versions (e.g., vX and vX.Y). See our own [phylum-dev/phylum-analyze-pr-action tags](https://github.com/phylum-dev/phylum-analyze-pr-action/tags) for an example of this. Currently, these three tags all point to the same hash: v2.1.1, v2.1, and v2
When v2.1.1 was released, the tags for v2.1 and v2 were overwritten to point away from the v2.1.0 release.

cd-work: But v2.1.0 is still the same, correct? If these are just git tags without any immutability assigned to them then that's quite honestly just very poor design.

cd-work (approved):

.github/workflows/cargo-update.yml
maxrake: Yes...v2.1.0 is still the same...but only by convention. Git tags can be updated at any time to point to new/different commit SHAs. This PR is meant to get away from the risk of that design where a tag can shift out from underneath us. The Dependabot method of tracking the version along with the tag's commit is just an added convenience.

PR #957, opened 2023-02-28
Use --dry-run output for poetry extension

### LINKED ISSUE #931: Parse `--dry-run` output for `poetry` extension
Currently the `poetry` extension uses `--lock` to update only the lockfile before analysis, without installing any packages or running any build scripts. This is somewhat problematic since `install` does not have a `--lock` parameter.

A preliminary fix simply removes the lockfile update step from the check for `poetry install` (https://github.com/phylum-dev/cli/pull/930), but this will break down when flags are passed to `poetry install` which modify the lockfile.

One alternative previously attempted was not passing any special flags during install, however this causes build scripts to be executed which can have dangerous side-effects and are very difficult to sandbox.

The most sensible solution I've found so far is parsing the output of `poetry --dry-run`. This option is available for `install`, `update`, and `add` and outputs which packages will be installed when running that command without `--dry-run`. It can be passed multiple times so there's no potential for conflict either.

The primary issue with `--dry-run` is that it requires parsing the STDOUT of `poetry` without any available machine-readable format. But it's still likely the best option.

This would also mean that only the parts of the lockfile which have changed are analyzed. While this is different from other extensions, this seems fine and could be argued to be an advantage too.

cd-work:
This fixes some issues with the poetry ecosystem extension by switching from updating the lockfile and parsing it over to parsing the `--dry-run` flag's output instead.

The primary issue with the `--lock` flag is that the `poetry install` subcommand does not have such a flag, only `poetry add`.

Besides `poetry install`, this patch also removes any necessity to write to the project directory since `--dry-run` is immutable. This allows further lockdown of our sandbox.

Closes #931.

### DIFF
--- a/extensions/poetry/PhylumExt.toml
+++ b/extensions/poetry/PhylumExt.toml
@@ -4,7 +4,7 @@ entry_point = "main.ts"
 
 [permissions]
 run = ["./", "/bin", "/usr/bin", "~/.pyenv", "~/.local/bin/poetry", "~/Library/Application Support/pypoetry", "~/.local/share/pypoetry", "~/.local/pipx"]
-write = ["~/.cache/pypoetry", "~/Library/Caches/pypoetry", "~/.pyenv"]
+write = ["./", "~/.cache/pypoetry", "~/Library/Caches/pypoetry", "~/.pyenv"]
 read = ["./", "~/.cache/pypoetry", "~/Library/Caches/pypoetry", "~/.pyenv", "~/Library/Preferences/pypoetry", "~/.config/pypoetry", "/etc/passwd"]
 net = true
 unsandboxed_run = ["poetry"]
--- a/extensions/poetry/PhylumExt.toml
+++ b/extensions/poetry/PhylumExt.toml
@@ -4,7 +4,7 @@ entry_point = "main.ts"
 
 [permissions]
 run = ["./", "/bin", "/usr/bin", "~/.pyenv", "~/.local/bin/poetry", "~/Library/Application Support/pypoetry", "~/.local/share/pypoetry"]
-write = ["./", "~/.cache/pypoetry", "~/Library/Caches/pypoetry", "~/.pyenv"]
+write = ["~/.cache/pypoetry", "~/Library/Caches/pypoetry", "~/.pyenv"]
 read = ["./", "~/.cache/pypoetry", "~/Library/Caches/pypoetry", "~/.pyenv", "~/Library/Preferences/pypoetry", "~/.config/pypoetry", "/etc/passwd"]
 net = true
 unsandboxed_run = ["poetry"]
--- a/extensions/poetry/PhylumExt.toml
+++ b/extensions/poetry/PhylumExt.toml
@@ -3,7 +3,7 @@ description = "Poetry package manager hooks"
 entry_point = "main.ts"
 
 [permissions]
-run = ["./", "/bin", "/usr/bin", "~/.pyenv", "~/.local/bin/poetry", "~/Library/Application Support/pypoetry", "~/.local/share/pypoetry"]
+run = ["./", "/bin", "/usr/bin", "~/.pyenv", "~/.local/bin/poetry", "~/Library/Application Support/pypoetry", "~/.local/share/pypoetry", "~/.local/pipx"]
 write = ["~/.cache/pypoetry", "~/Library/Caches/pypoetry", "~/.pyenv"]
 read = ["./", "~/.cache/pypoetry", "~/Library/Caches/pypoetry", "~/.pyenv", "~/Library/Preferences/pypoetry", "~/.config/pypoetry", "/etc/passwd"]
 net = true
--- a/extensions/poetry/main.ts
+++ b/extensions/poetry/main.ts
@@ -124,6 +124,7 @@ async function poetryCheckDryRun(
         "~/.local/pipx",
       ],
       write: [
+        "./",
         "~/.cache/pypoetry",
         "~/Library/Caches/pypoetry",
         "~/.pyenv",
@@ -155,22 +156,29 @@ async function poetryCheckDryRun(
     const installing_text = "• Installing ";
     const installing_index = line.indexOf(installing_text);
 
-    // Filter lines unrelated to new packages.
-    if (installing_index === -1) {
+    const updating_text = "• Updating ";
+    const updating_index = line.indexOf(updating_text);
+
+    // Strip text before package name.
+    let pkg;
+    if (installing_index !== -1) {
+      pkg = line.substring(installing_index + installing_text.length);
+    } else if (updating_index !== -1) {
+      pkg = line.substring(updating_index + updating_text.length);
+    } else {
+      // Filter lines unrelated to new packages.
       continue;
     }
 
-    // Extract name and version.
-    const pkg = line.substring(installing_index + installing_text.length);
-    const pkg_split = pkg.split(" ");
-    const name = pkg_split[0];
-    let version = pkg_split[1];
-
     // Strip suffix explaining why package install was skipped.
-    const colon_index = version.indexOf(":");
+    const colon_index = pkg.indexOf(":");
     if (colon_index !== -1) {
-      version = version.substring(0, colon_index);
+      pkg = pkg.substring(0, colon_index);
     }
+
+    // Extract name and version.
+    const name = pkg.substring(0, pkg.indexOf(" "));
+    let version = pkg.substring(name.length + 1);
 
     // Ensure the line is in the correct format.
     if (
@@ -184,6 +192,19 @@ async function poetryCheckDryRun(
 
     // Remove parenthesis from version.
     version = version.substring(1, version.length - 1);
+
+    // Extract target version from update.
+    if (updating_index !== -1) {
+      const version_split = version.split(" -> ");
+      if (version_split.length !== 2) {
+        console.error(
+          `[${red("phylum")}] Invalid version update: "${version}".\n`,
+        );
+        Deno.exit(124);
+      }
+
+      version = version_split[1];
+    }
 
     packages.push({ name, version });
   }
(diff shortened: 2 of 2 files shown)


maxrake (changes requested):
This is just the code review. A few suggestions and questions are provided, with varying levels of certainty given my experience with TypeScript (almost none). The extension will be tested locally next (after lunch) and any additional feedback will be provided then.

extensions/poetry/main.ts
```diff
@@ -146,74 +103,98 @@ Deno.exit(status.code);
...
+        "~/.pyenv",
+      ],
+      read: [
+        "./",
+        "~/.cache/pypoetry",
+        "~/Library/Caches/pypoetry",
+        "~/.pyenv",
+        "~/Library/Preferences/pypoetry",
+        "~/.config/pypoetry",
+        "/etc/passwd",
```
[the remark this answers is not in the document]
cd-work: Putting a comment here would not help users though? And developers should be aware of why an application might want to read `/etc/passwd`. So I'm not sure that's particularly helpful?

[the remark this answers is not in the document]
maxrake: Sure...just a suggestion...marking as resolved.

extensions/poetry/main.ts
```diff
@@ -146,74 +103,98 @@ Deno.exit(status.code);
...
+  });
+
+  // Ensure dry-run update was successful.
+  if (!result.success || result.stdout.length == 0) {
+    console.error(`[${red("phylum")}] Failed to determine new packages.\n`);
+    return status.code ?? 255;
+  }
+
+  // Parse dry-run output to look for new packages.
+  const packages = [];
```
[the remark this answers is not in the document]
cd-work: As far as I know, const is for variables which are never reassigned. Pushing to an array is not a reassignment, so defining it as `const` is correct (and our linter would warn otherwise I believe).

extensions/poetry/main.ts
```diff
@@ -146,74 +103,103 @@ Deno.exit(status.code);
...
+  const lines = result.stdout.split("\n");
+  for (const line of lines) {
+    const installing_text = "Installing ";
+    const installing_index = line.indexOf(installing_text);
+
+    // Filter lines unrelated to new packages.
+    if (installing_index === -1) {
+      continue;
```
maxrake, line 159: This method of finding and filtering lines based on the inclusion of `Installing ` is prone to false positives. That is, _any_ line with `Installing ` in it will pass through these checks and then may fail later when "Ensuring what we parsed is in a sensible format."

Indeed, testing this code resulted in one such error:

`[phylum] Invalid poetry output: Installing dependencies from lock file.`

This line is from the output of trying to run this command (shown executed without the extension):

```
../phylum-ci  6  17 on  main [?] is 📦 v0.24.1 via 🐍 v3.11.2 took 2s
❌ 125 ❯ poetry install --dry-run --sync --with test,qa
Installing dependencies from lock file

Package operations: 0 installs, 0 updates, 0 removals, 33 skipped

  • Installing attrs (22.2.0): Skipped for the following reason: Already installed
  • Installing backports-cached-property (1.0.2): Skipped for the following reason: Already installed
  • Installing cachetools (5.3.0): Skipped for the following reason: Already installed
  • Installing certifi (2022.12.7): Skipped for the following reason: Already installed
  • Installing cffi (1.15.1): Skipped for the following reason: Already installed
  • Installing chardet (5.1.0): Skipped for the following reason: Already installed
  • Installing charset-normalizer (3.0.1): Skipped for the following reason: Already installed
  • Installing colorama (0.4.6): Skipped for the following reason: Already installed
  • Installing commonmark (0.9.1): Skipped for the following reason: Already installed
  • Installing connect-markdown-renderer (2.0.1): Skipped for the following reason: Already installed
  • Installing cryptography (39.0.1): Skipped for the following reason: Already installed
  • Installing distlib (0.3.6): Skipped for the following reason: Already installed
  • Installing filelock (3.9.0): Skipped for the following reason: Already installed
  • Installing idna (3.4): Skipped for the following reason: Already installed
  • Installing iniconfig (2.0.0): Skipped for the following reason: Already installed
  • Installing packaging (23.0): Skipped for the following reason: Already installed
  • Installing mdurl (0.1.2): Skipped for the following reason: Already installed
  • Installing pluggy (1.0.0): Skipped for the following reason: Already installed
  • Installing pytest (7.2.1): Skipped for the following reason: Already installed
  • Installing pycparser (2.21): Skipped for the following reason: Already installed
  • Installing pyproject-api (1.5.0): Skipped for the following reason: Already installed
  • Installing tomli (2.0.1): Skipped for the following reason: Already installed
  • Installing markdown-it-py (2.2.0): Skipped for the following reason: Already installed
  • Installing rich (12.6.0): Skipped for the following reason: Already installed
  • Installing platformdirs (3.0.0): Skipped for the following reason: Already installed
  • Installing requests (2.28.2): Skipped for the following reason: Already installed
  • Installing tox (4.4.6): Skipped for the following reason: Already installed
  • Installing types-urllib3 (1.26.25.8): Skipped for the following reason: Already installed
  • Installing ruamel-yaml (0.17.21): Skipped for the following reason: Already installed
  • Installing urllib3 (1.26.14): Skipped for the following reason: Already installed
  • Installing virtualenv (20.19.0): Skipped for the following reason: Already installed
  • Installing types-requests (2.28.11.15): Skipped for the following reason: Already installed
  • Installing pygments (2.14.0): Skipped for the following reason: Already installed

Installing the current project: phylum (0.24.1)
```

Further, it is predicted that the same error would hit for the last line of the output:

`Installing the current project: phylum (0.24.1)`

extensions/poetry/main.ts
```diff
@@ -146,74 +103,103 @@ Deno.exit(status.code);
...
+    exceptions: {
+      run: [
+        "./",
+        "/bin",
+        "/usr/bin",
+        "~/.pyenv",
+        "~/.local/bin/poetry",
+        "~/Library/Application Support/pypoetry",
+        "~/.local/share/pypoetry",
+      ],
```
maxrake, line 124:
```suggestion
        "~/.local/share/pypoetry",
        "~/.local/pipx/venvs/poetry",
      ],
```
This additional entry is needed for `poetry` instances that are [installed with `pipx`](https://python-poetry.org/docs/#installing-with-pipx). The same entry is needed in the `PhylumExt.toml` manifest file. At least, this entry is known to work on a macOS system with `pipx`; it may be different for other environments.

maxrake (changes requested):
There are a few requested changes and a few questions.

extensions/poetry/PhylumExt.toml
```diff
@@ -3,8 +3,8 @@ description = "Poetry package manager hooks"
 entry_point = "main.ts"
 
 [permissions]
-run = ["./", "/bin", "/usr/bin", "~/.pyenv", "~/.local/bin/poetry", "~/Library/Application Support/pypoetry", "~/.local/share/pypoetry"]
-write = ["./", "~/.cache/pypoetry", "~/Library/Caches/pypoetry", "~/.pyenv"]
+run = ["./", "/bin", "/usr/bin", "~/.pyenv", "~/.local/bin/poetry", "~/Library/Application Support/pypoetry", "~/.local/share/pypoetry", "~/.local/pipx"]
+write = ["~/.cache/pypoetry", "~/Library/Caches/pypoetry", "~/.pyenv"]
```
maxrake, line 7:
```suggestion
write = ["./", "~/.cache/pypoetry", "~/Library/Caches/pypoetry", "~/.pyenv"]
```
It turns out the current directory _is_ needed...for the case when no Python virtual environment, as managed by `poetry`, already exists. At least, it is needed when the [`virtualenvs.in-project` setting](https://python-poetry.org/docs/configuration#virtualenvsin-project) is `true`...which is _very common_.

extensions/poetry/main.ts
```diff
@@ -146,74 +103,103 @@ Deno.exit(status.code);
...
+        "/bin",
+        "/usr/bin",
+        "~/.pyenv",
+        "~/.local/bin/poetry",
+        "~/Library/Application Support/pypoetry",
+        "~/.local/share/pypoetry",
+        "~/.local/pipx",
+      ],
+      write: [
+        "~/.cache/pypoetry",
```
maxrake, line 127:
```suggestion
      write: [
        "./",
        "~/.cache/pypoetry",
```
It turns out the current directory _is_ needed...for the case when no Python virtual environment, as managed by `poetry`, already exists. At least, it is needed when the [`virtualenvs.in-project` setting](https://python-poetry.org/docs/configuration#virtualenvsin-project) is `true`...which is _very common_.

maxrake (changes requested):
Updating an existing package by "adding" a new version does not work. The package is not included in the submission for analysis. For example, this "update" of the `cryptography` package was expected to fail since it has a known vulnerability that brings that risk domain down to a score of 20. This output is from running _outside_ of the extension...so the format of the `--dry-run` can be viewed:

```
../phylum-ci  6  17 on  main [!?] is 📦 v0.24.1 via 🐍 v3.11.2 took 2s
❯ poetry add --dry-run "cryptography==38.0.4"

Updating dependencies
Resolving dependencies... (0.2s)

Package operations: 0 installs, 1 update, 0 removals, 16 skipped

  • Installing backports-cached-property (1.0.2): Skipped for the following reason: Already installed
  • Installing certifi (2022.12.7): Skipped for the following reason: Already installed
  • Installing cffi (1.15.1): Skipped for the following reason: Already installed
  • Installing charset-normalizer (3.0.1): Skipped for the following reason: Already installed
  • Installing commonmark (0.9.1): Skipped for the following reason: Already installed
  • Installing connect-markdown-renderer (2.0.1): Skipped for the following reason: Already installed
  • Updating cryptography (38.0.3 -> 38.0.4)
  • Installing idna (3.4): Skipped for the following reason: Already installed
  • Installing markdown-it-py (2.2.0): Skipped for the following reason: Already installed
  • Installing mdurl (0.1.2): Skipped for the following reason: Already installed
  • Installing packaging (23.0): Skipped for the following reason: Already installed
  • Installing pycparser (2.21): Skipped for the following reason: Already installed
  • Installing pygments (2.14.0): Skipped for the following reason: Already installed
  • Installing requests (2.28.2): Skipped for the following reason: Already installed
  • Installing rich (12.6.0): Skipped for the following reason: Already installed
  • Installing ruamel-yaml (0.17.21): Skipped for the following reason: Already installed
  • Installing urllib3 (1.26.14): Skipped for the following reason: Already installed
```

maxrake (approved):
The changes were tested locally and all appear to work. LGTM.

PR #959, opened 2023-03-03
Switch default subcommand from `list` to `help`

### LINKED ISSUE #555: Confusing results of argument ordering
Our use of subcommands means that arguments are parsed differently at different levels. And often, arguments at a higher level are completely ignored, which can lead to confusing situations.

## Example 1: Listing projects
The results of `phylum project list --group foo` are different than `phylum project --group foo list`:

```
❯ phylum project list --group x
Project Name                 Project ID
foo                          b0f27770-ff5e-46f3-a53d-afac51e52308

❯ phylum project --group x list
Project Name                 Project ID
bar                          71ad1d65-ebe9-4c9f-a435-def73cf36fcc
```

## Example 2: Creating a project

First, I list the projects in a group:
```
❯ phylum project --group x
Project Name                 Project ID
foo                          b0f27770-ff5e-46f3-a53d-afac51e52308
```

Then I attempt to add a project. For simplicity, I just hit the up arrow and type `create bar` onto the end of the last command:
```
❯ phylum project --group x create bar
✅ Successfully created new project, bar
```

Somewhat confusingly, this did not create a new project in that group:
```
❯ phylum project --group x           
Project Name                 Project ID
foo                          b0f27770-ff5e-46f3-a53d-afac51e52308
```

### LINKED ISSUE #812: Inconsistent JSON vs table output
The `phylum project list` command has these cases:
* `phylum project list --json` outputs JSON
* `phylum project --json list` outputs JSON

The `phylum group list` command has these cases:
* `phylum group list --json` outputs JSON
* `phylum group --json list` outputs a table

Additionally, both of these commands are accessible by omitting the `list` arguments:
* `phylum project --json` outputs JSON
* `phylum group --json` outputs JSON

There are many options for fixing this inconsistency:
1. Always check for `--json` anywhere in the chain
2. Make `--json` conflict with subcommands (i.e., [`Command::args_conflicts_with_subcommands`](https://docs.rs/clap/latest/clap/builder/struct.Command.html#method.args_conflicts_with_subcommands)
3. Remove the `list` subcommands

And there may be others that I have not yet considered. Further discussion is needed.

_Originally posted by @kylewillmon in https://github.com/phylum-dev/cli/pull/809#discussion_r1024230422_

cd-work:
Currently commands like `phylum project` default to `phylum project
list` when no subcommand is specified. This means that all flags present
in `phylum project list` also need to be available on `phylum project`,
which makes it confusing to know where the user is supposed to specify
them.

To reduce the potential for confusion, the command `phylum project` is
now just an alias for `phylum project help`. Having fewer ways to do the
same thing will hopefully make it more clear what the correct approach
is without having to rely on extensive documentation and error messages.

While it would be possible to just print an error listing the available
subcommands, this patch chooses to print the help instead since it's a
less aggressive and more helpful output, especially considering the user
likely made this mistake because he's not aware of how the command
should be used in general.

Since the `phylum auth` command was relying on a custom implementation
for essentially the same functionality, it has been migrated to the
built-in clap flag too.

Closes #812.
Closes #555.


maxrake (approved):
The code was reviewed at my level of understanding and nothing stood out. The changes were tested locally and work as intended. These fixes are very helpful in simplifying things!

PR #963, opened 2023-03-09
Add SPDX SBOM parser

ejortega:
Added a package type of `SBOM`. Looks at the `ExternalRefs` for each package to identify the purl and extract the ecosystem and package namespace and name.

As discussed, the parsers expect a single type, so I had to work around that. Which I'm not sure was the best way but seemed minimal. Looking for feedback on implementation.

## Checklist


cli/Cargo.toml
```diff
@@ -37,7 +37,7 @@ maplit = "1.0.2"
 open = "3.0.1"
 phylum_lockfile = { path = "../lockfile" }
 phylum_project = { path = "../phylum_project" }
-phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "development" }
+phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "add-sbom" }
```
ejortega, line 40: Tagging to update back to development branch.

lockfile/Cargo.toml
```diff
@@ -8,12 +8,13 @@ rust-version = "1.63.0"
 
 [dependencies]
 anyhow = "1.0.44"
+ignore = "0.4.20"
 log = "0.4.6"
 nom = "7.1.1"
-phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "development" }
+packageurl = "0.3.0"
+phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "add-sbom" }
```
ejortega, line 15: Tagging to update back to development branch.

phylum_project/Cargo.toml
```diff
@@ -9,7 +9,7 @@ edition = "2021"
 rust-version = "1.64.0"
 
 [dependencies]
-phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "development" }
+phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "add-sbom" }
```
ejortega, line 12: Tagging to update back to development branch.

lockfile/src/lib.rs
```diff
@@ -161,7 +167,7 @@ pub struct Package {
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
 pub enum PackageVersion {
     /// Version from the ecosystem's first-party registry.
-    FirstParty(String),
+    FirstParty(FirstPartyVersion),
```
kylewillmon, line 170: Adding the `PackageType` here is quite a large design decision. We should meet to discuss this. Meeting invite coming soon.

Cc: @cd-work

lockfile/src/sbom.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        Ok(Package { name, version, package_type })
+    }
+}
+
+pub struct Sbom;
+
+impl Parse for Sbom {
+    fn parse(&self, data: &str) -> anyhow::Result<Vec<Package>> {
+        let mut lock: Spdx = serde_json::from_str(data).unwrap_or(serde_yaml::from_str(data)?);
```
kylewillmon, line 123: To be clear, this is equivalent to this:

```rust
let yaml_lock: Spdx = serde_yaml::from_str(data)?
let mut lock: Spdx = serde_json::from_str(data).unwrap_or(yaml_lock);
```

Which means that:

1. Any file which isn't valid YAML is immediately rejected
2. Every file is sent through both the `serde_yaml` and `serde_json` parser

This technically works because JSON is a subset of YAML. But I don't think this is what was desired.

I think you wanted one of the following options:

## Parse as JSON first
```suggestion
        let mut lock: Spdx = serde_json::from_str(data).or_else(|| serde_yaml::from_str(data))?;
```

## Always parse as YAML
```suggestion
        // This also parses JSON files because YAML is insane
        let mut lock: Spdx = serde_yaml::from_str(data)?;
```

lockfile/src/sbom.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        let packages = lock
+            .packages
+            .drain(..)
+            .filter_map(|package_info| Package::try_from(&package_info).ok())
+            .collect::<Vec<Package>>();
```
kylewillmon, line 129: Silently ignoring packages has led to issues in the past (and ultimately led to #882).

Errors in `Package::try_from` should not be ignored.

```suggestion
        let packages = Vec::new();
        for package_info in lock.packages {
            packages.push(Package::try_from(&package_info)?);
        }
```

kylewillmon (changes requested):

cd-work:
Removing my review until @kylewillmon's review is done. If you still want my feedback after that, just add me back to the reviewers.

lockfile/src/sbom.rs
ejortega: Ah yes, I meant the first option. Didn't realize it was being eagerly evaluated. Thanks!

lockfile/src/sbom.rs
ejortega: There are some issues with allowing all packages.

1. Package `versionInfo` is optional. If it's provided in the external refs, then that's used.
2. There isn't a good way to determine which ecosystem the package comes from without a `PURL`. Although I can add some logic to look at the `downloadLocation`, which is always present but can be `NOASSERTION`, and compare URLs with first party registries. Without the `PURL` and maybe `downloadLocation`, there's no good way to determine the ecosystem.
3. Sboms will contain ecosystems we currently do not support. This will prevent anyone from submitting an sbom and analyzing packages from ecosystems we do support.

I built a container targeting all the ecosystems we currently support and scanned it with [syft](https://github.com/anchore/syft) to generate some of the test files. Guaranteed we'll be ingesting sboms generated by syft.


Ex1: This has a `PURL` so we can determine it's maven package but there's no version info.
```
  {
   "name": "ca-certificates-java",
   "SPDXID": "SPDXRef-Package-java-archive-ca-certificates-java-300cbdd37d289bad",
   "downloadLocation": "NOASSERTION",
   "checksums": [
    {
     "algorithm": "SHA1",
     "checksumValue": "0b1d4f423b4ec416cb14026a75c184bdb06c6914"
    }
   ],
   "sourceInfo": "acquired package info from installed java archive: /usr/share/ca-certificates-java/ca-certificates-java.jar",
   "licenseConcluded": "NONE",
   "licenseDeclared": "NONE",
   "copyrightText": "NOASSERTION",
   "externalRefs": [
    {
     "referenceCategory": "PACKAGE-MANAGER",
     "referenceType": "purl",
     "referenceLocator": "pkg:maven/org.debian.security.UpdateCertificates/ca-certificates-java"
    }
   ]
  },
  ```

Ex2: No `PURL` but there is a `downloadLocation`. It's also something we don't support at the moment.
```
- SPDXID: "SPDXRef-Package"
  annotations:
  - annotationDate: "2011-01-29T18:30:22Z"
    annotationType: "OTHER"
    annotator: "Person: Package Commenter"
    comment: "Package level annotation"
  attributionTexts:
  - "The GNU C Library is free software.  See the file COPYING.LIB for copying conditions,\
    \ and LICENSES for notices about a few contributions that require these additional\
    \ notices to be distributed.  License copyright years may be listed using range\
    \ notation, e.g., 1996-2015, indicating that every year in the range, inclusive,\
    \ is a copyrightable year that would otherwise be listed individually."
  builtDate: "2011-01-29T18:30:22Z"
  checksums:
  - algorithm: "MD5"
    checksumValue: "624c1abb3664f4b35547e7c73864ad24"
  - algorithm: "SHA1"
    checksumValue: "85ed0817af83a24ad8da68c2b5094de69833983c"
  - algorithm: "SHA256"
    checksumValue: "11b6d3ee554eedf79299905a98f9b9a04e498210b59f15094c916c91d150efcd"
  - algorithm: "BLAKE2b-384"
    checksumValue: "aaabd89c926ab525c242e6621f2f5fa73aa4afe3d9e24aed727faaadd6af38b620bdb623dd2b4788b1c8086984af8706"
  copyrightText: "Copyright 2008-2010 John Smith"
  description: "The GNU C Library defines functions that are specified by the ISO\
    \ C standard, as well as additional features specific to POSIX and other derivatives\
    \ of the Unix operating system, and extensions specific to GNU systems."
  downloadLocation: "http://ftp.gnu.org/gnu/glibc/glibc-ports-2.15.tar.gz"
  externalRefs:
  - referenceCategory: "SECURITY"
    referenceLocator: "cpe:2.3:a:pivotal_software:spring_framework:4.1.0:*:*:*:*:*:*:*"
    referenceType: "cpe23Type"
  - comment: "This is the external ref for Acme"
    referenceCategory: "OTHER"
    referenceLocator: "acmecorp/acmenator/4.1.3-alpha"
    referenceType: "http://spdx.org/spdxdocs/spdx-example-444504E0-4F89-41D3-9A0C-0305E82C3301#LocationRef-acmeforge"
  filesAnalyzed: true
  homepage: "http://ftp.gnu.org/gnu/glibc"
  licenseComments: "The license for this project changed with the release of version\
    \ x.y.  The version of the project included here post-dates the license change."
  licenseConcluded: "(LGPL-2.0-only OR LicenseRef-3)"
  licenseDeclared: "(LGPL-2.0-only AND LicenseRef-3)"
  licenseInfoFromFiles:
  - "GPL-2.0-only"
  - "LicenseRef-2"
  - "LicenseRef-1"
  name: "glibc"
  originator: "Organization: ExampleCodeInspect (contact@example.com)"
  packageFileName: "glibc-2.11.1.tar.gz"
  packageVerificationCode:
    packageVerificationCodeExcludedFiles:
    - "./package.spdx"
    packageVerificationCodeValue: "d6a770ba38583ed4bb4525bd96e50461655d2758"
  primaryPackagePurpose: "SOURCE"
  hasFiles:
  - "SPDXRef-Specification"
  - "SPDXRef-Specification"
  - "SPDXRef-CommonsLangSrc"
  - "SPDXRef-Specification"
  - "SPDXRef-CommonsLangSrc"
  - "SPDXRef-JenaLib"
  - "SPDXRef-Specification"
  - "SPDXRef-CommonsLangSrc"
  - "SPDXRef-JenaLib"
  - "SPDXRef-DoapSource"
  - "SPDXRef-Specification"
  - "SPDXRef-CommonsLangSrc"
  - "SPDXRef-JenaLib"
  - "SPDXRef-DoapSource"
  releaseDate: "2012-01-29T18:30:22Z"
  sourceInfo: "uses glibc-2_11-branch from git://sourceware.org/git/glibc.git."
  summary: "GNU C library."
  supplier: "Person: Jane Doe (jane.doe@example.com)"
  validUntilDate: "2014-01-29T18:30:22Z"
  versionInfo: "2.11.1"
  ```

I guess I'm looking for direction on what should we do with regards to missing versions, unable to determine ecosystems (do we add an "unknown"), and for ecosystems we don't support.

kylewillmon: I'll take the easy case first.

> * Sboms will contain ecosystems we currently do not support. This will prevent anyone from submitting an sbom and analyzing packages from ecosystems we do support.

This should be a **very loud** warning. Users need to be aware that we do not yet support these ecosystems.

> * Package `versionInfo` is optional. If it's provided in the external refs, then that's used.

If there is no indication of package version, I think it's fair to reject the SBOM. We (currently) do not even support version ranges. And **no** version at all is even worse. This should probably be a hard error.

> * There isn't a good way to determine which ecosystem the package comes from without a `PURL`. Although I can add some logic to look at the `downloadLocation`, which is always present but can be `NOASSERTION`, and compare URLs with first party registries. Without the `PURL` and maybe `downloadLocation`, there's no good way to determine the ecosystem.

This is a bit more of a gray area... But I also think this should be a hard error for now. If PURLs are required for us to parse an SBOM, then we should reject SBOMs that don't have PURLs.

I know it seems harsh to reject entire files for errors like these, but it's better than giving the green light to a file that we barely understand. Our SBOM support will evolve and improve with time.

louislang: The thing about SBOMs is that _most_ of the stuff in them is optional. It's kind of a useless file imo. So I think loud warnings and hard errors are entirely appropriate here.

lockfile/src/lib.rs
```diff
@@ -48,6 +50,7 @@ pub enum LockfileFormat {
     Msbuild,
     Go,
     Cargo,
+    Sbom,
```
kylewillmon, line 53: After looking into this, I think this should be `Spdx` and not `Sbom`.

If I understand the terms correctly, SPDX is the file format that is being used here. That's why the files are named with `.spdx.json` (or `.yaml` or `.yml`) extensions. Our parser then uses the SBOM contained in that file to list packages.

But, critically, there are other file formats that may contain the SBOM (i.e., CycloneDX files). And if we decide to support those files in the future, they would use a separate parser.

I know this comment comes a bit late in the process.... But I'd like to avoid another `msbuild`/`nuget` situation. So I think we need to deal with this now.

lockfile/src/sbom.rs
```diff
@@ -0,0 +1,355 @@
...
+    } else if url.starts_with("https://repo1.maven.org") {
+        Ok(PackageType::Maven)
+    } else if url.starts_with("https://api.nuget.org") {
+        Ok(PackageType::Nuget)
+    } else if url.starts_with("https://proxy.golang.org") {
+        Ok(PackageType::Golang)
+    } else if url.starts_with("https://crates.io") {
+        Ok(PackageType::Cargo)
+    } else {
+        Err(())
```
kylewillmon, line 87: This function could benefit from `anyhow::Error` as a return type.

```suggestion
        Err(anyhow!("Unknown download URL: {:?}", url))
```

lockfile/src/sbom.rs
```diff
@@ -0,0 +1,355 @@
...
+            version_info: None,
+            download_location: "NOASSERTION".to_string(),
+            external_refs: Vec::new(),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
+pub enum SbomError {
+    UnknownEcossytem(String),
```
kylewillmon, line 43: This variant name is a typo
```suggestion
    UnknownEcosystem(String),
```

But as mentioned in other comments, I think the `SbomError` type should be removed.

lockfile/src/sbom.rs
```diff
@@ -0,0 +1,355 @@
...
+pub struct ExternalRefs {
+    pub reference_category: ReferenceCategory,
+    pub reference_locator: String,
+    pub reference_type: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Clone)]
+#[serde(rename_all = "SCREAMING-KEBAB-CASE")]
+pub enum ReferenceCategory {
+    Other,
```
kylewillmon, line 59: Should the `Other` variant use `#[serde(other)]` so that it can deserialize unknown values?

```suggestion
    #[serde(other)]
    Other,
```

Another option is to explicitly add an `Unknown` variant for that.

```suggestion
    Other,
    #[serde(other)]
    Unknown,
```

kylewillmon (changes requested):

lockfile/src/lib.rs
ejortega: Sounds good, I'll update. I thought we'd handle all formats, starting with SPDX and others to follow, under `Sbom`, and add any additional [parsers](https://github.com/phylum-dev/cli/pull/978) that might be needed.

lockfile/src/sbom.rs
```diff
@@ -0,0 +1,355 @@
...
+            .ok_or(SbomError::Purl)?;
+
+        let purl = PackageUrl::from_str(pkg_url).map_err(|_| SbomError::Purl)?;
+        let purl_ty = purl.ty();
+        let package_type = PackageType::from_str(purl.ty())
+            .or_else(|_| type_from_url(&pkg_info.download_location))
+            .map_err(|_| SbomError::UnknownEcossytem(purl_ty.into()))?;
+        let name = match (package_type, purl.namespace()) {
+            (PackageType::Maven, Some(ns)) => format!("{}:{}", ns, purl.name()),
+            (_, Some(ns)) => format!("{}/{}", ns, purl.name()),
```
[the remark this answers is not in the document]
ejortega: You're right, this should be explicit.

lockfile/src/sbom.rs
ejortega: Reverted to using anyhow

kylewillmon (changes requested):
One final note, but this is looking really good now. Thanks for putting up with my review 😂

lockfile/src/spdx.rs
```diff
@@ -105,29 +84,24 @@ impl TryFrom<&PackageInformation> for Package {
...
-            .ok_or(SbomError::Purl)?;
+            .ok_or(anyhow!("Missing PURL for {}", pkg_info.name))?;
 
-        let purl = PackageUrl::from_str(pkg_url).map_err(|_| SbomError::Purl)?;
+        let purl = PackageUrl::from_str(pkg_url)?;
         let purl_ty = purl.ty();
         let package_type = PackageType::from_str(purl.ty())
             .or_else(|_| type_from_url(&pkg_info.download_location))
-            .map_err(|_| SbomError::UnknownEcossytem(purl_ty.into()))?;
+            .map_err(|_| anyhow!("Unsupported ecosystem {purl_ty}"))?;
```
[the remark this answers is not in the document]
ejortega: Ah yes! That's what I was attempting to do via the `SbomError`. This is much cleaner. Thanks!

kylewillmon (approved):

PR #966, opened 2023-03-09
Remove --force option from phylum analyze

kylewillmon:
Remove `--force` option from `phylum analyze` and `phylum batch`. This option was previously used to set the `is_user` value when submitting a job, but that value has no effect.

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -297,7 +297,6 @@ impl PhylumApi {
         &self,
         req_type: &PackageType,
         package_list: &[PackageDescriptor],
-        is_user: bool,
         project: ProjectId,
         label: Option<String>,
         group_name: Option<String>,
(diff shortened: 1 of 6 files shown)


cd-work (dismissed):

cli/src/api/mod.rs
```diff
@@ -297,15 +297,14 @@ impl PhylumApi {
...
-        is_user: bool,
         project: ProjectId,
         label: Option<String>,
         group_name: Option<String>,
     ) -> Result<JobId> {
         let req = SubmitPackageRequest {
             package_type: req_type.to_owned(),
             packages: package_list.to_vec(),
-            is_user,
+            is_user: true,
```
kylewillmon, line 307: If we wait for phylum-dev/phylum-types#68, we can remove this line completely.

cd-work (approved):

maxrake (approved):
LGTM

PR #967, opened 2023-03-10
cargo clippy --fix

kylewillmon:
This was run with rustc 1.68.0

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -4,7 +4,7 @@ version = "4.7.0"
 authors = ["Phylum, Inc. <engineering@phylum.io>"]
 license = "GPL-3.0-or-later"
 edition = "2021"
-rust-version = "1.64.0"
+rust-version = "1.65.0"
 autotests = false
 
 [[test]]
--- a/cli/tests/common.rs
+++ b/cli/tests/common.rs
@@ -16,17 +16,15 @@ use tempfile::TempDir;
 pub const API_URL: &str = "https://api.staging.phylum.io";
 const PROJECT_NAME: &str = "integration-tests";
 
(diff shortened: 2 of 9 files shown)


maxrake (dismissed):
Automated changes LGTM.

cli/tests/common.rs
```diff
@@ -16,18 +16,14 @@ use tempfile::TempDir;
 pub const API_URL: &str = "https://api.staging.phylum.io";
 const PROJECT_NAME: &str = "integration-tests";
 
+#[derive(Default)]
 enum Cwd {
     Path(PathBuf),
     TempDir,
+    #[default]
```
cd-work, line 23: Isn't this a 1.65.0 feature? Our CLI is 1.64.0+ min version.

Though I guess that's fine for tests.

kylewillmon: Good catch. Are you opposed to bumping the MSRV to 1.65?

cd-work: Not at all.

cli/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "4.7.0"
 authors = ["Phylum, Inc. <engineering@phylum.io>"]
 license = "GPL-3.0-or-later"
 edition = "2021"
-rust-version = "1.64.0"
+rust-version = "1.65.0"
```
maxrake, line 7: Is it okay to have different MSRV values throughout the repo? With this change, there will be three different values for `rust-version` between the four (4) `Cargo.toml` files in the `cli` repository.

maxrake (approved):

cd-work (approved):

PR #968, opened 2023-03-10
Remove `package_type()` from `Parse` trait

kylewillmon:
The `Parse` trait in the `lockfile` crate previously required a `package_type()` function that specified the package type for all packages returned by that parser. Going forward, we want to support lockfiles that contains multiple package types (e.g., SBOM).

This patch moves the `package_type` into the `Package` struct returned by the parser in order to support these future parsers.

### DIFF
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -361,8 +361,12 @@ async fn parse_lockfile(
     // Attempt to parse as requested lockfile type.
     let parsed = parse::parse_lockfile(lockfile, lockfile_type.as_deref())?;
 
+    // PackageLock should report LockfileFormat and not PackageType. This is just a
+    // placeholder until that can be updated.
+    let package_type = parsed.packages.first().map(|p| p.package_type).unwrap_or(PackageType::Npm);
+
     Ok(PackageLock {
-        package_type: parsed.package_type,
+        package_type,
         packages: parsed.packages.into_iter().map(PackageSpecifier::from).collect(),
     })
 }
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -16,6 +16,7 @@ use std::str::FromStr;
 use anyhow::{anyhow, Error, Result};
 use deno_runtime::deno_core::{op, OpDecl, OpState};
 use deno_runtime::permissions::Permissions;
+use phylum_lockfile::LockfileFormat;
 use phylum_project::ProjectConfig;
 use phylum_types::types::auth::{AccessToken, RefreshToken};
 use phylum_types::types::common::{JobId, ProjectId};
@@ -42,11 +43,16 @@ use crate::dirs;
 struct PackageSpecifier {
     name: String,
     version: String,
+    package_type: Option<PackageType>,
 }
 
 impl From<PackageDescriptor> for PackageSpecifier {
     fn from(descriptor: PackageDescriptor) -> Self {
-        Self { name: descriptor.name, version: descriptor.version }
+        Self {
+            name: descriptor.name,
+            version: descriptor.version,
+            package_type: Some(descriptor.package_type),
+        }
     }
 }
 
@@ -54,7 +60,7 @@ impl From<PackageDescriptor> for PackageSpecifier {
 #[derive(Serialize, Deserialize, Debug)]
 struct PackageLock {
     packages: Vec<PackageSpecifier>,
-    package_type: PackageType,
+    format: LockfileFormat,
 }
 
 /// New process to be launched.
@@ -148,7 +154,7 @@ struct ProcessOutput {
 #[op]
 async fn analyze(
     op_state: Rc<RefCell<OpState>>,
-    package_type: PackageType,
+    package_type: Option<PackageType>,
     packages: Vec<PackageSpecifier>,
     project: Option<String>,
     group: Option<String>,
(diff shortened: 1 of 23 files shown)


cli/src/commands/jobs.rs
```diff
@@ -138,7 +138,6 @@ pub async fn handle_submission(api: &mut PhylumApi, matches: &clap::ArgMatches)
                 );
             }
 
-            request_type = res.package_type;
```
kylewillmon, line 141: This means that all job submissions will use the package type from your config file (which defaults to NPM). But API ignores the value we put here, so that's fine.

cli/src/commands/extensions/api.rs
```diff
@@ -361,8 +361,12 @@ async fn parse_lockfile(
...
 
+    // PackageLock should report LockfileFormat and not PackageType. This is just a
+    // placeholder until that can be updated.
+    let package_type = parsed.packages.first().map(|p| p.package_type).unwrap_or(PackageType::Npm);
+
     Ok(PackageLock {
-        package_type: parsed.package_type,
+        package_type,
         packages: parsed.packages.into_iter().map(PackageSpecifier::from).collect(),
     })
```
kylewillmon, line 371: I'll create an issue to track this transition

cd-work: What prevents us from fixing this immediately?

cli/src/commands/jobs.rs
cd-work: Can we not submit the lockfile type here?

cd-work (changes requested):

cli/src/commands/extensions/api.rs
kylewillmon: It would be a breaking change for extensions.

After looking through the extensions, the only official or community extension that this would break is [`sbom`](https://github.com/phylum-dev/community-extensions/blob/da1e4b324c76b11f1ce9846574c75b34cc1e9a47/sbom/main.ts#L127). I'm surprised it's only one. But I'd still rather separate that change since it is breaking.

cli/src/commands/jobs.rs
kylewillmon: Because that's not a `PackageType`... I'm not sure I understand the suggestion...

cd-work: The API doesn't check it anyway, right? And that's ultimately what type we have? So shouldn't the goal be to submit the lockfile type?

kylewillmon: The API doesn't do anything with the value. But it does deserialize it with serde. So it has to be present and has to be a valid `PackageType`.

cli/src/commands/extensions/api.rs
kylewillmon: > How would you suggest migration?

That's the question I intend to answer in the separate issue... I don't have a clear answer right now and I didn't want to hold up this PR until I've figured that out.

cd-work (dismissed):
I'm not sure I'd personally merge this without having a concrete migration plan, but if you want to ensure the sbom work isn't blocked unnecessarily I'm not going to hold it up any longer.

kylewillmon:
Extending this update to the extension API was somewhat complicated. But I think 5b13851 should be a decent way to handle it. Let me know what you think.

extensions/phylum.ts
```diff
@@ -4,11 +4,12 @@
 type Package = {
   name: string;
   version: string;
+  package_type: string | undefined;
 };
 
 type Lockfile = {
-  package_type: string;
```
kylewillmon, line 10: This is technically a breaking change. So perhaps we should continue to return something here, even if it's just a placeholder value.

But none of our existing extensions actually use this value as a string. They just pass it back to `analyze` (which now accepts `undefined`). So I am hopeful that we can get away with removing this.

extensions/npm/main.ts
```diff
@@ -190,7 +190,7 @@ async function checkDryRun(subcommand: string, args: string[]) {
     return;
   }
 
-  const jobId = await PhylumApi.analyze("npm", lockfile.packages);
+  const jobId = await PhylumApi.analyze(undefined, lockfile.packages);
```
cd-work, line 193: Just to confirm: Did you test any/all of the existing extensions without making any changes to confirm this isn't a breaking change (to us at least)?

extensions/phylum.ts
cd-work: If we keep around `package_type` and just add the `format` in addition to the package type, it would still be a breaking change since the type changed, but it would only be an addition and not a change to existing data stored in this struct.

Might be worth it just to not break things? Ultimately we add the information anyway for the `format`.

I guess the problem would be with things like SBOM where the package type cannot be uniquely identified.

cd-work (changes requested):

extensions/npm/main.ts
kylewillmon: Yes. I have confirmed that the extensions work without modification.

extensions/phylum.ts
kylewillmon: > it would still be a breaking change since the type changed

That's correct. As far as I can tell, this can't be done without a breaking change.

> I guess the problem would be with things like SBOM where the package type cannot be uniquely identified.

This is exactly what led me to remove `package_type`. Using a value that doesn't match the packages could also break things or just confuse extension authors. So we might as well fix this now, even if it's technically a breaking change.

cd-work (dismissed):

cd-work (approved):

Issue #864, opened 2022-12-07
Allow deno to resolve phylum import


kylewillmon:
Deno provides a lot of useful tooling (e.g., `deno lint`, `deno fmt`, `deno lsp`). But much of that tooling relies on the ability to resolve imports (e.g., to analyze types in `deno lint` or provide documentation in an IDE via `deno lsp`).

The failure can be seen with `deno info main.ts` on any extension:

```
> deno info main.ts
Download ⠦ https://deno.land/std@0.150.0/fmt/colors.ts                                                                                                                                                                                                                                   local: /app/main.ts
type: TypeScript
dependencies: 1 unique
size: 17.31KB

file:///app/main.ts (5.68KB)
├── https://deno.land/std@0.150.0/fmt/colors.ts (11.63KB)
└── Relative import path "phylum" not prefixed with / or ./ or ../ (resolve error)
```

Deno cannot resolve the `phylum` import and therefore cannot provide any useful information or linting for our API.

kylewillmon:
I think the most sensible solution to this would be to host the `cli/src/extension_api.ts` file somewhere and allow extension authors to import it with a URL rather than with the fake `"phylum"` path.

Unfortunately, this approach would require us to deal with different versions of that file being used against different versions of the CLI, which may be complicated. But since we have full control over the runtime environment, we should be able to come up with a sensible solution to these versioning issues.

kylewillmon:
For this repo, using an import map (as in #878) would resolve this issue. But that does not solve the issue for external repos (including [phylum-dev/community-extensions](https://github.com/phylum-dev/community-extensions))

PR #969, opened 2023-03-10
Move extension API source to extension directory

cd-work:
Since it is required for this file to be available under the name `phylum.ts` for `https://deno.phylum.io/phylum.ts` to work, we need to first move this file before being able to setup the remote import in a follow-up PR.

See #864.


maxrake (changes requested):
It appears that there are at least five more spots where the file rename is needed:
image

maxrake (approved):
LGTM

PR #972, opened 2023-03-14
Improve sandboxed process failure message

### LINKED ISSUE #781: Throw exception for failure to create a process in `PhylumApi.runSandboxed()`

cd-work:
This provides a little improvement over the traditional "No such file or directory (os error 2)" error that is usually printed when the process specified in `PhylumApi.runSandboxed` does not exist.

Before:

```
❗ Error: No such file or directory (os error 2)
```

After:

```
Process "poetry" failed to start: No such file or directory (os error 2)
Extension error: Uncaught (in promise) Error: Process "poetry" failed to start
        return Deno.core.ops.run_sandboxed(process);
                             ^
    at Function.runSandboxed (https://deno.phylum.io/phylum.ts:353:30)
    at poetryCheckDryRun (file:///home/chris/programming/phylum/cli/extensions/poetry/main.ts:91:30)
    at file:///home/chris/programming/phylum/cli/extensions/poetry/main.ts:79:7
```

Note that the first line in the "After" output will only be printed when STDOUT is not redirected.

Some extensions also always run `Deno.run` before
`PhylumApi.runSandboxed` with the same command. This will still print the old error since it is generated by Deno directly.

Closes #781.

### DIFF
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -46,6 +46,8 @@ pub enum ExitCode {
     JsError,
     FailedThresholds,
     ConfirmationFailed,
+    SandboxStart,
+    SandboxStartCollision,
     Custom(i32),
 }
 
@@ -70,7 +72,21 @@ impl From<&ExitCode> for i32 {
             ExitCode::JsError => 16,
             ExitCode::ConfirmationFailed => 17,
             ExitCode::FailedThresholds => 100,
+            ExitCode::SandboxStart => 117,
+            ExitCode::SandboxStartCollision => 118,
             ExitCode::Custom(code) => *code,
         }
+    }
+}
+
+impl PartialEq<i32> for ExitCode {
+    fn eq(&self, code: &i32) -> bool {
+        i32::from(self).eq(code)
+    }
+}
+
+impl PartialEq<ExitCode> for i32 {
+    fn eq(&self, code: &ExitCode) -> bool {
+        self.eq(&i32::from(code))
     }
 }
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -78,15 +78,3 @@ impl From<&ExitCode> for i32 {
         }
     }
 }
-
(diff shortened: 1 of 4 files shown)


kylewillmon (approved):

cli/src/commands/mod.rs
```diff
@@ -70,7 +72,21 @@ impl From<&ExitCode> for i32 {
...
+    fn eq(&self, code: &i32) -> bool {
+        i32::from(self).eq(code)
+    }
+}
+
+impl PartialEq<ExitCode> for i32 {
+    fn eq(&self, code: &ExitCode) -> bool {
+        self.eq(&i32::from(code))
+    }
+}
```
kylewillmon, line 92: This sets up a non-transitive equality relationship.

Specifically:
* `ExitCode::Custom(117) == 117`
* `117 == ExitCode::SandboxStart`
* `ExitCode::Custom(117) != ExitCode::SandboxStart`

This seems odd to me. But I guess it's probably fine here because of the types.

kylewillmon (dismissed):

cli/src/commands/mod.rs
cd-work: I didn't think of custom. This is actually not fine according to Rust's official guidelines.

kylewillmon (approved):

PR #974, opened 2023-03-16
lockfile paths

### LINKED ISSUE #973: project files in parent directories not found
# Overview
The `phylum_project` crate has a function `get_current_project` which is used for locating the `.phylum_project` file. It's supposed to find the file if it is in the current directory or a parent directory (up to 32 levels up).

However, it starts with a path of `"."` and then navigates upwards by removing path components. Parent directories are never searched because successively removing the last components of `"."` yields only `""`.

# How To Reproduce
1. Go to the docs folder in this repository.
2. Run `phylum parse`
3. Get `❗ Error: Missing lockfile parameter`.

# Expected Behavior
The project file in the root of the project should be found and the lockfiles should be parsed.

matt-phylum:
This PR fixes a couple issues I noticed while looking at the `phylum_project` crate.

1. `ProjectConfig::lockfiles()` returns different paths before and after the first serialization. This shouldn't be a problem for CLI because I doubt the configuration is ever used this way without first being serialized and whether or not the path starts with an extra `"./"` is not normally a problem when operating on real filesystems, so I doubt it's causing any problems today. I removed the extra `"./"`.
2. `get_current_project()` calls `find_project_conf(".", true)` and then `find_project_conf` uses `Path::parent()` to navigate to parent directories. `Path::parent()` gets the parent path, not the parent directory, because it operates only on the textual representation of the path without hitting the filesystem. The parent of `"."` is not `".."` so whether you pass `true` or not for the second parameter to `find_project_conf` doesn't matter. I changed `get_current_project()` to canonicalize paths. Alternatively, we could convert to an absolute path, but the behavior would be different wrt symlinks etc (eg the test I've written would not work).

Closes #973

## Checklist

### DIFF
--- a/phylum_project/src/lib.rs
+++ b/phylum_project/src/lib.rs
@@ -93,18 +93,23 @@ pub fn find_project_conf(
     recurse_upwards: bool,
 ) -> Option<PathBuf> {
     let max_depth = if recurse_upwards { 32 } else { 1 };
-    let path = starting_directory.as_ref();
     // If `path` is like `.`, `path.parent()` is `None`.
-    let canonicalized = path.canonicalize();
-    let mut path = canonicalized.as_deref().ok().unwrap_or(path);
+    // Convert to a canonicalized path so removing components navigates up the
+    // directory hierarchy.
+    let mut path = starting_directory.as_ref().canonicalize().ok()?;
 
     for _ in 0..max_depth {
-        let conf_path = path.join(PROJ_CONF_FILE);
-        if conf_path.is_file() {
-            return Some(conf_path);
+        path.push(PROJ_CONF_FILE);
+        if path.is_file() {
+            return Some(path);
         }
 
-        path = path.parent()?;
+        // Remove conf file.
+        path.pop();
+        // Remove last directory component.
+        if !path.pop() {
+            return None;
+        }
     }
 
     if recurse_upwards {
--- a/phylum_project/src/lib.rs
+++ b/phylum_project/src/lib.rs
@@ -99,14 +99,11 @@ pub fn find_project_conf(
     let mut path = starting_directory.as_ref().canonicalize().ok()?;
 
     for _ in 0..max_depth {
-        path.push(PROJ_CONF_FILE);
-        if path.is_file() {
-            return Some(path);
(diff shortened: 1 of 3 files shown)


maxrake (approved):
Confirmed with local testing. The code changes appear fine to my non-rusty eye. The added tests are very welcome! Not sure how we missed this when adding the feature.
LGTM.

phylum_project/src/lib.rs
```diff
@@ -93,7 +93,10 @@ pub fn find_project_conf(
     recurse_upwards: bool,
 ) -> Option<PathBuf> {
     let max_depth = if recurse_upwards { 32 } else { 1 };
-    let mut path = starting_directory.as_ref();
+    let path = starting_directory.as_ref();
+    // If `path` is like `.`, `path.parent()` is `None`.
+    let canonicalized = path.canonicalize();
+    let mut path = canonicalized.as_deref().ok().unwrap_or(path);
```
cd-work, line 99: I don't think there's any point to `unwrap_or` here? Seems like this would be much easier by just doing `canonicalize().ok()?`.

cd-work (changes requested):

phylum_project/src/lib.rs
matt-phylum: `.ok()?` would cause the function to fail if the path does not exist, which the old code did not do.

maxrake (dismissed):
Tested again...still works. LGTM.

phylum_project/src/lib.rs
```diff
@@ -93,15 +93,23 @@ pub fn find_project_conf(
...
+    let mut path = starting_directory.as_ref().canonicalize().ok()?;
 
     for _ in 0..max_depth {
-        let conf_path = path.join(PROJ_CONF_FILE);
-        if conf_path.is_file() {
-            return Some(conf_path);
+        path.push(PROJ_CONF_FILE);
+        if path.is_file() {
+            return Some(path);
         }
```
cd-work, line 105: Why was this logic changed from join to push?

cd-work (changes requested):

cd-work (approved):

maxrake (approved):
Tested again...still works. LGTM.

PR #979, opened 2023-03-22
Use recursive lockfile search for `phylum init`

cd-work:
This uses the existing logic in the lockfiles crate to search for lockfiles when running `phylum init` without providing the lockfiles. This ensures consistency between the recursive lockfile search used without project and the project creation.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -196,12 +196,7 @@ pub fn lockfiles(
             let project_lockfiles = project.map(|project| project.lockfiles());
 
             // Fallback to walking the directory.
-            let lockfiles = project_lockfiles.unwrap_or_else(|| {
-                phylum_lockfile::find_lockfiles()
-                    .drain(..)
-                    .map(|(path, format)| LockfileConfig::new(path, format.to_string()))
-                    .collect()
-            });
+            let lockfiles = project_lockfiles.unwrap_or_else(|| find_lockfiles("."));
 
             // Ask for explicit lockfile if none were found.
             if lockfiles.is_empty() {
@@ -211,6 +206,14 @@ pub fn lockfiles(
             Ok(lockfiles)
         },
     }
+}
+
+/// Find lockfiles at or below the specified directory.
+pub fn find_lockfiles(directory: impl AsRef<Path>) -> Vec<LockfileConfig> {
+    phylum_lockfile::find_lockfiles_at(directory)
+        .drain(..)
+        .map(|(path, format)| LockfileConfig::new(path, format.to_string()))
+        .collect()
 }
 
 pub fn get_home_settings_path() -> Result<PathBuf> {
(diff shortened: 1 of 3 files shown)


cli/src/config.rs
```diff
@@ -196,12 +196,7 @@ pub fn lockfiles(
             let project_lockfiles = project.map(|project| project.lockfiles());
 
             // Fallback to walking the directory.
-            let lockfiles = project_lockfiles.unwrap_or_else(|| {
-                phylum_lockfile::find_lockfiles()
-                    .drain(..)
-                    .map(|(path, format)| LockfileConfig::new(path, format.to_string()))
-                    .collect()
-            });
+            let lockfiles = project_lockfiles.unwrap_or_else(|| find_lockfiles("."));
```
cd-work, line 199: We use "." as the search root here, though technically `project.root` would be correct. However the project's lockfiles are initialized to an empty array so they're always provided.

Because of this there's no way to differentiate between an empty lockfiles array and a missing one, so we need to treat both the same and not do recursive lockfile search.

maxrake (approved):
The changes were tested locally and appear to function as intended. LGTM.

cli/src/config.rs
kylewillmon: > Because of this there's no way to differentiate between an empty lockfiles array and a missing one, so we need to treat both the same and not do recursive lockfile search.

If I'm following correctly, this is when a `.phylum_project` file has been located, but it does not contain any lockfiles. Why is it better to error than to do a recursive search in that case? Even if they explicitly created the `.phylum_project` file without any lockfiles, it still seems like a recursive search is reasonable.

kylewillmon (changes requested):

cli/src/config.rs
cd-work: > If I'm following correctly, this is when a .phylum_project file has been located, but it does not contain any lockfiles. Why is it better to error than to do a recursive search in that case? Even if they explicitly created the .phylum_project file without any lockfiles, it still seems like a recursive search is reasonable.

The lockfiles field is `#[serde(default)] Vec<String>`. So there is no "located, but does not contain any lockfiles".

This closure is only called when there is no project file.

cd-work: > That's exactly my issue with it. A project file with an empty list (or no list) causes a guaranteed error because we don't do the recursive search.

Correct.

> We should probably also do a recursive search if the list is empty. That way we stay consistent and only error if the recursive search doesn't find any lockfiles.

I disagree. That would stop people from blacklisting all lockfiles in their repositories. Having a project file without lockfile right now is not supported, if we want to change that it should be done in a different PR. The expectation right now is that every project file has a list of lockfiles, if that list is empty, it means there are no lockfiles.

kylewillmon: > Having a project file without lockfile right now is not supported, if we want to change that it should be done in a different PR.

This is a good point. This PR is probably not the right place to argue about what a user intends when they do some unsupported thing.

kylewillmon (approved):

maxrake (approved):

PR #980, opened 2023-03-23
Add pip extension as official extension

kylewillmon:
This patch moves the `pip` extension into this repository as an official extension. A few updates were needed to resolve type checking, formatting, and to pass through error messages from the `pip install --dry-run` command

### Install blocked
```
> phylum pip install "lxml==4.6.3" 

[phylum] Analyzing packages…
[phylum] The operation caused a threshold failure.

    [critical] lxml@4.6.3 uses suspicious executables
    [low] lxml@4.6.3 uses unicode in a way that might be indicative of a Trojan Source attack
    [high] lxml@4.6.3 is vulnerable to lxml NULL Pointer Dereference allows attackers to cause a denial of service
```

### Install allowed
```
> phylum pip install cssselect==1.1.0

[phylum] Analyzing packages…
[phylum] All packages pass project thresholds.

Collecting cssselect==1.1.0
  Using cached cssselect-1.1.0-py2.py3-none-any.whl (16 kB)
Installing collected packages: cssselect
Successfully installed cssselect-1.1.0
```


kylewillmon:
After noticing [this comment](https://github.com/pypa/pip/issues/53#issuecomment-1166547157), I decided to try using `pip install --dry-run --report -` to get JSON output from `pip`. It works perfectly and is [well documented](https://pip.pypa.io/en/stable/cli/pip_install/#obtaining-information-about-what-was-installed). I am not sure why we didn't use this approach to begin with.

extensions/pip/main.ts
```diff
@@ -0,0 +1,214 @@
...
+  const cmd = Deno.run({ cmd: ["pip3", ...Deno.args] });
+  const status = await cmd.status();
+  Deno.exit(status.code);
+}
+
+// Analyze new dependencies with phylum before install/update.
+await checkDryRun();
+
+// Perform the package installation.
+const installStatus = PhylumApi.runSandboxed({
```
maxrake, line 54: Was the intention to "perform the package installation" here _outside_ of the sandbox? A `Deno.run()` was expected here instead of another `PhylumApi.runSandboxed()`.

extensions/pip/main.ts
```diff
@@ -0,0 +1,214 @@
...
+  packages: {
+    issues: {
+      severity: string;
+      title: string;
+    }[];
+  }[];
+};
+
+// Logs any identified issues to the screen.
+function logIssues(jobStatus: Record<string, unknown>) {
```
maxrake, line 93: Is it possible to update this function to display the issues in order of their criticality (e.g., Critical > High > Medium > Low)? The example given when attempting to install `lxml==4.6.3` shows three issues, but "out of order" (Critical, then Low, followed by High).

extensions/pip/main.ts
```diff
@@ -0,0 +1,214 @@
...
+};
+
+type DryRunReport = {
+  install: {
+    metadata: {
+      name: string;
+      version: string;
+    };
+  }[];
+};
```
maxrake, line 196: The [pip installation report specification](https://pip.pypa.io/en/stable/reference/installation-report/) says:
> version: the string 1. It will change only if and when backward incompatible changes are introduced, such as removing mandatory fields or changing the semantics or data type of existing fields. The introduction of backward incompatible changes will follow the usual pip processes such as the deprecation cycle or feature flags. Tools must check this field to ensure they support the corresponding version.

The fields expected to exist in this report are limited and therefore the chance of breaking changes is reduced, but even still, should there be such a check in this extension?

extensions/pip/main.ts
```diff
@@ -0,0 +1,214 @@
...
+// Analyze new packages.
+async function checkDryRun() {
+  console.log(`[${green("phylum")}] Finding new dependencies…`);
+
+  const status = PhylumApi.runSandboxed({
+    cmd: "pip3",
+    args: [...Deno.args, "--quiet", "--report", "-", "--dry-run"],
+    exceptions: {
+      run: ["./", "/bin", "/usr/bin", "/usr/local/bin", "~/.pyenv"],
+      write: [
```
maxrake, line 122: The `write` entries here don't match what is in the manifest. Is that intended?

extensions/pip/main.ts
```diff
@@ -0,0 +1,214 @@
...
+        "~/Library/Caches",
+        "~/.pyenv",
+        "~/.cache",
+        "~/.local/lib",
+        "/tmp",
+      ],
+      read: [
```
maxrake, line 129: The `read` entries here don't match what is in the manifest. Is that intended?

maxrake (changes requested):
This is the first pass review. It includes comments/questions from a code review only. Local testing will be performed next (or maybe after changes are made) to confirm expected operation and possibly identify unhandled edge cases.

extensions/pip/main.ts
kylewillmon: Yes. The `--dry-run` call doesn't require the same permissions as the actual install

extensions/pip/main.ts
kylewillmon: Yes. The `--dry-run` call doesn't require the same permissions as the actual install

extensions/pip/main.ts
kylewillmon: The install is intentionally sandboxed. This is a feature of this extension.

extensions/pip/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```console
+git clone https://github.com/phylum-dev/cli
+phylum extension install cli/extensions/pip
+```
+
+## Requirements
+
+This extension requires `pip` version 22.2 or greater (for the `--dry-run`
+flag). Check your version with `pip --version` and upgrade if necessary with
```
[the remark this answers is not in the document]
kylewillmon: I've added the `--report` flag to the note. I'm not worried about the experimental status, so v22.2 is still fine.

extensions/pip/main.ts
kylewillmon: A little tricky to compare the strings, but totally possible. Done.

maxrake (changes requested):
Submitting an updated review with code review comments and results from local testing. More local testing _may_ continue...

extensions/pip/main.ts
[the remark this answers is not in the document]
kylewillmon: That 3rd item applies to all of our current official extensions as well.

I don't think any of these are important enough to hold up this PR. We can open separate issues.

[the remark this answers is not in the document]
maxrake: > That 3rd item applies to all of our current official extensions as well.

The comparison made was to the current official `poetry` extension, which _does_ handle `--help` output:

image

> I don't think any of these are important enough to hold up this PR. We can open separate issues.

Roger that.

maxrake (approved):
Approved...with the understanding that the remaining issues found while testing this extension will be spun off into new issues.

maxrake (approved):

PR #986, opened 2023-03-27
Update Deno dependencies

cd-work:


cli/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "4.7.0"
 authors = ["Phylum, Inc. <engineering@phylum.io>"]
 license = "GPL-3.0-or-later"
 edition = "2021"
-rust-version = "1.65.0"
+rust-version = "1.68.0"
```
cd-work, line 7: Very high MSRV, due to requirement of the `pin!` macro. But I do believe it should be fine for CLI since we build it ourselves anyway and very few people need to build from source.

kylewillmon (changes requested):

cli/Cargo.toml
```diff
@@ -65,6 +65,9 @@ deno_ast = { version = "0.25.0", features = ["transpiling"] }
 birdcage = { version = "0.2.0" }
 libc = "0.2.135"
 
+[target.'cfg(windows)'.build-dependencies]
+winapi = { version = "0.3.9", features = ["winbase", "processenv"] }
```
kylewillmon, line 69: It's probably worth adding an issue (or at least a comment) explaining that this can be removed after the next `deno_io` release

kylewillmon (approved):

PR #993, opened 2023-03-30
pip: Ignore local and URL dependencies

kylewillmon:
The pip extension needs to distinguish local and URL dependencies from those that are requested by name and version. Submitting the name and version of a local dependency is no good and will likely cause failure or incorrect results.

This patch ignores these dependencies so that analysis can continue.

Sample of the new output:

```
> cat requirements.txt
./foo
packaging==21.3

> phylum pip install -r requirements.txt
[phylum] Cannot analyze dependency: foo-0.1.0

[phylum] Analyzing packages…
[phylum] All packages pass project thresholds.

Processing ./foo
 [... SNIP ...]
Successfully built foo
Installing collected packages: pyparsing, foo, packaging
Successfully installed foo-0.1.0 packaging-21.3 pyparsing-3.0.9
```


andreaphylum (approved):

maxrake (approved):
The changes were tested locally and appear to work as intended. One limitation was found in that local packages that are not in the current directory will fail the `checkDryRun()` with a error of the sort:
`OSError: [Errno 1] Operation not permitted: '/path/to/local/package'`

maxrake (approved):

PR #995, opened 2023-03-30
Fix `phylum project link` overwriting project file

### LINKED ISSUE #994: `phylum project link` overwrites existing project config
Using `phylum project link` overwrites the `.phylum_project` file. This may be unexpected as a user may run `phylum project link` simply to change the project that is being used.

# How To Reproduce
In this example, you can see that the `lockfiles` are lost:

```
> cat .phylum_project 
id: b21c6236-c031-456e-b903-decb6797cf4f
name: bar
created_at: 2023-03-30T11:58:14.104512-05:00
group_name: null
lockfiles:
- path: Cargo.lock
  type: cargo

> phylum project link foo
✅ Linked the current working directory to the project foo.

> cat .phylum_project 
id: 06babf58-62dd-4ffb-8f44-b65257b18b0c
name: foo
created_at: 2023-03-30T11:58:26.124790-05:00
group_name: null
lockfiles: []
```

# Expected Behavior
`phylum project link` should update the existing file rather than overwrite it.

cd-work:
Resolve an issue where `phylum project link <PROJECT>` would create a new config file, discarding all information stored in the `.phylum_project` file beyond the project information (i.e. lockfile data).

To ensure all auxiliary information is kept, the existing project is loaded when available and only the project-specific keys are overwritten.

Closes #994.

### DIFF
--- a/cli/src/commands/init.rs
+++ b/cli/src/commands/init.rs
@@ -56,7 +56,7 @@ pub async fn handle_init(api: &PhylumApi, matches: &ArgMatches) -> CommandResult
             let uuid = project::lookup_project(api, &project, group.as_deref())
                 .await
                 .context(format!("Could not find project {project:?}"))?;
-            ProjectConfig::new(uuid, project.into(), group)
+            ProjectConfig::new(uuid, project, group)
         },
         project_config => project_config.context("Unable to create project")?,
     };
--- a/cli/src/commands/init.rs
+++ b/cli/src/commands/init.rs
@@ -8,7 +8,7 @@ use clap::parser::ValuesRef;
 use clap::ArgMatches;
 use dialoguer::{Confirm, FuzzySelect, Input, MultiSelect};
 use phylum_lockfile::LockfileFormat;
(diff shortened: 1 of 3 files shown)


cli/src/commands/init.rs
```diff
@@ -8,7 +8,7 @@ use clap::parser::ValuesRef;
 use clap::ArgMatches;
 use dialoguer::{Confirm, FuzzySelect, Input, MultiSelect};
 use phylum_lockfile::LockfileFormat;
-use phylum_project::{LockfileConfig, PROJ_CONF_FILE};
+use phylum_project::{LockfileConfig, PROJ_CONF_FILE, ProjectConfig};
```
maxrake, line 11:
```suggestion
use phylum_project::{LockfileConfig, ProjectConfig, PROJ_CONF_FILE};
```
CI is failing w/o this change

maxrake (changes requested):

maxrake (changes requested):
The changes were tested locally and work as intended. However, CI is failing because `clippy` is now unhappy.

maxrake (approved):
LGTM

PR #997, opened 2023-03-31
Make `phylum package` type argument mandatory

### LINKED ISSUE #996: Make `--package-type` a required option for the `phylum package` command
# Overview
Not requiring the type yields confusing results as the CLI will just choose for you without context.

# Acceptance Criteria
- [ ] The `phylum package` command requires a proper `--package-type` input

cd-work:
Previously with the `--package-type` missing, the global configuration file's request type would be used instead. Realistically this would mean that `npm` was always used as a fallback without an ecosystem specified.

To ensure users don't get incorrect results when requesting non-npm packages, the package type option is now mandatory.

Closes #996.

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -182,7 +182,8 @@ pub fn add_subcommands(command: Command) -> Command {
                     .short('t')
                     .long("package-type")
                     .value_name("type")
-                    .help(r#"The type of the package ("npm", "rubygems", "pypi", "maven", "nuget", "golang", "cargo")"#),
+                    .help(r#"The type of the package ("npm", "rubygems", "pypi", "maven", "nuget", "golang", "cargo")"#)
+                    .required(true),
                 Arg::new("json")
                     .action(ArgAction::SetTrue)
                     .short('j')
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -173,16 +173,21 @@ pub fn add_subcommands(command: Command) -> Command {
         )
         .subcommand(
             Command::new("package").about("Retrieve the details of a specific package").args(&[
-                Arg::new("name").value_name("name").help("The name of the package.").required(true),
+                Arg::new("package-type")
+                    .index(1)
+                    .value_name("type")
+                    .help("Package ecosystem type")
+                    .value_parser(["npm", "rubygems", "pypi", "maven", "nuget", "golang", "cargo"])
+                    .required(true),
+                Arg::new("name")
+                    .index(2)
+                    .value_name("name")
+                    .help("The name of the package.")
+                    .required(true),
                 Arg::new("version")
+                    .index(3)
                     .value_name("version")
                     .help("The version of the package.")
-                    .required(true),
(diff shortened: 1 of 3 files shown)


docs/command_line_tool/phylum_package.md
```diff
@@ -7,7 +7,7 @@ hidden: false
 Retrieve the details of a specific package
 
 ```sh
-Usage: phylum package [OPTIONS] <name> <version>
+Usage: phylum package [OPTIONS] --package-type <type> <name> <version>
```
maxrake, line 10: Was `--package-type` intentionally left as an option instead of an argument? The expectation was that the signature would look more like:

```
Usage: phylum package [OPTIONS] <name> <version> <type>
```

It seems awkward to have a required option.

cli/src/app.rs
```diff
@@ -182,7 +182,8 @@ pub fn add_subcommands(command: Command) -> Command {
                     .short('t')
                     .long("package-type")
                     .value_name("type")
-                    .help(r#"The type of the package ("npm", "rubygems", "pypi", "maven", "nuget", "golang", "cargo")"#),
+                    .help(r#"The type of the package ("npm", "rubygems", "pypi", "maven", "nuget", "golang", "cargo")"#)
```
maxrake, line 185: Can a `.value_parser` with a `PossibleValuesParser` be used here to limit the options to valid choices like is done for the `--lockfile-type` argument of the `parse`, `analyze`, and `init` commands?

As it is now, the error is not terrible:

```
❯ target/debug/phylum package pyyaml 5.3.1 -t pype
❗ Error: HTTP request error (400 Bad Request):

The submitted package has an unsupported registry type.
```

...but it could be even better, like how `--lockfile-type` works:

```
❯ target/debug/phylum init -t corgo
error: invalid value 'corgo' for '--lockfile-type <type>'
  [possible values: yarn, npm, gem, pip, pipenv, poetry, mvn, gradle, nuget, go, cargo, spdx, auto]

  note: value 'cargo' exists

For more information, try '--help'.
```

maxrake (changes requested):
The code was tested locally and it works as advertised. Marking as `Request changes` mostly just to have a discussion about ways to make the solution better.

docs/command_line_tool/phylum_package.md
cd-work: I kept it as an argument to not unnecessarily break stuff, but we could change it. The original issue didn't suggest doing so either so I just went with what is "safe".

maxrake: Roger that...and understood. In this case, the assumption is that there are no consumers of this command to break...but maybe @furi0us333 has more insight there.

furi0us333: Impact should be minimal and I think if it is a mandatory field, it isn't an "option" anymore. Seems like argument makes the most sense to me. Sorry, I didn't mean my choice of words in the issue to be loaded.

maxrake (approved):

PR #1004, opened 2023-04-04
Remove `phylum project set-thresholds` subcommand

cd-work:
The latest Phylum API removed thresholds in favor of policies, this means that the `set-thresholds` subcommand is no longer functional and thus can be removed.

---

I initially wanted to just do all the policy stuff in a single PR, but this seems like a nice and simple thing to have separate.


docs/command_line_tool/phylum_project.md
```diff
@@ -4,7 +4,7 @@ category: 6255e67693d5200013b1fa3e
 hidden: false
 ---
 
-Create, list, link and set thresholds for projects
+Create, list, and link projects
```
maxrake, line 7:
```suggestion
Create, delete, list, and link projects
```

cli/src/app.rs
```diff
@@ -99,7 +99,7 @@ pub fn add_subcommands(command: Command) -> Command {
         .subcommand(
             Command::new("project")
                 .aliases(["projects"])
-                .about("Create, list, link and set thresholds for projects")
+                .about("Create, list, and link projects")
```
maxrake, line 102:
```suggestion
                .about("Create, delete, list, and link projects")
```

cli/src/commands/project.rs
```diff
@@ -28,8 +27,8 @@ pub async fn get_project_list(
 }
 
 /// Handle the project subcommand. Provides facilities for creating a new
-/// project, linking a current repository to an existing project, listing
-/// projects and setting project thresholds for risk domains.
+/// project, linking a current repository to an existing project, and listing
+/// projects.
```
maxrake, line 31:
```suggestion
/// project, linking a current repository to an existing project, listing
/// projects, and deleting projects.
```

maxrake (changes requested):
The changes (removal) appear to work. Requesting a few documentation updates and the possible removal of an unused exit code.

cli/src/commands/project.rs
```diff
@@ -88,110 +87,6 @@ pub async fn handle_project(api: &mut PhylumApi, matches: &clap::ArgMatches) ->
...
-                    "Set all thresholds for the {} project",
-                    style(&project_name).white()
-                );
-            },
-            Err(err) => {
-                print_user_failure!(
-                    "Failed to set thresholds for the {} project: {err}",
-                    style(&project_name).white(),
-                );
-                return Ok(ExitCode::SetThresholdsFailure.into());
```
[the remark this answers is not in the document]
cd-work: Great point, it should be safe to remove.

docs/command_line_tool/phylum_project.md
cd-work: I think this just demonstrates that this documentation isn't really maintainable. I've changed it to something more generic that shouldn't require frequent updates.

maxrake (approved):
LGTM.

Talking with @cd-work, we decided to hold this PR back and only merge it _after_ a CLI v4.8.0 release. That way, it will be included in v5.0.0 along with the larger breaking changes related to the shift away from thresholds and towards policies.

maxrake (approved):

PR #1006, opened 2023-04-04
Switch to policy endpoint for job results

### LINKED ISSUE #1002: Use policy endpoint
`phylum analyze` and `phylum history` should use the new [policy endpoint](https://api.staging.phylum.io/api/v0/swagger/index.html#/Jobs/evaluate_policy) (currently only available on staging).

For the first pass, simply printing the `report` field from the API response will be good enough. In a later pass, we may generate our own report from the other fields in the response in order to display the information in a more terminal-friendly way.

cd-work:
This patch switches to the new policy endpoint for `phylum analyze` and `phylum history <JOB_ID>`.

Since the thresholds-based analysis endpoint is not functional anymore, only the new endpoint is used and exposed to the extension API. This makes it a breaking change for all extensions using the `getJobStatus` function.

Closes #1002.


cli/src/bin/phylum.rs
```diff
@@ -177,34 +175,26 @@ async fn handle_commands() -> CommandResult {
...
-    Ok(ExitCode::Ok.into())
+    Ok(ExitCode::Ok)
 }
 
 #[tokio::main]
 async fn main() {
     match handle_commands().await {
-        Ok(CommandValue::Action(action)) => match action {
-            Action::None => ExitCode::Ok.exit(),
-            Action::Warn => exit_warn("Project failed threshold requirements!"),
```
kylewillmon, line 201: This was the only use of `exit_warn`. It's dead code now, but isn't detected because it is (pointlessly) `pub`.

You can delete `exit_warn` and remove the `pub` for `exit_fail`.

cli/src/commands/jobs.rs
```diff
@@ -3,66 +3,36 @@ use std::str::FromStr;
...
+    let status = match response {
+        Ok(status) => status,
         Err(err) if err.status() == Some(StatusCode::NOT_FOUND) => {
-            print_user_warning!(
-                "No results found. Submit a lockfile for processing:\n\n\t{}\n",
-                style("phylum analyze <lock_file>").blue()
-            );
-            return Ok(Action::None);
+            print_user_warning!("No results found for JobId {job_id}.");
+            return Ok(ExitCode::Ok);
```
kylewillmon, line 25: You've just kept the old behavior here. But why does a 404 response code exit successfully? Seems like that should be an error.

kylewillmon (changes requested):

cli/src/commands/extensions/api.rs
```diff
@@ -239,17 +240,18 @@ async fn get_refresh_token(op_state: Rc<RefCell<OpState>>) -> Result<RefreshToke
...
     op_state: Rc<RefCell<OpState>>,
     job_id: String,
-) -> Result<JobStatusResponse<PackageStatusExtended>> {
+) -> Result<PolicyEvaluationResponse> {
     let state = ExtensionState::from(op_state);
     let api = state.api().await?;
 
     let job_id = JobId::from_str(&job_id)?;
-    api.get_job_status_ext(&job_id).await.map_err(Error::from)
+    api.get_job_status(&job_id, []).await.map_err(Error::from)
```
andreaphylum, line 254: We should probably either add a new API, or add an optional argument here, but I think we can postpone that / track it in an issue considering that currently no extensions should require it just yet.

andreaphylum (approved):
Looks good to me. It's nice that we could get rid of all the `.into()`s.

andreaphylum (dismissed):

kylewillmon (approved):

PR #1007, opened 2023-04-05
Skip analysis with empty package list

cd-work:
When sending an analysis request with an empty list of packages, the API will return an error. Since we know ahead of time that this will be the case, it shouldn't be necessary to actually make the request.

This patchs adds a simple warning message telling the user clearly what the issue is and then exits with a zero exit code before any requests are made. This also reduces the noise in the output making it easier for users to understand the issue.


cli/src/commands/jobs.rs
```diff
@@ -193,6 +193,12 @@ pub async fn handle_submission(api: &mut PhylumApi, matches: &clap::ArgMatches)
         unreachable!();
     }
 
+    // Avoid request error without dependencies.
+    if packages.len() == 0 {
```
maxrake, line 197:
```suggestion
    if packages.is_empty() {
```
Clippy is failing in CI, with this suggested change.

maxrake (changes requested):

maxrake (approved):
LGTM

PR #1008, opened 2023-04-05
Add base option to `phylum analyze`

### LINKED ISSUE #1003: Add `--base` argument to `phylum analyze`
This is dependent on #1002.

The `phylum analyze` command should have a `--base` argument where users (specifically, `phylum-ci`) can provide a list of ignored packages (as a JSON file).

## Example usage
To analyze the new dependencies in the current branch against those in `main`:

```sh
branch=$(git branch --show-current)

# Parse base dependencies
git switch --detach ...main
phylum parse > base.json

# Analyze the new dependencies
git switch $branch
phylum analyze --base base.json
```

cd-work:
This patch adds the new `--base` option which allows passing a file
which contains a list of packages that will be ignored for analysis.
This can be used to analyze the delta between two package lists.

The expected file format is identical to the output of `phylum parse`,
so a base can be obtained using `phylum parse > base.json` and then
analyzed using `phylum analyze --base base.json`.

The option is hidden and not intended to be used by endusers, mainly to
avoid having to document the exact required file format.

Closes #1003.

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -275,6 +275,13 @@ pub fn add_subcommands(command: Command) -> Command {
                         .requires("lockfile")
                         .help("Lock file type used for all lock files (default: auto)")
                         .value_parser(PossibleValuesParser::new(parse::lockfile_types(true))),
+                    Arg::new("base")
+                        .short('b')
+                        .long("base")
+                        .value_name("FILE")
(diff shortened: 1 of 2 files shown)


maxrake (approved):
The changes were confirmed with local testing. LGTM.

PR #1010, opened 2023-04-05
Reformat `phylum history` output

kylewillmon:
The format of the `phylum history` command was very hard to read and littered with unlabeled or useless information. This patch brings a new simplified format.

The new format avoids long lines, but it does so by using more lines (currently 8 lines per job). This is a necessary trade-off. I first tried a table format, but the lines were 150+ characters long.

## Sample output
phylum-history

### DIFF
--- a/cli/src/format.rs
+++ b/cli/src/format.rs
@@ -2,7 +2,7 @@ use std::cmp;
 use std::io::{self, Write};
 use std::str::{self, FromStr};
 
-use chrono::NaiveDateTime;
+use chrono::{DateTime, NaiveDateTime, Utc};
 use console::style;
 use phylum_types::types::group::{
     GroupMember, ListGroupMembersResponse, ListUserGroupsResponse, UserGroup,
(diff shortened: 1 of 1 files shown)


cli/src/format.rs
```diff
@@ -102,54 +102,50 @@ impl Format for ListGroupMembersResponse {
...
-            let mut colored_score = style(&score).green().to_string();
-            let project_name = print::truncate(&job.project, 39);
-            let colored_project_name = style(project_name).white().bold().to_string();
-
-            if job.num_incomplete > 0 {
-                colored_score = format!("{}", style(&score).yellow());
-                state = format!("{}", style("INCOMPLETE").yellow());
+        for job in self {
+            let status = if job.num_incomplete > 0 {
+                style("INCOMPLETE").yellow().to_string()
```
maxrake, line 121: In the CI integrations, there is another status to indicate `INCOMPLETE WITH FAILURE`. Does it make sense to add that here? It could also be colored red.

cli/src/format.rs
```diff
@@ -102,54 +102,50 @@ impl Format for ListGroupMembersResponse {
...
+fn write_fields<W: Write>(fields: &[(&str, &str)], writer: &mut W) -> std::io::Result<()> {
+    let max_label_width = fields.iter().map(|f| f.0.len()).max().unwrap_or(0);
+
+    for field in fields {
+        writeln!(writer, "  {}  {}", style(leftpad(field.0, max_label_width)).blue(), field.1)?;
+    }
+    Ok(())
+}
+
 impl Format for Vec<JobDescriptor> {
```
maxrake, line 115: There is also an `impl Format for Vec<HistoryJob>` later in this same file, with a different output format. The user will see a different output format when the `--project` option is supplied (on the console with or without `--json` supplied).

This difference was introduced in https://github.com/phylum-dev/cli/pull/947 and may be intended since the discussion then was about how there were/are different endpoints for history, jobs, and groups. Is this still a restriction or can the output formats be consolidated now?

maxrake (approved):
The changes were verified with local testing. There is a question and a suggestion that should be addressed before merging.

cli/src/format.rs
kylewillmon: This restriction still exists

cli/src/format.rs
kylewillmon: Good question... I don't like `INCOMPLETE WITH FAILURE` simply because it is too long. I'd prefer:

1. Keep `INCOMPLETE` but use red text
1. Just use `FAIL` because that will be the final result anyways

Or maybe I'll think of another option before I come back to this tomorrow...

cd-work (dismissed):
Big fan of the output. I think this is also much better than any table-style output could have been and avoids relying on that dependency even more.

cli/src/format.rs
kylewillmon: :+1: I've updated the language to match our other integrations.

cli/src/format.rs
```diff
@@ -117,12 +117,11 @@ impl Format for Vec<JobDescriptor> {
...
-                style("INCOMPLETE").yellow().to_string()
-            } else if !job.pass {
-                style("FAIL").red().to_string()
-            } else {
-                style("PASS").green().to_string()
+            let status = match (job.num_incomplete == 0, job.pass) {
+                (false, false) => style("INCOMPLETE WITH FAILURE").red().to_string(),
+                (true, false) => style("FAILED").red().to_string(),
+                (false, true) => style("INCOMPLETE").yellow().to_string(),
+                (true, true) => style("SUCCESS").green().to_string(),
```
cd-work, line 124: Why convert `num_incomplete` to a boolean rather than just matching the number?

kylewillmon: That's fair. I've made that change now

maxrake (approved):
LGTM

PR #1016, opened 2023-04-11
Update PolicyEvaluationResponse to match API

kylewillmon:
The API is changing to use a count of incomplete dependencies rather than a bool for completion status. This patch updates CLI to match.

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -635,7 +635,7 @@ mod tests {
     async fn get_job_status() -> Result<()> {
         let body = PolicyEvaluationResponse {
             is_failure: true,
-            is_complete: true,
+            incomplete_count: 0,
             output: "output".into(),
             report: "report".into(),
         };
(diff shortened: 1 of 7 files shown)


maxrake (changes requested):
There are still five instances of `is_complete` left in the code: 1 in a docstring and the rest in extensions. It's odd though...the extensions don't complain loudly about the incorrect/missing attribute...they simply skip over it (or consider `jobStatus.is_complete` a false value).

image

kylewillmon:
Good catch. This should be fixed now.

> It's odd though...the extensions don't complain loudly about the incorrect/missing attribute...they simply skip over it (or consider `jobStatus.is_complete` a false value).

That's expected. If it's not there, `jobStatus.is_complete` will be `undefined`, which is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)

maxrake (approved):
The changes were tested locally.

LGTM.

Issue #990, opened 2023-03-29
`pip` extension does not handle editable package installs


maxrake:
# Overview

The `pip` extension does not handle editable package installs, either from a VCS project URL or a local directory. Attempting to do either results in a sandbox permissions issue.

# How To Reproduce
Steps to reproduce this behavior:

1. Installing an editable package with a VCS project URL results in a sandbox permissions issue:

<details><summary>Details</summary>
<p>

```console
../cli  15  19 on  pip [?] via 🦕 via 🐍 v3.11.2 (pip_ext_test) via 🦀 v1.68.0
❌ 2 ❯ target/debug/phylum pip list
Package    Version
---------- -------
pip        23.0.1
setuptools 65.5.0

../cli  15  19 on  pip [?] via 🦕 via 🐍 v3.11.2 (pip_ext_test) via 🦀 v1.68.0
❯ cat requirements-phylum-ci.txt
# This entry was created by running a `pip freeze` command on a checkout of the `phylum-ci` repo.
# The other entries (that weren't editable) were removed for the purposes of this demonstration.

-e git+ssh://git@github.com/phylum-dev/phylum-ci.git@9b5b2d46c4fce0458dc906b814795c5335fd85ae#egg=phylum

../cli  15  19 on  pip [?] via 🦕 via 🐍 v3.11.2 (pip_ext_test) via 🦀 v1.68.0
❯ target/debug/phylum pip install -r requirements-phylum-ci.txt
[phylum] Finding new dependencies…
ERROR: Could not install packages due to an OSError: [Errno 1] Operation not permitted: '/Users/maxrake/dev/phylum/localdev/cli/pip_ext_test/src'

[phylum] Pip dry-run failed.


../cli  15  19 on  pip [?] via 🦕 via 🐍 v3.11.2 (pip_ext_test) via 🦀 v1.68.0
❌ 1 ❯ python -m pip install -r requirements-phylum-ci.txt
Obtaining phylum from git+ssh://****@github.com/phylum-dev/phylum-ci.git@9b5b2d46c4fce0458dc906b814795c5335fd85ae#egg=phylum (from -r requirements-phylum-ci.txt (line 4))
  Cloning ssh://****@github.com/phylum-dev/phylum-ci.git (to revision 9b5b2d46c4fce0458dc906b814795c5335fd85ae) to ./pip_ext_test/src/phylum
[...]

PR #1017, opened 2023-04-11
Allow `phylum pip install -e .` on macOS

kylewillmon:
The command still does not work on Linux because it uses a cross-device link, which requires a newer version of landlock to support

Ref: #990


maxrake (approved):
The changes were verified with local testing.

LGTM.

maxrake (approved):

PR #1024, opened 2023-04-13
Rewrite CHANGELOG

kylewillmon:
Use the same standards in our main changelog as in the extension changelog

Old release notes were pulled from GitHub with the following command:

    gh api /repo/phylum-dev/cli/releases

And then with some manual tweaking of the data, I was able to produce a changelog that gets a lot closer to the standard.

## Future entries

I assume that future changelog entries will look a bit different since they will be written with the code.

For example, they likely won't contain any reference to the PR number (since the PR won't have been written yet). But it's easy enough for a developer to find the PR using a `git blame` on the changelog file.

Also, future entries may have multiple changelog entries in one PR. And they won't be written like PR titles... This is probably the biggest benefit of making this change.


maxrake (approved):
The format does not indicate `BREAKING CHANGES`, but it appears that the "Keep a Changelog" site does not say to do so. I suppose every change is a breaking one to somebody...

If it matters, the [v5.0.0 release notes](https://github.com/phylum-dev/cli/releases/tag/v5.0.0) were updated before this PR came out and maybe they should match what is in here?

LGTM.

kylewillmon:
> If it matters, the [v5.0.0 release notes](https://github.com/phylum-dev/cli/releases/tag/v5.0.0) were updated before this PR came out and maybe they should match what is in here?

Maybe? Or maybe the release notes shouldn't match the changelog? It's pretty quick to generate and tweak the release notes with GitHub and then you get links to all the relevant PRs in that release. So we could just have both...

CHANGELOG.md
```diff
@@ -0,0 +1,515 @@
...
+The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` and `Removed`.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [Unreleased]
+
+## [5.0.0] - 2023-04-13
+
+### Added
+* Add extension changelog by @cd-work (#1019)
```
cd-work, line 13: Is there a reason why you went for `*` instead of `-` for the lists here? Seems rather unconventional to me.

CHANGELOG.md
```diff
@@ -0,0 +1,515 @@
+# Changelog
+
+All notable changes are documented in this file.
+The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` and `Removed`.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [Unreleased]
+
+## [5.0.0] - 2023-04-13
```
cd-work, line 10: I honestly don't really like the brackets around the version, they just seem very noisy to me.

Isn't `## 5.0.0 - 2023-04-13` just as good, if not better? I know that you likely just followed `keepachangelog.com`, but I'm personally not a massive fan of this (it also makes interacting with the text in an editor more difficult imo).

CHANGELOG.md
```diff
@@ -0,0 +1,515 @@
...
+## [4.3.0] - 2023-01-17
+
+### Added
+* Add multi-lockfile support to `.phylum_project` by @kylewillmon (#902)
+* Make config file write atomic by @cd-work (#892)
+
+### Fixed
+* Fix sandbox executable path resolution by @cd-work (#905)
+
+
```
cd-work, line 102: Are the double newlines here intentional? It doesn't match the rest of the file as far as I can tell.

bump_version.sh
```diff
@@ -18,25 +18,28 @@ printf "version (w/o a leading 'v'): "
...
 git add cli/Cargo.toml
 
 printf "\nUpdating extension changelog...\n"
-sed -i'.bak' "s/\(## Unreleased\)/\1\n\n## ${version}/" extensions/CHANGELOG.md
+sed -i'.bak' "s/\(## Unreleased\)/\1\n\n## ${version} - ${today}/" extensions/CHANGELOG.md
 rm extensions/CHANGELOG.md.bak
 git add extensions/CHANGELOG.md
 
-commit_message="Bump to ${TAG} - ${changelog}"
+commit_message="Bump to ${TAG} - ${summary}"
```
cd-work, line 42: If we actually want brackets around the version, it should be done here too.

cd-work (changes requested):

CHANGELOG.md
maxrake: The brackets are for markdown links (defined at the bottom of the file).

cd-work:
> The format does not indicate BREAKING CHANGES, but it appears that the "Keep a Changelog" site does not say to do so. I suppose every change is a breaking one to somebody...

I personally find that marking things breaking or not is difficult when you're not just developing a semver-based library. But if we actually wanted to point this out explicitly it wouldn't be that difficult to prefix entries with `*Breaking*:` or have some other syntax indicator for it (like making breaking changes bold, which could be explained in the description at the top).

But generally I do think that most changes, especially in a CLI, would probably be considered "breaking" by someone.

CHANGELOG.md
kylewillmon: That is what is in our GitHub release notes, which this was generated from. And [Markdown doesn't care](https://spec.commonmark.org/0.30/#bullet-list-marker)... But I also don't care, so we can standardize on `-` since that seems to be more common.

CHANGELOG.md
kylewillmon: Not intentional. Good catch.

CHANGELOG.md
kylewillmon: Exactly what @maxrake said. It makes them links [when rendered](https://github.com/phylum-dev/cli/blob/changelog/CHANGELOG.md).

In fact, if you go far enough down in the file, some early versions don't have these links, because I didn't have enough information to include them (and didn't feel it was worth manually looking each one up).

cd-work (approved):

PR #1037, opened 2023-04-20
Handle null label in `phylum history --project`

kylewillmon:
The [job submission endpoint][1] does not allow null labels, but it does allow empty string. And when an empty label is submitted, it becomes a `null` label in the [project history endpoint][2].

[1]: https://api.phylum.io/api/v0/swagger/index.html#/Jobs/start_job
[2]: https://api.phylum.io/api/v0/swagger/index.html#/Projects/history_endpoint

This patch accounts for the potential `null` value in this field.


maxrake (approved):
LGTM.

PR #1038, opened 2023-04-20
Use package check endpoint in extensions

### LINKED ISSUE #1035: Expose check packages endpoint to extensions
The new [package check](https://api.phylum.io/api/v0/swagger/index.html#/Packages/check_packages_endpoint) endpoint should be added to the extension API as an alternative to `PhylumApi.analyze()` that doesn't require a project or a `.phylum_project` file.

kylewillmon:
Expose and use a `packageCheck()` function in the extension API backed by the new [package check endpoint][1]

[1]: https://api.phylum.io/api/v0/swagger/index.html#/Packages/check_packages_endpoint

Close #1035


andreaphylum (approved):

cli/src/api/endpoints.rs
```diff
@@ -203,6 +208,14 @@ mod test {
         );
     }
 
+    #[test]
+    fn package_check_is_correct() {
+        assert_eq!(
+            package_check(API_URI).unwrap().as_str(),
+            format!("{API_URI}/{API_PATH}data/packages/check"),
+        );
+    }
```
cd-work, line 217: I absolutely hate these tests personally. I'm kinda curious how you feel about them. To me they're just pointless since there's no real danger of breakage here and it always requires making changes in two locations.

extensions/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Added
+
+- `packageCheck` function to check a package against the default policy
```
cd-work, line 12: I think this is honestly kinda misleading. This sounds just like `phylum package` but it's not to check a package against the default policy, but to check **a list** of packages against the default policy.

Do we maybe want to rename this to `checkPackages`? I know the endpoint is `packages/check`, but I think having the verb first is more natural for function names and `packagesCheck` doesn't sound as good.

cd-work (changes requested):

cli/src/api/endpoints.rs
kylewillmon: I completely agree. I am happy to remove them (probably in a separate PR).

extensions/CHANGELOG.md
kylewillmon: I definitely meant to write "check a list of packages" and not "check a package".

And yes, `checkPackages` is a better name for the function. I'll make that change.

cd-work (approved):

cd-work (approved):

PR #1045, opened 2023-04-21
fix: `pip` parser fails for some lines containing comments

### LINKED ISSUE #1044: `requirements.txt` parser fails on valid files
# Overview
Some build systems generate `requirements.txt` files with comments on adjacent lines. 

`phylum analyze` and `phylum parse` fail to parse these files. 

# How To Reproduce
1. Using https://github.com/hypothesis/h/blob/main/requirements/requirements.txt as input, run `phylum analyze` 
2. Observe error (tested on Version v5.0.1-1-g0612262)
```
PS C:\projects\FirstBlood\REPOS\hypothesis\h\requirements> phylum analyze .\requirements.txt
❗ Error: Unable to locate any valid package in lockfile ".\\requirements.txt"

Caused by:
    0: Failed to parse lockfile ".\\requirements.txt"
    1: Failed to parse requirements file
    2: 0: at line 8, in AlphaNumeric:
           # via -r requirements/requirements.in
```

# Expected Behavior
`requirements.txt` files with comments should not fail parsing

maxrake:
The `pip` "Requirements File Format" is documented here: https://pip.pypa.io/en/stable/reference/requirements-file-format/

It allows for line continuations and a host of arguments to `pip install`, which this change does not address. Instead, this change accounts for lines containing comments where the comment character (`#`) is not the first character of the line. The comment line could be indented or comments can be inline (e.g., after a valid requirement). Both types of comments were added to the test fixture.

When updating the `CHANGELOG.md`, saving the file in the IDE caused an automatic reformatting of some lines...to remove trailing whitespace and ensure consistency for the list markers (`-` everywhere instead of `*`).

Fixes #1044

## Checklist

### DIFF
--- a/lockfile/src/parsers/pypi.rs
+++ b/lockfile/src/parsers/pypi.rs
@@ -16,9 +16,10 @@ use crate::{Package, PackageVersion};
 pub fn parse(input: &str) -> Result<&str, Vec<Package>> {
     let mut pkgs = Vec::new();
 
-    // Iterate over all non-empty lines.
-    for line in
-        input.lines().filter(|line| !line.trim().starts_with('#') && !line.trim().is_empty())
+    for line in input
+        .lines()
+        .map(|line| line.trim())
+        .filter(|line| !line.starts_with('#') && !line.is_empty())
     {
         let (_, line) = recognize(alt((take_until(" #"), not_line_ending)))(line)?;
         let (_, pkg) = package(line)?;
--- a/lockfile/src/parsers/pypi.rs
+++ b/lockfile/src/parsers/pypi.rs
@@ -17,7 +17,10 @@ pub fn parse(input: &str) -> Result<&str, Vec<Package>> {
     let mut pkgs = Vec::new();
 
     // Iterate over all non-empty lines.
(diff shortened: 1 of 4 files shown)


lockfile/src/parsers/pypi.rs
```diff
@@ -17,7 +17,10 @@ pub fn parse(input: &str) -> Result<&str, Vec<Package>> {
     let mut pkgs = Vec::new();
 
     // Iterate over all non-empty lines.
-    for line in input.lines().filter(|line| !line.starts_with('#') && !line.trim().is_empty()) {
+    for line in
+        input.lines().filter(|line| !line.trim().starts_with('#') && !line.trim().is_empty())
```
cd-work, line 21: Would probably be more efficient to assign `line.trim()` to a temporary variable. Seems like the kind of thing that you don't necessarily want to repeat.

cd-work (changes requested):

lockfile/src/parsers/pypi.rs
maxrake: I _think_ I was able to make the suggested change. Rust is still very much unnatural to me and so I welcome any idiomatic improvements.

cd-work (approved):

Issue #1046, opened 2023-04-21
Generate new lockfiles when analyzing manifests


cd-work:
When a user runs an analysis on a manifest explicitly without a lockfile present, we should generate a new lockfile for them and run the analysis on that.

PR #1047, opened 2023-04-21
Add generator for manifest to lockfile conversion

cd-work:
This patch adds the new `lockfile_generator` crate to the project, which
takes in a project directory and attempts to non-mutably generate a
lockfile for it.

This crate is used by the CLI's lockfile parser to automatically
generate a lockfile if the lockfile path it has received is a manifest
file rather than a lockfile.

The type of analysis required is determined in the following steps.

1) Format determination:
 - Try to determine lockfile type from CLI arguments
 - Try to determine lockfile type from lockfile name
 - Try to determine lockfile type from manifest name:
    - Find a format for the manifest that already has a lockfile
    - Fallback to first format that matches the manifest

2) Lockfile analysis:
 - If 1) did not hit the manifest fallback path, we analyze the lockfile
   with the detected format's parser.

3) Lockfile generation:
 - Check if original non-lockfile path is a manifest
 - Check if there is a generator available for the format
 - Generate lockfile at manifest's location using the format's generator
 - Analyze the generated lockfile

The split in analysis -> manifest determination -> analysis is necessary
since some formats have files that are commonly used as both lockfiles
and manifests (requirements.txt), thus it is necessary to first attempt
a lockfile analysis before being able to re-attempt parsing the
generated lockfile.

Closes https://github.com/phylum-dev/cli/issues/1046.

### DIFF
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -1,12 +1,10 @@
 //! `phylum parse` command for lockfile parsing
 
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::{fs, io};
 
 use anyhow::{anyhow, Context, Result};
-use phylum_lockfile::{
-    get_path_format, LockfileFormat, Package, PackageVersion, ThirdPartyVersion,
-};
+use phylum_lockfile::{LockfileFormat, Package, PackageVersion, Parse, ThirdPartyVersion};
 use phylum_types::types::package::PackageDescriptor;
 use walkdir::WalkDir;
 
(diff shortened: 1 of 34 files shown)


cd-work:
Right now running `phylum analyze -t cargo Cargo.toml` should work, but will first try to analyze the package as a lockfile.

Especially once support is added to `phylum init`, this might become a more common scenario so we probably don't want to waste time trying to parse in incorrect formats.

So it's probably a smart idea to check if the file itself is a manifest or lockfile first before running the lockfile analysis. It's likely possible to run these checks in the format determination path for the CLI parameters.

kylewillmon (changes requested):
The code will no longer parse a file if `.is_path_lockfile()` return `False`... So a user can't force a parser for a nonstandard filename.

This seems like a signficant regression.

```
> echo "pyyaml==5.3.1" > requirements-dev.txt

# Before
> phylum parse -t pip requirements-dev.txt 
[
  {
    "name": "pyyaml",
    "version": "5.3.1",
    "type": "pypi"
  }
]

# After
> phylum parse -t pip requirements-dev.txt 
❗ Error: "requirements-dev.txt" is neither a supported lockfile nor manifest
```

lockfile_generator/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+        // Generate lockfile at the target location.
+        let mut command = self.command();
+        command.current_dir(project_path);
+        command.stdin(Stdio::null());
+        command.stdout(Stdio::null());
+        command.stderr(Stdio::null());
+
+        // Provide better error message, including the failed program's name.
+        let mut child = command.spawn().map_err(|err| {
+            let program = command.get_program().to_string_lossy().into_owned();
```
kylewillmon, line 46: The `Debug` implementation for `Command` might be more useful here, since it will also print arguments.

From [the docs][1]:

> The default format approximates a shell invocation of the program along with its arguments.

[1]: https://doc.rust-lang.org/std/process/struct.Command.html#impl-Debug-for-Command

```suggestion
            let program = format!("{command:?}");
```

kylewillmon (changes requested):

cli/src/commands/parse.rs
```diff
@@ -45,37 +44,138 @@ pub fn handle_parse(matches: &clap::ArgMatches) -> CommandResult {
...
+    }
+
+    // If the path is neither a valid manifest nor lockfile, we abort.
+    if !is_manifest {
+        // Try to use existing parser error when available.
+        match lockfile_error {
+            Some(err) => return Err(err),
+            None => return Err(anyhow!("{path:?} is neither a supported lockfile nor manifest")),
+        }
+    }
```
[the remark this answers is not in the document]
cd-work: It's not unreachable though? This happens when lockfile parsing fails and it's not a valid manifest?

kylewillmon (dismissed):

kylewillmon (approved):

lockfile_generator/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+    ///
+    /// This will ignore all existing lockfiles and create a new lockfile based
+    /// on the current project configuration.
+    fn generate_lockfile(&self, project_path: &Path) -> Result<String> {
+        // Move files which interfere with lockfile generation.
+        let _relocators = self
+            .conflicting_files()
+            .drain(..)
+            .map(|file| FileRelocator::new(project_path.join(file)))
+            .collect::<Vec<_>>();
```
kylewillmon, line 43: This collects a `Vec<Result<Option<FileRelocator>>>` and doesn't do anything with potential errors (i.e., if the file could not be renamed).

Did you mean to collect a `Result<Vec<Option<FileRelocator>>>` so that you can propagate errors?

```suggestion
            .collect::<Result<Vec<_>>>()?;
```

kylewillmon (changes requested):

lockfile_generator/src/lib.rs
cd-work: I don't think it's **super** important in this case, but yeah your suggestion was my intention and should be better.

kylewillmon (approved):

kylewillmon (approved):

andreaphylum (approved):

PR #1050, opened 2023-04-21
Specify which changes should be noted in each changelog

kylewillmon:
Clarify changelogs and remove empty version


maxrake (approved):
LGTM.

PR #1055, opened 2023-04-24
Remove unnecessary OIDC parameters

kylewillmon:
The `scopes` parameters to the token endpoint had no effect. Additionally, the `profile` and `email` scopes are unused and do not change the token that is issued.

`Content-Type: application/json` was overridden by the use of `.form()` which sets the `Content-Type` to `application/x-www-form-urlencoded`


maxrake (approved):
LGTM.

PR #1060, opened 2023-04-25
Use `npm-shrinkwrap.json` for npm extension

cd-work:
This will make use of the `npm-shrinkwrap.json` instead of the `package-lock.json` when it is available in an NPM project.


maxrake (approved):
The change was tested locally and works as intended. There still may be an open question about `yarn.lock`, but that could be a separate issue. Also, the Windows runner appears to be failing for a reason unrelated to the changes here.

maxrake (approved):

PR #1061, opened 2023-04-25
Use --no-self-update for rustup on Windows

kylewillmon:
Self-update on Windows is hard... And rustup occasionally fails to self-update. There is a [related issue][1] to fix this behavior, but until then, using `--no-self-update` is a decent workaround.

[1]: https://github.com/rust-lang/rustup/issues/2441


maxrake (approved):
LGTM

PR #1062, opened 2023-04-26
Fix parsing with empty yarn lockfile

### LINKED ISSUE #1028: Error with default yarn lockfile
By default yarn generates a lockfile on `yarn init`. This lockfile is just a completely empty file.

Currently our parser throws an error when encountering this file, but it should be able to parse this and correctly notify the user that there are no packages to be analyzed.

cd-work:
This patch resolves an issue in both the v1 and v2 yarn lockfile parsers where a newly created lockfile would cause the parser to throw an error.

With yarn v1 the generated lockfile will only contain the lockfile header, thus parsing the file and checking for emptiness is sufficient to detect the edgecase.

With yarn v2 we cannot actually detect that a v2 yarn lockfile is present, since the lockfile is completely empty and does not contain any version info. However by making the version header optional in the v1 parser, it's still possible to parse the empty v2 lockfile using the v1 parser fallback.

Closes #1028.


maxrake (approved):
The changes were tested locally. I was unable to generate an empty v1 `yarn.lock` using `yarn init` (no yarn.lock was produced at all), but the added test for it does pass...

The empty v2 `yarn.lock` case was seen in local testing and worked as expected.

LGTM.

cd-work:
> The changes were tested locally. I was unable to generate an empty v1 yarn.lock using yarn init (no yarn.lock was produced at all), but the added test for it does pass...

Just for the record I tested this myself by running `yarn install` after `yarn init`. While v2 always generates a lockfile during init, v1 did not.

PR #1063, opened 2023-04-26
Fix path exists check in FileRelocator

kylewillmon:
The relocator should skip files that don't exist, but was accidentally only skipping the files that do exist...


cd-work (approved):
Oops.

PR #1066, opened 2023-05-01
Use Deno.Command instead of Deno.run

kylewillmon:
`Deno.run` is deprecated as of deno v1.33.0. This patch updates our extensions to use the preferred `Deno.Command` API

### DIFF
--- a/extensions/npm/main.ts
+++ b/extensions/npm/main.ts
@@ -69,8 +69,8 @@ if (
     "udpate".startsWith(Deno.args[0])
   )
 ) {
-  const cmd = Deno.run({ cmd: ["npm", ...Deno.args] });
-  const status = await cmd.status();
+  const cmd = new Deno.Command("npm", { args: Deno.args });
+  const status = await cmd.spawn().status;
   Deno.exit(status.code);
 }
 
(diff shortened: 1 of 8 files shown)


kylewillmon:
It's worth pointing out that we bundle deno v1.32.3 in CLI. And `Deno.run` wasn't deprecated until v1.33.0. We are only receiving the warnings because we use latest deno tooling in our CI.

Fortunately, those versions are close enough that we don't have any problems here (and I guess we could always disable the lint if it were actually a problem).

extensions/npm/main.ts
```diff
@@ -105,13 +105,13 @@ try {
 console.log(`[${green("phylum")}] Installing without build scripts…`);
 
 // Install packages without executing build scripts.
-const cmd = Deno.run({
-  cmd: ["npm", ...Deno.args, "--ignore-scripts"],
+const cmd = new Deno.Command("npm", {
+  args: [...Deno.args, "--ignore-scripts"],
```
maxrake, line 109: This is the only change in this PR where `...Deno.args` didn't become `Deno.args`. Is that intentional?

maxrake (changes requested):
`Deno.run` shows up in a few more spots where it _appears_ an update is in order:
* `cli/tests/extensions/permissions.rs`
* `docs/extensions/extension_manifest.md`
* `docs/extensions/extension_sandboxing.md`

extensions/npm/main.ts
kylewillmon: Yes. It's intentional. In other places, removing the command name left us with just `[...Deno.args]` which is the same (value) as `Deno.args`. So I changed it to use `Deno.args` directly and avoid the copy.

Here the copy is unavoidable (since `Deno.args` is immutable and we need to add `--ignore-scripts`)

maxrake (approved):
LGTM

Issue #1068, opened 2023-05-04
Pre-check/Birdcage permission error


furi0us333:
# Overview
I am getting the following error when trying to use phylum with pip: `PermissionError: [Errno 13] Permission denied: '/etc/debian_version'`

# How To Reproduce
Steps to reproduce this behavior:
1. Given the system information below, run `phylum pip install "pyyaml==5.3.1"`
2. See error

# Expected Behavior
Phylum pre-check/birdcage run without the exception.

# Additional Context
System details:
```sh
$ python3 -m pip list
Package       Version
------------- -------
Pillow        9.5.0
pip           23.1.2
pkg_resources 0.0.0
setuptools    44.0.0

$ python3 -c 'import sys; print(sys.base_exec_prefix)'
/usr

$ uname -a
Linux DESKTOP-RQ0PRHE 5.15.90.1-microsoft-standard-WSL2 #1 SMP Fri Jan 27 02:56:13 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

$ phylum version
✅ phylum (Version v5.0.1)

```

kylewillmon:
Unfortunately, it appears that `pip install "pyyaml==5.3.1"` goes through the process of compiling `pyyaml` from source...

To allow compiling C code, we would need to significantly weaken our sandbox.

louislang:
this seems like a pretty common thing people are likely to encounter. We should document this and perhaps consider adding some sort of escape hatch, e.g., `Sandbox prevented package from compiling. Compile outside of the sandbox? [y/N]: `

cd-work:
I feel like that would just completely circumvent the whole purpose of sandboxing it in the first place?

PR #1071, opened 2023-05-04
Allow mime types and OS information in default sandbox

kylewillmon:
These files should be harmless and can be added to the default exceptions

_Note: This should fix the immediate error in #1068, but more testing is needed to determine if there are other issues that need to be addressed._


maxrake (approved):
LGTM

PR #1072, opened 2023-05-04
Improve Birdcage errors on Linux

cd-work:
This patch provides a more verbose error message when creating the sandbox fails on Linux.

While it is possible that this error occurs while on Linux kernel 5.13+, this is the most likely culprit, so it is the most realistic way a user can figure out a solution by themselves.

### DIFF
--- a/cli/src/commands/sandbox.rs
+++ b/cli/src/commands/sandbox.rs
@@ -3,7 +3,10 @@
 use std::os::unix::process::ExitStatusExt;
 use std::process::Command;
 
-use anyhow::{anyhow, Context, Error, Result};
+use anyhow::{anyhow, Result};
+#[cfg(target_os = "linux")]
+use anyhow::{Context, Error};
+#[cfg(target_os = "linux")]
 use birdcage::error::Error as SandboxError;
 use birdcage::{Birdcage, Exception, Sandbox};
 use clap::ArgMatches;
--- a/cli/src/commands/sandbox.rs
+++ b/cli/src/commands/sandbox.rs
@@ -56,7 +56,7 @@ fn lock_process(matches: &ArgMatches) -> Result<()> {
         Ok(birdcage) => birdcage,
         #[cfg(target_os = "linux")]
         Err(err @ SandboxError::Ruleset(_)) => {
-            return Err(Error::from(err)).context("sandbox requires Linux 5.13+");
+            return Err(Error::from(err)).context("sandbox requires Linux kernel 5.13+");
         },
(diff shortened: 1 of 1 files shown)


chrisduerr:
This is the output when testing on 5.10:

```
❗ Error: sandbox requires Linux 5.13+

Caused by:
    landlock ruleset error: fully incompatible access-rights: BitFlags<AccessFs>(0b1111111111111, Execute | WriteFile | ReadFile | ReadDir | RemoveDir | RemoveFile | MakeChar | MakeDir | MakeReg | MakeSock | MakeFifo | MakeBlock | MakeSym)
```

cli/src/commands/sandbox.rs
```diff
@@ -44,8 +48,18 @@ pub async fn handle_sandbox(matches: &ArgMatches) -> CommandResult {
...
-        if matches.get_flag("strict") { Birdcage::new()? } else { permissions::default_sandbox()? };
+    let birdcage =
+        if matches.get_flag("strict") { Birdcage::new() } else { permissions::default_sandbox() };
+
+    // Provide additional error context.
+    let mut birdcage = match birdcage {
+        Ok(birdcage) => birdcage,
+        #[cfg(target_os = "linux")]
+        Err(err @ SandboxError::Ruleset(_)) => {
+            return Err(Error::from(err)).context("sandbox requires Linux 5.13+");
```
maxrake, line 59:
```suggestion
            return Err(Error::from(err)).context("sandbox requires Linux kernel 5.13+");
```

maxrake (approved):
One very small wording suggestion but still LGTM.

cli/src/commands/sandbox.rs
kylewillmon, line 59: The `return` gets a lot shorter if you capture the entire `Result`

```suggestion
        err @ Err(SandboxError::Ruleset(_)) => {
            return err.context("sandbox requires Linux 5.13+");
```

cd-work: Your suggestion doesn't work since the error isn't the same as the return type?

maxrake (approved):
LGTM

cli/src/commands/sandbox.rs
kylewillmon: Ah. Good point... Didn't test it because I'm not on Linux...

I guess that would have to be

```
        err @ Err(SandboxError::Ruleset(_)) => err.context("sandbox requires Linux 5.13+")?,
```

But that feels weird....

kylewillmon (approved):

PR #1074, opened 2023-05-04
Remove release summary

kylewillmon:
Now that we have a more detailed changelog, the one line summary is only used for the release commit and the tag message. But those aren't great places to explain a release.

### DIFF
--- a/tag.sh
+++ b/tag.sh
@@ -26,7 +26,6 @@ if [ -z "${TAG_COMMIT}" ]; then
 fi
 
 TAG=$(git log --pretty="format:%(trailers:key=release-version,valueonly)" -1 "${TAG_COMMIT}")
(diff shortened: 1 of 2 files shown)


tag.sh
```diff
@@ -52,7 +51,7 @@ if [ -z "${SKIP_ORIGIN_HEAD_CHECK:-}" ]; then
     fi
 fi
 
-git tag --sign -m "${TAG} - ${SUMMARY}" "${TAG}" "${TAG_COMMIT}"
+git tag --sign -m "Phylum CLI ${TAG}" "${TAG}" "${TAG_COMMIT}"
```
kylewillmon, line 54: I don't think anybody reads tag messages... But I'll take suggestions if there's something else I should put here.

maxrake: I think you are correct. This should be good enough.

maxrake (approved):
LGTM

PR #1079, opened 2023-05-08
Use `skip-build` mode for yarn lockfile generation

cd-work:
This patch adds the `--mode=skip-build` argument to the already existing `--mode=update-lockfile` argument when generating lockfiles for yarn.

While the output is identical for both commands, adding the `skip-build` mode does speed up generation time and thus likely provides benefits even with the lockfile mode already provided.


maxrake (approved):
The changes were tested locally. LGTM.

Issue #1011, opened 2023-04-06
Add parsing for additional SPDX locator formats


ejortega:
# Overview
The SPDX spec lists additional locator format other than `PURL` for the `PACKAGE-MONITOR` field. We should add this feature to the current SPDX parsers.

https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f3-package-manager

```PACKAGE-MANAGER <ecosystem> <package-name>@<package-version>```

# Acceptance Criteria
- [ ] Update json/yaml parser
- [ ] Update tag:value parser

# Additional Context
The [community extension](https://github.com/phylum-dev/community-extensions/blob/da1e4b324c76b11f1ce9846574c75b34cc1e9a47/sbom/main.ts#L81) outputs in this format as well.

PR #1083, opened 2023-05-09
Update SPDX parser for additional locator formats

ejortega:
## Checklist

Updates the SPDX parser to capture packages for non-PURL locator formats https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f3-package-manager

- [x] Have you updated `CHANGELOG.md` (or `extensions/CHANGELOG.md`), if applicable

Closes https://github.com/phylum-dev/cli/issues/1011


lockfile/src/spdx.rs
```diff
@@ -76,62 +79,84 @@ fn type_from_url(url: &str) -> anyhow::Result<PackageType> {
...
+fn from_locator(registry: &str, locator: &str) -> anyhow::Result<Package> {
+    lazy_static! {
+        static ref RE: Regex = Regex::new(r"^(([^:]*:)*[^@]+)@([^:]+)$").unwrap();
+    }
+
+    let package_type = PackageType::from_str(registry).map_err(|_| UnknownEcosystem)?;
+
+    let captures = RE.captures(locator).expect("Invalid locator");
+    let name = decode(captures.get(1).expect("Invalid package name").as_str())?;
+    let version = captures.get(3).expect("Invalid package name").as_str();
```
maxrake, line 133:
```suggestion
    let version = captures.get(3).expect("Invalid package version").as_str();
```
This is a guess, based on the context.

maxrake (changes requested):
It seems okay based on a non-rusty review and the tests pass.

There are two changes requested:

* Possible typo fix
* Add an entry to the `CHANGELOG.md` file

It might still be good to have another, rustier, team member complete this review. Perhaps @kylewillmon or @cd-work ...

cd-work (changes requested):

lockfile/src/spdx.rs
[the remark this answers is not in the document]
ejortega: From previous conversations with @kylewillmon and @louislang, we want to fail hard if missing any critical information when parsing SPDX SBOMS.

lockfile/src/spdx.rs
```diff
@@ -76,62 +79,84 @@ fn type_from_url(url: &str) -> anyhow::Result<PackageType> {
...
+        (PackageType::Maven, Some(ns)) => format!("{}:{}", ns, purl.name()),
+        (PackageType::Npm | PackageType::Golang, Some(ns)) => format!("{}/{}", ns, purl.name()),
+        _ => purl.name().into(),
+    };
+
+    let pkg_version = match (&pkg_info.version_info, purl.version()) {
+        (Some(v), _) => v.to_string(),
+        (None, Some(v)) => v.into(),
+        _ => bail!("Version not found for `{}`", pkg_info.name),
+    };
```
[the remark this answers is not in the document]
ejortega: Updated

lockfile/src/spdx.rs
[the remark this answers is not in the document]
cd-work: Panicking inside a library like this is a no-go imo and gives us zero control over how these errors are printed.

It also means this affects both CLI and the GitHub App unilaterally.

Failing parsing is one thing and we already do that in many of our parsers. But panicking is absolutely not an option imo.

maxrake (approved):
LGTM

lockfile/src/spdx.rs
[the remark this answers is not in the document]
ejortega: I updated to return anyhow error, is that acceptable?

maxrake (dismissed):

tests/fixtures/locator.spdx.json
```diff
@@ -0,0 +1,162 @@
...
+            "versionInfo": "9.3.7.0",
+            "supplier": "NOASSERTION",
+            "homepage": "https://central.sonatype.com/artifact/org.jruby/jruby-complete/9.3.7.0",
+            "externalRefs": [
+                {
+                    "referenceCategory": "PACKAGE-MANAGER",
+                    "referenceType": "maven",
+                    "referenceLocator": "org.jruby:jruby-complete@9.3.7.0"
+                }
+            ]
```
kylewillmon, line 32: Looking at the [spec][1], I think this should be `"referenceType": "maven-central"`...

[1]: https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f31-maven-central

tests/fixtures/locator.spdx.json
```diff
@@ -0,0 +1,162 @@
...
+            "licenseConcluded": "NONE",
+            "licenseDeclared": "NONE",
+            "copyrightText": "NOASSERTION",
+            "versionInfo": "13.0.1",
+            "sourceInfo": "acquired package info from dotnet project assets file: /root/.dotnet/sdk/7.0.202/DotnetTools/dotnet-user-secrets/7.0.4-servicing.23119.3/tools/net7.0/any/dotnet-user-secrets.deps.json",
+            "externalRefs": [
+                {
+                    "referenceCategory": "PACKAGE-MANAGER",
+                    "referenceType": "nuget",
+                    "referenceLocator": "Newtonsoft.Json@13.0.1"
```
kylewillmon, line 100: This doesn't match the [spec][1]... Shouldn't this be `"Newtonsoft.Json/13.0.1"`?

[1]: https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f33-nuget

tests/fixtures/locator.spdx.json
```diff
@@ -0,0 +1,162 @@
+{
+    "spdxVersion": "SPDX-2.2",
+    "documentNamespace": "https://localmaven",
+    "creationInfo": {
+        "creators": [
+            "Tool: github.com/phylum-dev/cli"
+        ],
```
kylewillmon, line 7: I see now... We're testing against an SPDX file that was created by [our own API][1]...

[1]: https://api.staging.phylum.io/api/v0/swagger/index.html#/Jobs/jobs_get_job_sbom

That makes some sense because we should be able to parse these files. But there are a few big problems with that:

1. Out API currently uses `"referenceType"` values that don't exist in the spec
2. Our API is currently working to switch to PURL, so these locators will be phased out.
3. Even for reference types that **are** defined in the spec, our API seems to deviate from the spec (e.g., `nuget`)

Can we use some other tool to generate our test files?

kylewillmon (changes requested):
The test file used seems to deviate from the spec, which makes it a troubling test...

lockfile/src/spdx.rs
```diff
@@ -76,62 +79,84 @@ fn type_from_url(url: &str) -> anyhow::Result<PackageType> {
...
                     if external.reference_type == "purl" {
-                        Some(&external.reference_locator)
+                        Some(from_purl(&external.reference_locator, pkg_info))
                     } else {
-                        None
+                        Some(from_locator(&external.reference_type, &external.reference_locator))
                     }
                 },
                 _ => None,
             })
```
kylewillmon, line 158: Is it possible to have multiple `PACKAGE-MANAGER` locators? If so, this will only ever attempt to parse the first one. What if we can't parse the first one, but we could parse one of the other locators?

tests/fixtures/locator.spdx.json
ejortega: I was having trouble finding anything not using PURL. The spec only has references for a few ecosystems. It seems abandoned in favor of PURL.

tests/fixtures/locator.spdx.json
kylewillmon: I just realized that the `"referenceLocator"` doesn't match the spec either...

Accoding to the spec, this should be `"referenceLocator": "org.jruby:jruby-complete:9.3.7.0"`

tests/fixtures/locator.spdx.json
ejortega: We can probably still support it given it's in the spec. I used https://github.com/spdx/tools-python to convert back and forth to help finalize this file (knowing need to update the API output).

lockfile/src/spdx.rs
ejortega: I don't think anything prevents it, so it's technically possible, but you should only have one to identify where the package actually came from?

lockfile/src/spdx.rs
```diff
@@ -76,62 +79,84 @@ fn type_from_url(url: &str) -> anyhow::Result<PackageType> {
...
+            _ => None,
+        })
+        .unwrap_or(PackageVersion::FirstParty(pkg_version.into()));
+
+    Ok(Package { name, version, package_type })
+}
+
+fn from_locator(registry: &str, locator: &str) -> anyhow::Result<Package> {
+    lazy_static! {
+        static ref RE: Regex = Regex::new(r"^(([^:]*:)*[^@:]*@?[^@:/]*)[:@/]([^:]+)$").unwrap();
```
kylewillmon, line 126: This regex has become pretty complicated... It would probably be more readable to just check `package_type` and handle each case. There aren't that many of them and they aren't complicated when taken one at a time.

i.e.,

```rust
    let package_type = PackageType::from_str(registry).map_err(|_| UnknownEcosystem)?;
    let (name, version) = match package_type {
        PackageType::Npm => locator.split_once('@'),
        PackageType::Nuget => locator.split_once('/'),
        PackageType::Maven => locator.rsplit_once(':').filter(|(name, _)| name.contains(':')),
        _ => {
            // Not in the spec, but included for compatibility with our API
            locator.split_once('@')
        },
    }
    .ok_or(anyhow!("invalid locator: {}", locator))?;
```

lockfile/src/spdx.rs
```diff
@@ -76,62 +79,84 @@ fn type_from_url(url: &str) -> anyhow::Result<PackageType> {
...
+    }
+
+    let package_type = PackageType::from_str(registry).map_err(|_| UnknownEcosystem)?;
+    let decoded_locator = decode(locator).unwrap();
```
kylewillmon, line 130: Two questions:

1. Do all SPDX tools use URL encoding? I don't see it in the spec... But the spec also implies that `npm` package names won't contain a `@` (which they often do...)
2. It seems like the URL decode should happen **after** the regex and not before... For example, `%40npmcli/fs@2.1.2` is easier to parse as a locator than `@npmcli/fs@2.1.2`.

kylewillmon (changes requested):

CHANGELOG.md
maxrake, line 1: With the CLI v5.2.0 release that just went out, this file needs to be updated.

maxrake (changes requested):

lockfile/src/spdx.rs
```diff
@@ -538,11 +538,6 @@ mod tests {
         let pkgs = Spdx.parse(include_str!("../../tests/fixtures/locator.spdx.json")).unwrap();
 
         let expected_pkgs = [
-            Package {
-                name: "@npmcli/fs".into(),
-                version: PackageVersion::FirstParty("2.1.2".into()),
-                package_type: PackageType::Npm,
-            },
```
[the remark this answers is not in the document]
ejortega: Oh I accidentally removed it. Added back.

lockfile/src/spdx.rs
ejortega: This is so much simpler. Thanks!

CHANGELOG.md
ejortega: Updated

maxrake: There is still an entry in the 5.2.0 section that needs to be removed.

ejortega: Oops fixed now

maxrake (dismissed):

lockfile/src/spdx.rs
ejortega: 1. I've seen the PURL ones use encoding. I just took a defensive approach.
2. Moved to just decode the name, but I can remove entirely

maxrake:
I have approved the changes I requested. The other reviewers who requested changes should also be given a chance to review.

lockfile/src/spdx.rs
```diff
@@ -76,62 +77,88 @@ fn type_from_url(url: &str) -> anyhow::Result<PackageType> {
...
+        PackageType::Nuget => locator.rsplit_once('/'),
+        PackageType::Maven => locator.rsplit_once(':').filter(|(name, _)| name.contains(':')),
+        _ => {
+            // Not in the spec, but included for compatibility with our API
+            locator.rsplit_once('@')
+        },
+    }
+    .ok_or(anyhow!("Invalid locator: {}", locator))?;
+
+    let name = decode(name).map_err(|e| anyhow!("Error: {:?} when decoding {}", e, name))?;
```
kylewillmon, line 135: Don't throw away the underlying error. This should just add context

```suggestion
    let name = decode(name).with_context(|| anyhow!("URL decode failed: {:?}", name))?;
```

kylewillmon (changes requested):

kylewillmon (approved):

ejortega:
@cd-work Did you have any remaining feedback?

cd-work (approved):

PR #1075, opened 2023-05-05
Fix use of `pip-compile`

kylewillmon:
This patch updates our usage of `pip-compile` to:

- Specify the manifest file to compile
- Handles `requirements.in`, `setup.cfg`, `setup.py`, and `pyproject.toml`

PR #1087, opened 2023-05-09
Improve generator process creation error

### LINKED ISSUE #1084: Poor command not found error in lockfile generator
When the underlying command is not found, the error given is not great.
```
❯ phylum parse
Generating lockfile for manifest "/Users/maxrake/dev/delme/manifest_test/./pip/requirements.txt" using Pip…
❗ Error: failed to run command "cd \"/Users/maxrake/dev/delme/manifest_test/pip\" && \"pip-compile\" \"-o\" \"requirements-locked.txt\" \"/Users/maxrake/dev/delme/manifest_test/./pip/requirements.txt\"": No such file or directory (os error 2)
```

This has a few issues:

1. The `cd <dir> && ` is just noise that hides the actual issue
2. All of the escaped quotes in the string are distracting

A better message could look something like this:
```
❯ phylum parse
Generating lockfile for manifest "/Users/maxrake/dev/delme/manifest_test/./pip/requirements.txt" using Pip…
❗ Error: failed to run command: "pip-compile" "-o" "requirements-locked.txt" "/Users/maxrake/dev/delme/manifest_test/./pip/requirements.txt"

Caused by:
  No such file or directory (os error 2)
```

cd-work:
This patch changes the error generated during lockfile generation when the ecosystem's package manager subprocess could not be spawned successfully.

Instead of printing the full debug formatting of the command, just the program and its argument are printed.

Closes #1084.

---

New error message example:

```
Generating lockfile for manifest "./requirements.txt" using Pip…
❗ Error: failed to spawn command "pip-compile": No such file or directory (os error 2)
```

### DIFF
--- a/lockfile_generator/src/lib.rs
+++ b/lockfile_generator/src/lib.rs
@@ -1,5 +1,4 @@
 use std::ffi::OsString;
-use std::fmt::Write;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 use std::{fs, io};
@@ -52,13 +51,8 @@ pub trait Generator {
 
         // Provide better error message, including the failed program's name.
         let mut child = command.spawn().map_err(|err| {
-            // Collect program and args for error message.
-            let mut cmdline = format!("{:?}", command.get_program());
-            for arg in command.get_args() {
-                let _ = write!(cmdline, " {arg:?}");
-            }
-
-            Error::ProcessCreation(cmdline, err)
+            let program = format!("{:?}", command.get_program());
+            Error::ProcessCreation(program, err)
         })?;
 
         // Ensure generation was successful.
--- a/lockfile_generator/src/lib.rs
+++ b/lockfile_generator/src/lib.rs
@@ -117,7 +117,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 pub enum Error {
     #[error("I/O error: {0}")]
     Io(#[from] io::Error),
-    #[error("failed to run command {0}: {1}")]
+    #[error("failed to spawn command {0}: {1}")]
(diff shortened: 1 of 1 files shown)


lockfile_generator/src/lib.rs
```diff
@@ -50,8 +51,15 @@ pub trait Generator {
...
-        let mut child =
-            command.spawn().map_err(|err| Error::ProcessCreation(format!("{command:?}"), err))?;
+        let mut child = command.spawn().map_err(|err| {
+            // Collect program and args for error message.
+            let mut cmdline = format!("{:?}", command.get_program());
+            for arg in command.get_args() {
+                let _ = write!(cmdline, " {arg:?}");
+            }
+
+            Error::ProcessCreation(cmdline, err)
```
kylewillmon, line 61: After some more thought and a conversation with @furi0us333, I actually think that including the args is unnecessary (and potentially misleading)

```suggestion
            Error::ProcessCreation(format!("{:?}", command.get_program()), err)
```

For example, looking at this error (from this branch):

```
❗ Error: failed to run command "pip-compile" "-o" "requirements-locked.txt": No such file or directory (os error 2)
```

Without context, a user may think "No such file" refers to `requirements-locked.txt` (which indeed does not exist). It's cleaner without the args:

```
❗ Error: failed to run command "pip-compile": No such file or directory (os error 2)
```

And as far as I know, the args will never cause `spawn()` to fail. So I think it's safe to ignore them here.

kylewillmon (changes requested):

maxrake (approved):
The changes here were applied to the `pip-tools` branch (#1075) and tested locally. The results are as advertised.

kylewillmon (approved):

PR #1088, opened 2023-05-10
Fix npm install from other node_modules

kylewillmon:
Some projects install packages from a top level `node_modules` folder, resulting in dependencies like this one:

```
  "typescript": "file:../node_modules/typescript",
```

This was confusing our `package-lock.json` parser and causing an error such as: `Dependency 'typescript' is missing "resolved" key`

_Note_: While this patch fixes the error, this is still a filesystem dependency and will not be analyzed. Projects that use this pattern must ensure that they also submit the top-level `package-lock.json` (which contains details about where the package came from) in order to avoid missing dependencies.

### DIFF
--- a/lockfile/src/javascript.rs
+++ b/lockfile/src/javascript.rs
@@ -53,6 +53,10 @@ impl Parse for PackageLock {
                 // Since we care more about the name of a local dependency than its package, we
                 // discard the version here and include the package later when it's mentioned by
                 // name.
+                if !name.starts_with("node_modules/") {
+                    continue;
+                }
+
                 let name = match name.rsplit_once("node_modules/") {
                     Some((_, name)) => name,
                     None => continue,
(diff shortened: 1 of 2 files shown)


lockfile/src/javascript.rs
```diff
@@ -53,6 +53,10 @@ impl Parse for PackageLock {
                 // Since we care more about the name of a local dependency than its package, we
                 // discard the version here and include the package later when it's mentioned by
                 // name.
+                if !name.starts_with("node_modules/") {
+                    continue;
+                }
+
                 let name = match name.rsplit_once("node_modules/") {
                     Some((_, name)) => name,
                     None => continue,
```
kylewillmon, line 62: This is technically unreachable now, but `continue` is harmless and less scary that `.unwrap()` or `unreachable!()`

maxrake (approved):
The added test case appears to cover the gap and the tests pass with the new code. Since this is a fix, a `CHANGELOG.md` entry might be expected.

cd-work (approved):
This feels a bit scary considering how generic the conditional is, but I think it should work for all packge-lock.json variations I'm familiar with. I can definitely see format changes causing problems that fail silently here in the future though.

PR #1113, opened 2023-05-31
Add reachability analysis

cd-work:
This patch adds support for outputting vulnerability reachability data
when users perform a `phylum analyze`.

Currently the analysis is only performed if the user has the `vulnreach`
role.

By default, the only output is a list with vulnerabilities and whether
they are reachable or not. With verbose output the reachability paths
will also be printed.


cli/src/commands/jobs.rs
```diff
@@ -188,13 +190,61 @@ pub async fn handle_submission(api: &mut PhylumApi, matches: &clap::ArgMatches)
...
+/// Perform vulnerability reachability analysis.
+async fn vulnreach(
+    api: &mut PhylumApi,
+    matches: &clap::ArgMatches,
+    packages: Vec<PackageDescriptor>,
+    job_id: String,
+) -> Result<()> {
+    // Skip requests early if we know user doesn't have the required role.
+    if !api.roles().contains(&RealmRole::Vulnreach) {
+        return Ok(());
```
maxrake, line 215: This is an undocumented pre-requisite. Perhaps a `debug` or `warn` log message would help here?

Local testing was unable to proceed past this point. The early return path was taken for both STAGING and PRODUCTION, both before and after a `phylum auth login` (to update/refresh the refresh token). To be certain, the code was modified to provide a bit more output:

```
    // Skip requests early if we know user doesn't have the required role.
    log::warn!("Checking for `Vulnreach` role in refresh token ...");
    if !api.roles().contains(&RealmRole::Vulnreach) {
        log::warn!("NO `Vulnreach` role in refresh token");
        return Ok(());
    }
    log::warn!("`Vulnreach` role found in refresh token");
```

...and the result was always the same:

```
../phylum-ui  6  18 on  develop [✘?] via 🐳 desktop-linux via  v20.2.0 took 4s
❌ 100 ❯ ~/dev/phylum/localdev/cli/target/debug/phylum analyze -l reacher_test -p delme_ui -v
✅ Successfully parsed lockfile "./yarn.lock" as type: yarn
✅ Job ID: 41ef3aeb-3e5e-4b6e-83d0-ee2f4edf983a
[2023-06-01T01:37:24Z WARN  phylum_cli::commands::jobs] Checking for `Vulnreach` role in refresh token ...
[2023-06-01T01:37:24Z WARN  phylum_cli::commands::jobs] NO `Vulnreach` role in refresh token
# Phylum OSS Supply Chain Risk Analysis - FAILED

This repository analyzes the risk of new dependencies. An
administrator of this repository has set requirements via Phylum policy.

If you see this comment, one or more dependencies have failed Phylum's risk analysis.

### Package: `is-typed-array@1.1.10` failed.

#### is-typed-array@1.1.10 may be a typosquatted package

- **Risk Domain**: Malicious Code
- **Risk Level**: high
- **Reason**: risk level cannot exceed medium

### Package: `yaml@2.1.3` failed.

#### yaml@2.1.3 is vulnerable to Uncaught Exception

- **Risk Domain**: Software Vulnerability
- **Risk Level**: high
- **Reason**: risk level cannot exceed medium

[View this project in the Phylum UI](https://app.phylum.io/projects/42abb122-d6c3-4305-b9e1-3cfe00be87dc?label=reacher_test)
```

(That is PROD, but STG was the same). Are there other requirements for getting vulnreach output? Is there a recommended testing methodology to confirm the behavior in this PR?

maxrake (changes requested):
The code was reviewed but should be examined more closely by @kylewillmon or @andreaphylum .

Local testing was unable to confirm the new behavior added in this PR. Suggestions for how to do so are requested.

cli/src/commands/jobs.rs
kylewillmon: A debug log might make sense, but this should definitely not warn. Access to the vuln reach API is controlled by the role, so skipping this code when the user doesn't have the role is expected behavior.

cd-work: Yeah I'd like to avoid any output here since this is basically our form of "feature detection" and we'd always shove pointless warnings into people's faces if we would log this.

cli/src/auth/jwt.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/// Available Phylum JWT realm roles.
+#[derive(Deserialize, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+pub enum RealmRole {
+    #[serde(rename = "pro-account")]
+    Pro,
+    #[serde(rename = "vulnreach")]
+    Vulnreach,
+    #[serde(rename = "offline-access")]
```
maxrake, line 48:
```suggestion
    #[serde(rename = "offline_access")]
```
A decoded bearer JWT from STAGING was used to confirm this.

maxrake (changes requested):

cli/src/commands/jobs.rs
```diff
@@ -237,3 +299,22 @@ impl JobsProject {
...
+    dependencies: Vec<PackageDescriptor>,
+    imports: Vec<String>,
+    analysis_job_id: String,
+) -> Job {
+    let dependencies = dependencies
+        .into_iter()
+        .map(|package| JobPackage {
+            name: package.name,
+            version: package.version,
+            ecosystem: "npm".into(),
```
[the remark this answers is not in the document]
cd-work: I believe this is currently ignored.

cli/src/auth/jwt.rs
cd-work: I just removed it to be safe, we don't need it anyway and API also doesn't have it iirc.

maxrake (approved):
It all appears okay. Local testing was not fully completed since there are minimal packages available to test with, but since this is now under a feature flag the intent is to do more thorough testing when the flag is removed.

There may still be a CI failure...something to do with mismatched types.

cli/Cargo.toml
```diff
@@ -38,6 +39,9 @@ open = "4.0.0"
 phylum_lockfile = { path = "../lockfile", features = ["generator"] }
 phylum_project = { path = "../phylum_project" }
 phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "development" }
+vuln-reach = { git = "https://github.com/phylum-dev/vuln-reach", optional = true }
+vulnreach_types = { path = "../vulnreach_types", optional = true }
+jsonwebtoken = { version = "8.3.0", optional = true }
```
kylewillmon, line 44: Nothing in the current code uses this crate (unless I've missed something). This should be removed.

cli/src/auth/jwt.rs
```diff
@@ -0,0 +1,79 @@
...
+    #[error("JWT missing signature part")]
+    MissingSignature,
+    #[error("JWT cannot have more than 3 parts")]
+    UnexpectedComponent,
+    #[error("invalid base64: {0}")]
+    InvalidBase64(#[from] DecodeError),
+    #[error("invalid UTF-8: {0}")]
+    InvalidUtf8(#[from] FromUtf8Error),
+    #[error("invalid JSON: {0}")]
+    InvalidJson(#[from] JsonError),
```
kylewillmon, line 66: All of these will double print the inner error (since it's included in both `Display` and `.source()`). It should only be in one or the other. Personally, I prefer `.source()`.

```suggestion
    #[error("invalid base64")]
    InvalidBase64(#[from] DecodeError),
    #[error("invalid UTF-8")]
    InvalidUtf8(#[from] FromUtf8Error),
    #[error("invalid JSON")]
    InvalidJson(#[from] JsonError),
```

kylewillmon (changes requested):

cli/src/auth/jwt.rs
```diff
@@ -0,0 +1,79 @@
...
+use base64::prelude::*;
+use base64::DecodeError;
+use serde::Deserialize;
+use serde_json::Error as JsonError;
+use thiserror;
+
+/// Get user roles from a bearer token without performing validation.
+pub fn user_roles(bearer: &str) -> Result<HashSet<RealmRole>, JwtError> {
+    // Extract the base64 payload.
+    let (_, payload_base64, _) = parts(bearer)?;
```
maxrake, line 15: If the signature is thrown away here and no validation performed, couldn't someone update the payload to claim any role they wanted? That may not matter for `vulnreach` in particular, but this code could easily be relied on to restrict functionality by role for other/future features.

kylewillmon: > but this code could easily be relied on to restrict functionality by role for other/future features

CLI will never use roles as a security control because CLI is an open source client. The only purpose for looking at user roles in CLI is optimization (to avoid making requests that we know will fail).

And there is little risk of anyone copying this code to use it in a place where security actually matters. First, because the function comment clearly states "without performing validation". And second, because we already have an internal crate dedicated to this purpose of validating these tokens.

kylewillmon (approved):

maxrake (approved):
Tests are passing now...LGTM.

PR #1123, opened 2023-06-15
Fix analysis report output

### LINKED ISSUE #1116: Terminal friendly job report
The [/policy/evaluate/raw][1] endpoint is now available on staging. This should allow the CLI to create a more friendly report for job results.

Additionally, we can include more or less details with or without the `-v` option

[1]: https://api.staging.phylum.io/api/v0/swagger/index.html#/Jobs/jobs_evaluate_policy_raw

cd-work:
This patch uses the new raw policy endpoint to fix the output of the analysis command to be properly formatted for a terminal emulator rather than just printing the default markdown.

Closes #1116.

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -29,7 +29,10 @@ use crate::auth::{
     fetch_oidc_server_settings, handle_auth_flow, handle_refresh_tokens, jwt, AuthAction, UserInfo,
 };
 use crate::config::{AuthInfo, Config};
-use crate::types::{HistoryJob, PingResponse, PolicyEvaluationRequest, PolicyEvaluationResponse};
+use crate::types::{
+    HistoryJob, PingResponse, PolicyEvaluationRequest, PolicyEvaluationResponse,
+    PolicyEvaluationResponseRaw,
+};
 
 pub mod endpoints;
 
@@ -301,6 +304,16 @@ impl PhylumApi {
     ) -> Result<PolicyEvaluationResponse> {
         let body = PolicyEvaluationRequest { ignored_packages: ignored.into() };
         self.post(endpoints::get_job_status(&self.config.connection.uri, job_id)?, body).await
+    }
+
+    /// Get the status of a previously submitted job.
+    pub async fn get_job_status_raw(
+        &self,
+        job_id: &JobId,
+        ignored: impl Into<Vec<PackageDescriptor>>,
+    ) -> Result<PolicyEvaluationResponseRaw> {
+        let body = PolicyEvaluationRequest { ignored_packages: ignored.into() };
+        self.post(endpoints::get_job_status_raw(&self.config.connection.uri, job_id)?, body).await
     }
 
     /// Check a set of packages against the default policy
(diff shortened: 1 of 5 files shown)


cli/src/api/mod.rs
```diff
@@ -684,6 +697,28 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn get_job_status_raw() -> Result<()> {
+        let body = todo!();
```
kylewillmon, line 702: WIP?

cd-work: Forgot to push the latest changes, on second.

cli/src/format.rs
```diff
@@ -95,6 +95,65 @@ impl Format for PolicyEvaluationResponse {
...
+        // Write summary for each issue.
+        for package in &self.dependencies {
+            for rejection in package.rejections.iter().filter(|rejection| !rejection.suppressed) {
+                let severity = match rejection.source.domain.as_str() {
+                    "author" => "AUT",
+                    "engineering" => "ENG",
+                    "malicious_code" | "malicious" => "MAL",
+                    "vulnerability" => "VUL",
+                    "license" => "LIC",
+                    _ => "UNK",
```
kylewillmon, line 135: I think it would be safer to use the underlying value if we find something we're not aware of:

```suggestion
                    domain => domain,
```

I don't think that will cause lifetime issues... But I haven't compiled it...

cli/src/format.rs
```diff
@@ -95,6 +95,65 @@ impl Format for PolicyEvaluationResponse {
...
+                };
+                let message = format!("[{severity}] {}", rejection.title);
```
kylewillmon, line 137: Hmm... You're counting on `rejection.title` to contain the name of the package...

That's definitely not guaranteed for policy-based rejections. But it seems to be true for every issue-based rejection I can find...

We probably need to discuss this with @louislang before relying on it.

cli/src/format.rs
```diff
@@ -95,6 +95,65 @@ impl Format for PolicyEvaluationResponse {
...
+
+                let colored = match rejection.source.severity.as_str() {
+                    "low" | "info" => style(message).green(),
```
kylewillmon, line 140: Severity is weird...

If the policy rejects a `low` or `info` issue, the result is a failed job. Printing that in green seems counterintuitive.

In an ideal world, I think the policy should tell us the severity. But we're not there yet... It might be better to simply print every rejection in red since that's what we've got right now.

kylewillmon (changes requested):

cli/src/format.rs
```diff
@@ -95,6 +95,65 @@ impl Format for PolicyEvaluationResponse {
...
+            let _ = writeln!(writer, "{}", textwrap::fill(&incomplete_message, 80));
+        }
+
+        // Write summary for each issue.
+        for package in &self.dependencies {
+            for rejection in package.rejections.iter().filter(|rejection| !rejection.suppressed) {
+                let severity = match rejection.source.domain.as_str() {
+                    "author" => "AUT",
+                    "engineering" => "ENG",
+                    "malicious_code" | "malicious" => "MAL",
```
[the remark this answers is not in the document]
cd-work: I just copied the old code. I can force just the new version if we want to.

cli/src/format.rs
cd-work: It will completely mess up formatting though.

cli/src/format.rs
cd-work: I just don't want everything duplicated. The UI currently does the same. This PR's format is inspired by what the UI shows at the moment. I'm happy to change things in the future.

cli/src/format.rs
cd-work: The UI also shows things as green even when they're technically an "issue".

I think this is just consistent with the rest of our platform.

cli/src/format.rs
kylewillmon: I believe that will change when the UI is updated to use this endpoint. And maybe the titles will no longer contain the package names?

kylewillmon: I believe that will change when the UI is updated to use this endpoint. And maybe the titles will no longer contain the package names?

cli/src/format.rs
louislang: We're working on changing this in the UI atm. Basically a transition to triaging at the package level, instead of issue level. So I'd say we should do whatever is right here, and let the rest of the platform catch up.

kylewillmon: After discussing with @louislang, I'm going to resolve this conversation and we can leave this as is.

There is still quite a bit of uncertainty about how this will look when the UI is updated. So matching the UI today and tweaking this code if/when the UI changes is a good strategy moving forward.

kylewillmon (approved):

kylewillmon (approved):

PR #1126, opened 2023-06-19
Fix line continuations and --hash in python parser

### LINKED ISSUE #1117: pip lockfiles with hashes are treated as manifest files
# Overview
When parsing a python lockfile generated with `pip-compile --generate-hashes` the Phylum CLI treats it as a manifest file and attempts to generate a lockfile. This extra step adds more places for failure and isn't needed as the provided file is already a lockfile.

# How To Reproduce
1. Create a simple loose requirements file `echo flask > requirements.in`
2. Use `pip-compile` to generate the lockfile with hashes `pip-compile --generate-hashes requirements.in`
3. Parse the resultant lockfile `phylum parse requirements.txt`
4. See the Phylum CLI generate a lockfile

# Expected Behavior
Since the resultant file is a lockfile, I would expect the Phylum CLI tool to use it and not attempt to generate another one.

# Additional Context
While this may not seem like much of an issue, it does create additional places for things to go wrong. Two examples were observed so far.
1. A user attempted to analyze one of these files via `phylum-ci` using the current `v.5.2.0` and did not have `pip-tools` installed on their runner causing a failure. This should be solved by the deprecation of the `pip-tools` requirement in `v.5.3.0-rc1` (once a full release is made), but was an unnecessary failure.
2. When testing this I ran into a `pip` issue. While this isn't a Phylum issue, it does break the parse command and the generation step is unnecessary, so this error should be avoidable.

cd-work:
This fixes the handling of line continuations in the requirements.txt file.

It also resolves an issue where the `--hash` argument was not accepted by the lockfile parser and caused the parsing of the file to fail.

Closes #1117.

### DIFF
--- a/lockfile/src/parsers/mod.rs
+++ b/lockfile/src/parsers/mod.rs
@@ -28,7 +28,7 @@ fn take_till_blank_line(input: &str) -> Result<&str, &str> {
 ///
 /// This supports both `\n` and `\r\n`. It also skips line continuations
 /// (`\\\n`, `\\\r\n`) and stops on EOF.
-fn take_continued_line(mut input: &str) -> Result<&str, &str> {
+fn take_continued_line(mut input: &str) -> Result<&str, ()> {
     loop {
         // Get everything up to the next NL or EOF.
         let (new_input, line) = recognize(alt((take_till_line_end, rest)))(input)?;
@@ -40,7 +40,7 @@ fn take_continued_line(mut input: &str) -> Result<&str, &str> {
         }
     }
 
-    Ok((input, ""))
+    Ok((input, ()))
 }
 
 type Result<T, U> = IResult<T, U, VerboseError<T>>;
--- a/lockfile/src/parsers/mod.rs
+++ b/lockfile/src/parsers/mod.rs
@@ -1,10 +1,10 @@
 use nom::branch::alt;
-use nom::bytes::complete::{tag, take, take_until};
-use nom::character::complete::{anychar, line_ending, multispace0, space0};
-use nom::combinator::{eof, opt, recognize};
+use nom::bytes::complete::{tag, take_until};
+use nom::character::complete::{line_ending, multispace0, not_line_ending, space0};
+use nom::combinator::{eof, opt, recognize, rest};
 use nom::error::{context, VerboseError};
 use nom::multi::{count, many1, many_till};
-use nom::sequence::{delimited, tuple};
+use nom::sequence::{delimited, terminated, tuple};
 use nom::{AsChar, IResult};
 
 pub mod gem;
@@ -16,17 +16,12 @@ pub mod yarn;
 
 /// Consume everything until the next `\n` or `\r\n`.
 fn take_till_line_end(input: &str) -> Result<&str, &str> {
-    recognize(tuple((alt((take_until("\n"), take_until("\r\n"))), take(1usize))))(input)
+    recognize(terminated(not_line_ending, line_ending))(input)
 }
 
 /// Consume everything until the next `\n\n` or `\r\n\r\n`.
 fn take_till_blank_line(input: &str) -> Result<&str, &str> {
     recognize(alt((take_until("\n\n"), take_until("\r\n\r\n"))))(input)
-}
-
-/// Consume everything until the input is empty.
-fn take_till_eof(input: &str) -> Result<&str, &str> {
-    recognize(many_till(anychar, eof))(input)
 }
 
 /// Consume the next line.
@@ -36,7 +31,7 @@ fn take_till_eof(input: &str) -> Result<&str, &str> {
 fn take_continued_line(mut input: &str) -> Result<&str, &str> {
     loop {
         // Get everything up to the next NL or EOF.
-        let (new_input, line) = recognize(alt((take_till_line_end, take_till_eof)))(input)?;
+        let (new_input, line) = recognize(alt((take_till_line_end, rest)))(input)?;
         input = new_input;
 
         // Stop consuming lines once there are no continuations.
--- a/lockfile/src/parsers/mod.rs
+++ b/lockfile/src/parsers/mod.rs
@@ -1,6 +1,6 @@
 use nom::branch::alt;
 use nom::bytes::complete::{tag, take, take_until};
-use nom::character::complete::{line_ending, multispace0, space0};
+use nom::character::complete::{anychar, line_ending, multispace0, space0};
 use nom::combinator::{eof, opt, recognize};
 use nom::error::{context, VerboseError};
 use nom::multi::{count, many1, many_till};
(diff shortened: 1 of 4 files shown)


lockfile/src/python.rs
```diff
@@ -33,7 +33,10 @@ fn is_requirements_file(path: &Path) -> bool {
 impl Parse for PyRequirements {
     /// Parses `requirements.txt` files into a vec of packages
     fn parse(&self, data: &str) -> anyhow::Result<Vec<Package>> {
-        let (_, entries) = pypi::parse(data)
+        // Ignore newlines with line continuation characters.
+        let input = data.replace(" \\\n", " ").replace(" \\\r\n", " ");
```
cd-work, line 37: This is absolutely atrocious and I hate that I'm even suggesting this. But the only alternative to this is to rewrite the entire main loop of the Python parser to not split on ` \\\n` (probably by trying to use nom for this), then pass these full strings to our parsing functions and replace `space0` and `space1` with a function that treats ` \\\n` as normal whitespace.

At least that's the only alternative I could come up with and it wasn't something I'd like to implement to fix this issue (though maybe I should?).

maxrake (changes requested):
Local testing showed success but the tests in CI are showing a panic around some of the `nom` parsing code:

```
---- commands::parse::tests::it_can_identify_lock_file_types stdout ----
thread 'commands::parse::tests::it_can_identify_lock_file_types' panicked at 'range end index 669810688 out of range for slice of length 52425'
```

maxrake (approved):
The code was reviewed but not fully understood. The changes were confirmed to work with local testing. LGTM.

lockfile/src/parsers/pypi.rs
```diff
@@ -2,36 +2,62 @@ use std::path::PathBuf;
...
+        let (new_input, line) = line(input)?;
+        input = new_input;
+
+        // Ignore empty lines.
+        if line.is_empty() {
+            continue;
+        }
+
+        // Strip comments.
+        let (_, line) = recognize(alt((take_until(" #"), parsers::take_till_eof)))(line)?;
```
kylewillmon, line 30: 1. I think `recognize` is unnecessary here since `alt` should already return `&str` in this case
2. As mentioned in another comment, you can use `rest`
   ```suggestion
           let (_, line) = alt((take_until(" #"), rest))(line)?;
   ```
3. This may be a case where it's simpler to do without `nom`
   ```suggestion
           let line = line.split(" #").next().unwrap();
   ```

lockfile/src/parsers/pypi.rs
```diff
@@ -2,36 +2,62 @@ use std::path::PathBuf;
...
 
     Ok((input, pkgs))
 }
 
+/// Parse one line in the lockfile.
+fn line(input: &str) -> Result<&str, &str> {
+    // Take everything until the next newline.
+    //
+    // This takes line continuation characters into account.
+    let (input, mut line) = recognize(parsers::take_continued_line)(input)?;
```
kylewillmon, line 45: You shouldn't need `recognize` here.

```suggestion
    let (input, mut line) = parsers::take_continued_line(input)?;
```

lockfile/src/parsers/pypi.rs
```diff
@@ -2,36 +2,62 @@ use std::path::PathBuf;
...
+        line = "";
+    }
+
+    Ok((input, line))
+}
+
 fn package(input: &str) -> Result<&str, Package> {
     // Ignore everything after `;`.
-    let (_, input) = recognize(alt((take_until(";"), not_line_ending)))(input)?;
+    let (_, input) = recognize(alt((take_until(";"), parsers::take_till_eof)))(input)?;
```
kylewillmon, line 60: Another unnecessary `recognize`

```suggestion
    let (_, input) = alt((take_until(";"), parsers::take_till_eof))(input)?;
```

lockfile/src/parsers/pypi.rs
```diff
@@ -151,16 +177,56 @@ fn identifier_list(input: &str) -> Result<&str, &str> {
...
+/// Newline-aware space1.
+///
+/// This automatically handles " \\\n" and treats it as normal space.
+fn nl_space1(input: &str) -> Result<&str, &str> {
+    recognize(many1(alt((space1, line_continuation))))(input)
+}
+
+/// Recognize line continuations.
+fn line_continuation(input: &str) -> Result<&str, &str> {
+    recognize(tuple((tag("\\"), opt(tag("\r")), tag("\n"))))(input)
```
kylewillmon, line 231: It doesn't make much difference, but this could be written with `line_ending`

```suggestion
    recognize(tuple((tag("\\"), line_ending)))(input)
```

kylewillmon (changes requested):

lockfile/src/parsers/mod.rs
```diff
@@ -14,12 +14,38 @@ pub mod pypi;
 pub mod spdx;
 pub mod yarn;
 
+/// Consume everything until the next `\n` or `\r\n`.
 fn take_till_line_end(input: &str) -> Result<&str, &str> {
     recognize(tuple((alt((take_until("\n"), take_until("\r\n"))), take(1usize))))(input)
```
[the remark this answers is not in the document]
kylewillmon: OH! I see now why this isn't a bug....

The `take_until("\r\n")` is unreachable... If the input contains `"\r\n"`, it also contains the substring `"\n"`, which means the first side of the `alt` (`take_until("\n")`) succeeds and we only need to `take(1)` to consume the `"\n"`...

So the code will never malfunction, but it's still very misleading...

lockfile/src/parsers/pypi.rs
cd-work: I do need recognize here because the function itself isn't actually capable of finding the entire line. If you look at it you can see that it just throws everything away.

To my knowledge there isn't another way to do it, since we skip over it in chunks.

lockfile/src/parsers/mod.rs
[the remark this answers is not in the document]
cd-work: Working exactly how I planned it. :sweat_smile:

lockfile/src/parsers/pypi.rs
kylewillmon: Ah. I hadn't noticed that...

If `take_continued_line` isn't returning anything, that should be reflected in the return type (i.e., `Result<&str, ()>`). I'll add suggestions to the appropriate lines.

lockfile/src/parsers/mod.rs
```diff
@@ -14,12 +14,33 @@ pub mod pypi;
...
+/// Consume everything until the next `\n\n` or `\r\n\r\n`.
 fn take_till_blank_line(input: &str) -> Result<&str, &str> {
     recognize(alt((take_until("\n\n"), take_until("\r\n\r\n"))))(input)
 }
 
+/// Consume the next line.
+///
+/// This supports both `\n` and `\r\n`. It also skips line continuations
+/// (`\\\n`, `\\\r\n`) and stops on EOF.
+fn take_continued_line(mut input: &str) -> Result<&str, &str> {
```
kylewillmon, line 31:
```suggestion
fn take_continued_line(mut input: &str) -> Result<&str, ()> {
```

lockfile/src/parsers/mod.rs
```diff
@@ -14,12 +14,33 @@ pub mod pypi;
...
+        let (new_input, line) = recognize(alt((take_till_line_end, rest)))(input)?;
+        input = new_input;
+
+        // Stop consuming lines once there are no continuations.
+        if !line.ends_with("\\\n") && !line.ends_with("\\\r\n") {
+            break;
+        }
+    }
+
+    Ok((input, ""))
```
kylewillmon, line 43:
```suggestion
    Ok((input, ()))
```

kylewillmon (changes requested):

lockfile/src/parsers/mod.rs
cd-work: Yeah that's a great idea.

kylewillmon (approved):

PR #1129, opened 2023-06-21
Rename parsers::Result to parsers::IResult

kylewillmon:
Calling this type `Result` leads to confusion because it has two generic parameters just like `std::result::Result`. But the parameters are for input and output, not ok and error.

Using `IResult` is less confusing.

### DIFF
--- a/lockfile/src/parsers/yarn.rs
+++ b/lockfile/src/parsers/yarn.rs
@@ -1,9 +1,10 @@
+use nom::InputTakeAtPosition;
 use phylum_types::types::package::PackageType;
 
 use super::*;
 use crate::{Package, PackageVersion};
 
-pub fn parse(input: &str) -> Result<&str, Vec<Package>> {
+pub fn parse(input: &str) -> IResult<&str, Vec<Package>> {
     let (input, _) = yarn_lock_header(input)?;
 
     // Ignore empty lockfiles.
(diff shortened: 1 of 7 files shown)


lockfile/src/parsers/yarn.rs
```diff
@@ -39,24 +40,20 @@ fn parse_entry(input: &str) -> Result<&str, Package> {
...
     let version_key = tuple((tag(r#"version"#), opt(tag(r#"""#)), tag(r#" ""#)));
     context("version", delimited(version_key, is_version, tag(r#"""#)))(i)
 }
 
-fn is_version<T, E: nom::error::ParseError<T>>(input: T) -> IResult<T, T, E>
-where
-    T: nom::InputTakeAtPosition,
-    <T as nom::InputTakeAtPosition>::Item: AsChar,
-{
+fn is_version(input: &str) -> IResult<&str, &str> {
```
kylewillmon, line 56: This was the only place that we were using `nom::IResult` directly instead of using our alias. But it wasn't necessary, so I changed it to be like the rest of our code and use the alias.

Why was it written that way if it wasn't necessary? My bet is that the person who wrote this function just copied [`alphanumeric1`](https://docs.rs/nom/7.1.3/src/nom/character/complete.rs.html#583-589) and tweaked the body to fit our use case.

cd-work (approved):

maxrake (approved):
LGTM

PR #1130, opened 2023-06-21
Add missed CHANGELOG entries

kylewillmon:
These CHANGELOG entries are for changes made in b78793e and ee167b3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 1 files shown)


maxrake (approved):
LGTM.

This is a recurring issue (missing changelog entries). Perhaps we should be more deliberate in ensuring the PR template "Checklist" is complete...maybe by keeping that checklist on the PR so reviewers can see what is and isn't done.

cd-work:
> This is a recurring issue (missing changelog entries). Perhaps we should be more deliberate in ensuring the PR template "Checklist" is complete

I feel like this is just an issue due to inconsistencies between projects and getting used to things.

I worked on vulnreach mainly recently where there are no changelogs, so that's probably why I forgot.

PR #1132, opened 2023-06-22
Add label to extension analyze API

cd-work:
This adds the new `label` option as the last optional parameter of the `analyze` function, allowing `phylum-ci` to use the extension API for its analysis.

### DIFF
--- a/extensions/ci/main.ts
+++ b/extensions/ci/main.ts
@@ -1,0 +1,37 @@
+import { PhylumApi } from 'phylum';
+
+// Ensure required arguments are present.
+if (Deno.args.length < 4 || Deno.args.length > 5) {
+  console.error("Usage: phylum ci <PROJECT> [--group <GROUP>] <LABEL> <BASE> <LOCKFILE>");
+  Deno.exit(1);
+}
+
+// Find optional groups argument.
+let group = undefined;
+let groupArgsIndex = Deno.args.indexOf("--group");
+if (groupArgsIndex != -1) {
+    const groupArgs = Deno.args.splice(groupArgsIndex, groupArgsIndex);
+    group = groupArgs[1];
+}
+
+// Parse remaining arguments.
+const project = Deno.args[0];
+const label = Deno.args[1];
+const base = Deno.args[2];
+const lockfile = Deno.args[3];
+
+// Parse new lockfile.
+const lockfileDeps = await PhylumApi.parseLockfile(lockfile);
+
+// Deserialize base dependencies.
+const baseDepsJson = await Deno.readTextFile(base);
+const baseDeps = JSON.parse(baseDepsJson);
+
+// Submit analysis job.
+const jobID = await PhylumApi.analyze(lockfileDeps.packages, project, group, label);
+
+// Get analysis job results.
+const jobStatus = await PhylumApi.getJobStatus(jobID, baseDeps);
+
+// Output results as JSON.
+console.log(JSON.stringify(jobStatus));
--- a/extensions/phylum.ts
+++ b/extensions/phylum.ts
@@ -88,6 +88,7 @@ export class PhylumApi {
    * @param packages - List of packages to analyze
    * @param project - Project name. If undefined, the `.phylum_project` file will be used
    * @param group - Group name
+   * @param label - Analysis label for this job
    *
    * @returns Analyze Job ID, which can later be queried with `getJobStatus`.
    */
@@ -95,6 +96,7 @@ export class PhylumApi {
     packages: Package[],
     project?: string,
     group?: string,
+    label?: string,
   ): Promise<string> {
     return DenoCore.opAsync(
       "analyze",
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -140,6 +140,7 @@ async fn analyze(
     packages: Vec<PackageDescriptor>,
     project: Option<String>,
     group: Option<String>,
+    label: Option<String>,
 ) -> Result<JobId> {
     let state = ExtensionState::from(op_state);
     let api = state.api().await?;
@@ -155,7 +156,7 @@ async fn analyze(
         },
     };
 
-    let job_id = api.submit_request(&packages, project, None, group.map(String::from)).await?;
+    let job_id = api.submit_request(&packages, project, label, group.map(String::from)).await?;
 
     Ok(job_id)
 }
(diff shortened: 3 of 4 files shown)


cd-work:
I plan to remove the extension itself before merging this, since it's probably better to have it in the `phylum-ci` repository. However this should make testing easier.

extensions/phylum.ts
```diff
@@ -88,13 +88,15 @@ export class PhylumApi {
...
    * @param group - Group name
+   * @param label - Analysis label for this job
    *
    * @returns Analyze Job ID, which can later be queried with `getJobStatus`.
    */
   static analyze(
     packages: Package[],
     project?: string,
     group?: string,
+    label?: string,
```
maxrake, line 99: This needs to also be added to the `DenoCore.opAsync()` call below.

extensions/ci/main.ts
```diff
@@ -0,0 +1,44 @@
...
+if (Deno.args.length < 4 || Deno.args.length > 5) {
+  console.error(
+    "Usage: phylum ci <PROJECT> [--group <GROUP>] <LABEL> <BASE> <LOCKFILE>",
+  );
+  Deno.exit(1);
+}
+
+// Find optional groups argument.
+let group = undefined;
+let groupArgsIndex = Deno.args.indexOf("--group");
```
maxrake:
```suggestion
const groupArgsIndex = Deno.args.indexOf("--group");
```
To satisfy `deno lint`.

extensions/ci/main.ts
```diff
@@ -0,0 +1,44 @@
+import { PhylumApi } from "phylum";
+
+// Ensure required arguments are present.
+if (Deno.args.length < 4 || Deno.args.length > 5) {
+  console.error(
+    "Usage: phylum ci <PROJECT> [--group <GROUP>] <LABEL> <BASE> <LOCKFILE>",
```
maxrake, line 6: The `phylum analyze` command, and the [current implementation of `phylum-ci`](https://github.com/phylum-dev/phylum-ci/blob/main/src/phylum/ci/ci_base.py#L411-L445), allows for multiple lockfiles to be specified. Can this (sample) extension be modified to also allow for multiple lockfiles? Javascript/Typescript is not something I'm terribly familiar with.

maxrake (changes requested):
Local testing has not been performed yet...but it will be if/when the suggested changes are made.

extensions/phylum.ts
cd-work: Already is.

extensions/ci/main.ts
cd-work: Already done.

extensions/ci/main.ts
cd-work: Should be trivial, yeah.

extensions/ci/main.ts
```diff
@@ -0,0 +1,55 @@
...
+    "Usage: phylum ci <PROJECT> [--group <GROUP>] <LABEL> <BASE> <LOCKFILE...>",
+  );
+  Deno.exit(1);
+}
+
+// Find optional groups argument.
+let group = undefined;
+const groupArgsIndex = args.indexOf("--group");
+if (groupArgsIndex != -1) {
+  const groupArgs = args.splice(groupArgsIndex, groupArgsIndex + 1);
```
maxrake, line 16:
```suggestion
  const groupArgs = args.splice(groupArgsIndex, 2);
```
It would work as written when `--group` was the second argument (i.e., `groupArgsIndex == 1`)
This does not need to be changed in this review since this entire file is expected to be removed...just pointing it out in case it helps anyone else reading this.

maxrake (approved):
The changes were tested locally and found to work with `phylum-ci` after a few changes were made in that repo. The real trick is making a `ci` extension available in the `phylum-ci` Docker image.

The current method for making `phylum` available in the image relies on the custom/hidden `--global-install` option. Something similar may have to be done for this extension since the expectation is that using `phylum extension install` will result in an extension installed in the `root` user's directory and then not be available to non-root users.

Perhaps the extension could be copied manually to a globally accessible directory (like the binary is) and then accessed with a `phylum extension run` command instead of a `phylum ci` command.

cd-work:
> Something similar may have to be done for this extension since the expectation is that using phylum extension install will result in an extension installed in the root user's directory and then not be available to non-root users.

Can't you just install it as root?

extensions/ci/main.ts
maxrake: Yes...the position of `--group` is not fixed. This is actually how this bug was discovered. The inclusion of this extension in `phylum-ci` for testing was done such that the `phylum-ci` code called it as `phylum ci <PROJECT> <LABEL> [--group <GROUP>] <BASE> <LOCKFILE...>` (moved `<LABEL>` to come before `--group`).

The [docs for `.splice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) show the syntax for a two argument call as `splice(start, deleteCount)`, where `deleteCount` means:

> An integer indicating the number of elements in the array to remove from `start`.

So, in the example above, `start` would be the value of `groupArgsIndex`, which is 2 and then `deleteCount` would be (2+1), which is 3 and one more than expected.

maxrake:
> Can't you just install it as root?

Maybe...if the permissions can also be changed to allow access by non-root users. There are some environments...like Azure Pipelines...where a container is created from the image with a non-root user. From [previous issue covering this](https://github.com/phylum-dev/phylum-ci/issues/118):

> This is a problem for some CI environments. For instance, Azure Pipelines creates containers from a given image with a user named `vsts_azpcontainer` and an id of 1001, in a group named `azure_pipelines_sudo`. Tasks/scripts/commands run from that container are done so with this user, which doesn't have access to the `/root/.local/bin` directory where the `phylum` Python package script entry points are located.

extensions/ci/main.ts
cd-work: Oh, duh.

maxrake (approved):
The example `ci` extension has been tested and shown to work in the `phylum-ci` repo. It can be removed from this PR now if that is still the plan. The rest of the code changes in this PR are desired in a release candidate for use in a new `phylum-ci` release.

cd-work:
I've moved the extension to https://github.com/phylum-dev/phylum-ci/pull/267 for now.

maxrake (approved):
LGTM.

PR #1133, opened 2023-06-23
Disable lockfile generation for yarn v1

### LINKED ISSUE #1086: Lockfile generation install packages for yarn v1
When running yarn on a v1 project, the `--mode=update-lockfile` argument is ignored and packages are actually installed during lockfile generation.

See the outputs here:

<details>
<summary>yarn v1</summary>

```
> ls                 
index.js     package.json yarn.lock

> phylum parse -t yarn package.json  
Generating lockfile for manifest "package.json" using Yarn…
[
  {
    "name": "lodash",
    "version": "4.17.21",
    "type": "npm"
  }
]                                                                                                                                                                                                                                                                                        

> ls
index.js     node_modules package.json yarn.lock
```
</details>

<details>
<summary>yarn v2</summary>

```
> yarn set version berry
➤ YN0000: Retrieving https://repo.yarnpkg.com/3.5.1/packages/yarnpkg-cli/bin/yarn.js
➤ YN0000: Saving the new release in .yarn/releases/yarn-3.5.1.cjs
➤ YN0000: Done in 0s 555ms

> ls                               
index.js     package.json yarn.lock

> phylum parse -t yarn package.json
Generating lockfile for manifest "package.json" using Yarn…
[
  {
    "name": "lodash",
    "version": "4.17.21",
    "type": "npm"
  }
]                                                                                                                                                                                                                                                                                        

> ls                               
index.js     package.json yarn.lock
```
</details>

We need to find a way to detect this to avoid installing packages before analysis has run.

cd-work:
Since yarn v1 isn't able to generate a lockfile without installing the dependencies and executing their build code, it could be dangerous to automatically generate lockfiles for the user.

To fix this issue, the lockfile generator will now return an error when lockfile generation is attempted with the yarn generator and only version 1 is used by the project.

While the lockfile generation with yarn v1 still worked and it would provide protection against vulnerable code in the target executable, it is safer to leave the lockfile generation up to the user so they're in full control.

Closes #1086.

### DIFF
--- a/lockfile_generator/src/yarn.rs
+++ b/lockfile_generator/src/yarn.rs
@@ -1,9 +1,10 @@
 //! JavaScript yarn ecosystem.
 
+use std::fs;
 use std::path::Path;
 use std::process::Command;
 
-use crate::Generator;
+use crate::{Error, Generator, Result};
 
 pub struct Yarn;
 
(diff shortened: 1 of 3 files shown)


maxrake (approved):
LGTM.

cd-work:
I'd like to wait for @kylewillmon before merging this. Since I'm not sure if just removing support completely was his ideal outcome for this issue.

kylewillmon (approved):

lockfile_generator/src/yarn.rs
```diff
@@ -21,4 +22,30 @@ impl Generator for Yarn {
     fn tool(&self) -> &'static str {
         "Yarn"
     }
+
+    fn check_prerequisites(&self, manifest_path: &Path) -> Result<()> {
+        let yarn_version = yarn_version(manifest_path)?;
+        if yarn_version.starts_with("1.") {
+            let version = yarn_version.trim().into();
+            return Err(Error::UnsupportedCommandVersion("yarn", "2.0.0+", version));
```
kylewillmon, line 30: The error message here becomes something like:

```
unsupported yarn version 1.22.19, expected 2.0.0+
```

We are leaving it up to the user to figure out how to upgrade. Which isn't trivial because `yarn` is complicated...

But also, I don't understand `yarn` well enough to know what to suggest here... So I guess this will have to do.

kylewillmon:
> I'd like to wait for @kylewillmon before merging this. Since I'm not sure if just removing support completely was his ideal outcome for this issue.

This is exactly what I had in mind for #1086. We can always add yarn v1 support later if requested.

lockfile_generator/src/yarn.rs
cd-work: I feel like "how to update yarn from v1 to v2" is something that should be trivial to find online with extremely detailed information on how to go about it.

So I don't think it's going to be worth it to try and provide our own documentation here.

PR #1136, opened 2023-06-27
Filter setup.py with pyproject.toml present

### LINKED ISSUE #1102: Python projects have multiple manifests
Now that we have expanded our Python manifest support, many (or perhaps even most) Python projects have multiple lockfiles and manifest files, which results in odd behavior when using automatic lockfile detection.

For example, let's look at the [requests][1] project:

[1]: https://github.com/psf/requests

```
> rm requirements-dev.txt 

> phylum status
Project: null
Group: null
Project Root: null
Lockfiles:
 - path: ./docs/requirements.txt
   type: pip
 - path: ./pyproject.toml
   type: poetry
 - path: ./setup.py
   type: pip
 - path: ./setup.cfg
   type: pip
```

_Note: I first remove `requirements-dev.txt` because it gets picked up as a lockfile, which prevents the manifests from being included... This is a separate mistake and possibly deserves its own issue_

Phylum has picked up 3 separate manifests... `pyproject.toml`, `setup.py`, and `setup.cfg`. But because of some details, all three of these are effectively the same manifest.

<details>
<summary>Details for those that care</summary>

1. The `pyproject.toml` takes priority. But according to [PEP 517][]:
   > If the `pyproject.toml` file is absent, or the `build-backend` key is missing, the source tree is not using this specification, and tools should revert to the legacy behaviour of running `setup.py` (either directly, or by implicitly invoking the `setuptools.build_meta:__legacy__` backend).
2. The build system invokes `setup.py`, which calls the `setup()` function in [`setuptools`][setuptools].
3. `setuptools` [detects][] `setup.cfg` and uses these values to build the package

[PEP 517]: https://peps.python.org/pep-0517/
[setuptools]: https://setuptools.pypa.io/en/latest/index.html
[detects]: https://setuptools.pypa.io/en/latest/userguide/declarative_config.html

</details>

cd-work:
This patch changes the behavior for loading Python manifest files by ignoring `setup.py` files when a `pyproject.toml` is present at or above the `setup.py`'s directory.

This is in line with what pip will do by itself when loading manifest files, ensuring that our behavior is consistent with pip and we do not unnecessarily generate the same lockfile twice.

Closes #1102.

---

One noteworthy thing about this patch is that it ignores `setup.py` even if a `pyproject.toml` is present in the parent directory. However currently we do not ignore a `setup.py` if another `setup.py` is present in the parent directory. So while we do this for lockfiles already, it's unique for manifest files.

### DIFF
--- a/lockfile/src/lib.rs
+++ b/lockfile/src/lib.rs
@@ -270,8 +270,8 @@ pub fn find_lockfiles_at(root: impl AsRef<Path>) -> Vec<(PathBuf, LockfileFormat
 ///
 /// Paths excluded by gitignore are automatically ignored.
 pub fn find_lockable_files_at(root: impl AsRef<Path>) -> Vec<(PathBuf, LockfileFormat)> {
+    let mut manifests: Vec<(PathBuf, _)> = Vec::new();
     let mut lockfiles = Vec::new();
-    let mut manifests = Vec::new();
 
     let walker = WalkBuilder::new(root).max_depth(Some(MAX_LOCKFILE_DEPTH)).build();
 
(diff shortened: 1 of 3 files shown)


lockfile/src/lib.rs
```diff
@@ -440,4 +457,55 @@ mod tests {
...
+        // Find manifest files.
+        let lockable_files = find_lockable_files_at(tempdir.path());
+
+        // Compare results.
+        let expected = vec![
+            (tempdir.path().join("b/pyproject.toml").to_path_buf(), LockfileFormat::Pip),
+            (tempdir.path().join("a/setup.py").to_path_buf(), LockfileFormat::Pip),
+            (tempdir.path().join("setup.py").to_path_buf(), LockfileFormat::Pip),
+        ];
+        assert_eq!(lockable_files, expected);
```
kylewillmon, line 509: It seems less than ideal that the test requires a specific ordering of the results.

Maybe we should sort `lockable_files` before comparing it against `expected` (We would also need to change the order of `expected` to match)?

```rust
lockable_files.sort_by(|(a, _), (b, _)| a.cmp(b));
```

In this case, you could even sort by length if you want to keep the `expected` list pretty:

```rust
// Sort the list, longest path first
lockable_files.sort_by_key(|(p, _)| usize::MAX - p.as_os_str().len());
```

kylewillmon (approved):
Some minor thoughts on how the test was written. Feel free to ignore if you disagree. Code looks good.

lockfile/src/lib.rs
kylewillmon: Looks like this issue is actually causing the tests to fail...

kylewillmon (approved):

PR #1139, opened 2023-06-28
Fix ecosystem extensions without subcommand

### LINKED ISSUE #1025: Allow top-level `--help` output in extensions
# Overview

Help output is not printed when `--help` is the first/only flag provided (and there are no sub-commands specified) to some extensions. This affects the `pip`, `poetry`, and `yarn` extensions, but not the `npm` extension:

```console
~/dev/delme/npm_init  1  3 is 📦 v1.0.0 via  v19.9.0 took 26s
❯ ~/dev/phylum/localdev/cli/target/debug/phylum npm --help
npm <command>

Usage:

npm install        install all the dependencies in your project
npm install <foo>  add the <foo> dependency to your project
npm test           run this project's tests
npm run <foo>      run the script named <foo>
npm <command> -h   quick help on <command>
npm -l             display usage info for all commands
npm help <term>    search for help on <term>
npm help npm       more involved overview

All commands:

    access, adduser, audit, bugs, cache, ci, completion,
    config, dedupe, deprecate, diff, dist-tag, docs, doctor,
    edit, exec, explain, explore, find-dupes, fund, get, help,
    help-search, hook, init, install, install-ci-test,
    install-test, link, ll, login, logout, ls, org, outdated,
    owner, pack, ping, pkg, prefix, profile, prune, publish,
    query, rebuild, repo, restart, root, run-script, search,
    set, shrinkwrap, star, stars, start, stop, team, test,
    token, uninstall, unpublish, unstar, update, version, view,
    whoami

Specify configs in the ini-formatted file:
    /Users/maxrake/.npmrc
or on the command line via: npm <command> --key=value

More configuration info: npm help config
Configuration fields: npm help 7 config

npm@9.6.3 /opt/homebrew/lib/node_modules/npm
```

```console
../cli  14  24 on  extension_help [?] via 🦕 via 🐍 v3.11.3 via 🦀 v1.68.2
❯ target/debug/phylum pip --help
[phylum] This extension does not support arguments before the first subcommand. Please open an issue if "--help" is not an argument.
```

```console
~/dev/delme/npm_init  1  3 is 📦 v1.0.0 via  v19.9.0
❯ ~/dev/phylum/localdev/cli/target/debug/phylum yarn --help
[phylum] This extension does not support arguments before the first subcommand. Please open an issue if "--help" is not an argument.
```

```console
../phylum-ci  9  22 on  main [?] via 🐳 desktop-linux is 📦 v0.28.1 via 🐍 v3.11.3
❯ ~/dev/phylum/localdev/cli/target/debug/phylum poetry --help
[phylum] This extension does not support arguments before the first subcommand. Please open an issue if "--help" is not an argument.
```

Functional top-level `--help` usage is desired because the documentation still suggests users alias these package manager commands and if they do, they will almost certainly be surprised when basic help output is not available.

# Acceptance Criteria
- [x] `phylum pip --help` prints the expected help output
- [x] `phylum poetry --help` prints the expected help output
- [x] `phylum yarn --help` prints the expected help output

cd-work:
This fixes an issue where ecosystem extension invocations wouldn't work unless they also had a subcommand specified. This specifically prevented the common `--help` flag from being used on its own.

Closes #1025.

---

I think this is a way simpler and more reliable way to prevent options before subcommands, however it still isn't a bulletproof design. In this scenario something like `yarn add install` would fail for example, but I don't think any of the package manager subcommands have packages with the same name that are popular enough to make this an issue.

The other alternative would be a more complicated argument parser that actually checks for at least `-` at the start of arguments, which might require some edgecase handling too, but I'd like to avoid that unless necessary. If we did add something like this, it would be nice to have some abstraction in place so we only have to write it once for all ecosystem extensions.

### DIFF
--- a/extensions/poetry/main.ts
+++ b/extensions/poetry/main.ts
@@ -24,41 +24,17 @@ async function findRoot(manifest: string): Promise<string | undefined> {
   return undefined;
 }
 
-// List with all of poetry's subcommands.
-const knownSubcommands = [
-  "about",
-  "add",
-  "build",
-  "check",
-  "config",
-  "export",
-  "help",
-  "init",
-  "install",
-  "list",
-  "lock",
-  "new",
-  "publish",
-  "remove",
-  "run",
-  "search",
-  "shell",
-  "show",
-  "update",
-  "version",
-  "cache",
-  "debug",
-  "env",
-  "self",
-  "source",
-];
-
-// Ensure the first argument is a known subcommand.
+// Ensure no arguments are passed before `add`/`update`/`install` subcommand.
 //
 // This prevents us from skipping the analysis when an argument is passed before
 // the first subcommand (i.e.: `poetry --no-color add package`).
 const subcommand = Deno.args[0];
-if (Deno.args.length != 0 && !knownSubcommands.includes(subcommand)) {
+if (
+  Deno.args.length != 0 &&
+    (Deno.args.includes("add") && subcommand !== "add") ||
+  (Deno.args.includes("update") && subcommand !== "update") ||
+  (Deno.args.includes("install") && subcommand !== "install")
+) {
   console.error(
     `[${
       red("phylum")
--- a/extensions/poetry/main.ts
+++ b/extensions/poetry/main.ts
@@ -24,21 +24,18 @@ async function findRoot(manifest: string): Promise<string | undefined> {
   return undefined;
 }
 
-// Ensure no arguments are passed before `add`/`update`/`install` subcommand.
+// Ensure no arguments are passed before a subcommand.
 //
 // This prevents us from skipping the analysis when an argument is passed before
 // the first subcommand (i.e.: `poetry --no-color add package`).
-const subcommand = Deno.args[0];
-if (
-  Deno.args.length != 0 &&
-    (Deno.args.includes("add") && subcommand !== "add") ||
-  (Deno.args.includes("update") && subcommand !== "update") ||
-  (Deno.args.includes("install") && subcommand !== "install")
-) {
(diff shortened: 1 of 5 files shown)


maxrake:
Changelog entry?

extensions/poetry/main.ts
```diff
@@ -24,41 +24,17 @@ async function findRoot(manifest: string): Promise<string | undefined> {
...
-// Ensure the first argument is a known subcommand.
+// Ensure no arguments are passed before `add`/`update`/`install` subcommand.
 //
 // This prevents us from skipping the analysis when an argument is passed before
 // the first subcommand (i.e.: `poetry --no-color add package`).
 const subcommand = Deno.args[0];
-if (Deno.args.length != 0 && !knownSubcommands.includes(subcommand)) {
+if (
+  Deno.args.length != 0 &&
+    (Deno.args.includes("add") && subcommand !== "add") ||
```
kylewillmon, line 34: This now triggers on:

```
phylum poetry help add
```

kylewillmon:
> In this scenario something like `yarn add install` would fail for example

This case seems easy to solve. If `subcommand` is one of the commands that we care about (in this case, `add`), we should never fail

cd-work:
I'm honestly not sure why I decided to go for such complicated solutions. I think my current version should perform far better than all the previous ones and is much closer to what we're actually trying to do.

I've also noticed that `npm` did still prevent `npm --help` and fixed that too.

kylewillmon (approved):

maxrake (approved):
The code looks good. The changes were tested locally and work as intended. The Acceptance Criteria from the linked issue was marked complete. LGTM.

PR #1140, opened 2023-06-29
Add support for NuGet lockfiles

### LINKED ISSUE #1056: Recognize `packages.lock.json` as NuGet lockfile
# Overview

NuGet lockfiles have been mis-identified in [the documentation](https://docs.phylum.io/docs/analyzing-dependencies) and `phylum_lockfile` crate. For C# projects using the NuGet package manager, `*.csproj` files are *manifests* while `packages.lock.json` is the *lockfile*. This distinction matters a lot more now that work is ongoing to distinguish between manifests and their corresponding lockfiles.

## References

* [MS Documentation for NuGet](https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#enabling-the-lock-file)
* [MS DevBlog: "Enable repeatable package restores using a lock file"](https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/)
* [NuGet notes (internal)](https://www.notion.so/phylum/NuGet-b741ffff820f4e709c97f920cc6a21da?pvs=4)

## Acceptance Criteria
- [ ] `phylum_lockfile` crate recognizes `packages.lock.json` as a NuGet lockfile and `*.csproj` files as manifests
- [ ] Tests and/or test fixtures are updated
- [ ] Documentation is updated
  - [ ] [Analyzing Dependencies page](https://docs.phylum.io/docs/analyzing-dependencies)

cd-work:
This patch adds the ability for CLI to parse NuGet's `packages.lock.json` lockfiles.

Currently this only supports the format itself and does no generation from `*.csproj` files to `packages.lock.json` using `nuget`.

Closes #1056.

### DIFF
--- a/lockfile/src/lib.rs
+++ b/lockfile/src/lib.rs
@@ -4,7 +4,7 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 pub use cargo::Cargo;
-pub use csharp::CSProj;
+pub use csharp::{CSProj, PackagesLock};
 pub use golang::GoSum;
 use ignore::WalkBuilder;
 pub use java::{GradleLock, Pom};
@@ -50,9 +50,9 @@ pub enum LockfileFormat {
     Gradle,
     // This is historically called "nuget" but it's actually for MSBuild project files.
     // Nuget has its own file formats that are not currently supported.
-    #[serde(rename = "nuget")]
-    #[serde(alias = "msbuild")]
+    #[serde(alias = "nuget")]
     Msbuild,
+    NugetLock,
     Go,
     Cargo,
     Spdx,
@@ -91,7 +91,8 @@ impl LockfileFormat {
             LockfileFormat::Poetry => "poetry",
             LockfileFormat::Maven => "mvn",
             LockfileFormat::Gradle => "gradle",
-            LockfileFormat::Msbuild => "nuget",
+            LockfileFormat::Msbuild => "msbuild",
+            LockfileFormat::NugetLock => "nugetlock",
             LockfileFormat::Go => "go",
             LockfileFormat::Cargo => "cargo",
             LockfileFormat::Spdx => "spdx",
@@ -111,6 +112,7 @@ impl LockfileFormat {
             LockfileFormat::Maven => &Pom,
             LockfileFormat::Gradle => &GradleLock,
             LockfileFormat::Msbuild => &CSProj,
+            LockfileFormat::NugetLock => &PackagesLock,
             LockfileFormat::Go => &GoSum,
             LockfileFormat::Cargo => &Cargo,
             LockfileFormat::Spdx => &Spdx,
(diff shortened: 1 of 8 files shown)


lockfile/src/lib.rs
```diff
@@ -49,9 +49,9 @@ pub enum LockfileFormat {
     Gradle,
     // This is historically called "nuget" but it's actually for MSBuild project files.
     // Nuget has its own file formats that are not currently supported.
-    #[serde(rename = "nuget")]
-    #[serde(alias = "msbuild")]
+    #[serde(alias = "nuget")]
     Msbuild,
+    NugetLock,
```
cd-work, line 54: I'd personally much prefer to call this just `nuget` because `nugetlock` vs `nuget` is incredibly confusing. However I'm not sure we can currently make this change without breaking user's setups.

It might be best to leave this as `nugetlock` for a while until project files are transitioned to the new default `msbuild` name for `MsBuild`, so we can then claim the `nuget` identifier for this.

maxrake (changes requested):
Preliminary review found the following:

Documentation updates needed:
* `docs/knowledge_base/analyzing-dependencies.md`
* Regenerate existing docs to satisfy the broken tests in CI

> Currently this only supports the format itself and does no generation from `*.csproj` files to `packages.lock.json` using `nuget`.

Creating a new issue would help to track the work of lockfile generation from `*.csproj` files since #1056 was created before lockfile generation was available.

cd-work:
> Creating a new issue would help to track the work of lockfile generation

https://github.com/phylum-dev/cli/issues/1142

lockfile/src/lib.rs
matt-phylum: There is another format where old NuGet projects use a packages.config file. Some people still use it.

cd-work: I think that's more of a manifest than lockfile, right?

maxrake: Sure enough...it appears to still be an [additional supported format](https://learn.microsoft.com/en-us/nuget/reference/packages-config). It looks mostly like a lockfile, but the `allowedVersions` field might introduce some additional complexity. That field is used for _updates_ and not _install_ or _restore_ operations.

maxrake (commented):
Addressing the current merge conflicts plus addressing the format support question are enough for me to hold off on a review for a bit longer until @cd-work has a chance to weigh in. Plus, I need to get smarter on NuGet to have a better chance at testing this locally.

lockfile/src/lib.rs
matt-phylum: `packages.config` isn't a lockfile but it normally functions as much like a lockfile as you can get for `pip`. It contains the transitive dependencies and it normally contains the exact versions of dependencies that are used.

`allowedVersions` is just a limited version selector in a different format that isn't too hard to convert, but we probably don't care about it.

cd-work:
> Addressing the current merge conflicts plus addressing the format support question are enough for me to hold off on a review for a bit longer until @cd-work has a chance to weigh in.

The merge conflicts are solely documentation changes and exist only **because** nobody reviewed this PR in so long.

There is no additional format to support. We talked about `packages.config` about a week ago when I initially implemented this.

> Agreed...it is confusing. Can MSBuild be removed entirely from here since it maps to *.csproj manifests with no matching lockfile? *.csproj manifests would still be allowed, they would just only map to NuGet since that is the package manager that can take those manifests and turn them into packages.lock.json lockfiles.

No it cannot. At the very least csproj needs to be kept until lockfile generation is implemented.

maxrake (dismissed):
It looks like my last review had some stale comments (drafted last week and then weren't submitted until the overall review was). Sorry about that. The discussions from last week are coming back now...we did decide as a team to work #1142 separately.

The new test fixture and passing tests LGTM.

maxrake (approved):

Issue #1138, opened 2023-06-28
Support pnpm lockfiles


kylewillmon:
Add support for `pnpm-lock.yaml` (from [pnpm](https://pnpm.io/))

This was originally requested in phylum-dev/phylum-ci#270

PR #1143, opened 2023-06-30
Add support for pnpm

cd-work:
This adds support for the new `pnpm-lock.yaml` lockfile type and its
package manager `pnpm`. Supported are both parsing the lockfile itself
and generating a lockfile from the `package.json`.

Closes https://github.com/phylum-dev/cli/issues/1138.

### DIFF
--- a/lockfile/src/javascript.rs
+++ b/lockfile/src/javascript.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::ffi::OsStr;
 use std::path::Path;
 
@@ -5,12 +6,15 @@ use anyhow::{anyhow, Context};
 #[cfg(feature = "generator")]
 use lockfile_generator::npm::Npm as NpmGenerator;
 #[cfg(feature = "generator")]
+use lockfile_generator::pnpm::Pnpm as PnpmGenerator;
+#[cfg(feature = "generator")]
 use lockfile_generator::yarn::Yarn as YarnGenerator;
 #[cfg(feature = "generator")]
 use lockfile_generator::Generator;
 use nom::error::convert_error;
 use nom::Finish;
 use phylum_types::types::package::PackageType;
+use serde::Deserialize;
 use serde_json::Value as JsonValue;
 use serde_yaml::Value as YamlValue;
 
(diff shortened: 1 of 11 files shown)


lockfile/src/javascript.rs
```diff
@@ -263,6 +267,69 @@ impl Parse for YarnLock {
...
+
+impl PnpmLock {
+    /// Get all packages in the lockfile.
+    fn packages(&self) -> anyhow::Result<Vec<Package>> {
+        let mut packages = Vec::new();
+
+        // Convert `/@org/package@1.2.3(test@3.2.1)` to package.
+        for name in self.packages.keys() {
+            // Strip `/` prefix.
+            let name = &name[1..];
```
kylewillmon, line 311: It's probably safer to confirm that `name[0]` is a `/` before stripping it. Something like:

```rust
let name = name.strip_prefix("/").ok_or_else(|| anyhow!("Invalid package: {name}"))?;
```

kylewillmon (changes requested):
I know that `pnpm` (like `yarn`) supports [patching dependencies](https://pnpm.io/feature-comparison).

I assume they also support filesystem dependencies and VCS dependencies. But the current code doesn't seem to handle any of that. Do those not show up in the `packages` list?

maxrake (dismissed):
The code was reviewed and looked okay to my eye. The changes were tested locally with a few real-world `pnpm-lock.yaml` files and worked as expected.

The requested change is to add a change log entry.

cd-work:
> I assume they also support filesystem dependencies and VCS dependencies.

I tested this and filesystem dependencies indeed do not show up in the packages list.

VCS dependencies I didn't take into account and they do indeed seem to be supported with a slightly different syntax.

cd-work:
I've added support for tarballs. Neither git nor path dependencies seem to show up in the `packages` section of the lockfile. They do show up at the top as dependencies, but I don't think it's worth trying to parse it considering we don't analyze them anyway.

cd-work:
Patched dependencies show up as their original dependency with some minor patch information associated. I think it makes sense to just treat them as firstparty dependencies since that we can actually analyze.

kylewillmon (approved):

PR #1146, opened 2023-07-06
Split header initialization into separate function

kylewillmon:
The code was failing `deno check` because the type system was confused about the type of `fetchInit.headers`. This patch removes the confusion.

This patch also fixes the conditional around the `Content-Type` header, which was wrong.

### DIFF
--- a/extensions/phylum.ts
+++ b/extensions/phylum.ts
@@ -20,6 +20,23 @@ export type ProcessOutput = {
   code: number | null;
 };
 
+async function requestHeaders(headersInit?: HeadersInit): Promise<Headers> {
+  const headers = new Headers(headersInit);
+
+  // Set Authorization header if it is missing.
+  if (!headers.has("Authorization")) {
+    const token = await PhylumApi.getAccessToken();
+    headers.set("Authorization", `Bearer ${token}`);
+  }
+
+  // Set Content-Type header if it is missing.
+  if (!headers.has("Content-Type")) {
+    headers.set("Content-Type", "application/json");
+  }
+
+  return headers;
+}
+
 export class PhylumApi {
   /**
    * Send a request to the Phylum REST API.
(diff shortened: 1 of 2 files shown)


extensions/phylum.ts
```diff
@@ -37,19 +54,7 @@ export class PhylumApi {
...
-    // Set Authorization header if it is missing.
-    if (!fetchInit.headers.has("Authorization")) {
-      const token = await PhylumApi.getAccessToken();
-      fetchInit.headers.set("Authorization", `Bearer ${token}`);
-    }
-
-    // Set Content-Type header if it is missing.
-    if (fetchInit.headers.has("Content-Type")) {
-      fetchInit.headers.set("Content-Type", "application/json");
-    }
```
kylewillmon, line 52: If has, then set.... Whoops. We missed a `!` here.

extensions/phylum.ts
```diff
@@ -37,19 +54,7 @@ export class PhylumApi {
...
+    fetchInit.headers = await requestHeaders(fetchInit.headers);
 
     // Set Content-Type header if it is missing.
```
maxrake, line 59: Extra comment was left in here from before as well.

maxrake (approved):
Might want a changelog entry since there was a bug fix as well.

LGTM.

kylewillmon:
It bothered me that the type system was so easily confused, so I tracked it down to the recent [`d632cce`](https://github.com/denoland/deno/commit/d632cce129cb7025a34cf0aa7262a38fb12f47c4) commit in the `deno` repository. It looks like `HeadersInit` no longer includes the `Headers` type in the union... I believe this is a bug and will be submitting a ticket to `deno`.

kylewillmon:
I've created denoland/deno#19746, but I'm going to move forward with this PR so that we can move on.

maxrake (approved):
LGTM

extensions/phylum.ts
maxrake: Good catch!

PR #1147, opened 2023-07-06
Bump deno versions

kylewillmon:
This patch updates the `deno` crates to latest. Only one change to account for, which was the removal of `.force_op_registration()`. This update happened in `deno` commit [`34dac6c`](https://github.com/denoland/deno/commit/34dac6c6efa75f38c29031a65db1ee3332a67259) because the function is no longer required.

This should fix the issues seen in #1144


cd-work (approved):
Looks good, though it seems there might be some Deno API changes we need to fix.

PR #1148, opened 2023-07-07
Remove Windows CI

kylewillmon:
Our builds are too large for GitHub's Windows runners and they keep running out of disk space. And since we don't officially support native Windows, it's time for these builds to go.

After all, 2023 is the Year of the Linux Desktop.


cd-work (approved):

maxrake (approved):
LGTM

PR #1151, opened 2023-07-11
Switch documentation hosting from Readme.com to Docusaurus

### LINKED ISSUE #1141: Host docs with Docusaurus
# Overview

As part of the larger [documentation overhaul story](https://github.com/phylum-dev/roadmap/issues/344), the CLI repo will host it's own documentation. The documentation should be generated as a static site by Docusaurus and hosted on GitHub Pages. The content should be organized in a way that makes sense as a standalone reference even though it will be linked to from the main documentation site at https://docs.phylum.io

## Additional Context

* ~Do not remove the documentation publication step for Readme.com~
  * ~That step will be removed in a different issue, after the new CLI sub-domain is linked to the main docs.phylum.io site~
  * Update: the Readme.com site can continue with stale CLI content for the time it takes to get the new Docusaurus site up; it is okay to remove the `rdme` publication step now
* Work with @louislang to configure the custom domain with the DNS provider
  * Do this **after** adding the custom domain to the GitHub pages site
  * Possible (preferred) sub-domain: `cli.phylum.io`
* The Docusaurus project `package-lock.json` should be included for analysis and dep bumps
* References
  * [Docusaurus docs](https://docusaurus.io/docs)
  * [About GitHub Pages](https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages)
  * [Managing a custom domain for a GitHub pages site](https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site)

## Acceptance Criteria

* [x] CLI documentation is available at a sub-domain of phylum.io (e.g., https://cli.phylum.io)
* [x] Documentation is generated during the release workflow

maxrake:
This is part of a larger effort to move <https://docs.phylum.io> from Readme.com
to Docusaurus, as a statically generated site. The main site will host all the
documentation _except_ for the CLI docs, which will be hosted on a separate
statically generated site, built from this repository and served at
<https://cli.phylum.io>

Actions taken include:

* Add docusaurus site in `website` directory
  * Remove the blog content/config and disable the blog plugin
  * Remove unused files and docusaurus default pages
  * Remove the README.com docs that have been replaced by Docusaurus
* Add docusaurus website project's lockfile to `.phylum_project`
* Update docusaurus configuration
  * Update and add new values for docusaurus config
    * Replace defaults with Phylum CLI specific values
  * Configure site for "docs-only" mode
  * Import Phylum logo and favicon
  * Update README for `npm` based commands
  * Import quickstart guide as the root level page
    * Update the sidebar positions to make it first
  * Populate the footer content and links
  * Add custom CSS for a GitHub link icon in the navbar
  * Ensure broken links and duplicated routes throw errors when building
  * Switch the default mode to dark
  * Add an announcement bar
  * Create custom sidebar
* Add CNAME record pointing to `cli.phylum.io`
  * Coordinate with HitHub Pages usage and configuration for the repo
* Reorganize top level docs
  * Create a new `supported_lockfiles` doc
    * Populate it with the content from the `analyzing_dependencies` doc
  * Rename files
* Format docs
  * Remove `extension_howto` doc
    * Put it's content in a callout tip in the `extension_overview` doc
  * Move and modify the "alternate install" doc
  * Add admonitions
  * Format for `markdownlint` adherence
  * Update all reference links throughout the repo to point to new URL
  * Ensure internal links use relative file paths
    * It is better to use relative _file_ links than any other kind offered
    * Ref: <https://docusaurus.io/docs/markdown-features/links>
  * Move extensions, analyzing dependencies, and lockfile generation pages
  * Change Readme.com language to Docusaurus in CONTRIBUTING and README files
  * Change all instances of 'lock file' to 'lockfile'
  * Change instances of `console` to `shellsession`
* Update `analyzing_dependencies.md` to current output
* Create a `Commands` category for the sidebar to replace `command_line_tool`
  * Auto generate the items to allow for dynamic command updates
* Remove all doc template files that only existed to specify a category
  * The sub-categories are removed in favor of simpler command updates
* Update `gendocs` xtask
  * Sort the subcommands so they show in same order as sidebar entries
    * The autogenerated sidebar is easier to work with
    * Keeping the order consistent makes browsing and comparing easier
  * Wrap angle brackets in markdown output with backticks
    * Site generation interprets `<foo>` as an unclosed tag and fails
    * Do the same for square brackets for visual consistency
    * Do the same for long and short options for visual consistency
  * Change the output directory for generated docs
  * Remove Readme.com front matter in the default header
  * Regenerate all docs with `cargo xtask gendocs`
* Update the test workflow
  * Add documentation site building job
  * Add all jobs to the test rollup job
    * Some were missing and therefore not enforced by the repo settings
* Update workflows for automatic updates
  * Update Dependabot config to add NPM updates for docusaurus
  * Rename `cargo-update.yml` workflow to `auto_updates.yml`
    * Add comments
    * Change the language to be less Cargo specific
    * Add docusaurus dependency update steps
* Manually update the docusaurus dependencies to latest versions
* Add a `docs` workflow for updating the documentation website
  * Make it reusable so it can be called from within the same repo
  * Allow it to be called manually from the "Actions" tab in GitHub
  * Call the docs workflow from within the release workflow
  * Ensure the docs and release workflows only allow one run at a time
* Correct template extension for `phylum group delete` command

Closes #1141 

## Checklist

- [x] ~~Have you updated CHANGELOG.md (or extensions/CHANGELOG.md), if applicable~~

## Screenshots

The changes in this PR have been built and are already hosted with GitHub Pages via settings in this repository. Check out the site here: https://cli.phylum.io


docs/quickstart.md
```diff
@@ -1,19 +1,11 @@
...
+# Quickstart
 
----
-
-# Introduction
-
-The command line interface (CLI) allows users to submit their project package dependencies to [Phylum's](https://phylum.io) API for analysis. Currently [pre-built binaries](https://github.com/phylum-dev/cli/releases) for Linux and macOS are available. On Windows, we recommend using the Linux binaries under [WSL](https://learn.microsoft.com/en-us/windows/wsl/). For more options, see [Alternate Installation Methods](https://docs.phylum.io/docs/alternate_install).
+The Phylum command line interface (CLI) allows users to submit their project package dependencies to [Phylum's](https://phylum.io) API for analysis. Currently [pre-built binaries](https://github.com/phylum-dev/cli/releases) for Linux and macOS are available. On Windows, we recommend using the Linux binaries under [WSL](https://learn.microsoft.com/en-us/windows/wsl/). For more options, see [Alternate Installation Methods](./alternate_install.md).
 
 [![asciicast](https://asciinema.org/a/431262.svg)](https://asciinema.org/a/431262)
```
maxrake, line 10: @andreaphylum provided feedback to suggest changing this animation so it is not so small and also so that it is current.

maxrake:
There are a lot of commits on this branch, but rest assured the PR will be squash merged to include the commit message from the description of this PR.

doc_templates/README.md
```diff
@@ -1,34 +1,27 @@
 # CLI documentation templates
 
-These templates provide extra information for the automatically generated CLI
-documentation.
+These templates provide extra information for the automatically generated CLI documentation.
```
cd-work, line 3: Don't really like that formatting was changed here. It certainly makes the review significantly more difficult.

Just for reference, the previous column width was 80, which I think is the best choice for reading plain text.

doc_templates/phylum_auth_set-token.md
```diff
@@ -1,9 +1,4 @@
----
-title: phylum auth set-token
-category: 6255e67693d5200013b1fa3e
-parentDoc: 625758b12a1e9d0186416451
-hidden: false
----
+{PH-HEADER}
```
cd-work, line 1: If this header is removed everywhere, do we even still need it as an option?

docs/quickstart.md
maxrake: The animation was removed entirely because it was outdated and provided no additional content/value over the text that is already on the Quickstart/README pages.

website/.gitignore
cd-work, line 1: Not a fan of having gitignores outside the root. That just makes things so much more difficult to troubleshoot.

cd-work (changes requested):
I'm not sure I like the idea of adding a website to this repository. That just seems entirely unrelated.

It might be better to move this to a separate place?

website/.gitignore
maxrake: That makes sense. This file was auto-generated by Docusaurus when the site was created...but it likely doesn't need to be a standalone file. I'll look at merging it with the top-level `.gitignore` file.

doc_templates/phylum_auth_set-token.md
maxrake: It think it is still useful, for a few reasons:

* It contains the `{PH-TITLE}` placeholder, which is a way to get the page title set correctly
  * The other way is to use the first header of the doc (`#` and maybe also `##`) when the `title` frontmatter is _not_ set
  * That can be tricky when the first header either doesn't exist or is meant to be something _other_ than the page title
* It can be used for future frontmatter entries, which [Docusaurus has many](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#markdown-front-matter)

cd-work: Fair enough I wasn't sure if docasaurus supported headers in similar fashion. Otherwise just using `{PH-TITLE}` should have been sufficient.

doc_templates/README.md
maxrake: This was a tough change to make because I also had previously been using a fixed line length as well...albeit a larger one than 80. That is kind of the point...there is no enforced standard for markdown and it would/could change between authors. Here are some reasons why hard wrapping markdown text was removed in favor of (possibly) longer lines and soft wrapping:

* Hard wrapping requires more author effort, especially when modifying existing text
* Lines with link/image reference definitions are harder to split without breaking the URL
* The markdown source does not look like the rendered output, in which there are no line breaks
* Manual line breaking can make the Markdown more readable than the rendered output, which is bad because it gives a false sense of readability encouraging less readable long paragraphs
* Markdown format characters count against the line limit even if they wouldn't in rendered form
* Individual developers can choose to set their editors and diff tools to view the source/diffs the way they prefer
* GitHub's web interface as well as many desktop Git tools feature substring change highlighting within lines and side-by-side or similar version comparison so there is no need for artificial line breaks
* A quick survey of other popular Docusaurus powered sites revealed they do not enforce hard wrapping
  * [Docusaurus itself](https://github.com/facebook/docusaurus/blob/main/website/docs/deployment.mdx)
  * [Redis](https://github.com/redis-developer/redis-developer.github.io/blob/master/docs/create/redis-stack/index-redis-stack.mdx)
  * [React](https://github.com/facebook/react-native-website/blob/main/docs/introduction.md?plain=1)
  * [Supabase](https://github.com/supabase/supabase/blob/master/apps/docs/pages/guides/auth.mdx) (although they seem to hard break on sentence boundaries)

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 - Support for NuGet's `packages.lock.json` lockfiles
 
+### Changed
+- Switch documentation hosting from Readme.com to Docusaurus
```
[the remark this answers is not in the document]
maxrake: It will be removed.

.github/workflows/test.yml
```diff
@@ -23,6 +23,12 @@ jobs:
       - name: Checkout the repo
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
 
+      # Workaround to provide additional free space for testing.
+      # Reference: https://github.com/actions/runner-images/issues/2840
+      - name: Free disk space
+        if: startsWith(runner.name, 'GitHub Actions') && startsWith(matrix.os, 'ubuntu')
+        run: sudo rm -rf /usr/share/dotnet
```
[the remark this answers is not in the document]
maxrake: Since yesterday...at least within this branch. There is a larger discussion about this in Slack. Bottom line is that the runner images are only guaranteed to have _at least_ 14GB of disk space available, they often have more, but can dip down to 14GB _at any time_.

.github/workflows/auto_updates.yml
```diff
@@ -0,0 +1,71 @@
...
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+
+      # This GPG key is for the `phylum-bot` account and used in order to ensure commits are signed/verified
+      - name: Import GPG key for bot account
+        uses: crazy-max/ghaction-import-gpg@72b6676b71ab476b77e676928516f6982eef7a41 # v5.3.0
```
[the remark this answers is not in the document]
maxrake: This action was reviewed and vetted the first time it was used in the `phylum-dev/phylum-ci` repository. It was found to be safe and useful for ensuring commits made in workflows are signed, without having to recreate the same features in script blocks and post-workflow jobs/steps.

One of it's nicest features is how it will automatically purge the imported GPG key and cache information and also kill the agent from the runner. It is possible to do this manually, but much more error prone. Here is an example of what it looks like for the single action to take care of both the setup and tear down:

image

---

Further, the action is pinned to a specific hash and updates are only made in a Dependabot PR, providing a chance to ensure nothing sketchy is added to newer versions. "crazy-max" is a bit of an odd choice for someone asking folks to run code from a stranger...but this developer has _a lot_ of popular GitHub actions.

maxrake:
> I'm not sure I like the idea of adding a website to this repository. That just seems entirely unrelated.
> 
> It might be better to move this to a separate place?

Extensive conversation about the various options to consider for the overall "Documentation Overhaul" effort already took place in https://github.com/phylum-dev/documentation/issues/24 and consensus was reached and approved by the affected team leads and product owners. This PR is merely the manifestation of that plan.

For some background, Docusaurus was pre-selected as the content generation solution and GitHub Pages as the content hosting solution. One of the limitations with static site generators (SSGs) like Docusaurus is that they require all content for a given site to be in one place at the time of generation. There are various options for making that happen (outlined in the phylum-dev/documentation#24) but the option to keep CLI docs in the `cli` repo was selected...with a separate `cli.phylum.io` site hosted there.

website/.gitignore
maxrake: The `website/.gitignore` file was merged with the root level `.gitignore`.

FYI...for troubleshooting there is always the `git-check-ignore` command available to determine if and where/how a given file is ignored by `git`, taking into account multiple ignore sources:

```
../cli  17  16 on  docusaurus via 🦕 via 🦀 v1.70.0
❯ git check-ignore --verbose website/npm-debug.log
.gitignore:47:npm-debug.log*	website/npm-debug.log

../cli  17  16 on  docusaurus via 🦕 via 🦀 v1.70.0
❯ git check-ignore --verbose website/a/b/npm-debug.log
.gitignore:47:npm-debug.log*	website/a/b/npm-debug.log

../cli  17  16 on  docusaurus via 🦕 via 🦀 v1.70.0
❯ git check-ignore --verbose website/package-lock.json

../cli  17  16 on  docusaurus via 🦕 via 🦀 v1.70.0
❌ 1 ❯ git check-ignore --verbose website/does/not/exist.txt

../cli  17  16 on  docusaurus via 🦕 via 🦀 v1.70.0
❌ 1 ❯
```

doc_templates/README.md
andreaphylum: I have not used it personally, and I'm not sure it could be helpful to auto-enforce Markdown formatting, but I think we could consider checking out [editorconfig](https://editorconfig.org/).

I have checked over time for Markdown auto-formatting solutions but I have never found anything that really fit the bill in an acceptable way. It would definitely be a huge value add for a case like this, though.

.github/workflows/auto_updates.yml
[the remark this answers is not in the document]
cd-work: > Further, the action is pinned to a specific hash and updates are only made in a Dependabot PR, providing a chance to ensure nothing sketchy is added to newer versions.

Do we actually have someone willing or capable of reviewing this action though? Did we actually review it to ensure no existing vulnerabilities are present?

I do not think I'd be capable of auditing updates for this without **significant** effort, so if someone else is able to I'm happy to leave these dependabot PRs up to them.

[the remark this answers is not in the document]
maxrake: The action was reviewed initially and then incrementally as it was updated via Dependabot PRs. It is true that this is a particularly sensitive action, but it is also true that any of the other actions or dependencies in use can similarly "go bad." Pinning actions/dependencies and reviewing their changes...through automation or manually...is the price to pay for vigilance.

Even so, further discussion with @louislang revealed that there is _not_ a requirement to have all intermediate commits signed. With the squash merge strategy that this repo has, most (all?) commits coming from PRs merged to `main` will show as signed by GitHub and it's verified signature (currently with a GPG key id of `4AEE18F83AFDEB23`) instead of the individual author(s) of the PR.

Therefore, the action has been removed in favor of making unsigned commits in workflows with the `phylum-bot` account.

doc_templates/README.md
kylewillmon: The conversation has derailed from the original comment.

The debate between hard breaks and soft breaks can wait. This PR is about the move to Docusaurus. Significant and unrelated formatting changes belong in a separate PR.

kylewillmon:
> Extensive conversation about the various options to consider for the overall "Documentation Overhaul" effort already took place in phylum-dev/documentation#24 and consensus was reached and approved by the affected team leads and product owners.

When that decision was made, I don't think it was clear that keeping the CLI docs in this repo would require adding a node.js project to this repo... We may need to discuss other options...

doc_templates/README.md
maxrake: Agreed...it should have been done separately. There was an intermediate commit, after initializing the Docusaurus site, where the `docs` directory was moved to `website/docs`. That is when the formatting changes were made b/c it was expected that the diff would not look pretty anyway. Later, it was discovered that the `docs` directory could live elsewhere and so it was changed back...but the formatting kept.

I'll create a separate issue to discuss markdown standardization...for this repo and ideally as a company standard. The formatting changes made here will be reverted in a best effort approach.

maxrake: The changes have been reverted and https://github.com/phylum-dev/cli/issues/1154 was created.
The https://cli.phylum.io site was deployed with these latest changes for assisting with the review.

maxrake:
The test failure is unrelated to the changes in this PR. The same failure is happening in `main` - https://github.com/phylum-dev/cli/actions/runs/5559623193

andreaphylum:
> The test failure is unrelated to the changes in this PR. The same failure is happening in `main` - https://github.com/phylum-dev/cli/actions/runs/5559623193

Tracking the failure in #1157.

.github/workflows/auto_updates.yml
```diff
@@ -0,0 +1,62 @@
...
+      run:
+        shell: bash
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+
+      - name: Set up Node
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
+        with:
+          node-version: 18
```
cd-work, line 26: A fixed node version here will probably cause maintenance overhead? Isn't node installed on every runner anyway?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Configure git user
+        run: |
+          git config user.name 'phylum-bot'
+          git config user.email '69485888+phylum-bot@users.noreply.github.com'
+
+      - name: Set up Node
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
+        with:
+          node-version: 18
```
cd-work, line 40: Another concrete node version here.

.github/workflows/release.yml
```diff
@@ -3,6 +3,8 @@
 ---
 name: Release
 
+concurrency: release
```
cd-work, line 6: Is this change necessary? Did something break?

.github/workflows/test.yml
```diff
@@ -84,12 +90,49 @@ jobs:
...
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+      - name: Set up Node
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
+        with:
+          node-version: 18
```
cd-work, line 105: Another concrete node version.

.github/workflows/test.yml
```diff
@@ -84,12 +90,49 @@ jobs:
...
+        working-directory: ./website
+        run: npm run build
+
+  # This job reports the results of the test jobs above and is used to enforce status checks in
+  # the repo settings without needing to update those settings everytime the test jobs are updated.
   test:
     if: always()
-    needs: test-matrix
     runs-on: ubuntu-latest
+    needs: [test-matrix, deno-checks, shellcheck, doc-checks]
```
cd-work, line 120: Was the old job broken? Deno checks failed the build correctly even though it wasn't explicitly listed here or in the conditional.

.github/workflows/test.yml
```diff
@@ -84,12 +90,49 @@ jobs:
...
+          echo "At least one test job was not successful"
+          exit 1
+
+      - name: Confirm test jobs success
+        if: >
+          (needs.test-matrix.result == 'success')
+          && (needs.deno-checks.result == 'success')
+          && (needs.shellcheck.result == 'success')
+          && (needs.doc-checks.result == 'success')
+        run: echo "All test jobs were successful"
```
cd-work, line 138: Doing the same thing twice just to print a different message is pretty pointless imo. Developers should be able to distinguish a failed build from a successful one.

doc_templates/phylum_extension_new.md
```diff
@@ -1,16 +1,7 @@
...
 {PH-MARKDOWN}
 
 ### Details
 
-This command will create a "Hello, World!" extension skeleton that can be
-modified for a specific use. See the [Extension Quickstart] guide for more
-detail.
-
-[Extension Quickstart]: https://docs.phylum.io/docs/extension_quickstart
+This command will create a "Hello, World!" extension skeleton that can be modified for a specific use. See the [Extension Quickstart](../extensions/extension_quickstart.md) guide for more detail.
```
cd-work, line 7: Don't want to be overly pedantic but this is technically already applying your preferred markdown style. Should probably be reverted.

docs/extensions/extension_example.md
```diff
@@ -106,7 +102,8 @@ this function is asynchronous, we need to `await` it.
 
 The lockfile object will look something like this:
 
-```text
+```js
```
cd-work, line 105: This applies to all the codeblocks that were changed in this file:

I'm not sure calling these `js` is the best idea. I've intentionally marked them as `text` because they represent the text output of `console.log` for these structs. They might technically be in a format that is a valid JavaScript object, but it's not really JavaScript code.

docs/extensions/extension_quickstart.md
```diff
@@ -43,7 +39,7 @@ phylum extension install ./my-extension
 Once successfully installed, our extension can be executed by using its name as
 a subcommand for the phylum CLI:
 
-```console
+```sh
```
cd-work, line 42: This is wrong. No line in this codeblock is valid `sh`.

docs/lockfile_generation.md
```diff
@@ -34,18 +30,22 @@ The Phylum CLI can generate a lockfile when it is given a manifest file.
 [go]: https://go.dev
 [cargo]: https://www.rust-lang.org
 
+:::tip
```
cd-work, line 33: What is this?

docs/lockfile_generation.md
```diff
@@ -34,18 +30,22 @@ The Phylum CLI can generate a lockfile when it is given a manifest file.
...
 
 This can be overridden on the command line with the `--lockfile-type` (`-t`) option. For example:
 
-```
+```sh
 phylum analyze -t yarn package.json
 ```
 
-### Lockfile detection
+:::
```
cd-work, line 46: I'd assume this is similar to the `:::tip` thing?

docs/supported_lockfiles.md
```diff
@@ -0,0 +1,35 @@
+# Supported Lockfiles
+
+The Phylum CLI supports processing many different lockfiles:
+
+| Lockfile type | Filenames                                                              |
```
cd-work, line 5: You renamed the other table from filenames to Manifests. Would make sense to rename this from filenames to Lockfiles then, right?

cd-work (changes requested):
I don't think the issue of where the docs should live is resolved yet, but I've given everything a detailed review to ensure wherever it goes, it doesn't require another full review.

The only stuff I skimmed is the JavaScript website. I'd assume that's mostly from some template or standard stuff.

.github/workflows/auto_updates.yml
maxrake: > Isn't node installed on every runner anyway?

Node is [installed and cached on the Ubuntu 20.04 runners](https://github.com/actions/runner-images/blob/ubuntu20/20230710.1/images/linux/Ubuntu2004-Readme.md#nodejs), but the [documentation for the `actions/setup-node` action](https://github.com/actions/setup-node) states:

> The `node-version` input is optional. If not supplied, the node version from PATH will be used. However, it is recommended to always specify Node.js version and don't rely on the system one.

The idea was to specify a version that would use a cached version of node that is already available on the runner because that is quick...only 2 seconds to run the "Set up Node" step:

image

A more detailed reading of the `actions/setup-node` action documentation shows that there are options to satisfy our shared desires:

* Setting the value for `node-version` to `*` will "get the latest locally-cached Node.js version"
* `node-version-file` (e.g., `.node-version`) can be used to specify the node version instead of using `node-version`
  * That might reduce the maintenance overhead by providing a single reference source
  * `website/package.json` could be used since it already already contains the `engines.node` value [as requested in the docs](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file)

.github/workflows/docs.yml
maxrake: This will be updated with one of the proposed solutions for staying fast and DRY.

.github/workflows/test.yml
maxrake: This will be updated with one of the proposed solutions for staying fast and DRY.

.github/workflows/release.yml
maxrake: Nothing broke...but it could now that there is both a `docs` and `release` workflow that each can update the `cli.phylum.io` website. Adding `concurrency: release` to each workflow ensures that only one such workflow can run at a time. See the [GitHub Actions documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) for more detail.

.github/workflows/test.yml
maxrake: It has to do with branch protection rules. The current ones for this repo have enabled the setting to enforce specific status checks before allowing a merge to `main`:

image

These are the same status checks that have a `required` label by them on the conversation view of the PR:

image

The old job was not broken in the sense that the overall build would show a failure if one of the status checks failed. It was just that it was still possible to merge the PR to `main` if one of the failed status checks was not `required`.

With that in mind, the initial idea for a "rollup" test report job like this to exist in the first place was to reduce the amount of coordination required when adding or modifying jobs in the test workflow. That is, it is much easier to update the rollup job to add new test jobs than it is to add a new job (or change an existing job's name) and then _also_ remember to update the repo settings...assuming permissions exist to make that change.

.github/workflows/test.yml
maxrake: That is a good point. This will be changed.

docs/extensions/extension_example.md
maxrake: Okay...they will be changed back.

docs/supported_lockfiles.md
maxrake: Sure...it will be renamed.

docs/supported_lockfiles.md
```diff
@@ -0,0 +1,35 @@
...
+If needed, this can be overridden with the `--lockfile-type` (`-t`) option.
+
+:::
+
+:::tip Manifest Support
+
+Lockfiles can also automatically be generated for certain manifest files.
+See [lockfile_generation](./lockfile_generation.md) for details.
+
+:::
```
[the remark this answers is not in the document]
maxrake: > Are we just abandoning Markdown at this point?

Yes? At least, for the Docusaurus rendered pages since they have some additional, non-standard, [Markdown Features](https://docusaurus.io/docs/markdown-features).

docs/lockfile_generation.md
maxrake: The triple colon syntax is a [Docusaurus admonition](https://docusaurus.io/docs/markdown-features/admonitions)

docs/lockfile_generation.md
maxrake: Yes...that is the closing identifier for a [Docusaurus admonition](https://docusaurus.io/docs/markdown-features/admonitions)

docs/extensions/extension_quickstart.md
maxrake: `console` is also wrong because it is not a [supported language for Docusaurus code blocks](https://docusaurus.io/docs/markdown-features/code-blocks#supported-languages). Basically, the choice of code block language needs to come from [this list of supported languages](https://prismjs.com/#supported-languages). Which one from there is preferred?

doc_templates/phylum_extension_new.md
maxrake: Oops...missed this one during the revert. It will be changed.

.github/workflows/auto_updates.yml
cd-work: > node-version-file (e.g., .node-version) can be used to specify the node version instead of using node-version 

That's not what we want though. We want to always update it automatically so we don't have to worry about it. It's a tiny project, it should never need any maintenance intervention.

The node version installed on the runner should always work.

.github/workflows/release.yml
cd-work: That just seems like a workflow issue tbh. Why wouldn't we just run the `docs` workflow automatically on release, rather than having two places where the same workflow can be run from?

docs/lockfile_generation.md
cd-work: I ***really*** don't like using platform-specific syntax like this.

We've already migrated between documentation providers once and it was pretty simple since it's all just markdown. I don't want the next time to be more difficult.

docs/extensions/extension_quickstart.md
kylewillmon: > `console` is also wrong because it is not a [supported language for Docusaurus code blocks](https://docusaurus.io/docs/markdown-features/code-blocks#supported-languages).

Does docusaurus enforce this? Because that's not really how Markdown works...

Here is the relevant excerpt from [the CommonMark spec][1]:

[1]: https://spec.commonmark.org/0.30/#code-fence

> An [info string](https://spec.commonmark.org/0.30/#info-string) can be provided after the opening code fence. Although this spec doesn’t mandate any particular treatment of the info string, the first word is typically used to specify the language of the code block. In HTML output, the language is normally indicated by adding a class to the code element consisting of language- followed by the language name.

So when you say this is "wrong"... Do you just mean that no syntax highlighting is applied? That's not really a problem.

docs/supported_lockfiles.md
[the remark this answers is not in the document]
cd-work: > additional, non-standard, [Markdown Features](https://docusaurus.io/docs/markdown-features).

Which require training for new developers and vendor-lock us into the platform.

.github/workflows/auto_updates.yml
maxrake: Sounds like a vote for setting the value for `node-version` to `*`. Change coming.

.github/workflows/release.yml
maxrake: > Why wouldn't we just run the `docs` workflow automatically on release, rather than having two places where the same workflow can be run from?

Keeping the `Update-Documentation` job in the `release` workflow ensures that it can be made dependent on the `Release` job. Otherwise, the documentation would be updated _before_ it is known that the release is successful...which could be messy.

docs/extensions/extension_quickstart.md
maxrake: GitHub uses Linguist to perform language detection and syntax highlighting. `console` [appears to be an alias for `shellsession`](https://github.com/github-linguist/linguist/blob/master/lib/linguist/languages.yml#L6476-L6487) (and `bash session`).

Docusaurus uses PrismJS to perform code block syntax highlighting. `console` is not [one of the supported languages](https://prismjs.com/#supported-languages) it detects, but `shellsession` and `shell` (with an `sh` alias) both are...making either of them a common choice.

Docusaurus comes with a subset of [commonly used languages](https://github.com/FormidableLabs/prism-react-renderer/blob/v1.3.0/src/vendor/prism/includeLangs.js). Bash is included in that list (with accepted aliases of `bash`, `sh`, and `shell`)...making the change here a natural substitution for `console`.

If that is still not desired, then additional languages, taken from the list of PrismJS languages, can be used if they are added to the `docusaurus.config.js` file, in the `themeConfig.prism.additionalLanguages` list. This PR still needs to add that setting with the additional code block languages in use...currently only `toml` but maybe an additional entry for whatever is decided here to use for representing a shell/console command/session.

Once it does, "enforcement" happens when trying to build the site since incorrect entries in that list will generate a build error. So...yes...currently these code fence "info strings" are ignored when they don't match one of the configured languages and no syntax highlighting is the result.

kylewillmon: > Docusaurus comes with a subset of [commonly used languages](https://github.com/FormidableLabs/prism-react-renderer/blob/v1.3.0/src/vendor/prism/includeLangs.js). Bash is included in that list (with accepted aliases of `bash`, `sh`, and `shell`)...making the change here a natural substitution for `console`.

You're confusing two very different "languages"...

1. Shell scripts (i.e., `sh`, `shell`, `bash`, etc.)
2. Console output (i.e., `console`, `shellsession`, `shell-session`, etc.)

We are in the second group here and we need to stay in that group. Annoyingly, PrismJS has [rejected `console`](https://github.com/PrismJS/prism/pull/2209) as an alias for `shell-session`. But it seems like `shellsession` is a widely supported value.

```suggestion
```shellsession
```

maxrake:
Despite the perceived consensus reached in https://github.com/phylum-dev/documentation/issues/24, it was decided that hosting a _separate_ Docusaurus site at `cli.phylum.io` from the `cli` repo would be undesirable. Instead, a _single_ Docusaurus site at `docs.phylum.io` will be hosted from the `documentation` repo. Source content from the `cli` repo will continue to be hosted there so the docs can live along side the code and be more likely to be updated. A different method for getting that content into the `documentation` repo will be pursued.

Closing this PR now that the changes worth keeping have been captured with #1226 ...

PR #1158, opened 2023-07-17
test: update test for phylum website change

### LINKED ISSUE #1157: Fix extension network permissions test
Currently, CI is failing due to a breakage in this [extension test](https://github.com/phylum-dev/cli/blob/main/cli/tests/extensions/permissions.rs#L59).

The reason is that the test extension [loads https://www.phylum.io](https://github.com/phylum-dev/cli/blob/main/cli/tests/fixtures/extensions/permissions/correct-net-perms/main.ts#L1), which no longer contains the lowercase `<!doctype html>`. We should adjust the predicate.

maxrake:
The `extensions::permissions::correct_net_permission_successful_install_and_run` test started failing after the main Phylum website was updated. It no longer contains the lowercase text `<!doctype html>` and now it is uppercase: `<!DOCTYPE html>`.

A workaround is also added to the `Test` workflow, to free up disk space. Without it, the linking step causes an error due to running out of disk space.

Fixes #1157

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,12 @@ jobs:
       - name: Checkout the repo
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
 
+      # Workaround to provide additional free space for testing.
+      # Reference: https://github.com/actions/runner-images/issues/2840
+      - name: Free disk space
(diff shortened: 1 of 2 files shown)


cd-work (approved):

maxrake:
Dang...the out of disk space issue is happening here. The proposed fix for that currently only exists in the `docusaurus` branch.

cd-work:
Feel free to add it here, I don't think that's a big problem.

cd-work (approved):

PR #1159, opened 2023-07-17
Run tests in parallel

kylewillmon:
Faster tests with less disk space required

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,7 +43,7 @@ jobs:
           rustup default stable
 
       - name: Clippy
-        run: cargo clippy --locked -- -D warnings
+        run: cargo clippy --locked --all-features --all-targets -- -D warnings
 
   build-and-test:
     strategy:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,8 @@ jobs:
       - name: Format check
         run: cargo +nightly fmt --all -- --check
 
-  test-matrix:
+  clippy:
+    if: github.event_name != 'schedule'
     strategy:
       fail-fast: false
       matrix:
@@ -41,12 +42,26 @@ jobs:
           rustup toolchain install --no-self-update stable --profile minimal -c clippy
           rustup default stable
 
+      - name: Clippy
+        run: cargo clippy --locked -- -D warnings
+
+  build-and-test:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04, [self-hosted, macos, arm64]]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install --no-self-update stable --profile minimal
+          rustup default stable
+
       - name: Build
         run: cargo build --locked
-
-      - name: Clippy
-        if: github.event_name != 'schedule'
-        run: cargo clippy --locked -- -D warnings
 
       - name: Test
         run: cargo test --locked
@@ -106,7 +121,7 @@ jobs:
   # This job reports the results of the matrixes above
   test:
     if: always()
-    needs: [test-matrix, all-features, oldstable]
+    needs: [clippy, build-and-test, all-features, oldstable]
     runs-on: ubuntu-latest
     steps:
       - if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,15 +56,33 @@ jobs:
       - name: Test
         run: cargo test --locked
 
+  oldstable:
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, [self-hosted, macos, arm64]]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+
       - name: Oldstable
         run: |
           oldstable=$(cat "./cli/Cargo.toml" | grep "rust-version" | sed 's/.*"\(.*\)".*/\1/')
           rustup toolchain install --profile minimal "$oldstable"
           cargo "+$oldstable" check
 
-      # Skip this step when the secret is unavailable
-      - if: github.secret_source == 'Actions'
-        name: All Features
+  all-features:
+    # Skip this job when the secret is unavailable
+    if: github.secret_source == 'Actions'
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, [self-hosted, macos, arm64]]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+
+      - name: All Features
         env:
           PHYLUM_API_KEY: ${{ secrets.PHYLUM_TOKEN_STAGING }}
         run: cargo test --all-features
@@ -93,9 +111,9 @@ jobs:
   # This job reports the results of the test matrix above
   test:
     if: always()
-    needs: test-matrix
+    needs: [test-matrix, all-features, oldstable]
     runs-on: ubuntu-latest
     steps:
-      - if: needs.test-matrix.result != 'success'
+      - if: contains(needs.*.result, 'failure')
         name: Fail the build
         run: exit 1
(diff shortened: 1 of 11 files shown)


.github/workflows/test.yml
```diff
@@ -87,9 +105,9 @@ jobs:
   # This job reports the results of the test matrix above
   test:
     if: always()
-    needs: test-matrix
+    needs: [test-matrix, all-features, oldstable]
     runs-on: ubuntu-latest
     steps:
-      - if: needs.test-matrix.result != 'success'
+      - if: contains(needs.*.result, 'failure')
```
kylewillmon, line 111: Here are [the docs](https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context) for anyone wondering what the possible states are.

There is no reason to fail the build for `skipped`. But maybe we should test the `cancelled` case?

cd-work: Yeah cancelled probably should fail. You don't want to allow people to cancel all the stops to bypass the merge checks.

cd-work (changes requested):
Is the goal to split the other steps in separate PRs? That seems like a lot of effort and I'd prefer having it all in one.

If there are no other steps planned, I'd suggest at least moving the `rustfmt` check to a separate job. It uses its own toolchain and only needs to run on a single platform. So that should allow speeding things up a good chunk.

.github/workflows/test.yml
```diff
@@ -50,21 +49,39 @@ jobs:
         run: cargo clippy --locked -- -D warnings
 
       - name: Test generating shell completion files
-        if: matrix.os == 'ubuntu-latest'
+        if: matrix.os == 'ubuntu-20.04'
         run: cargo xtask gencomp
```
cd-work, line 53: I guess we're keeping this here because it depends on CLI?

Though tbh the step confuses me a bit since it just generates the completions, which doesn't really seem helpful to me.

cd-work (approved):

.github/workflows/test.yml
cd-work: Just to be clear what I mean here: For docs generation we have an actual test that ensures the docs are up-to-date. That seems much more sensible.

kylewillmon: Good point. I've written an actual test to replace this.

.github/workflows/test.yml
```diff
@@ -90,12 +103,12 @@ jobs:
         if: github.event_name != 'schedule'
         run: shellcheck -o all -S style -s sh $(find . -iname "*.sh")
 
-  # This job reports the results of the test matrix above
+  # This job reports the results of the matrixes above
   test:
     if: always()
-    needs: test-matrix
+    needs: [test-matrix, all-features, oldstable]
```
maxrake, line 109:
```suggestion
    needs: [rustfmt, test-matrix, all-features, oldstable, deno-checks, shellcheck]
```
Shouldn't all test jobs in this workflow be included here...to keep from having to add them individually to the required status checks list in the `main` branch protection rule?

.github/workflows/test.yml
```diff
@@ -23,48 +36,48 @@ jobs:
...
-      - name: Format check
-        if: github.event_name != 'schedule'
-        run: cargo +nightly fmt --all -- --check
-
       - name: Build
         run: cargo build --locked
 
       - name: Clippy
         if: github.event_name != 'schedule'
         run: cargo clippy --locked -- -D warnings
```
maxrake, line 49:
```suggestion
        run: cargo lint
```
Not exactly what this PR is looking to change...but an easy change. The suggestion is to use the form that is more thorough and also more likely to be used locally. From the `.cargo/config` file in this repository:

`lint = "clippy --locked --all-features --all-targets -- -D warnings"`

maxrake (changes requested):
Time has been saved with the changes made so far, but it looks like more can be done. From the status checks that ran when this review was submitted:

image

That `test-matrix (ubuntu-20.04)` job really sticks out, taking over twice the time as the next longest job. Drilling down into it shows where the time is spent:

image

The `Build`, `Clippy`, and `Test` steps each account for a non-trivial run time. Is there a reason they are left in the same job instead of split out to parallel jobs of their own?

A very unscientific local test has shown that there is no real time savings in running them serially. For that matter, it also didn't appear to be better to run any two of the three serially. It really does appear that the fastest approach is to split out each step into it's own job...but maybe there are other factors to consider that aren't visible to this untrained author.

.github/workflows/test.yml
kylewillmon: Just adding it to `.cargo/config` doesn't make it more likely to be used locally... It just makes it more confusing by making us look elsewhere to figure out what's actually being run...

As for `--all-features --all-targets` , that may require code changes and is beyond the scope of this PR.

kylewillmon:
> A very unscientific local test has shown that there is no real time savings in running them serially. For that matter, it also didn't appear to be better to run any two of the three serially.

My very unscientific local tests show the exact opposite... `cargo test` runs significantly faster after `cargo build`

```
# cargo clean
# cargo build
       61.06 real       427.54 user        42.14 sys
# cargo test
       53.26 real       284.97 user        18.79 sys

# cargo clean
# cargo test
       74.58 real       472.08 user        47.19 sys
```

However, putting `cargo clippy` in the middle seems to slow things down (I suspect this is because `cargo clippy` uses the nightly compiler internally).

So I've split `cargo clippy` into its own test accordingly.

.github/workflows/test.yml
```diff
@@ -13,7 +13,21 @@ on:
...
+        run: rustup toolchain install --no-self-update nightly --profile minimal -c rustfmt
+
+      - name: Format check
+        run: cargo +nightly fmt --all -- --check
+
+  clippy:
+    if: github.event_name != 'schedule'
     strategy:
       fail-fast: false
       matrix:
```
kylewillmon, line 33: I'm slightly torn about running `cargo clippy` as a matrix... It feels excessive to run it on multiple targets, and it probably is. But we do use conditional compilation in a few files:

```
./cli/tests/sandbox.rs:12:    #[cfg(target_os = "linux")]
./cli/tests/sandbox.rs:14:    #[cfg(not(target_os = "linux"))]
./cli/src/update/unix.rs:109:    let arch = if cfg!(target_arch = "x86_64") {
./cli/src/update/unix.rs:111:    } else if cfg!(target_arch = "aarch64") {
./cli/src/update/unix.rs:116:    let os = if cfg!(target_os = "linux") {
./cli/src/update/unix.rs:118:    } else if cfg!(target_os = "macos") {
./cli/src/commands/sandbox.rs:7:#[cfg(target_os = "linux")]
./cli/src/commands/sandbox.rs:9:#[cfg(target_os = "linux")]
./cli/src/commands/sandbox.rs:57:        #[cfg(target_os = "linux")]
```

Does anybody feel strongly about this one way or the other?

andreaphylum: I think running clippy against all targets is going to be worth it when we want to be fully thorough, i.e. right before a merge or a release, otoh I find it a bit overkill for day-to-day activity as we don't really touch platform specific code very often.

xtask/src/main.rs
```diff
@@ -80,6 +80,20 @@ mod gencomp {
...
+
+        use super::*;
+
+        #[test]
+        fn generate_completions() {
+            let tmp_dir = tempdir().unwrap();
+
+            copy_completions(tmp_dir.path()).unwrap();
+        }
+    }
```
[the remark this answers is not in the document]
kylewillmon: Oh. You want to commit completion files to the repo? What benefit does that bring?

For background: The `cargo xtask gencomp` "test" was originally added in #741 after a `clap` migration caused shell completion file generation to panic. And because of our workflows, the panic went unnoticed until we attempted to cut a release. So the only purpose of this test was to ensure that this does not panic.

.github/workflows/test.yml
cd-work: > Just adding it to .cargo/config doesn't make it more likely to be used locally... It just makes it more confusing by making us look elsewhere to figure out what's actually being run...

100% agree, I've also never used `cargo lint` and I've said it in other PRs that I'd rather get rid of that than change this CI step to use it.

> As for --all-features --all-targets , that may require code changes and is beyond the scope of this PR.

That is fair though the changes are absolutely minimal so it should be fine tbh.

.github/workflows/test.yml
cd-work: I don't feel **strongly** about either direction, however I do think it provides value to run clippy on all target operating systems.

It's also probably not worth it to bikeshed over reducing number of build steps until we're likely to run into issues. And as far as I understand it we're far below any limit of parallel GitHub runners? So my vote's to just keep it because we can and it doesn't hurt.

cd-work:
> My very unscientific local tests show the exact opposite... cargo test runs significantly faster after cargo build

Important to note that running significantly faster after a `cargo build` doesn't necessarily mean it runs significantly faster than both in parallel.

Especially with the slow low-core runners, it's going to be beneficial to run things in parallel.

That said, a `build` step actually isn't necessary at all technically. A test always includes a build anyway? I never include build steps for my personal projects.

maxrake (dismissed):
A ~10min CI pipeline...now we're cooking with gas! 🚀 

There appears to be one small `cargo lint` fix to make (err... `cargo clippy --locked --all-features --all-targets -- -D warnings` fix) 😛

maxrake (approved):
LGTM

PR #1160, opened 2023-07-18
Bump to v5.5.0

kylewillmon:
CLI v5.4.0 was rejected before release due to a race condition in deno. The race condition was fixed in #1149 and now we can release CLI v5.5.0.

The `v5.4.0` tag will remain, but mention of the release has been removed from the Changelogs so that they reflect our actual releases.

Release-Version: v5.5.0


maxrake (approved):
LGTM

PR #1161, opened 2023-07-19
Ensure correct cargo is used

kylewillmon:
This patch fixes an issue that was occurring on our self-hosted runner. Since this runner also builds Homebrew bottles, it has `cargo` installed via Homebrew.

Unfortunately, this can lead to the wrong `cargo` being used, which means that we are not getting the version that we expect in our stable test and our oldstable tests fail outright because the +<toolchain> syntax does not work with `cargo` from Homebrew.

To ensure that we are using `cargo` from `rustup`, this patch explicitly calls `rustup run`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,12 +38,10 @@ jobs:
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
 
       - name: Install Rust toolchain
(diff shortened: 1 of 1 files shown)


andreaphylum (approved):

kylewillmon:
Using `rustup run` is apparently not good enough... So this needs to be fixed by changing the `$PATH` on the runner, which has now been done.

PR #1162, opened 2023-07-19
Use std::io::IsTerminal instead of atty

kylewillmon:
See the related [GitHub security advisory](https://github.com/advisories/GHSA-g98v-hv3f-hcfr). This patch removes our dependency on `atty`, because it is unmaintained.

Alternately, I could have used the [`is-terminal`](https://crates.io/crates/is-terminal) crate and avoided the MSRV bump. But rust 1.70.0 is already available in [Homebrew](https://formulae.brew.sh/formula/rust#default), which is the only package manager that we currently target. So there is no real need to support anything older.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3558,7 +3558,6 @@ version = "5.5.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
- "atty",
  "base64 0.21.2",
  "birdcage",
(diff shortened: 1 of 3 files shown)


maxrake (approved):
LGTM

cd-work (approved):

PR #1163, opened 2023-07-20
Add failure reason for ecosystem extensions

### LINKED ISSUE #1093: Provide failure report in pre-check extensions
# Overview
Each of the current package manager extensions offered by the CLI (`npm`, `pip`, `poetry`, and `yarn`) will report that there was a failure when analyzing the new package(s) to be installed. However, the report is simply that there was a failure and no further detail is provided. Provide more detail as to why the analysis failed.

## Additional Context

Here is an example of the current output:

```
❯ phylum pip install "pyyaml==5.3.1"

[notice] A new release of pip is available: 23.0.1 -> 23.1.2
[notice] To update, run: pip install --upgrade pip

[phylum] Analyzing packages…
[phylum] Supply Chain Risk Analysis - FAILURE
```

It _may_ be acceptable to provide a link to the project in the UI, to view the results of the analysis there. Check with @louislang before assuming that is _all_ that is required.

Providing a better report output should be possible when the API provides job evaluation results in a machine readable format (issue and PR are internal, but in progress).

## Acceptance Criteria
- [ ] Each package manager extension provides more analysis failure detail
  - [ ] `pip` extension
  - [ ] `npm` extension
  - [ ] `poetry` extension
  - [ ] `yarn` extension

cd-work:
This adds the new `getJobStatusRaw` and `checkPackagesRaw` extension APIs which allow retrieving a more detailed analysis report.

The ecosystem extensions are also switched over to `checkPackagesRaw`, allowing the output to match the CLI's analysis output, which provides additional details on failure.

Closes #1093.

### DIFF
--- a/extensions/npm/main.ts
+++ b/extensions/npm/main.ts
@@ -219,27 +219,13 @@ async function checkDryRun(subcommand: string, args: string[]) {
     return;
   }
 
-  const result = await PhylumApi.checkPackages(lockfile.packages);
+  const result = await PhylumApi.checkPackagesRaw(lockfile.packages);
+  logPackageAnalysisResults(result);
 
-  if (!result.is_failure && result.incomplete_count == 0) {
-    console.log(`[${green("phylum")}] Supply Chain Risk Analysis - SUCCESS\n`);
-  } else if (!result.is_failure) {
-    console.warn(
-      `[${yellow("phylum")}] Supply Chain Risk Analysis - INCOMPLETE`,
-    );
-    console.warn(
-      `[${
-        yellow(
-          "phylum",
-        )
-      }] Unknown packages were submitted for analysis, please check again later.\n`,
-    );
-    Deno.exit(126);
-  } else {
-    console.error(
-      `[${red("phylum")}] Supply Chain Risk Analysis - FAILURE\n`,
-    );
-    Deno.exit(127);
+  if (result.is_failure) {
+      Deno.exit(127);
+  } else if (result.incomplete_packages_count !== 0) {
+      Deno.exit(126);
   }
 }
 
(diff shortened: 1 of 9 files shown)


extensions/phylum.ts
```diff
@@ -20,6 +20,37 @@ export type ProcessOutput = {
...
+  source: RejectionSource;
+};
+
+export type RejectionSource = {
+  type: string;
+  tag: string | null;
+  domain: string | null;
+  severity: string | null;
+  description: string | null;
+  reason: string | null;
```
kylewillmon, line 51: Hmm... I expected `string | undefined` rather than `string | null`. And what the API sends **is** `string | undefined`... But [our type](https://github.com/phylum-dev/cli/blob/3d9d4158da7a54d9d6900dab406e443be9f31077/cli/src/types.rs#L102-L112) doesn't use `#[skip_serializing_if = "Option::is_none"]`, which means that these fields will be given a value of `null` when serializing...

I think we should update our type to serialize the same way that API does. And then we can update these fields to use the `?` syntax for optional dependencies.

```suggestion
  tag?: string;
  domain?: string;
  severity?: string;
  description?: string;
  reason?: string;
```

extensions/phylum.ts
```diff
@@ -178,6 +222,49 @@ export class PhylumApi {
...
+   *
+   * @returns Raw job analysis results
+   *
+   * Job analysis results example:
+   * ```
+   * {
+   *   is_failure: false,
+   *   incomplete_packages_count: 0,
+   *   help: "The analysis contains 346 package(s) Phylum has not yet processed,\npreventing a complete risk analysis. Phylum is processing these\npackages currently and should complete soon.\nPlease wait for up to 30 minutes, then re-run the analysis.\n\nThis repository analyzes the risk of new dependencies. An\nadministrator of this repository has set requirements via Phylum policy.\n\nIf you see this comment, one or more dependencies have failed Phylum's risk analysis.",
+   *   job_link?: "https://phylum.io/projects/62a91e89-65cb-4597-8d4b-6ef119b29c5c",
```
kylewillmon, line 239: Why does the example have `?`s on these fields?

```suggestion
   *   job_link: "https://phylum.io/projects/62a91e89-65cb-4597-8d4b-6ef119b29c5c",
```

extensions/phylum.ts
```diff
@@ -178,6 +222,49 @@ export class PhylumApi {
...
+   *     rejections: [
+   *       title: "Commercial license risk detected in birdcage@0.2.1",
+   *       suppressed: false,
+   *       source: {
+   *         type: "Issue",
+   *         tag?: "HL0030",
+   *         domain?: "license",
+   *         severity?: "high",
+   *         description?: "### Overview\nThis package uses the **GPL-3.0-or-later** license, which is a **high** risk level to commercial use.",
+   *         reason?: "risk level cannot exceed medium"
```
kylewillmon, line 254: Same as above...

```suggestion
   *         tag: "HL0030",
   *         domain: "license",
   *         severity: "high",
   *         description: "### Overview\nThis package uses the **GPL-3.0-or-later** license, which is a **high** risk level to commercial use.",
   *         reason: "risk level cannot exceed medium"
```

kylewillmon (changes requested):

extensions/phylum.ts
cd-work: Yeah I much prefer the `?` syntax anyway. I just didn't use that because other feels seemed to use `| null` already.

kylewillmon: They are different. `foo: string | null` implies that `foo` will always exist in the object, but might be `null`.

On the other hand `foo?: string` (which is just fancy syntax for `foo: string | undefined`) means that `foo` may not exist in the object.

extensions/npm/main.ts
```diff
@@ -257,3 +243,82 @@ async function restoreBackup() {
...
+
+// Write the analysis result status to STDOUT/STDERRR.
+function logPackageAnalysisResults(result: PolicyEvaluationResponseRaw) {
+  if (result.is_failure) {
+    console.error(
+      `[${red("phylum")}] Phylum Supply Chain Risk Analysis - FAILURE\n`,
+    );
+  } else if (result.incomplete_packages_count > 0) {
+    console.warn(
+      `[${yellow("phylum")}] Phylum Supply Chain Risk Analysis - INCOMPLETE`,
```
kylewillmon, line 255: Should this end with `\n` like the others?

```suggestion
      `[${yellow("phylum")}] Phylum Supply Chain Risk Analysis - INCOMPLETE\n`,
```

kylewillmon (changes requested):

kylewillmon (approved):

PR #1166, opened 2023-07-25
Use locksmith tokens

kylewillmon:
This patch updates CLI to generate and use Locksmith tokens (aka "API keys")

For backwards compatibility, OIDC refresh tokens are still supported in the `settings.yaml` or via the `PHYLUM_API_KEY` environment variable. But they are no longer generated.

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -2,7 +2,6 @@ use std::collections::HashSet;
 use std::time::Duration;
 
 use anyhow::{anyhow, Context};
-use phylum_types::types::auth::TokenResponse;
 use phylum_types::types::common::{JobId, ProjectId};
 use phylum_types::types::group::{
     CreateGroupRequest, CreateGroupResponse, ListGroupMembersResponse, ListUserGroupsResponse,
@@ -164,18 +163,21 @@ impl PhylumApi {
     pub async fn new(mut config: Config, request_timeout: Option<u64>) -> Result<Self> {
         // Do we have a refresh token?
         let ignore_certs = config.ignore_certs();
-        let tokens: TokenResponse = match &config.auth_info.offline_access() {
-            Some(refresh_token) => {
-                handle_refresh_tokens(refresh_token, ignore_certs, &config.connection.uri)
-                    .await
-                    .context("Token refresh failed")?
+        let refresh_token = match config.auth_info.offline_access() {
+            Some(refresh_token) => refresh_token.clone(),
+            None => {
+                let refresh_token =
+                    handle_auth_flow(AuthAction::Login, ignore_certs, &config.connection.uri)
+                        .await
+                        .context("User login failed")?;
+                config.auth_info.set_offline_access(refresh_token.clone());
+                refresh_token
             },
-            None => handle_auth_flow(AuthAction::Login, ignore_certs, &config.connection.uri)
-                .await
-                .context("User login failed")?,
         };
 
-        config.auth_info.set_offline_access(tokens.refresh_token.clone());
+        let tokens = handle_refresh_tokens(&refresh_token, ignore_certs, &config.connection.uri)
+            .await
+            .context("Token refresh failed")?;
 
         let mut headers = HeaderMap::new();
         // the cli runs a command or a few short commands then exits, so we do
@@ -210,8 +212,8 @@ impl PhylumApi {
         reauth: bool,
     ) -> Result<AuthInfo> {
         let action = if reauth { AuthAction::Reauth } else { AuthAction::Login };
-        let tokens = handle_auth_flow(action, ignore_certs, api_uri).await?;
-        auth_info.set_offline_access(tokens.refresh_token);
+        let refresh_token = handle_auth_flow(action, ignore_certs, api_uri).await?;
+        auth_info.set_offline_access(refresh_token);
         Ok(auth_info)
     }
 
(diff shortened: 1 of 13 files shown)


cli/src/commands/auth.rs
```diff
@@ -29,30 +41,46 @@ async fn handle_auth_login(
...
     match user_info {
         Ok(user) => {
             print_user_success!(
-                "Currently authenticated as '{}' with long lived refresh token",
-                user.email
+                "Currently authenticated as '{}<{}>' via {}",
+                user.name.map_or(" ".into(), |mut n| {
+                    n.push(' ');
+                    n
+                }),
```
andreaphylum, line 81: If I interpret this correctly, this should output `...as ' <email>' via...` if `user.name` is `None`.
Wouldn't that be equivalent to having `...as '{} <{}>' via...` in the format string + `unwrap_or` on the field? Maybe an empty string was intended in the first argument?

Also it's super minor and I'm not sure it is not already getting optimized away anyway, but considering that `"".into()` allocates a string, I'd use `map_or_else` instead.

```suggestion
                user.name.map_or_else(|| "".into(), |mut n| {
                    n.push(' ');
                    n
                }),
```

andreaphylum (changes requested):

cli/src/commands/auth.rs
kylewillmon: Yep. This should have been empty string... I've updated the code to use `Default::default()` instead.

andreaphylum (approved):

cli/src/commands/auth.rs
```diff
@@ -29,30 +41,46 @@ async fn handle_auth_login(
...
+            return Ok(ExitCode::NotAuthenticated);
+        },
+    };
 
     // Create a client with our auth token attached.
     let api = PhylumApi::new(config, timeout).await?;
 
     let user_info = api.user_info().await;
 
+    debug!("User info reponse: {:?}", user_info);
```
cd-work, line 72: Just making sure this new log is intentional and not left over from debugging.

cli/src/app.rs
```diff
@@ -188,16 +188,33 @@ pub fn add_subcommands(command: Command) -> Command {
...
-                        Arg::new("reauth")
-                            .action(ArgAction::SetTrue)
-                            .short('r')
-                            .long("reauth")
-                            .help("Force a login prompt"),
+                    Command::new("register").about("Register a new account").arg(
+                        Arg::new("token-name")
+                            .action(ArgAction::Set)
+                            .short('n')
+                            .long("token-name")
```
cd-work, line 196: It actually took me a second to figure out what "token-name" means in this context. I confused it with the actual locksmith token for a second.

I probably got myself confused a bit, but it might be worth taking into account and maybe expanding the help a bit?

cli/src/api/mod.rs
```diff
@@ -164,26 +164,30 @@ impl PhylumApi {
...
-            None => handle_auth_flow(AuthAction::Login, ignore_certs, &config.connection.uri)
-                .await
-                .context("User login failed")?,
         };
 
-        config.auth_info.set_offline_access(tokens.refresh_token.clone());
+        let access_token =
+            handle_refresh_tokens(&refresh_token, ignore_certs, &config.connection.uri)
+                .await
+                .context("Token refresh failed")?;
```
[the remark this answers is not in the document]
cd-work: Does it **have** to be called `handle_`? I find that a rather useless verb to have in a function name to begin with anyway.

cli/src/auth/mod.rs
```diff
@@ -6,3 +6,7 @@ mod ip_addr_ext;
 pub mod jwt;
 mod oidc;
 mod server;
+
+pub fn is_locksmith_token(token: impl AsRef<str>) -> bool {
+    token.as_ref().starts_with("ph0_")
```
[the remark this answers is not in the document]
kylewillmon: That crate is private. This one is public. We cannot depend on token store here.

cli/src/auth/oidc.rs
```diff
@@ -175,18 +186,42 @@ pub async fn fetch_oidc_server_settings(
...
+
 fn build_grant_type_auth_code_post_body(
     redirect_url: &Url,
     authorization_code: &AuthorizationCode,
     code_verfier: &CodeVerifier,
+    token_name: Option<String>,
 ) -> Result<HashMap<String, String>> {
     let body = hashmap! {
-        "client_id".to_owned() => OIDC_CLIENT_ID.to_owned(),
+        "client_id".to_owned() => LOCKSMITH_CLIENT_ID.to_owned(),
```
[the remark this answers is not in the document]
kylewillmon: Changing it has no effect. But it's still a required parameter. I could put a bogus value here for now. But we would need to change the Locksmith API to allow omitting it.

cli/src/app.rs
kylewillmon: Yeah. "token-name" was the best I could come up with. We definitely use many different names for the token itself (e.g., "token", "API Key", "API Access Token", etc.)...

I'll try to make the help text more descriptive.

cli/src/auth/mod.rs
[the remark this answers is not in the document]
cd-work: Duh.

cli/src/auth/oidc.rs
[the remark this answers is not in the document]
cd-work: We probably should.

cli/src/app.rs
cd-work: I think token name is fine, I'd just make the help text more verbose.

cli/src/commands/auth.rs
kylewillmon: Yes. It's intentional

Issue #1169, opened 2023-08-01
Lockfile paths should be relative


kylewillmon:
After ae05e29, CLI is now including lockfile paths when submitting jobs. Unfortunately, the file paths being used are absolute. For example:

```sh-session
> cat .phylum_project 
id: b21c6236-c031-456e-b903-decb6797cf4f
name: cli
created_at: 2023-08-01T14:49:47.493642-05:00
group_name: null
lockfiles:
- path: ./requirements.txt
  type: pip
- path: ./Cargo.lock
  type: cargo
 
> phylum analyze
✅ Successfully parsed lockfile "/Users/kyle/code/cli/./requirements.txt" as type: pip
✅ Successfully parsed lockfile "/Users/kyle/code/cli/./Cargo.lock" as type: cargo
✅ Job ID: 7204e8d6-dd1b-4435-9930-b9cacbe2b93b
```

And those absolute paths show in the UI.

![Absolute path in UI](https://github.com/phylum-dev/cli/assets/616067/4aea6a75-2dc9-418e-8ec0-946774c37f68)

These paths should be relative to avoid leaking user directory information.

cd-work:
Note that relative here should mean relative to the `.phylum_project` file, not relative to some working directory.

PR #1172, opened 2023-08-04
Use relative paths when project root is set

ejortega:
The lockfile paths were being submitted with an absolute path if the project root was set.

This can be caused by using relative lockfile paths in in the `.phylum_project`.

Closes https://github.com/phylum-dev/cli/issues/1169

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 - Include lockfile paths when analyzing projects
+- Remove root path from lockfile paths
 
 ## [5.5.0] - 2023-07-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Include lockfile paths when analyzing projects
 - Generate and use API Keys instead of OpenID Connect tokens
-- Use relative paths for project lockfile path submissions
 
 ### Fixed
 - Search for manifests' lockfiles in parent, rather than child directories
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Include lockfile paths when analyzing projects
 - Generate and use API Keys instead of OpenID Connect tokens
-- Remove root path from lockfile paths
+- Use relative paths for project lockfile path submissions
 
 ### Fixed
 - Search for manifests' lockfiles in parent, rather than child directories
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -39,7 +39,17 @@ impl IntoIterator for ParsedLockfile {
     type Item = PackageDescriptorAndLockfile;
 
     fn into_iter(self) -> Self::IntoIter {
-        ParsedLockfileIterator { path: self.path, packages: self.packages.into_iter() }
+        // Get .phylum_project path
+        let root = phylum_project::get_current_project().map(|p| p.root().to_owned());
+        // Strip root path when set
+        let relative_path = match root {
+            Some(base) => match self.path.strip_prefix(&base) {
+                Ok(rel_path) => rel_path.to_path_buf(),
+                Err(_) => self.path.to_owned(),
+            },
+            None => self.path.to_owned(),
+        };
+        ParsedLockfileIterator { path: relative_path, packages: self.packages.into_iter() }
     }
 }
 
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -84,7 +84,7 @@ pub fn parse_lockfile(
     let project_root = project_root.to_owned();
 
     // Attempt to strip root path
-    let path = strip_root_path(path, &project_root)?;
+    let path = strip_root_path(path, &project_root);
 
     // Attempt to parse with all known parsers as fallback.
     let (format, lockfile) = match format {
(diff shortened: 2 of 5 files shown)


cd-work (changes requested):
We should also strip prefixes from absolute paths to avoid storing host and user names.

cli/src/commands/parse.rs
```diff
@@ -14,6 +14,7 @@ use crate::{config, print_user_warning};
 
 pub struct ParsedLockfile {
     pub path: PathBuf,
+    pub project_root: Option<PathBuf>,
```
kylewillmon, line 17: This seems overly complicated... Why not just strip the project root while creating the `ParsedLockfile` struct instead. That way `path` is always exactly what we submit as the lockfile path.

cli/src/commands/parse.rs
```diff
@@ -275,6 +285,23 @@ where
     None
 }
 
+/// Strip root prefix from lockfile paths
```
cd-work, line 288:
```suggestion
/// Strip root prefix from lockfile paths.
```

cli/src/commands/parse.rs
```diff
@@ -275,6 +285,23 @@ where
...
+fn strip_root_path(path: PathBuf, project_root: &Option<PathBuf>) -> PathBuf {
```
cd-work, line 289: We should probably do canonicalization. Especially in the case where the project root exists already, we don't want to be relative to the current directory.

cd-work (changes requested):

cli/src/commands/parse.rs
ejortega: I think I covered the issue you brought up. Previously running analyze in a subdirectory was returning path relative to that directory vs the project.

cli/src/commands/parse.rs
```diff
@@ -357,4 +413,65 @@ mod tests {
...
+        // Create a sample project directory named "sample" inside the "projects"
+        // directory. Also create a "Cargo.lock" file inside the "sample"
+        // directory.
+        let sample_dir = all_projects.join("sample");
+        let lockfile_path = sample_dir.join("Cargo.lock");
+        fs::create_dir_all(&sample_dir).unwrap();
+        File::create(&lockfile_path).unwrap();
+
+        // Change the current directory to the "sample" project directory.
+        env::set_current_dir(&sample_dir).unwrap();
```
kylewillmon, line 437: I'm a bit nervous about changing the current working directory inside of a unit test. These tests run in parallel, so if there is any other test that relies on the current working directory, this could cause issues...

I wonder if it would be beneficial to refactor a bit so that we can test a function that doesn't rely on the current working directory.

kylewillmon (changes requested):

cli/tests/parse.rs
```diff
@@ -25,3 +25,22 @@ fn parse_with_project_lockfile() {
...
+    let config = "id: 00000000-0000-0000-0000-000000000000\nname: test\ncreated_at: \
+                  2000-01-01T00:00:00.0Z\nlockfile_path: ./package-lock.json\nlockfile_type: npm";
+    fs::write(temp_path.join(".phylum_project"), config).unwrap();
+
+    // Copy lockfile to temp dir.
+    fs::copy("../tests/fixtures/package-lock.json", temp_path.join("./package-lock.json")).unwrap();
+
+    let not_contain_var_folders =
+        predicate::function(|x: &str| !x.contains("\"lockfile\": \"/private/var/folders"));
+    test_cli.cmd().args(["parse"]).assert().success().stdout(not_contain_var_folders);
```
kylewillmon, line 45: This test relies on `/private/var/folders`, which is specific to MacOS. To make this test work across platforms, create a temp folder with a specific name and check for that instead:

```suggestion
    // Setup CLI with temp dir.
    let tempdir = TempDir::new().unwrap();
    let sensitive_dir = tempdir.path().join("sensitive_dir_name");
    fs::create_dir_all(&sensitive_dir).unwrap();
    let test_cli = TestCli::builder().cwd(sensitive_dir.clone()).build();

    // Write .phylum_project to temp dir.
    let config = "id: 00000000-0000-0000-0000-000000000000\nname: test\ncreated_at: \
                  2000-01-01T00:00:00.0Z\nlockfile_path: ./package-lock.json\nlockfile_type: npm";
    fs::write(sensitive_dir.join(".phylum_project"), config).unwrap();

    // Copy lockfile to temp dir.
    fs::copy("../tests/fixtures/package-lock.json", sensitive_dir.join("./package-lock.json"))
        .unwrap();

    let not_contain_var_folders = predicate::str::contains("sensitive_dir_name").not();
    test_cli.cmd().args(["parse"]).assert().success().stdout(not_contain_var_folders);
```

kylewillmon (changes requested):

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Include lockfile paths when analyzing projects
 - Generate and use API Keys instead of OpenID Connect tokens
+- Use relative paths for project lockfile path submissions
```
kylewillmon, line 14: There is no need for a changelog entry in the PR. These changes are fully covered by the changelog entry above (which is still unreleased):

```
- Include lockfile paths when analyzing projects
```

So you can remove this line

```suggestion
```

cli/tests/parse.rs
```diff
@@ -29,18 +30,20 @@ fn parse_with_project_lockfile() {
...
+    fs::write(sensitive_dir.join(".phylum_project"), config).unwrap();
 
     // Copy lockfile to temp dir.
-    fs::copy("../tests/fixtures/package-lock.json", temp_path.join("./package-lock.json")).unwrap();
+    fs::copy("../tests/fixtures/package-lock.json", sensitive_dir.join("./package-lock.json"))
+        .unwrap();
 
-    let not_contain_var_folders =
-        predicate::function(|x: &str| !x.contains("\"lockfile\": \"/private/var/folders"));
+    let not_contain_var_folders = predicate::str::contains("sensitive_dir_name").not();
```
kylewillmon: I just realized that I didn't change this variable name...

kylewillmon (approved):

kylewillmon (approved):

cd-work (approved):

Issue #1170, opened 2023-08-01
Lockfile detection choosing wrong lockfile


kylewillmon:
When I run `phylum analyze` on `Cargo.toml` in the root of CLI, the lockfile detection code chooses to analyze `vulnreach_types/Cargo.lock` instead...

```sh-session
> phylum analyze /Users/kyle/code/cli/Cargo.toml 
⚠️  "/Users/kyle/code/cli/Cargo.toml" is not a lockfile, using "/Users/kyle/code/cli/vulnreach_types/Cargo.lock" instead
✅ Successfully parsed lockfile "/Users/kyle/code/cli/vulnreach_types/Cargo.lock" as type: cargo
⚠️  No packages found in lockfile
```

### Expected behavior

Lockfile detection should instead detect and use the `Cargo.lock` that is in the same directory as the `Cargo.toml` that I provided.

PR #1173, opened 2023-08-04
fix: update lockfile detection to search root directory first

ejortega:
The current search could potentially miss lock files in the current path.
This changes to search the current path first before sub directories.

Either way I imagine could still choose an incorrect lockfile related to another ecosystem if present.

Closes https://github.com/phylum-dev/cli/issues/1170

## Checklist

### DIFF
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -181,7 +181,26 @@ fn find_manifest_format(path: &Path) -> Option<(LockfileFormat, Option<PathBuf>)
 
     // Look for formats which already have a lockfile generated.
     let manifest_lockfile = formats.find_map(|format| {
+        // Check the root directory for a lockfile
+        let root_dir_entries = match fs::read_dir(manifest_dir) {
+            Ok(entries) => entries,
+            Err(_) => return None,
+        };
+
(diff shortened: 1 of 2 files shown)


andreaphylum (approved):
LGTM. We may want to rethink the detection priority at some point (e.g. if top-level `Cargo.lock` is not found, what makes the one in `vulnreach_types` the next best choice?) but that's a larger (and not necessarily urgent) discussion.

cli/src/commands/parse.rs
```diff
@@ -181,7 +181,26 @@ fn find_manifest_format(path: &Path) -> Option<(LockfileFormat, Option<PathBuf>)
...
+        for entry_result in root_dir_entries {
+            match entry_result {
+                Ok(entry) => {
+                    if format.parser().is_path_lockfile(&entry.path()) {
+                        return Some((format, entry.path()));
+                    }
+                },
+                Err(_) => continue,
+            }
+        }
```
cd-work, line 199: This is not a fix to the problem. Now you have the same issue just one level down. There's no guarantee that the lockfile will actually be at the root level.

ejortega: Okay. I reopened the issue.

PR #1174, opened 2023-08-04
Reverse search direction for manifests' lockfiles

cd-work:
Previously when a manifest was specified as the target for an analyze or
parse operation, the directory would be walked downwards depth-first to
find a lockfile for this manifest. This would both prefer lockfiles
farther away from the manifest and allow lockfiles from subdirectories.

This patch reverses the search direction for lockfiles, since they're
generally at or above their corresponding manifest. The new logic
prefers lockfiles closest to the manifest and searches the directory
tree upwards from there.

Closes https://github.com/phylum-dev/cli/issues/1170.

### DIFF
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -7,7 +7,6 @@ use std::{fs, io};
 use anyhow::{anyhow, Context, Result};
 use phylum_lockfile::{LockfileFormat, Package, PackageVersion, Parse, ThirdPartyVersion};
 use phylum_types::types::package::{PackageDescriptor, PackageDescriptorAndLockfile};
-use walkdir::WalkDir;
 
 use crate::commands::{CommandResult, ExitCode};
 use crate::{config, print_user_warning};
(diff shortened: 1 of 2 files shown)


cd-work (commented):
Technically this will go beyond the phylum project root to search for the lockfile, however I don't think this should be a problem in practice.

It would be fairly simple to limit this by setting a "stop" at the depth of the `.phylum_project` file, I just don't think this is worth it personally.

ejortega (dismissed):
LGTM. It probably should stop at the project level, but it might not be an issue in most instances.

ejortega (approved):

cli/src/commands/parse.rs
```diff
@@ -251,8 +249,36 @@ fn filter_packages(mut packages: Vec<Package>) -> Vec<PackageDescriptor> {
...
+    mut origin: &Path,
+    mut predicate: P,
+) -> Option<PathBuf>
+where
+    P: FnMut(&Path) -> bool,
+{
+    for _ in 0..=MAX_DEPTH {
+        for entry in fs::read_dir(origin).ok()?.flatten() {
+            let path = entry.path();
+            if predicate(&path) {
```
kylewillmon, line 267: Is it a problem that this would also find a directory named `Cargo.lock`?

The previous code also didn't distinguish files or directories (or symlinks). So we can ignore it for this PR. And I don't **think** it will cause any real world issues, so maybe we don't need to address it at all.

kylewillmon (approved):

cli/src/commands/parse.rs
cd-work: Honestly that should probably be checked in the `is_path_lockfile` functions if anything.

Feel free to open an issue about this if we ever have absolutely **nothing** else to do, but I think this scenario is **very** unlikely to ever cause problems.

Issue #1176, opened 2023-08-08
List and revoke API keys


kylewillmon:
There are not currently commands in CLI to list or revoke API keys (i.e., locksmith tokens). These commands would certainly be useful to some users.

Here is an example of what these might look like:

```
> phylum token list

Name        Created           Last Used
JRR Token   2023-07-31 20:32  2023-08-08 17:01
Token Ring  2023-07-31 21:50  never
```

```
> phylum token revoke "Token Ring"

✅ Successfully revoked token!
```

kylewillmon:
Marking as https://github.com/phylum-dev/cli/labels/low%20priority because this functionality is readily available in the UI and there is currently no indication that users want this in CLI.

PR #1180, opened 2023-08-09
Add `auth list-tokens` subcommand

cd-work:
This patch adds a new `list-tokens` subcommand to the existing `auth` subcommand which lists all locksmith tokens associated with the logged in user.

See #1176.

### DIFF
--- a/cli/src/api/endpoints.rs
+++ b/cli/src/api/endpoints.rs
@@ -4,7 +4,11 @@ use url::{ParseError, Url};
 
 use super::JobId;
 
+/// Phylum API base path.
 const API_PATH: &str = "api/v0/";
+
+/// Locksmith API base path.
+const LOCKSMITH_PATH: &str = "locksmith/v1/";
 
 // This wrapper provides important context to the user when their configuration
 // has a bad URL. Without it, the message can be something like "empty host".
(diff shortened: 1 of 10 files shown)


cd-work:
This is what the new endpoint's output looks like: 

tmp

cli/src/api/endpoints.rs
```diff
@@ -191,6 +200,10 @@ fn get_api_path(api_uri: &str) -> Result<Url, BaseUriError> {
     Ok(parse_base_url(api_uri)?.join(API_PATH)?)
 }
 
+fn get_locksmith_path(api_uri: &str) -> Result<Url, BaseUriError> {
```
kylewillmon, line 203: Might as well use this for `locksmith_discovery()` above as well

cli/src/types.rs
```diff
@@ -111,6 +111,15 @@ pub struct RejectionSource {
     pub reason: Option<String>,
 }
 
+/// Locksmith token details accessible after creation.
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Debug)]
+pub struct UserToken {
+    pub name: String,
+    pub creation_time: DateTime<Utc>,
+    pub access_time: Option<DateTime<Utc>>,
+    pub expiry: Option<DateTime<Utc>>,
```
kylewillmon, line 120: Is this just optional because we haven't deployed expiry to prod? Are we going to come back later and remove the `Option` here?

cd-work: Yeah basically just to be safe. I don't think there's any drawbacks to having it optional though.

cli/src/api/endpoints.rs
cd-work: You mean to avoid the proxy? I suppose that should work just fine.

kylewillmon (dismissed):

maxrake (approved):
Tested locally and works as expected/intended. Some questions were asked out of band and a suggested change for the CHANGELOG was already implemented. LGTM.

PR #1187, opened 2023-08-14
Fix reported lines in extension errors

### LINKED ISSUE #961: TypeScript extensions report incorrect error locations
When running into an error with a Deno extension written in TypeScript, the error locations reported are often incorrect. Not only is this unhelpful while debugging, it might actively lead the developer into a direction that is completely wrong.

I'm not sure if this is also the case with deno, since the culprit is likely the transpilation process, but if we cannot fix this we should probably omit the lines of the error messages completely.

cd-work:
This fixes an issue with extensions where the reported line for an error did not match the supplied code, since the error was based on the transpiled JavaScript.

To make sure the error can be mapped from JavaScript back to the original TypeScript, a transpiler cache has been introduced which is used to retrieve the original source map when an error is encountered.

Fixes #961.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -68,6 +68,7 @@ birdcage = { version = "0.2.0" }
 libc = "0.2.135"
 ignore = { version = "0.4.20", optional = true }
 uuid = "1.4.1"
+dashmap = "5.5.0"
 
 [dev-dependencies]
 assert_cmd = "2.0.4"
--- a/cli/src/deno.rs
+++ b/cli/src/deno.rs
@@ -108,12 +108,15 @@ fn print_js_error(error: Error) -> CommandResult {
 /// See https://github.com/denoland/deno/blob/main/core/examples/ts_module_loader.rs.
 struct ExtensionsModuleLoader {
     extension_path: Rc<PathBuf>,
-    transpiler: Rc<Transpiler>,
+    source_mapper: Rc<SourceMapper>,
 }
 
 impl ExtensionsModuleLoader {
     fn new(extension_path: PathBuf) -> Self {
-        Self { extension_path: Rc::new(extension_path), transpiler: Rc::new(Transpiler::new()) }
+        Self {
+            extension_path: Rc::new(extension_path),
+            source_mapper: Rc::new(SourceMapper::new()),
+        }
     }
 
     async fn load_from_filesystem(extension_path: &Path, path: &Url) -> Result<String> {
@@ -164,12 +167,12 @@ impl ModuleLoader for ExtensionsModuleLoader {
     ) -> Pin<Box<ModuleSourceFuture>> {
         let module_specifier = module_specifier.clone();
         let extension_path = self.extension_path.clone();
-        let transpiler = self.transpiler.clone();
+        let source_mapper = self.source_mapper.clone();
 
         Box::pin(async move {
             // Inject Phylum API module.
             if module_specifier.as_str() == "deno:phylum" {
-                return phylum_module(&transpiler);
+                return phylum_module(&source_mapper);
             }
 
             // Determine source file type.
@@ -207,8 +210,10 @@ impl ModuleLoader for ExtensionsModuleLoader {
 
             if should_transpile {
                 let transpiled =
-                    transpiler.transpile(module_specifier.to_string(), code, media_type)?;
+                    source_mapper.transpile(module_specifier.to_string(), code, media_type)?;
                 code = transpiled.text.clone();
+            } else {
+                source_mapper.source_cache.insert(module_specifier.to_string(), code.clone());
             }
 
             Ok(ModuleSource::new(module_type, code.into(), &module_specifier))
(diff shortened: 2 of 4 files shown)


cli/src/deno.rs
```diff
@@ -199,35 +206,86 @@ impl ModuleLoader for ExtensionsModuleLoader {
...
+                text_info: SourceTextInfo::from_string(code.into()),
+                specifier: specifier.clone(),
+                capture_tokens: false,
+                scope_analysis: false,
+                maybe_syntax: None,
+                media_type,
+            })?;
+
+            // Transpile to JavaScript.
+            let options = EmitOptions { inline_source_map: false, ..EmitOptions::default() };
```
cd-work, line 267: Disabling the inlining of the source map here is probably the change I'm most uncertain about, which is why I asked about ideal solutions on the Deno repo:

https://github.com/denoland/deno/issues/19617#issuecomment-1678031648

I do believe either way should work just fine though and I doubt it makes any difference for our usecase.

cli/Cargo.toml
```diff
@@ -68,6 +68,7 @@ birdcage = { version = "0.2.0" }
 libc = "0.2.135"
 ignore = { version = "0.4.20", optional = true }
 uuid = "1.4.1"
+dashmap = "5.5.0"
```
cd-work, line 71: I was pondering if the `dashmap` dependency was really worth it or if a simple locking `HashMap` would be better (there should be zero contention anyway). But I noticed that we depend on it already anyway, so I feel like it shouldn't be an issue.

cli/src/deno.rs
```diff
@@ -199,35 +206,86 @@ impl ModuleLoader for ExtensionsModuleLoader {
...
+        let transpiled = self.transpiler.cache.get(file_name)?;
+        transpiled.source_map.clone().map(|map| map.into_bytes())
+    }
+
+    fn get_source_line(&self, file_name: &str, line_number: usize) -> Option<String> {
+        let source_map = self.get_source_map(file_name)?;
+
+        let text = String::from_utf8(source_map).ok()?;
+
+        text.lines().nth(line_number).map(|line| line.to_owned())
```
kylewillmon, line 230: This gets a particular line of the source map... Not the source itself. This leads to very weird errors:

```sh-session
> cat foo/main.ts
throw new Error();

> ./target/debug/phylum extension run -y ./foo
Extension error: Uncaught Error
{"version":3,"sources":["file:///Users/kyle/code/cli/foo/main.ts"],"sourcesContent":["throw new Error();\n"],"names":[],"mappings":"AAAA,MAAM,IAAI"}
      ^
    at file:///Users/kyle/code/cli/foo/main.ts:1:7
```

kylewillmon (changes requested):

cli/src/deno.rs
kylewillmon: For what it's worth, it looks like the deno CLI keeps the inline source map and then [extracts it from the code](https://github.com/denoland/deno/blob/625bd3905047173390eeaffd0ffcbb0bf340bb52/cli/util/text_encoding.rs#L57-L71) when needed.

cli/src/deno.rs
```diff
@@ -199,35 +206,89 @@ impl ModuleLoader for ExtensionsModuleLoader {
...
-    })?;
-    Ok(parsed.transpile(&Default::default())?.text)
+impl SourceMapGetter for ExtensionsModuleLoader {
+    fn get_source_map(&self, file_name: &str) -> Option<Vec<u8>> {
+        let transpiled = self.transpiler.transpiled_cache.get(file_name)?;
+        transpiled.source_map.clone().map(|map| map.into_bytes())
+    }
+
+    fn get_source_line(&self, file_name: &str, line_number: usize) -> Option<String> {
+        let source = self.transpiler.source_cache.get(file_name)?;
```
kylewillmon, line 226: Unfortunately, this fails for any source that we didn't need to transpile (i.e., JavaScript files)... So for those files, we would need to cache the source with no source map.

You can see an example of this regression here:

```sh-session
> cat foo/main.ts
import { foo } from "./foo.js";

foo();

> cat foo/foo.js
export function foo() {
  throw new Error();
}

> phylum extension run -y ./foo
Extension error: Uncaught Error
  throw new Error();
        ^
    at foo (file:///Users/kyle/code/cli/foo/foo.js:2:9)
    at file:///Users/kyle/code/cli/foo/main.ts:2:1

> ./target/debug/phylum extension run -y ./foo
Extension error: Uncaught Error
    at foo (file:///Users/kyle/code/cli/foo/foo.js:2:9)
    at file:///Users/kyle/code/cli/foo/main.ts:3:1
```

kylewillmon (changes requested):

kylewillmon (approved):

cli/src/deno.rs
cd-work: Yeah that's what I expected, which I think is definitely worse. Though it might be the only option in some scenarios?

PR #1191, opened 2023-08-15
Add .editorconfig file

### LINKED ISSUE #1154: SPIKE: Identify a Markdown formatting standard
This repository has a large number of markdown files (`*.md`) and no established standard for formatting and linting those files. This issue is meant to be a research spike only. Possible solutions should be identified, discussed, and documented for reference. The desired end state is a solution reached through consensus, but not implemented. Implementation will happen in a different issue, assuming there is a solution that requires changes.

Goals:

* Standards for formatting and linting documentation files (Markdown) are documented
* Standards are adopted/enforced for the entire company
* Standards are able to be enforced with automation, both locally and in CI
* It might be nice to also account for `*.mdx` files, which are another option for Docusaurus docs

Resources/links:

* Blog: [To Wrap or Not to Wrap: One of Life's Greatest Questions](https://about.gitlab.com/blog/2016/10/11/wrapping-text/)
* Example Markdown style guides
  * [GitLab UI](https://docs.gitlab.com/ee/development/documentation/styleguide/#markdown)
  * [for about.gitlab.com](https://about.gitlab.com/handbook/markdown-guide)
* [Example config file for markdownlint](https://gitlab.com/gitlab-org/gitlab/-/blob/master/.markdownlint.yml)

Some possible tools to consider include:

* [`markdownlint`](https://github.com/DavidAnson/markdownlint)
* [`markdownlint-cli`](https://github.com/igorshubovych/markdownlint-cli)
* [`markdownlint-cli2`](https://github.com/DavidAnson/markdownlint-cli2)
* [EditorConfig](https://editorconfig.org/)
* Others (this list is not exhaustive)

kylewillmon:
Many of the items in this file are already enforced by other tools (i.e., `.gitattributes` for line endings, `rustfmt` for `.rs` files). But this file will assist in areas where strict automated tools do not exist or are not wanted (e.g., Markdown).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.{rs,toml}]
+indent_style = space
+indent_size = 4
+
+[*.md]
+max_line_length = 80


.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.{rs,toml}]
+indent_style = space
+indent_size = 4
+
+[*.md]
+max_line_length = 80
```
kylewillmon, line 14: I picked 80 to match the limit currently being enforced on `extensions/**/*.md` by `deno fmt`.

cd-work (approved):

maxrake (changes requested):
The `.editorconfig` solution proposed in this PR is not functional with VS Code. At least, personal trial and error has not resulted in a configuration that works to format markdown files with VS Code and the `prettier` extension.

This PR _may_ be okay on it's own but it should not be linked to #1154 for closing since that is a spike/research issue that states:

> The desired end state is a solution reached through consensus, but not implemented. Implementation will happen in a different issue, assuming there is a solution that requires changes.

There are other goals (e.g., enforcement and standards documentation) that should also be addressed within the spike ticket before any implementation is brought forward. Basically, the request here is to propose solutions in #1154 and complete that issue before bringing forward any PRs related to it.

kylewillmon:
> The `.editorconfig` solution proposed in this PR is not functional with VS Code. At least, personal trial and error has not resulted in a configuration that works to format markdown files with VS Code and the `prettier` extension.

That is because `prettier` does not wrap prose in Markdown by default. And configuring the extension with `"prettier.proseWrap": "always"` does not help because the extension sees that we have an `.editorconfig` file and ignores the extension config (as seen in line from the extension log):

```
["INFO" - 10:53:51 AM] Detected local configuration (i.e. .prettierrc or .editorconfig), VS Code configuration will not be used
```

So to get prettier to wrap prose from the VS Code extension, we would need to add a `.prettierrc` file containing `{"proseWrap": "always"}` to the root of the repo.

But even that has major issues... Because `prettier` only supports formatting entire documents (for Markdown, at least). And the goal of using `.editorconfig` was to **avoid** locking developers into a specific tool.

So I think we have no choice but to give `prettier` the boot on this.

As an alternative for VS Code, I would recommend adding `"[markdown]": { "editor.rulers": [80] }` to your config to give a visual indication. The editorconfig extension has plans to add this automatically (See editorconfig/editorconfig-vscode#272), but that is not yet done.

And, to supplement, the [Rewrap extension](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) can wrap to a ruler.

These solutions are annoyingly complicated... But that is because VS Code has made this complicated... Other editors do not have this issue and we can't let the choices of a single editor slow us down.

kylewillmon:
> This PR _may_ be okay on it's own but it should not be linked to #1154

I have removed the link

maxrake (approved):
The VS Code integration complication is a shame. Visual rulers are in place now, which does help a bit. LGTM.

PR #1195, opened 2023-08-15
Update READMEs for ecosystem extensions

### LINKED ISSUE #1186: Update default extension docs
The default extensions have READMEs with lines of this sort:

> If the analysis is unsuccessful because some of the new dependencies don't
  meet the required project thresholds, the command will fail.

Project thresholds have been replaced with policies and the language here should change accordingly.

_Originally posted by @kylewillmon in https://github.com/phylum-dev/cli/pull/1182#discussion_r1290740020_

---

The READMEs should also be updated to be more like the new `bundle` extension (in the "How it works" section).

cd-work:
Closes #1186.

### DIFF
--- a/extensions/bundle/README.md
+++ b/extensions/bundle/README.md
@@ -33,7 +33,7 @@ bundle add my-package  # This will be checked by Phylum!
 
 ## How it works
 
-Wher running the package manager through this extension, subcommands which would
+When running the package manager through this extension, subcommands which would
 install new packages will trigger a Phylum analysis first. Once that analysis
 passes Phylum's default policy, the installation is performed. If it did not
 pass the analysis, the command will return early with an error.
--- a/extensions/bundle/README.md
+++ b/extensions/bundle/README.md
@@ -33,16 +33,14 @@ bundle add my-package  # This will be checked by Phylum!
 
 ## How it works
 
-When invoking `phylum bundle`, subcommands that would install gems will trigger
-a Phylum analysis first.
+Wher running the package manager through this extension, subcommands which would
+install new packages will trigger a Phylum analysis first. Once that analysis
+passes Phylum's default policy, the installation is performed. If it did not
+pass the analysis, the command will return early with an error.
 
-- If the analysis is successful, the corresponding changes will be applied.
-- If the analysis is unsuccessful because some of the new dependencies don't
-  meet the required project thresholds, the command will fail.
-- If the analysis is waiting for Phylum to process one or more of the submitted
-  packages, the command will fail and the changes will _not_ be applied.
-- Commands that modify neither `Gemfile`, nor `Gemfile.lock` will be passed
-  through to `bundle` directly.
+In cases where Phylum still needs to process some of the packages, the command
+will exit with a warning **without** installing the packages. Once the analysis
+is complete, another attempt can be made.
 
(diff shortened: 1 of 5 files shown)


extensions/bundle/README.md
```diff
@@ -33,16 +33,14 @@ bundle add my-package  # This will be checked by Phylum!
...
-- If the analysis is successful, the corresponding changes will be applied.
-- If the analysis is unsuccessful because some of the new dependencies don't
-  meet the required project thresholds, the command will fail.
-- If the analysis is waiting for Phylum to process one or more of the submitted
-  packages, the command will fail and the changes will _not_ be applied.
-- Commands that modify neither `Gemfile`, nor `Gemfile.lock` will be passed
-  through to `bundle` directly.
+Wher running the package manager through this extension, subcommands which would
+install new packages will trigger a Phylum analysis first. Once that analysis
+passes Phylum's default policy, the installation is performed. If it did not
```
maxrake, line 38: A link to the [policy documentation](https://docs.phylum.io/docs/policy) would be nice here.

maxrake (changes requested):
It might help to expand on the "install new packages" language to also include something like "executing arbitrary code," which is also what we are trying to protect against.

extensions/bundle/README.md
maxrake: Fair...but reading from a user's perspective, a natural question is going to be: "What is Phylum's default policy?" Heading that question off at the pass is the goal with this suggestion.

kylewillmon: I think the point of mentioning the default policy here is just to clarify for anyone that is using custom policies for their projects.

For users that are completely unfamiliar with our policy framework (or free users that cannot set custom policies), the default policy is all they know. A pass is a pass. A fail is a fail.

To adequately answer "What is Phylum's default policy?", you have to answer "What makes an issue high risk?"... It is unfortunately very subjective and I don't believe we even attempt to answer that question anywhere in our documentation.

maxrake: > ...I don't believe we even attempt to answer that question anywhere in our documentation.

Agreed...couldn't find it. Okay, leaving this unchanged is fine.

cd-work:
> It might help to expand on the "install new packages" language to also include something like "executing arbitrary code," which is also what we are trying to protect against.

I'd rather just keep it simple? I don't see a benefit of this. Executing arbitrary code is implied by installing packages.

extensions/npm/README.md
```diff
@@ -33,17 +33,14 @@ npm install my-package  # This will be checked by Phylum!
...
-analysis.
-
-- If the analysis is successful, the corresponding changes will be applied.
-- If the analysis is unsuccessful because some of the new dependencies don't
-  meet the required project thresholds, the command will fail.
-- If the analysis is waiting for Phylum to process one or more of the submitted
-  packages, the command will fail and the changes will _not_ be applied.
-- Commands that modify neither `package.json`, `npm-shrinkwrap.json`, nor
-  `package-lock.json` will be passed through to `npm` directly.
+Wher running the package manager through this extension, subcommands which would
```
maxrake, line 36:
```suggestion
When running the package manager through this extension, subcommands which would
```

maxrake (changes requested):

maxrake (approved):
LGTM

PR #1198, opened 2023-08-16
Allow compile time override of default data dir

### LINKED ISSUE #822: Ship default extensions via Homebrew
Installing the Phylum CLI with Homebrew should install the default extensions.

kylewillmon:
This patch is an attempt to fix #822

Allowing Homebrew installs to override the default data directory would allow us to use a custom path (likely `$HOMEBREW_PREFIX/var/lib/phylum/extensions`) for extensions when installed via homebrew.

### DRAFT

Feedback is welcome, but I am still testing this. A companion PR will be needed in phylum-dev/homebrew-cli to fully accomplish this.

### DIFF
--- a/cli/src/dirs.rs
+++ b/cli/src/dirs.rs
@@ -5,12 +5,18 @@ use anyhow::{anyhow, Result};
 
 /// Resolve XDG data directory.
 pub fn data_dir() -> Result<PathBuf> {
-    xdg_dir("XDG_DATA_HOME", ".local/share")
+    if let Some(path) = env_path("XDG_DATA_HOME") {
+        return Ok(path);
+    }
+    Ok(home_dir()?.join(".local/share"))
 }
 
 /// Resolve XDG config directory.
 pub fn config_dir() -> Result<PathBuf> {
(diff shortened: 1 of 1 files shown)


kylewillmon:
> I can't imagine every application has their own environment variable for global configuration files.

The environment variable here only exists at compile time. It is not read (and will not be present) at runtime.

How other tools handle this depends completely on the build system... A project using autotools, for example, would probably use `./configure --prefix=/opt/homebrew` when building for Homebrew.

cd-work:
> The environment variable here only exists at compile time. It is not read (and will not be present) at runtime.

But we're not the only application being compiled. Every application is.

> How other tools handle this depends completely on the build system... A project using autotools, for example, would probably use ./configure --prefix=/opt/homebrew when building for Homebrew.

There is no uniform environment variable or anything like that? A fakeroot to build during homebrew? This seems ridiculous.

kylewillmon:
Abandoning this due to https://github.com/phylum-dev/cli/issues/822#issuecomment-1684422450

Issue #1190, opened 2023-08-15
Generate new tokens


kylewillmon:
A command to generate a new token **without storing it in the config** would be useful in some situations. The command would look something like this:

```sh-session
> phylum auth new-token
ph0_abcdefghijklmnopqrstuvwxyz0123456789ABCDEFG
```

Such a command should also include options to set the name and expiry for the token.

It may also be useful to ensure that the generated token is the only output on stdout when running the command. That would allow use in scripting:

```sh
MY_TOKEN=$(phylum auth new-token)
```

PR #1199, opened 2023-08-16
Add `phylum auth create-token` subcommand

cd-work:
Closes https://github.com/phylum-dev/cli/issues/1190.


cli/src/commands/auth.rs
```diff
@@ -208,6 +208,28 @@ pub async fn handle_auth_revoke_token(
...
+
+    let token_name = matches.get_one::<String>("token-name").unwrap().to_string();
+    let ignore_certs = api.config().ignore_certs();
+    let api_uri = &api.config().connection.uri;
+
+    let token =
+        auth::handle_auth_flow(AuthAction::Login, Some(token_name), ignore_certs, api_uri).await?;
+
+    eprintln!();
+    print_user_success!("API token: {token}");
```
maxrake, line 228: Formatting the output like this does not allow for the suggestion in #1190:

> It may also be useful to ensure that the generated token is the only output on stdout when running the command. That would allow use in scripting:
>
> ```sh
> MY_TOKEN=$(phylum auth new-token)
> ```

Was this overlooked or is there a reason the output can't be made this way?

maxrake (changes requested):
The code was reviewed and tested locally. While it does create a token, it does not appear to meet the (implied) acceptance criteria from #1190, which states:

> Such a command should also include options to set the name and expiry for the token.

The name is required in this PR. Is that necessary or can it be optional?

There is no expiry option in this PR. Was that overlooked or is there a reason not to include it?

It appears the creation causes a web browser to open, even when already logged in:

```
../cli  16  17 on  create_token [!?] via 🦕 via 🦀 v1.71.1
❯ target/debug/phylum auth status
✅ Currently authenticated as 'Charles Coggins <charles@phylum.io>' via API key

../cli  16  17 on  create_token [!?] via 🦕 via 🦀 v1.71.1 took 2s
❯ target/debug/phylum auth create-token create_token_test_2
Please use browser window to complete login process
If browser window doesn't open, you can use the link below:
    https://login.phylum.io/realms/phylum/protocol/openid-connect/auth?client_id=locksmith&code_challenge=<REDACTED>

✅ API token: ph0_<REDACTED>
```

That, plus the output related to the browser request text, interferes with the desire to "ensure that the generated token is the only output on stdout when running the command."

cd-work:
> The name is required in this PR. Is that necessary or can it be optional?

It's mandatory.

> There is no expiry option in this PR. Was that overlooked or is there a reason not to include it?

I'll add it to this PR.

> It appears the creation causes a web browser to open, even when already logged in:

Correct, that's intended and there's no way around that.

> That, plus the output related to the browser request text, interferes with the desire to "ensure that the generated token is the only output on stdout when running the command."

If we didn't emit the browser link, users wouldn't have a way to open it manually.

cli/src/app.rs
```diff
@@ -254,6 +254,21 @@ pub fn add_subcommands(command: Command) -> Command {
...
+                            .value_name("TOKEN_NAME")
+                            .action(ArgAction::Set)
+                            .required(true)
+                            .help("Unique name to identify the new token"),
+                        Arg::new("expiry")
+                            .value_name("DAYS")
+                            .short('e')
+                            .long("expiry")
+                            .action(ArgAction::Set)
+                            .help("Number of days the token will be valid for"),
```
maxrake, line 270:
```suggestion
                            .help("Number of days the token will be valid"),
```

maxrake (changes requested):
The code was minimally reviewed...but growing past the point of my Rust comprehension. Adding @kylewillmon for code review and acceptance criteria clarification...

Some observations from local testing and requested changes:

* When creating a token with an expiry set, the expiry does not show when viewing the tokens
  * It is blank in the `expiry` column of the `auth list-tokens` output in the CLI
  * It shows as `Invalid DateTime` in the UI
* Creating a token with a name that has already been used results in generic error message:

```
../cli  16  17 on  create_token [!?] via 🦕 via 🦀 v1.71.1
❯ target/debug/phylum auth create-token create_token_test_exists
Please use browser window to complete login process
If browser window doesn't open, you can use the link below:
    https://login.phylum.io/realms/phylum/protocol/openid-connect/auth?client_id=locksmith&code_challenge=<REDACTED>
❗ Error: HTTP status server error (500 Internal Server Error) for url (https://api.phylum.io/locksmith/v1/token)

Caused by:
    internal server error
```

kylewillmon:
> If we didn't emit the browser link, users wouldn't have a way to open it manually.

Can stderr be used for this?

kylewillmon:
> Creating a token with a name that has already been used results in a generic error message:

This appears to be a bug in the locksmith service. I've opened a ticket there.

cd-work:
> Can stderr be used for this?

I mean it **could**, but I don't see how that would change much. If anything I'd just say that makes the command output more confusing.

Is your goal to make this pipeable? If so then `print_user_success` can't be used since it also prints a checkmark.

kylewillmon:
> Is your goal to make this pipeable? If so then `print_user_success` can't be used since it also prints a checkmark.

That is exactly the goal. We're printing a token with the expectation that the user will put that token somewhere. Making it pipeable just seems like a nice way to assist users in that process.

We avoid `print_user_success` in other contexts like `phylum auth token` for this exact reason.

I'm open to opposition here if there is reason that it shouldn't be pipeable.

cd-work:
> I'm open to opposition here if there is reason that it shouldn't be pipeable.

No it makes some sense, not for piping between commands but for redirecting to a file.

maxrake (commented):
The stderr redirection appears to work now. The browser opening (and requesting login after a period of inactivity) may still hinder the desired use case in scripts...but perhaps that is simply unavoidable.

The expiry testing was able to proceed after realizing (thanks @cd-work) that it would need to happen in STG instead of PROD. The "Expires" column in the CLI and UI output appears correct. Some bounds testing revealed a few things.

It is possible to specify `--expiry 0`, which has some interesting side effects:

* The token is created but not shown to the user
  * An HTTP 500 status error is shown to the user
* The expires date is the same as the created date
* The token shows up in an `auth list-tokens` command
* The token can not be revoked, from the CLI or the UI

It is also possible to specify an expiry value in days greater than 3650 (the default "10 years" when not specified at all).

All of this brings up some interesting questions:

* Is it even possible to revoke an expired token?
* Does it make sense to allow a token expiry value greater than 10 years?
* Should there be bounds enforced on the expiry (e.g., min of 1 day and max of 10 years)

Leaving a comment instead of approving or requesting changes b/c it is not clear whether any changes are needed...here, in Locksmith, or at all...

maxrake (approved):

cd-work:
> Is it even possible to revoke an expired token?

I'm not sure expired tokens should ever show up to the user. I thought I removed them from list tokens explicitly. So that might be worth looking into.

> Does it make sense to allow a token expiry value greater than 10 years?

Yes, the value of 10 years has been chosen completely arbitrarily. Imo if someone wants to make a token valid for 9999999 years, that's perfectly fine. It's essentially the same thing anyway.

> Should there be bounds enforced on the expiry (e.g., min of 1 day and max of 10 years)?

I don't see why. On the upper end it doesn't make a difference because 10 years and 100 years are both basically "never expire". On the lower end there's only 0 days which doesn't make sense and in that scenario I honestly don't really care how we react as long as no valid token is created. We could short-circuit in the CLI, but backend should be able to handle it anyway. And if no "valid" token is produced then that's fine because the user never asked for one.

maxrake:
@cd-work, thanks for the explanations. In that case, this PR LGTM and the current approval stands.

maxrake (approved):
LGTM

Issue #1188, opened 2023-08-14
Support parsing CycloneDX SBOMs


ejortega:
# Overview
Users should be able to submit CycloneDX Software Bill of Materials (SBOMs) v1.5 for analysis.

# Acceptance Criteria
- [ ] Parsing of common CycloneDX SBOM file formats (JSON, XML)
- [ ] Extract the relevant dependencies for ecosystems we currently support

## Additional Context
https://cyclonedx.org/docs/1.5/json/
https://cyclonedx.org/docs/1.5/xml/

PR #1200, opened 2023-08-16
Add support for CycloneDX SBOMs

ejortega:
Add support for parsing CycloneDX SBOMs.

Documentation on `v1.5 `specs
https://cyclonedx.org/docs/1.5/json/
https://cyclonedx.org/docs/1.5/xml/

Added test files for `v1.3`, `v1.2` and `v1.5`.

Note: `optional` dependencies are skipped for now. @louislang and @furi0us333 suggested adding a flag to include those does when specified by the user.

Closes https://github.com/phylum-dev/cli/issues/1188

## Checklist


lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+// Define the generic trait for components.
+trait Component {
+    fn component_type(&self) -> &str;
+    fn name(&self) -> &str;
+    fn version(&self) -> &str;
+    fn scope(&self) -> Option<&str>;
+    fn purl(&self) -> Option<&str>;
+    fn components(&self) -> Option<&[Self]>
+    where
+        Self: Sized;
```
cd-work, line 27: It's probably easier to just add this requirement to the trait directly? That way it doesn't have to be specified every time during implementation.

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+// Represents a single XML component.
+#[derive(Clone, Debug, Deserialize)]
+struct XmlComponent {
+    #[serde(rename = "type")]
+    component_type: String,
+    name: String,
+    version: String,
+    scope: Option<String>,
+    purl: Option<String>,
+    components: Option<Components<XmlComponent>>,
```
cd-work, line 53: Do I see it correctly that the only reason for having separate structs between XML and JSON is this difference in array handling of XML?

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+
+    #[test]
+    fn parse_cyclonedx_1_5_json() {
```
cd-work, line 232: Is there a specific reason why the json in this test is inlined rather than using a fixture?

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+        .unwrap_or(PackageVersion::FirstParty(pkg_version.into()));
+
+    Ok(Package { name, version, package_type })
+}
+
+pub struct CycloneDX;
+
+impl Parse for CycloneDX {
+    fn parse(&self, data: &str) -> anyhow::Result<Vec<Package>> {
+        if let Ok(lock) = serde_json::from_str::<serde_json::Value>(data) {
```
cd-work, line 201: Should use a match here.

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+            let parsed: Bom<Vec<JsonComponent>> = serde_json::from_value(lock)?;
+            parsed.components.map_or(Ok(vec![]), |comp| {
```
cd-work, line 203: Should use `Vec::new()` over `vec![]`. Relying on macro expansion for something so simple seems excessive.

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+// Represents a single JSON component.
+#[derive(Clone, Debug, Deserialize)]
+struct JsonComponent {
+    #[serde(rename = "type")]
+    component_type: String,
+    name: String,
+    version: String,
+    scope: Option<String>,
+    purl: Option<String>,
+    components: Option<Vec<JsonComponent>>,
```
cd-work, line 94: Why is this an option? Can't we just default to empty vec? We're unwrapping to an empty vec later anyway.

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+        .flat_map(|comp| {
+            let nested_iter = match comp.components() {
+                Some(nested) => filter_components(nested).collect::<Vec<_>>(),
+                None => Vec::new(),
+            };
+            std::iter::once(comp).chain(nested_iter.into_iter())
+        })
+}
+
+// Convert a component's Package URL (PURL) into a Package object.
```
cd-work, line 152:
```suggestion
/// Convert a component's package URL (PURL) into a package object.
```

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+fn from_purl<T: Component>(component: &T) -> anyhow::Result<Package> {
+    let purl_str = component
+        .purl()
+        .ok_or_else(|| anyhow!("Missing purl for {}:{}", component.name(), component.version()))?;
```
cd-work, line 156: Is this required by specification? Generally if it is, why wouldn't we enforce that in the deserialization?

Providing better error messages is nice, but that's only useful if people are expected to run into them.

And if it's actually optional, should we maybe not error out?

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+    let purl = GenericPurl::<String>::from_str(purl_str)?;
+    let package_type = PackageType::from_str(purl.package_type()).map_err(|_| UnknownEcosystem)?;
+
+    // Determine the package name based on its type and potential namespace.
```
cd-work, line 160:
```suggestion
    // Determine the package name based on its type and namespace.
```

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+    let name = match (package_type, purl.namespace()) {
+        (PackageType::Maven, Some(ns)) => format!("{}:{}", ns, purl.name()),
+        (PackageType::Npm | PackageType::Golang, Some(ns)) => format!("{}/{}", ns, purl.name()),
+        _ => purl.name().into(),
+    };
```
cd-work, line 165: This looks **very** familiar. Is it possible we did this in the sbom parser already?

If so, isn't it maybe worth it to create an abstraction for this?

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+            "vcs_url" => {
+                if value.starts_with("git+") {
+                    Some(PackageVersion::Git(value.to_string()))
+                } else {
+                    None
+                }
+            },
+            _ => None,
+        })
+        .unwrap_or(PackageVersion::FirstParty(pkg_version.into()));
```
cd-work, line 192: This also seems like a concept that applies to more than just CycloneDX. I'm not sure if this was necessary in spdx already, but probably worth it to at least extract it into a method so it can be moved into a more generic position if we ever need to reuse it.

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+// Filter components based on their type and scope.
+fn filter_components<T: Component>(components: &[T]) -> impl Iterator<Item = &'_ T> {
+    components
+        .iter()
+        .filter(|&comp| {
+            let type_check = comp.component_type() == "application"
+                || comp.component_type() == "framework"
+                || comp.component_type() == "library";
+
+            // Check if the scope is "required" or not specified (required)
```
cd-work, line 135: This comment confused me and I would not have understood it without reading the code.

It basically translates to "Check if the scope is required or required", which doesn't make sense because that's obviously the same in both scenarios. But it fails to mention that it's a tri-state, not just two different possibilities.

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+                let component_iter = filter_components(&comp.components);
+                component_iter.map(from_purl).collect()
+            })
+        }
+    }
+
+    fn is_path_lockfile(&self, path: &Path) -> bool {
+        path.file_name() == Some(OsStr::new("bom.json"))
+            || path.file_name() == Some(OsStr::new("bom.xml"))
+    }
```
maxrake, line 219: This function should be synced with the changes made in #1202, depending on which PR gets merged first.

lockfile/src/cyclonedx.rs
ejortega: It's optional by specification however for us it wouldn't be optional. A similar conversation occurred with SPDX as I initially just ignored packages missing a PURL. The outcome was that we could not determine the ecosystem otherwise and could lead to providing consumers incorrect results (we do ignore ecosystems we don't support), so we fail the parsing and give the user an error.

lockfile/src/cyclonedx.rs
cd-work: I was going to add this to my review but am hoping that we can figure this PR out before #1202.

lockfile/src/cyclonedx.rs
```diff
@@ -0,0 +1,313 @@
...
+    fn name(&self) -> &str;
+    fn version(&self) -> &str;
+    fn scope(&self) -> Option<&str>;
+    fn purl(&self) -> Option<&str>;
+    fn components(&self) -> Option<&[Self]>
+    where
+        Self: Sized;
+}
+
+// CycloneDX BOM
```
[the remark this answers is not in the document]
ejortega: Oh I didn't know this. Thanks!

lockfile/src/cyclonedx.rs
cd-work: Hm that certainly doesn't seem like a great solution, but I'm not familiar with sboms and there probably just isn't any good resolution.

lockfile/src/cyclonedx.rs
ejortega: Yes, something to do with the attribute. Is there's a better way to handle that?

Seems to be something to do with the attribute.
```<component type="library"```

lockfile/src/cyclonedx.rs
ejortega: It was more that I couldn't find a good `v1.5` example. All tools seems to still generate `v1.4`. Granted `v1.5` was released fairly recently. The fields used for parsing packages are the same.

I can just move that example to a file to make it more consistent.

lockfile/src/cyclonedx.rs
cd-work: Honestly I'm not **entirely** sure. It would be possible to share the fields that are in common I suppose.

lockfile/src/cyclonedx.rs
cd-work: Sounds good.

lockfile/src/cyclonedx.rs
ejortega: I think we can pass on parsing external refs for now for the first release. There needs to be additional work for the optional dependencies. Can wrap that work into next iteration.

maxrake:
> I did add test files for `v1.4` and `v1.3` specs but couldn't find a `v1.5` example.

Can one be made and then included here? It appears the [`cyclonedx-npm` tool provided by CycloneDX](https://github.com/CycloneDX/cyclonedx-node-npm) has a `--spec-version` option that includes v1.5.

ejortega:
Thanks @maxrake, updated with those examples.

cd-work (changes requested):
300k lines of lockfiles seems a bit excessive to me. Do these files really bring 300k lines of benefit? I don't think we gain much by having excessively large lockfiles that just repeat the same thing over and over again and it would be really unfortunate to bloat up our repo through these.

cd-work (approved):

PR #1203, opened 2023-08-17
Add cargo ecosystem extension

### LINKED ISSUE #1196: Official `cargo` extension
Create a `phylum cargo` extension to pre-check builds for Rust projects.

Sandboxing `cargo` is probably impractical because `build.rs` files and proc macros can do all sorts of untold things... (e.g., database access in `sqlx`, downloads and/or C++ build in `v8`, etc.)

cd-work:
Closes #1196.

### DIFF
--- a/extensions/cargo/main.ts
+++ b/extensions/cargo/main.ts
@@ -1,0 +1,237 @@
+import {
+  green,
+  red,
+  yellow,
+} from "https://deno.land/std@0.150.0/fmt/colors.ts";
+import { PhylumApi, PolicyEvaluationResponseRaw } from "phylum";
+
+class FileBackup {
+  readonly fileName: string;
+  fileContent: string | null;
+
+  constructor(fileName: string) {
+    this.fileName = fileName;
+    this.fileContent = null;
+  }
+
+  async backup() {
+    try {
+      this.fileContent = await Deno.readTextFile(this.fileName);
+    } catch (_e) { /* Do nothing */ }
+  }
+
+  async restoreOrDelete() {
+    try {
+      if (this.fileContent != null) {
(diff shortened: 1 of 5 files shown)


extensions/cargo/main.ts
```diff
@@ -0,0 +1,237 @@
...
+const status = PhylumApi.runSandboxed({
+  cmd: "cargo",
+  args: Deno.args,
+  exceptions: {
+    run: ["cargo", "~/.rustup"],
+    read: ["./", "~/.cargo", "~/.rustup", "/etc/passwd"],
+    write: ["./", "~/.cargo"],
+    net: true,
+  },
+});
```
cd-work, line 107: Technically I'm not sure there's any reason for this sandbox, but since we only run this for `add` and `update` I suppose it shouldn't be a problem either.

extensions/cargo/main.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// Ignore all commands that shouldn't be intercepted.
+if (
+  Deno.args.length == 0 ||
+  !(
+    "add".startsWith(Deno.args[0]) ||
+    "update".startsWith(Deno.args[0])
+  )
+) {
```
kylewillmon, line 73: This ignores far too many commands...

We need to check `cargo build` and friends since they will run the `build.rs` files of the projects dependencies.

In fact, `cargo add` and `cargo update` are probably two commands that we **don't** need to intercept. Because they will never execute untrusted code.

kylewillmon (changes requested):

extensions/cargo/main.ts
cd-work: Going to close this PR because I'm not sure how to implement this.

Issue #1177, opened 2023-08-08
Directories detected as lockfiles


kylewillmon:
Lockfile detection believes that directories might be lockfiles.

# How To Reproduce
```sh-session
> mkdir requirements.txt

> ls
requirements.txt

> phylum parse
Generating lockfile for manifest "./requirements.txt" using Pip…
❗ Error: Lockfile generation failed! For details, see: https://docs.phylum.io/docs/lockfile-generation

Caused by:
    package manager exited with error code 1:

    ERROR: Could not open requirements file: [Errno 21] Is a directory: '/path/to/requirements.txt'

    [notice] A new release of pip is available: 23.0.1 -> 23.2.1
    [notice] To update, run: pip install --upgrade pip
```

This error message isn't very helpful....

```sh-session
> mkdir Cargo.lock

> phylum parse
❗ Error: Is a directory (os error 21)
```

cd-work:
https://github.com/phylum-dev/cli/pull/1202 as a solution was discarded, since even checking for `!is_directory` still causes syscalls to lookup metadata for every lockfile, even if it doesn't cause failure for the GitHub App. Since this has never caused any issues in practice, there's no reason to slow down the GitHub App.

PR #1204, opened 2023-08-18
Improve lockfile parsing errors

kylewillmon:
As shown in #1177, some lockfile parsing errors result in poor error messages because the lockfile path is not always known.

This patch adds the lockfile path as context on the error to ensure that it will be present in the error message.

**Before**
```
> phylum parse
❗ Error: Is a directory (os error 21)
```

**After**
```
> phylum parse
❗ Error: could not parse lockfile: ./Cargo.lock

Caused by:
    Is a directory (os error 21)
```


cd-work (approved):
LGTM

PR #1206, opened 2023-08-21
Add cargo ecosystem extension

### LINKED ISSUE #1196: Official `cargo` extension

cd-work:
This patch adds a Rust ecosystem extesnion for `cargo`, automatically triggering an analysis before any build/check/test/clippy command.

Build commands being triggered much more frequently than npm/yarn/poetry/...'s add/install commands makes it necessary to cache the success for cargo, ensuring the development pace isn't slowed down unnecessarily. This cache resides in the extension's directory (`~/.local/share/phylum/extensions/cargo/cache`).

Closes #1196.


extensions/cargo/PhylumExt.toml
```diff
@@ -0,0 +1,10 @@
+name = "cargo"
+description = "Cargo package manager hooks"
+entry_point = "main.ts"
+
+[permissions]
+write = ["./", "~/.cargo", "~/.local/share/phylum/extensions/cargo"]
```
cd-work, line 6: I've added an extension function to retrieve the extensions dir, but had to hardcode the permissions here. In theory we could just hardcode it in both places and skip the update to our extension interface?

Alternatively we could also give every extension read/write access to its own extension directory by default.

kylewillmon: > Alternatively we could also give every extension read/write access to its own extension directory by default.

🤔 Would this allow an extension to change its own permissions without user interaction by writing to its own installed `PhylumExt.toml`?

But `deno` also has `deny_write` now. So we could allow the folder, but deny writing to `PhylumExt.toml`.

cd-work: > 🤔 Would this allow an extension to change its own permissions without user interaction by writing to its own installed PhylumExt.toml?

Yes it would, I forgot we don't prompt for installed extensions.

kylewillmon (changes requested):
I'd like to see us either:

1. Give read/write access to the extension directory by default
2. Hard code the path to the extension directory everywhere and remove the new API function
3. Enable and use the Web Storage API

The current code is somewhere in between 1 and 2 because it's hard coded in the manifest, but using the API function everywhere else.

extensions/cargo/README.md
```diff
@@ -0,0 +1,47 @@
...
+git clone https://github.com/phylum-dev/cli
+phylum extension install cli/extensions/cargo
+```
+
+## Basic usage
+
+Prepend `phylum` to your `cargo` command invocations:
+
+```console
+phylum cargo add my-package  # This will be checked by Phylum!
```
kylewillmon, line 24: This example (and the second one below) needs to be changed

extensions/cargo/main.ts
```diff
@@ -0,0 +1,333 @@
...
+    await restoreBackup();
+    return;
+  }
+
+  console.log(`[${green("phylum")}] Parsing lockfile…`);
+  let lockfile;
+  try {
+    lockfile = await PhylumApi.parseLockfile("./Cargo.lock", "cargo");
+  } catch (_e) {
+    console.warn(`[${yellow("phylum")}] No lockfile present.\n`);
```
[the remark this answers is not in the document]
cd-work: I've changed it to include the error. Especially since we ran `generate-lockfile` without an error, this should **always** succeed.

cargo_perms
kylewillmon, line 1: Pretty sure this file was included by mistake.

cd-work:
@kylewillmon I'd assume you want to hold off on this until the local storage stuff is figured out?

Do you want me to add it in this PR?

kylewillmon:
> @kylewillmon I'd assume you want to hold off on this until the local storage stuff is figured out?

Yes. I should have a PR up soon to enable the Web Storage API for extensions.

kylewillmon (approved):

extensions/cargo/PhylumExt.toml
kylewillmon, line 6: This directory isn't needed anymore

```suggestion
write = ["./", "~/.cargo"]
```

kylewillmon (approved):

bjorn3:
Do I understand it correctly that this doesn't actually sandbox any compilation?

louislang:
Hi @bjorn3! The allowed paths in the sandbox are defined on a per-extension basis. In some cases, broader disk access is required to allow the package manager to operate (we're transitioning this to a blacklist instead of a whitelist, as it makes this problem more tractable). The currently allowed paths for npm are defined [here](https://github.com/phylum-dev/cli/blob/357a54abe042db4b01861e6c531430e3e841d628/extensions/npm/main.ts#L140-L145).

Network is still heavily restricted for `npm`, so any attempt to reach out and exfiltrate data _should_ fail. And if it doesn't, please let us know!

If you can, please create new tickets for any problems instead of closed issues where we might miss them. We also have [a Discord](https://discord.gg/Fe6pr5eW6p) that we are active in if you'd like to chat about the sandbox!

maxrake:
We've also just enabled [discussions](https://github.com/phylum-dev/cli/discussions) on this repository...which is another good place to ask questions.

PR #1207, opened 2023-08-21
Fix find-permissions extension

cd-work:
This patch adds several improvements for the `find-permissions` extension.

Executables requiring environment variables would always fail to run since the extension was written before the introduction of the `env` permission. To fix that, reading from `env` is now always allowed.

The pre/post scripts also incorrectly specified the `run` permission as `execute`, causing it to fail at launching these scripts.

Additionally the `--allow-run` and `--allow-write` flags were added to allow adding exceptions ahead of time. This can help speed up the process when some directories are known to be required anyway.


cd-work:
I intentionally did not add a changelog for this because I'm not sure these changes can be summarized succinctly in a fashion endusers would actually benefit from. Considering that they should mostly just be small UX improvements, users should be able to just notice and benefit by themselves.

extensions/find-permissions/main.ts
```diff
@@ -241,10 +276,27 @@ function test(directories: string[]): boolean {
...
+      continue;
+    }
+
+    if (offset <= 0) {
+      optionIndex = i;
+      break;
+    }
+
+    offset -= 1;
+  }
```
kylewillmon, line 296: Pretty sure there is a function for this... [`indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf)

```suggestion
  const optionIndex = Deno.args.indexOf(option, offset);
```

kylewillmon (changes requested):

extensions/find-permissions/main.ts
cd-work: This is subtly different. Here the `offset` is based on array index position, while the offset in my function is based on number of occurrences.

PR #1208, opened 2023-08-21
Use deno_core::extension! macro

kylewillmon:
All of deno's internal and example extensions are declared using the `extension!` macro. This patch updates our extension to use that macro.

### DIFF
--- a/cli/src/deno.rs
+++ b/cli/src/deno.rs
@@ -10,7 +10,7 @@ use dashmap::DashMap;
 use deno_ast::{EmitOptions, MediaType, ParseParams, SourceTextInfo, TranspiledSource};
 use deno_runtime::deno_core::error::JsError;
 use deno_runtime::deno_core::{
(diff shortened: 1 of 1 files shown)


kylewillmon:
As far as I can tell, the only "advantage" that the macro brings is that it provides two functions for initializing an extension. `init_ops_and_esm()` fully initializes the extension while `init_ops()` only initializes the ops and does not include the JS files.

This is useful for deno extensions because they make heavy use of v8 snapshots. So if an extension is already included in their startup snapshot, they can use the faster `init_ops()` at runtime and avoid the cost of running the JS files during startup.

We do not currently use a custom snapshot, so this advantage isn't important to us. Additionally, the deno authors have shown in denoland/deno@c1c8eb3 that you don't really need this distinction because they now sometimes use `init_ops_and_esm()` and then go back and remove the JS files....

In any case, this is not an improvement for our code, so I will abandon this PR.

PR #1209, opened 2023-08-22
Use absolute paths for install check

kylewillmon:
This patch updates the install check to correctly handle relative paths. Before this, it was possible to fool the check by using a relative path on the command line.

```
> phylum extension install -y ~/.local/share/phylum/extensions/npm
Installing extension npm...
❗ Error: extension path and installation path are identical, skipping

> phylum extension install -y ./.local/share/phylum/extensions/npm
Installing extension npm...
❗ Error: IO error for operation on ./.local/share/phylum/extensions/npm: No such file or directory (os error 2)
```

I believe the previous check could also be fooled with an absolute path on the command line if `XDG_DATA_HOME` contained a relative path. But that scenario is very unlikely, so I did not test it.

### DIFF
--- a/cli/src/commands/extensions/extension.rs
+++ b/cli/src/commands/extensions/extension.rs
@@ -49,6 +49,7 @@ impl ExtensionManifest {
 
 #[derive(Clone, Debug)]
 pub struct Extension {
+    /// Absolute path to the extension.
     path: PathBuf,
     manifest: ExtensionManifest,
 }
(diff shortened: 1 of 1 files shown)


kylewillmon:
This edge case is likely too obscure to warrant a CHANGELOG entry. So I did not include one.

cd-work (approved):
I'm **pretty** sure switching the extension path to be absolute should be fine, so LGTM.

PR #1211, opened 2023-08-22
Enable Web Storage API for extensions

kylewillmon:
This patch allows extensions to use `localStorage` to store state. Each extension has separate state and it is removed when the extension is uninstalled. Additionally, only installed extensions are allowed to persist state.

_Note:_ This also enables usage of `sessionStorage`, but the session lasts only as long as the process, so this does not provide any additional value.

### DIFF
--- a/cli/src/commands/extensions/extension.rs
+++ b/cli/src/commands/extensions/extension.rs
@@ -124,12 +124,12 @@ impl Extension {
             return Err(anyhow!("extension {} is not installed, skipping", self.name()));
         }
 
-        fs::remove_dir_all(&self.path)?;
-
         if let Some(state_path) = self.state_path() {
             // Ignore errors since this may not exist
             let _ = fs::remove_dir_all(state_path);
         }
+
+        fs::remove_dir_all(&self.path)?;
 
         println!("Extension {} uninstalled successfully", self.name());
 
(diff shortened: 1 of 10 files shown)


cli/src/commands/extensions/extension.rs
```diff
@@ -150,6 +155,15 @@ impl Extension {
         self.path.clone()
     }
 
+    /// A directory where installed extensions can store state.
+    pub fn state_path(&self) -> Option<PathBuf> {
+        if self.installed() {
+            extension_state_path(self.name()).ok()
+        } else {
+            None
+        }
```
cd-work, line 164: Could use `.then(|| )` if you wanted to shorten this a bit.

cli/src/commands/uninstall.rs
```diff
@@ -42,23 +42,38 @@ fn purge() -> Result<()> {
...
 }
 
+/// Return Ok for file not found errors. Pass everything else through.
+fn ignore_not_found(err: io::Error) -> io::Result<()> {
+    if err.kind() == ErrorKind::NotFound {
+        Ok(())
+    } else {
+        Err(err)
+    }
+}
```
cd-work, line 52: Seems a bit excessive for a single usage?

cd-work (dismissed):

cd-work (approved):

cd-work (approved):

PR #1214, opened 2023-08-24
Bump to v5.7.0

maxrake:
Release-Version: v5.7.0


lockfile/src/cyclonedx.rs
maxrake, line 1: The change in this file was made due to a new clippy error finding that first hit in this PR.

cd-work (approved):

PR #1221, opened 2023-08-30
Don't print package with no active rejections

kylewillmon:
Packages with all rejections suppressed should not be displayed in the output report. This patch corrects the filter to account for that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+- Package header printed even when all issues were suppressed
+
 ## [5.7.0] - 2023-08-24
 
 ### Added
--- a/cli/src/format.rs
+++ b/cli/src/format.rs
@@ -124,7 +124,11 @@ impl Format for PolicyEvaluationResponseRaw {
         }
 
         // Write summary for each issue.
-        for package in self.dependencies.iter().filter(|package| !package.rejections.is_empty()) {
+        for package in self
+            .dependencies
+            .iter()
+            .filter(|package| package.rejections.iter().any(|rejection| !rejection.suppressed))
+        {
             let _ = writeln!(writer, "[{}] {}@{}", package.registry, package.name, package.version);
 
             for rejection in package.rejections.iter().filter(|rejection| !rejection.suppressed) {


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Absolute paths submitted when analyzing manifest files
 - Ecosystem extensions not pre-checking `remove`/`uninstall` operations
 - Disabled update and uninstall commands in completion for Homebrew users
+- Package header printed even when all issues were suppressed
```
cd-work, line 23: This will be put in the wrong location once you rebase.

cd-work (changes requested):

maxrake (changes requested):
The changes were tested locally and appear to work for `phylum-ci`, which has a single package that has been suppressed (er..."allowed policy violation" now?)

image

---

However, the output does not quite look right for testing with this (`cli`) repo, which currently has four (4) packages with policy violations:

image

image

It appears the packages with a vulnerability issue were retained while the others were filtered out. Or, maybe it is that the packages with malicious issues were filtered out. Perhaps there is another bug here...

kylewillmon:
> Perhaps there is another bug here...

That is clearly a separate issue as they don't show up in either output in your screenshot. So they are not affected by this change.

maxrake:
> That is clearly a separate issue as they don't show up in either output in your screenshot.

The first output for the `cli` repo analysis, using the currently released version of the Phylum CLI includes the `md4@0.10.2` and `unicode-id@0.3.3` packages whereas the second output, using the changes from this PR, do not.

It still may be an unrelated bug...

kylewillmon:
> The first output for the `cli` repo analysis, using the currently released version of the Phylum CLI includes the `md4@0.10.2` and `unicode-id@0.3.3` packages whereas the second output, using the changes from this PR, do not.

Right. The package name is printed. But the issue (or rejection) is not. So the output is unhelpful. That is exactly what this PR fixes.

If (as you claim) there are issues missing that should be printed, that is an unrelated issue.

maxrake:
Oooh...I get it now 🐢
I'll look into this behavior a bit more and create a different issue for it.

maxrake (approved):
LGTM

cd-work (approved):

PR #1223, opened 2023-08-31
Bump Birdcage to v0.3.1

cd-work:
This fixes a regression introduced in 041be47 which would prevent local sockets from working when the network sandbox was enabled.


maxrake (approved):
...and we're back!

LGTM

kylewillmon (approved):

PR #1228, opened 2023-09-08
Ignore extraneous npm dependencies

### LINKED ISSUE #1227: Extraneous packages in npm lockfiles fail to parse
# Overview
When running Phylum CLI on a lockfile that contains an extranneous package, the parser fails resulting in an inability to submit a job.

# How To Reproduce
1. Locate a lockfile that contains an extranneous package
2. Run `phylum analyze` on it
3. See error

# Expected Behavior
Since we cannot gather information about the extranneous package from the lockfile, we should skip over it and print a warning to the user letting them know they should run `npm prune` to clean up these packages.

# Additional Context
The parser fails with the following error: 

````
phylum analyze
:exclamation: Error: Unable to locate any valid package in lockfile "/private/tmp/sites/web/./package-lock.json"

Caused by:
    0: Failed to parse lockfile
    1: Dependency 'netlify-cli/tools/lint-rules' is missing "resolved" key
````

Example section causing the issue:

```
"node_modules/netlify-cli/tools/lint-rules": {
    "name": "eslint-plugin-workspace",
        "extraneous": true
},
```

See [`npm prune`](https://docs.npmjs.com/cli/v7/commands/npm-prune) for additional details.

### PR DESCRIPTION
**Pull request by kylewillmon**
Fix #1227 by ignoring extraneous packages

### DIFF
--- a/lockfile/src/javascript.rs
+++ b/lockfile/src/javascript.rs
@@ -71,6 +71,11 @@ impl Parse for PackageLock {
                     continue;
                 }
 
+                // Ignore extraneous dependencies.
+                if keys.get("extraneous").is_some() {
+                    continue;
+                }
+
                 // Get dependency type.
                 let resolved = get_field(keys, "resolved")
                     .ok_or_else(|| anyhow!("Dependency '{name}' is missing \"resolved\" key"))?;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 - Package header printed even when all issues were suppressed
+- Parsing fails for extraneous npm packages in package-lock.json
 
 ## [5.7.0] - 2023-08-24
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: lockfile/src/javascript.rs [hunk 1]
```diff
@@ -71,6 +71,11 @@ impl Parse for PackageLock {
                     continue;
                 }
 
+                // Ignore extraneous dependencies.
+                if keys.get("extraneous").is_some() {
+                    continue;
```
maxrake, line 76: Skipping over it looks good, but what about the second part of the issue's expected behavior:

> we should skip over it and print a warning to the user letting them know they should run npm prune to clean up these packages.

maxrake (approved):
The changes were tested locally and work as expected to skip these `extraneous` packages. There is still a question about informing users about the skipped package and what to do about it.

The inclusion of a representative package in a test fixture was great to see.

lockfile/src/javascript.rs
kylewillmon: We cannot print warnings directly from the parsers, so that gets a bit tricky.

Silently ignoring them matches the behavior of other cases, such as bundled or vendored dependencies. With no install step, they pose minimal security risk.

PR #1234, opened 2023-09-15
Add extension API for Linux sandbox control

cd-work:
This patch adds two new fields to the `runSandboxed` extension API which allow controlling the minimum required Linux landlock ABI version and whether the command should still be executed when sandboxing is not possible.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -64,7 +64,7 @@ once_cell = "1.12.0"
 deno_runtime = { version = "0.122.0" }
 deno_core = { version = "0.199.0" }
 deno_ast = { version = "0.27.2", features = ["transpiling"] }
-birdcage = { version = "0.3.1" }
+birdcage = { version = "0.3.1", path = "../../birdcage" }
 libc = "0.2.135"
 ignore = { version = "0.4.20", optional = true }
 uuid = "1.4.1"
--- a/cli/src/commands/extensions/api.rs
+++ b/cli/src/commands/extensions/api.rs
@@ -78,6 +78,10 @@ struct ProcessException {
     net: bool,
     #[serde(default)]
     strict: bool,
+    #[serde(default)]
+    ignore_sandboxing_errors: bool,
+    #[serde(default)]
+    minimum_landlock_version: u8,
 }
 
 #[cfg(unix)]
@@ -393,6 +397,30 @@ async fn parse_lockfile(
     Ok(PackageLock { packages: parsed.packages, format: parsed.format })
 }
 
+/// Return error when trying to sandbox on Windows.
+#[op]
+#[cfg(not(unix))]
+fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
+    if !process.exceptions.best_effort {
+        Err(anyhow!("Extension sandboxing is not supported on this platform"))
+    }
+
+    let output = Command::new(process.cmd)
+        .args(process.args)
+        .stdin(process.stdin)
+        .stdout(process.stdout)
+        .stderr(process.stderr)
+        .output()?;
+
+    Ok(ProcessOutput {
+        stdout: String::from_utf8_lossy(&output.stdout).into_owned(),
+        stderr: String::from_utf8_lossy(&output.stderr).into_owned(),
+        success: output.status.success(),
+        signal: output.status.signal(),
+        code: output.status.code(),
+    })
+}
+
 /// Run a command inside a sandbox.
 ///
 /// This runs the supplied command in a sandbox, without restricting the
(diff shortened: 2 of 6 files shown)


cli/src/commands/extensions/permissions.rs
```diff
@@ -188,40 +190,6 @@ impl Permissions {
             && self.unsandboxed_run.get().is_none()
     }
 
-    /// Build a sandbox matching the requested permissions.
-    #[cfg(unix)]
-    pub fn build_sandbox(&self) -> Result<Birdcage> {
```
cd-work, line 193: It didn't look like this was ever called anywhere, so I removed it.

cli/src/commands/extensions/permissions.rs
```diff
@@ -276,9 +244,24 @@ impl From<&Permissions> for PermissionsOptions {
...
+        2 => LANDLOCK_ABI::V2,
+        _ => LANDLOCK_ABI::V1,
+    };
+    #[cfg(target_os = "linux")]
+    let mut birdcage = Birdcage::new_with_version(abi)?;
+
+    // Do not add any default exceptions for strict sandboxes.
+    if strict {
+        return Ok(birdcage);
+    }
```
cd-work, line 263: Moved this inside the function since I think it's better to handle it all in one place.

cli/src/commands/sandbox.rs
```diff
@@ -48,13 +50,21 @@ pub async fn handle_sandbox(matches: &ArgMatches) -> CommandResult {
...
+    let birdcage = permissions::default_sandbox(strict, min_landlock_abi);
 
     // Provide additional error context.
     let mut birdcage = match birdcage {
         Ok(birdcage) => birdcage,
         #[cfg(target_os = "linux")]
+        Err(_) if best_effort => {
+            log::debug!("Landlock v{min_landlock_abi} is not supported, skipping sandbox");
+            return Ok(());
+        },
```
cd-work, line 66: Technically sandboxing can still fail when `add_exception` fails or `lock` fails. However this shouldn't be a landlock ABI error at that point so I don't think `best_effort` should apply?

There might be some scenarios we want to include into `best_effort` in the future, we could for example make ignore an error in `lock` when both seccomp and namespaces are blocked. But with our current behavior we should support pretty much every system so I don't think there's a need to relax things even more.

cli/Cargo.toml
```diff
@@ -64,7 +64,7 @@ once_cell = "1.12.0"
 deno_runtime = { version = "0.122.0" }
 deno_core = { version = "0.199.0" }
 deno_ast = { version = "0.27.2", features = ["transpiling"] }
-birdcage = { version = "0.3.1" }
+birdcage = { version = "0.3.1", path = "../../birdcage" }
```
maxrake, line 67: Is this path entry left over from local development?

maxrake (changes requested):
Even though these new features were added as arguments to a hidden `phylum sandbox` command, should there still be a `CHANGELOG.md` entry?

Local testing has not happened yet since there are stray entries in `cli/Cargo.toml` preventing building. Plus, it looks like this is a draft now and there might be more changes coming. Local testing will happen after this PR comes out of draft.

cli/Cargo.toml
cd-work: That's why this is still a draft, will need to wait for the birdcage PR to be merged.

cli/src/commands/extensions/api.rs
```diff
@@ -393,6 +397,30 @@ async fn parse_lockfile(
     Ok(PackageLock { packages: parsed.packages, format: parsed.format })
 }
 
+/// Return error when trying to sandbox on Windows.
+#[op]
+#[cfg(not(unix))]
+fn run_sandboxed(process: Process) -> Result<ProcessOutput> {
+    if !process.exceptions.best_effort {
+        Err(anyhow!("Extension sandboxing is not supported on this platform"))
+    }
```
kylewillmon, line 406: Missing `return`...

```suggestion
    if !process.exceptions.best_effort {
        return Err(anyhow!("Extension sandboxing is not supported on this platform"));
    }
```

cli/src/commands/extensions/api.rs
```diff
@@ -393,6 +397,30 @@ async fn parse_lockfile(
...
+
```
kylewillmon, line 407: We've got a debug message when the min landlock version isn't met. We should probably have a similar message here (and anywhere else that we use `best_effort` in the future).

```suggestion
    log::debug!("Sandboxing unsupported on this platform, skipping sandbox")

```

kylewillmon (changes requested):

cli/src/commands/extensions/permissions.rs
```diff
@@ -276,9 +244,24 @@ impl From<&Permissions> for PermissionsOptions {
...
-pub fn default_sandbox() -> SandboxResult<Birdcage> {
+pub fn default_sandbox(strict: bool, min_landlock_abi: u8) -> SandboxResult<Birdcage> {
+    #[cfg(not(target_os = "linux"))]
     let mut birdcage = Birdcage::new()?;
 
+    #[cfg(target_os = "linux")]
+    let abi = match min_landlock_abi {
+        3 => LANDLOCK_ABI::V3,
+        2 => LANDLOCK_ABI::V2,
+        _ => LANDLOCK_ABI::V1,
```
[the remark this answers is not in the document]
cd-work: I just wanted to avoid having to error out, but I suppose I could just do that.

cd-work:
Made obsolete with https://github.com/phylum-dev/birdcage/pull/47.

Issue #1236, opened 2023-09-19
Fix lockfile generation in workspaces


cd-work:
Currently workspace generation does not work properly with workspaces in many ecosystems, since it always expects the manifest and lockfile to be generated in the same directory.

The following ecosystems are missing support:
 - [X] Cargo
 - [x] Yarn
 - [x] Npm
 - [x] Pnpm
 - [x] Bundler (no workspaces)
 - [x] .NET
 - [x] Gradle
 - [x] Maven
 - [x] Go
 - [x] Pip (no workspaces)
 - [x] Poetry (no workspaces)
 - [x] Pipenv (no workspaces)

cd-work:
Go seems to be a bit special, there is workspace support and when running `go install` it will place a `go.work.sum` in the root of the project, however our current lockfile generation uses `go get -d`, which always generates the `go.sum` inside the manifest's directory.

So I don't think we need to do anything for Go workspace support, since our solution sidesteps it. Due to our automatic cleanup of generated files, this also shouldn't leave anything behind.

Should we still need it, then the following patch should work:

<details>

<summary>Patch</summary>

```
    fn lockfile_path(&self, manifest_path: &Path) -> Result<PathBuf> {
        let project_path = manifest_path
            .parent()
            .ok_or_else(|| Error::InvalidManifest(manifest_path.to_path_buf()))?;

        let output =
            Command::new("go").current_dir(&project_path).args(["env", "GOWORK"]).output()?;

        // Ensure command was successful.
        if !output.status.success() {
            let stderr = String::from_utf8_lossy(&output.stderr);
            return Err(Error::NonZeroExit(output.status.code(), stderr.into()));
        }

        // Get workspace root path.
        let stdout = String::from_utf8(output.stdout).map_err(Error::InvalidUtf8)?;
        let workspace_manifest = stdout.trim();

        if workspace_manifest.is_empty() {
            // If no workspace path is set, assume there is no workspace.
            Ok(project_path.join("go.sum"))
        } else {
            let workspace_manifest = PathBuf::from(workspace_manifest);
            let workspace_root = workspace_manifest
                .parent()
                .ok_or_else(|| Error::InvalidManifest(manifest_path.to_path_buf()))?;
            Ok(workspace_root.join("go.work.sum"))
        }
    }
```

</details>

cd-work:
I've confirmed that Gradle also generates a `gradle.lockfile` in the manifest directory when executing it in a multi-project workspace.

[...]

PR #1237, opened 2023-09-19
Fix JavaScript workspace lockfile generation

cd-work:
This patch fixes the lockfile generation for workspaces with npm, yarn, and pnpm.

See #1236.

### DIFF
--- a/lockfile_generator/src/lib.rs
+++ b/lockfile_generator/src/lib.rs
@@ -145,6 +145,7 @@ pub enum Error {
     InvalidManifest(PathBuf),
     PipReportVersionMismatch(&'static str, String),
     UnsupportedCommandVersion(&'static str, &'static str, String),
+    UnexpectedOutput(&'static str, String),
     Anyhow(anyhow::Error),
     NoLockfileGenerated,
 }
@@ -184,8 +185,9 @@ impl Display for Error {
             Self::UnsupportedCommandVersion(command, expected, received) => {
                 write!(f, "unsupported {command:?} version {received:?}, expected {expected:?}")
             },
-            Self::NoLockfileGenerated => write!(f, "no lockfile was generated"),
             Self::Anyhow(err) => write!(f, "{err}"),
+            Self::UnexpectedOutput(cmd, output) => write!(f, "unexpected output for {cmd:?}: {output}"),
+            Self::NoLockfileGenerated => write!(f, "no lockfile was generated"),
         }
     }
 }
--- a/lockfile_generator/src/lib.rs
+++ b/lockfile_generator/src/lib.rs
@@ -186,7 +186,9 @@ impl Display for Error {
                 write!(f, "unsupported {command:?} version {received:?}, expected {expected:?}")
             },
             Self::Anyhow(err) => write!(f, "{err}"),
-            Self::UnexpectedOutput(cmd, output) => write!(f, "unexpected output for {cmd:?}: {output}"),
+            Self::UnexpectedOutput(cmd, output) => {
+                write!(f, "unexpected output for {cmd:?}: {output}")
+            },
             Self::NoLockfileGenerated => write!(f, "no lockfile was generated"),
         }
     }
--- a/lockfile_generator/src/lib.rs
+++ b/lockfile_generator/src/lib.rs
@@ -145,8 +145,8 @@ pub enum Error {
     InvalidManifest(PathBuf),
     PipReportVersionMismatch(&'static str, String),
     UnsupportedCommandVersion(&'static str, &'static str, String),
-    UnexpectedOutput(&'static str, String),
     Anyhow(anyhow::Error),
+    UnexpectedOutput(&'static str, String),
     NoLockfileGenerated,
 }
 
--- a/lockfile_generator/src/lib.rs
+++ b/lockfile_generator/src/lib.rs
@@ -1,7 +1,5 @@
-use std::error::Error as StdError;
 use std::ffi::OsString;
-use std::fmt::{self, Display, Formatter};
-use std::path::{Path, PathBuf};
+use std::path::{Path, PathBuf, StripPrefixError};
 use std::process::{Command, Stdio};
 use std::string::FromUtf8Error;
 use std::{fs, io};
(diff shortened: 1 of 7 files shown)


lockfile_generator/src/pnpm.rs
```diff
@@ -25,3 +43,18 @@ impl Generator for Pnpm {
...
+        let dir = fs::read_dir(path).ok()?;
+
+        for dir_entry in dir.into_iter().flatten().map(|entry| entry.path()) {
+            if dir_entry.file_name().map_or(false, |name| name == WORKSPACE_MANIFEST_FILENAME) {
+                return Some(path.into());
+            }
+        }
+
+        path = path.parent()?;
+    }
```
kylewillmon, line 59: You could consider using `path.ancestors()` here instead.

```rust
for path in path.ancestors() {
   // ...
}
None
```

It avoids `mut` and is perhaps slightly more descriptive.

I guess this entire function could even be `path.ancestors().find(...).map(PathBuf::from)`

lockfile_generator/src/yarn.rs
```diff
@@ -13,7 +15,38 @@ impl Generator for Yarn {
...
+        // Sort by longest location path, to prefer longer matches.
+        workspaces.sort_by(|a, b| b.location.len().cmp(&a.location.len()));
+
+        // Find workspace root by stripping the first matching location path.
+        let project_str = project_path
+            .to_str()
+            .ok_or_else(|| Error::InvalidManifest(manifest_path.to_path_buf()))?;
+        let workspace_root = workspaces
+            .into_iter()
+            .find_map(|project| project_str.strip_suffix(&project.location))
```
kylewillmon, line 46: A contrived example:

```sh-session
> pwd
/Users/kyle/tmp-js/foo

> yarn workspaces list --json
{"location":".","name":"tmp-js"}
{"location":"foo","name":"foo"}
{"location":"bar","name":"bar"}
{"location":"js/foo","name":"js-foo"}
```

This would incorrectly pick `/Users/kyle/tmp-/yarn.lock` as the lockfile location.

I don't think `yarn workspaces list --json` actually gives enough information to correctly find the root workspace...

lockfile_generator/src/lib.rs
```diff
@@ -184,8 +185,11 @@ impl Display for Error {
             Self::UnsupportedCommandVersion(command, expected, received) => {
                 write!(f, "unsupported {command:?} version {received:?}, expected {expected:?}")
             },
-            Self::NoLockfileGenerated => write!(f, "no lockfile was generated"),
             Self::Anyhow(err) => write!(f, "{err}"),
+            Self::UnexpectedOutput(cmd, output) => {
+                write!(f, "unexpected output for {cmd:?}: {output}")
+            },
+            Self::NoLockfileGenerated => write!(f, "no lockfile was generated"),
```
kylewillmon, line 192: If the goal was to keep these in the same order as above, you'll need to flip `Anyhow` and `UnexpectedOutput`

```suggestion
            Self::UnexpectedOutput(cmd, output) => {
                write!(f, "unexpected output for {cmd:?}: {output}")
            },
            Self::Anyhow(err) => write!(f, "{err}"),
            Self::NoLockfileGenerated => write!(f, "no lockfile was generated"),
```

kylewillmon (changes requested):

lockfile_generator/src/yarn.rs
cd-work: Right that's why I originally used `ends_with` on the path, which I promptly forgot about apparently.

Would you suggest removing yarn completely then? It's the best I found.

kylewillmon: It's not bad enough to drop support for yarn... This still improves the status quo. We just have to accept the possibility of this annoying edge case.

Using `ends_with` to ensure that the break happens between path components would certainly be a welcome improvement.

lockfile_generator/src/yarn.rs
```diff
@@ -13,7 +15,38 @@ impl Generator for Yarn {
...
+
+        // Convert output to actual valid json.
+        let stdout = String::from_utf8(output.stdout).map_err(Error::InvalidUtf8)?;
+        let workspaces_json = format!("[{}]", stdout.trim_end().replace('\n', ","));
+
+        // Parse workspace list.
+        let mut workspaces: Vec<Workspace> = serde_json::from_str(&workspaces_json)?;
+
+        // Sort by longest location path, to prefer longer matches.
+        workspaces.sort_by(|a, b| b.location.len().cmp(&a.location.len()));
```
cd-work, line 38: This should probably use `components.len()` rather than the path length.

lockfile_generator/src/yarn.rs
maxrake: For another approach, take a look at the [`find-yarn-workspace-root` project](https://github.com/square/find-yarn-workspace-root/tree/master), which itself [adapted the algorithm directly from `yarn`](https://github.com/square/find-yarn-workspace-root/blob/master/index.js#L11), here:

https://github.com/yarnpkg/yarn/blob/ddf2f9ade211195372236c2f39a75b00fa18d4de/src/config.js#L612

Granted, this reference is for the `yarn` v1.x branch and the same code could not be found [in the current v3.x branch](https://github.com/yarnpkg/berry), but the algorithm appears to still be valid.

maxrake (changes requested):
The changes were tested locally and found to work for `npm` (using a self-created workspace setup) and `pnpm` (using a checkout of [SvelteKit](https://github.com/sveltejs/kit)).

For `yarn`, a checkout of [Facebook's `react` repo](https://github.com/facebook/react) was used as an example of a modern yarn v3.x workspace project. The changes in this PR failed to parse the `package.json` file from within the `packages/react-art` directory (which was just selected as a quick test). 

```
../react/packages/react-art  2  6 on  main is 📦 v18.2.0 via  v20.6.1
❯ /Users/maxrake/dev/phylum/localdev/cli/target/debug/phylum parse -t yarn ./package.json > parse.json
Generating lockfile for manifest "package.json" using Yarn…
❗ Error: could not parse lockfile: ./package.json

Caused by:
    0: Lockfile generation failed! For details, see: https://docs.phylum.io/docs/lockfile_generation
    1: package manager exited with error code 1:
```

lockfile_generator/src/yarn.rs
cd-work: Doesn't look too bad, I can probably adapt that. Though I hate having to rely on an ever-changing implementation like that.

maxrake (dismissed):
The code was reviewed and the changes tested locally. This time, for `yarn`, the [`jest` project](https://github.com/jestjs/jest/tree/main) was used as a true representative example of a modern Yarn project. The changes worked as expected, for both the project root and for individual workspaces.

LGTM.

lockfile_generator/src/npm.rs
```diff
@@ -33,3 +39,107 @@ impl Generator for Npm {
...
+        .canonicalize()?;
+
+    // Search parent directories for workspace manifests.
+    let mut path = original_root.as_path();
+    for _ in 0..WORKSPACE_ROOT_RECURSION_LIMIT {
+        path = match path.parent() {
+            Some(root_dir) => root_dir,
+            None => break,
+        };
+
```
kylewillmon, line 58: This loop could be simplified with `.ancestors()`

```suggestion
    // Search parent directories for workspace manifests.
    for path in original_root.ancestors().skip(1).take(WORKSPACE_ROOT_RECURSION_LIMIT) {
```

kylewillmon (approved):

kylewillmon (approved):

PR #1238, opened 2023-09-21
Fix allow all networking permissions

kylewillmon:
This patch fixes a regression from e166fce that was causing `net = true` permissions to be handled like `net = false`. This meant than an extension using `net = true` would only be able to access the Phylum API domains from their extension code. Subprocesses and the network sandbox were unaffected by this regression.

### DIFF
--- a/cli/src/commands/extensions/permissions.rs
+++ b/cli/src/commands/extensions/permissions.rs
@@ -44,6 +44,17 @@ impl Permission {
         }
     }
 
+    /// Add resource to allow list
+    pub fn allow_resource(&mut self, resource: String) {
+        match self {
+            Self::Boolean(true) => {},
+            Self::Boolean(false) => {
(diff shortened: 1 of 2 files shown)


maxrake (approved):
The changes make sense and the added test is very welcome. LGTM.

cli/src/commands/extensions/permissions.rs
```diff
@@ -44,6 +44,17 @@ impl Permission {
         }
     }
 
+    /// Add resource to allow list
+    pub fn allow_resource(&mut self, resource: String) {
+        match self {
+            Self::Boolean(true) => {},
```
cd-work, line 50:
```suggestion
            Self::Boolean(true) => (),
```

cd-work (approved):

maxrake (approved):
LGTM

PR #1244, opened 2023-10-02
Fix issues with cargo-update workflow

kylewillmon:
This patch addresses many little issues:

- The backtick (\`) character was being interpreted as shell expansion in the `git commit` command
- The `noreply` email for `phylum-bot` did not include the user ID
- This repository does not use Conventional Commits, so the `build:` prefix was unnecessary

### DIFF
--- a/.github/workflows/cargo-update.yml
+++ b/.github/workflows/cargo-update.yml
@@ -22,8 +22,8 @@ jobs:
         continue-on-error: true
         run: |
           git config user.name 'Phylum Bot'
-          git config user.email 'phylum-bot@users.noreply.github.com'
-          git commit -a -m "build: Bump `Cargo.lock` dependencies"
+          git config user.email '69485888+phylum-bot@users.noreply.github.com'
+          git commit -a -m "Bump dependencies"
           git push --force origin HEAD:auto-cargo-update
 
       - name: Create Pull Request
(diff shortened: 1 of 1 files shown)


.github/workflows/cargo-update.yml
```diff
@@ -22,8 +22,8 @@ jobs:
         continue-on-error: true
         run: |
           git config user.name 'Phylum Bot'
-          git config user.email 'phylum-bot@users.noreply.github.com'
-          git commit -a -m "build: Bump `Cargo.lock` dependencies"
+          git config user.email '69485888+phylum-bot@users.noreply.github.com'
+          git commit -a -m "Bump dependencies"
```
kylewillmon, line 26: Also considered:

```suggestion
          git commit -a -m 'Bump `Cargo.toml` dependencies'
```

But I decided simpler was better

.github/workflows/cargo-update.yml
```diff
@@ -22,8 +22,8 @@ jobs:
         continue-on-error: true
         run: |
           git config user.name 'Phylum Bot'
```
maxrake, line 24:
```suggestion
          git config user.name 'phylum-bot'
```
Not sure if this matters as well...just a suggestion.

maxrake (approved):
LGTM.

.github/workflows/cargo-update.yml
kylewillmon: Good catch. It matters if we want these to perfectly match the commits created by the GitHub web flow, so we might as well make the change. But I don't think it matters for user matching or display.

maxrake (approved):
LGTM

PR #1246, opened 2023-10-04
Fix go lockfile generation

cd-work:
While lockfile generation was already implemented, it was never actually enabled in the `lockfile` crate. This patch enables the Go generator, which has been tested both with workspaces and normal go projects.


kylewillmon (approved):

maxrake:
It is kind of odd how the Phylum GitHub App reported no dependency changes in this PR but still put a comment on it showing a (new) high vulnerability finding for the `webpki@0.22.1` package.

PR #1252, opened 2023-10-11
Ignore obj directory of the test csproj fixture

kylewillmon:
These files are unnecesary and they contain system-specific file paths that change when the tests are run. This patch removes and ignores these files.


maxrake (approved):
LGTM

PR #1257, opened 2023-10-19
Unify clap `value_name` case

cd-work:
Previously the case for the value names of CLI arguments was mixed lower and upper case. Since the standard on unix systems is usually uppercase, all options have been changed to comply with this format.


maxrake (approved):
LGTM

PR #1259, opened 2023-10-20
Fix path dependencies for pnpm lockfile parser

### LINKED ISSUE #1258: Workspace entries in `pnpm` lockfiles fail to parse
# Overview

Lockfiles for workspace `pnpm` projects containing `file:` entries cause parsing failures.

## How To Reproduce

Using the `isildurs_bane` repo:

```
../isildurs_bane/pnpm_workspace  1  4 on  master [?] is 📦 v1.0.0 via  v20.8.1
❯ phylum --version
phylum v5.7.3

../isildurs_bane/pnpm_workspace  1  4 on  master [?] is 📦 v1.0.0 via  v20.8.1
❯ ls -alh
total 32
drwxr-xr-x   7 maxrake  staff   224B Oct  5 11:17 .
drwxr-xr-x  24 maxrake  staff   768B Oct 19 16:10 ..
-rw-r--r--   1 maxrake  staff   1.2K Oct  5 11:17 package-lock.json
-rw-r--r--   1 maxrake  staff   336B Oct  5 11:17 package.json
-rw-r--r--   1 maxrake  staff   890B Oct  5 11:17 pnpm-lock.yaml
-rw-r--r--   1 maxrake  staff    27B Oct  5 11:17 pnpm-workspace.yaml
drwxr-xr-x   3 maxrake  staff    96B Oct  5 11:17 projects

../isildurs_bane/pnpm_workspace  1  4 on  master [?] is 📦 v1.0.0 via  v20.8.1
❯ phylum parse -t pnpm pnpm-lock.yaml
❗ Error: could not parse lockfile: pnpm-lock.yaml

Caused by:
    0: Failed to parse lockfile
    1: Dependency 'file:projects/workspace_member' is missing '/' prefix
```

## Expected Behavior

Lockfiles created with `pnpm`, from workspace projects, are able to be parsed.

## Additional Context

The `pnpm-lock.yaml` file was re-created with the `package.json` file by using the command `pnpm install --lockfile-only --ignore-scripts` and it ended up being exactly the same as the one that was already in the repo.

cd-work:
Closes #1258.

### DIFF
--- a/tests/fixtures/pnpm-lock.yaml
+++ b/tests/fixtures/pnpm-lock.yaml
@@ -482,6 +482,4 @@ packages:
   file:projects/workspace_member:
     resolution: {directory: projects/workspace_member, type: directory}
     name: workspace_member
-    dependencies:
-      typescript: 5.2.2
     dev: false
--- a/tests/fixtures/pnpm-lock.yaml
+++ b/tests/fixtures/pnpm-lock.yaml
@@ -478,3 +478,10 @@ packages:
     name: testing
     version: 1.0.0
     dev: false
+
+  file:projects/workspace_member:
+    resolution: {directory: projects/workspace_member, type: directory}
+    name: workspace_member
+    dependencies:
+      typescript: 5.2.2
+    dev: false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+- Path dependencies for pnpm lockfiles
+
 ## [5.7.3] - 2023-10-17
 
 ### Fixed
--- a/lockfile/src/javascript.rs
+++ b/lockfile/src/javascript.rs
@@ -312,11 +312,14 @@ impl PnpmLock {
 
         for (name, package) in self.packages.into_iter() {
             // Parse package based on available fields.
+            let directory = package.resolution.directory;
             let tarball = package.resolution.tarball;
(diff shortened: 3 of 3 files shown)


tests/fixtures/pnpm-lock.yaml
```diff
@@ -478,3 +478,10 @@ packages:
     name: testing
     version: 1.0.0
     dev: false
+
+  file:projects/workspace_member:
+    resolution: {directory: projects/workspace_member, type: directory}
+    name: workspace_member
+    dependencies:
+      typescript: 5.2.2
```
maxrake, line 486: Since this entry would normally also be in the same lockfile, it might be worth adding it to this test lockfile...so it doesn't trip someone up (like it did me). It would be this block added:

```
  /typescript@5.2.2:
    resolution: {integrity: sha512-mI4WrpHsbCIcwT9cF4FZvr80QUeKvsUsUvKDoR+X/7XHQH98xYD8YHZg7ANtz2GtZt/CBq2QJ0thkGJMHfqc1w==}
    engines: {node: '>=14.17'}
    hasBin: true
    dev: false
```

If so, the number of packages parsed in the test would need to be bumped up by one.

maxrake (approved):
The changes here work to parse the same file from #1258

There is one suggestion...but it is not enough to hold things up here. LGTM.

tests/fixtures/pnpm-lock.yaml
cd-work: You'll find that many of our test lockfiles are written by hand and not 100% correct. But their purpose is to test our parser not necessarily to be correct lockfiles.

I've still fixed this just because you asked so nicely. :)

maxrake (approved):
LGTM

PR #1264, opened 2023-10-24
Add missing changelog entries for repository URL

cd-work:


maxrake (approved):
I forgot to mention it before, but there is an entry in the 5.7.3 section that could also be updated:

```
Automatic manifest resolution with subcommands `init`/`parse`/`analyze` will...
```

...to add `status`

PR #1267, opened 2023-10-26
Fix `find-permissions` for symlinks

cd-work:
This fixes an issue where dependencies on symlinked directories or files would report overly coarse results.


cd-work:
Converting to draft until we get the latest Birdcage changes released, since these significantly change symlink handling.

cd-work:
Updated locally and confirmed that this still fixes symlink handling.

kylewillmon (approved):

PR #1269, opened 2023-10-27
Add support for build.gradle.kts manifests

cd-work:
This fixes an issue where build.gradle files written in Kotlin wouldn't be detected as manifests.


kylewillmon (dismissed):

maxrake (changes requested):
A new entry in the [lockfile_generation](https://docs.phylum.io/docs/lockfile_generation) page is needed for this additional manifest name. That is in the `docs/lockfile_generation.md` file.

Also need a CHANGELOG entry.

maxrake (approved):
LGTM

PR #1270, opened 2023-10-27
Fix pnpm lockfile generation

cd-work:
This fixes an issue where the pnpm lockfile generator would fall back to `./pnpm-lock.yaml` if no workspace was present, instead of the correct `<MANIFEST_PATH>/../pnpm-lock.yaml`.


kylewillmon (dismissed):

maxrake (changes requested):
CHANGELOG entry needed for this bug fix.

maxrake (approved):
LGTM

maxrake (approved):
LGTM

PR #1276, opened 2023-11-06
Bump deno version

kylewillmon:
This includes migration to the new `op2` macro because `op` is now deprecated (See denoland/deno_core#279). The new macro requires explicit annotation for string and serde parameters and return values because of the performance hit. We use these heavily (and don't care about the performance).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4075,9 +4075,9 @@ dependencies = [
 
 [[package]]
 name = "rsa"
-version = "0.9.3"
+version = "0.9.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "86ef35bf3e7fe15a53c4ab08a998e42271eab13eb0db224126bc7bc4c4bad96d"
+checksum = "af6c4b23d99685a1408194da11270ef8e9809aff951cc70ec9b17350b087e474"
 dependencies = [
  "const-oid",
  "digest",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,14 +34,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
  "crypto-common",
- "generic-array 0.14.7",
+ "generic-array",
+]
+
+[[package]]
+name = "aead-gcm-stream"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4a09ecb526d53de2842cc876ee5c9b51161ee60399edeca4cf74892a01b48177"
+dependencies = [
+ "aead",
+ "aes",
+ "cipher",
+ "ctr",
+ "ghash",
+ "subtle",
 ]
 
 [[package]]
 name = "aes"
-version = "0.8.2"
+version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


cli/src/commands/extensions/api.rs
```diff
@@ -496,19 +514,22 @@ fn add_permission_args<'a>(
 }
 
 /// Return error when trying to sandbox on Windows.
-#[op]
 #[cfg(not(unix))]
-fn run_sandboxed(_process: Process) -> Result<ProcessOutput> {
+#[op2]
+#[serde]
+fn run_sandboxed(#[serde] _process: Process) -> Result<ProcessOutput> {
```
kylewillmon, line 520: Just have to hope I did this one right... We don't test native Windows builds anymore in CI (and I don't have a Windows host to test on)...

Maybe I can convince @furi0us333 to test this for me... Or we should discuss using [windows-xl][1] runners (at cost) periodically to test this sort of stuff.

[1]: https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners

kylewillmon:
Looks like the build requires `protoc` now, which is causing CI to fail....

I'll investigate later today to see if I can avoid the dependency

cli/src/commands/extensions/api.rs
furi0us333: I am happy to test something if needed, though I always try and push people to using WSL.

kylewillmon: Me too. We don't officially support Windows outside of WSL.

But we've historically made an attempt to at least build successfully on native Windows. So I was just curious if that still worked or not...

kylewillmon:
It looks like there is no opt out. `protoc` has been required to build deno since denoland/deno@2d9298f

I'll update the workflows to restore CI.

cd-work (dismissed):

cd-work (approved):

Cargo.lock
```diff
@@ -4418,6 +4471,22 @@ version = "1.0.15"
...
+
+[[package]]
+name = "saffron"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "03fb9a628596fc7590eb7edbf7b0613287be78df107f5f97b118aad59fb2eea9"
+dependencies = [
+ "chrono",
+ "nom 5.1.3",
+]
```
kylewillmon, line 4488: This project is (basically) dead, but [deno is aware][1]. Hopefully, they'll replace it or get PRs merged to update it soon.

[1]: https://github.com/denoland/deno/pull/21019#discussion_r1377019647

kylewillmon:
I've opened littledivy/aead-gcm-stream#2 to fix the build error on Rust 1.71.0

Hoping to get a quick response there. I'd rather not bump MSRV beyond the latest 3 minor releases.

cd-work (approved):

cli/src/deno.rs
```diff
@@ -49,12 +49,8 @@ pub async fn run(
     let main_module =
         deno_core::resolve_path(&extension.entry_point().to_string_lossy(), &PathBuf::from("."))?;
 
-    let bootstrap = BootstrapOptions {
-        args,
-        runtime_version: env!("CARGO_PKG_VERSION").into(),
```
kylewillmon, line 54: This was putting the Phylum CLI version in `Deno.version.deno`. That variable is now blank (set [here][1] in the `deno_runtime` snapshot).

[1]: https://github.com/denoland/deno/blob/9444bab18647bc871b63716a3a70b2aebaaf3821/runtime/build.rs#L38-L40

We could consider injecting the Phylum CLI version into the extension runtime somewhere else. But that would be a separate PR.

PR #1279, opened 2023-11-07
Move CHANGELOG entries to correct section

kylewillmon:
The entries from #1269 and #1270 were accidentally put in the wrong section. The patch fixes that.

It looks like #1269 was based on an old `main`, so the entry ended up in the wrong spot when merged. And since #1270 was done on the same day, it was natural to put that changelog entry right next to the other one.


maxrake (approved):
LGTM

PR #1281, opened 2023-11-09
Add option to disable lockfile generation

### LINKED ISSUE #1094: Provide ability to opt-out of lockfile generation
# Overview
Manifest support was added in CLI v5.1.0 to allow analyzing manifest files by generating lockfiles on-demand. This issue is to provide (more?) mechanisms for opting out of this behavior.

## Additional Context

If there are already means to opt-out of the lockfile generation (e.g., specifying explicit entries in the `.phylum_project` file), they should be documented more clearly. For example, that could be in the form of an update to a current page on the https://docs.phylum.io/ site or maybe a new page (e.g., KB article for "manifests vs. lockfiles" or perhaps a new Troubleshooting/Tips/FAQ page).

## Acceptance Criteria
- [x] Automatic lockfile generation is retained as the default behavior
- [x] Opting out of automatic lockfile generation is possible for use cases that do not involve the `.phylum_project` file
- [x] Documentation is created/updated to be more clear about the options for opt-out

cd-work:
Closes #1094

### DIFF
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -114,8 +114,12 @@ pub fn parse_lockfile(
         // Return the original lockfile parsing error.
         match lockfile_error {
             Some(err) => return Err(err),
-            // If it's neither manifest nor lockfile, `try_get_packages` is used instead.
-            None => unreachable!("neither lockfile nor manifest"),
+            None => {
+                return Err(anyhow!(
+                    "Parsing {path:?} requires lockfile generation, but it was disabled through \
+                     the CLI"
+                ))
+            },
         }
     }
 
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -47,6 +47,8 @@ pub fn lockfile_types(add_auto: bool) -> Vec<&'static str> {
 
 pub fn handle_parse(matches: &clap::ArgMatches) -> CommandResult {
     let sandbox_generation = !matches.get_flag("skip-sandbox");
+    let generate_lockfiles = !matches.get_flag("no-generation");
+
     let project = phylum_project::get_current_project();
     let project_root = project.as_ref().map(|p| p.root());
     let lockfiles = config::lockfiles(matches, project.as_ref())?;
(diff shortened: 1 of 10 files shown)


maxrake (changes requested):
First pass quick review:
* CHANGELOG entry needed
* Docs need to be regenerated
* There is a deno check failure in CI
* Add more detail to the [lockfile generation](https://docs.phylum.io/docs/lockfile_generation) page for explaining the options for opt-out

cli/src/commands/parse.rs
```diff
@@ -98,8 +103,8 @@ pub fn parse_lockfile(
         }
     }
 
-    // If the path is neither a valid manifest nor lockfile, we abort.
-    if !parser.is_path_manifest(&path) {
+    // Abort if generation is disabled or path is neither lockfile nor manifest.
+    if !generate_lockfiles || !parser.is_path_manifest(&path) {
```
maxrake, line 107: There is a problem with the logic in this block. The `unreachable` macro is triggered with the changes from this PR. Here is an example from [a library project](https://github.com/deepcausality-rs/deep_causality) where there are only manifests:

```
../deep_causality  9  19 on  main [?] via 🦀 v1.73.0
❯ /Users/maxrake/dev/phylum/localdev/cli/target/debug/phylum --version
phylum v5.8.1-4-ge70c5c1

../deep_causality  9  19 on  main [?] via 🦀 v1.73.0
❯ /Users/maxrake/dev/phylum/localdev/cli/target/debug/phylum parse --no-generation ./Cargo.toml
thread 'main' panicked at cli/src/commands/parse.rs:112:21:
internal error: entered unreachable code: neither lockfile nor manifest
```

maxrake (changes requested):
The code was reviewed. The doc updates and changelog entries look good. Local testing revealed an issue.

cli/src/commands/parse.rs
cd-work: Great catch, thanks.

maxrake (dismissed):
LGTM

maxrake (approved):
Code reviewed and tested locally. LGTM.

PR #1287, opened 2023-11-15
Fix package-lock.json deps with aliased names

### LINKED ISSUE #1285: Incorrect parsing for renamed npm packages
The `package-lock.json` parser seems to misinterpret renamed dependencies.

For example, see [this entry][1] in `@isaacs/cliui`'s `package.json`:

[1]: https://github.com/isaacs/cliui/blob/aa397fedbd0550c9925af6b62f970de663285641/package.json#L55

```
    "strip-ansi-cjs": "npm:strip-ansi@^6.0.1",
```

That causes a section in `package-lock.json` like this:
```json
    "node_modules/strip-ansi-cjs": {
      "name": "strip-ansi",
      "version": "6.0.1",
      "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-6.0.1.tgz",
      "integrity": "sha512-Y38VPSHcqkFrCpFnQ9vuSXmquuv5oXOKpGeT6aGrr3o3Gc9AlVa6JBfUSOCnbxGGZF+/0ooI7KrPuUSztUdU5A==",
      "dependencies": {
        "ansi-regex": "^5.0.1"
      },
      "engines": {
        "node": ">=8"
      }
    },
```

The package name here is `strip-ansi`. It has just been locally renamed to `strip-ansi-cjs`. Unfortunately, `phylum parse` is fooled:

```
> phylum parse | jq '.[] | select(.name=="strip-ansi-cjs")'
{
  "name": "strip-ansi-cjs",
  "version": "6.0.1",
  "type": "npm",
  "lockfile": "package-lock.json"
}
```

Instead, `phylum parse` should detect this as a package named `strip-ansi`.

cd-work:
Closes #1285.

### DIFF
--- a/lockfile/src/javascript.rs
+++ b/lockfile/src/javascript.rs
@@ -30,15 +30,15 @@ impl Parse for PackageLock {
         let parsed: JsonValue = serde_json::from_str(data)?;
 
         // Get a field as string from a JSON object.
-        let get_field = |value: &JsonValue, key| {
+        fn get_field(value: &JsonValue, key: &str) -> Option<String> {
             value.get(key).and_then(|value| value.as_str()).map(|value| value.to_string())
-        };
+        }
 
         // Get version field from JSON object.
-        let get_version = |value, name| {
+        fn get_version(value: &JsonValue, name: &str) -> anyhow::Result<String> {
             get_field(value, "version")
                 .ok_or_else(|| anyhow!("Failed to parse version for '{name}' dependency"))
-        };
+        }
 
         if let Some(deps) = parsed.get("packages").and_then(|v| v.as_object()) {
             // Parser for package-lock.json >= v7.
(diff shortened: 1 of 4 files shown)


lockfile/src/javascript.rs
```diff
@@ -80,9 +80,12 @@ impl Parse for PackageLock {
                 let resolved = get_field(keys, "resolved")
                     .ok_or_else(|| anyhow!("Dependency '{name}' is missing \"resolved\" key"))?;
 
+                // Handle aliased dependencies.
+                let name = get_field(keys, "name").unwrap_or_else(|| name.into());
```
mathew-horner, line 84: Will the entry even be valid if there is no `name` field? I wouldn't imagine that npm is parsing the name out of the key as a fallback, but who knows 🤷

cd-work: Yeah you can check the test file for this, having no `name` field is actually the "normal" case. Only for aliases the name field is even populated.

It's very unfortunate but otherwise I wouldn't even do all this ridiculous path splitting.

mathew-horner (approved):

PR #1290, opened 2023-11-16
Update and correct CHANGELOGS

maxrake:


cd-work (approved):
I noticed the missing entry when rebasing `no-generation`, but good catch on the extension stuff.

PR #1293, opened 2023-11-17
Add exit code for disabled lockfile generation

### LINKED ISSUE #1291: Return unique error code for disabled lockfile generation
The new `--no-generation` option for the `parse`/`analyze` commands is nice in that it tells you why the command failed:

```
❯ phylum parse -t cargo --no-generation ./Cargo.toml
❗ Error: could not parse lockfile: ./Cargo.toml

Caused by:
    Parsing "./Cargo.toml" requires lockfile generation, but it was disabled through the CLI

❯ echo $?
1
```

It would be even nicer if the exit code was unique instead of `1` like it is for other parse failures. That way it would be possible, when iterating over a range of dependency files, to recognize the `--no-generation` situation and skip over the manifests while allowing lockfiles to continue.

cd-work:
Closes #1291.


cli/src/commands/parse.rs
```diff
@@ -113,13 +138,8 @@ pub fn parse_lockfile(
...
-            Some(err) => return Err(err),
-            None => {
-                return Err(anyhow!(
-                    "Parsing {path:?} requires lockfile generation, but it was disabled through \
-                     the CLI"
-                ))
-            },
+            Some(err) => return Err(err.into()),
+            None => return Err(ParseError::ManifestWithoutGeneration(path)),
         }
```
mathew-horner, line 143: nit, but I think this could be:
```rust
return Err(lockfile_error.map_or(ParseError::ManifestWithoutGeneration(path), Into::into));
```

mathew-horner (approved):

cli/src/commands/parse.rs
cd-work: You can't call `Into::into` on the result itself, it needs to be called on the error.

maxrake (approved):
Tested locally and works as expected. LGTM.

cli/src/commands/parse.rs
mathew-horner: Not sure what you mean. `lockfile_error` isn't a `Result`, it is an instance of `anyhow::Error`, which implements `Into<ParseError>` via your `thiserror` implementation.

PR #1304, opened 2023-11-23
Improve yarn v1 parser

### LINKED ISSUE #1299: Yarn v1 renamed dependencies not handled correctly
The issue found in #1285 also exists for yarn v1 lockfiles.

```
    "@isaacs/cliui": "^8.0.2"
```

When using yarn v1, this incorrectly finds a dependency named `strip-ansi-cjs`.

```
> yarn --version
1.22.21
> yarn install
[ ... SNIP ... ]
> phylum parse | jq '.[] | select(.name=="strip-ansi-cjs")'
{
  "name": "strip-ansi-cjs",
  "version": "6.0.1",
  "type": "npm",
  "lockfile": "yarn.lock"
}
```

Here is the relevant section from `yarn.lock`:

```
"strip-ansi-cjs@npm:strip-ansi@^6.0.1", strip-ansi@^6.0.0, strip-ansi@^6.0.1:
  version "6.0.1"
  resolved "https://registry.yarnpkg.com/strip-ansi/-/strip-ansi-6.0.1.tgz#9e26c63d30f53443e9489495b2105d37b67a85d9"
  integrity sha512-Y38VPSHcqkFrCpFnQ9vuSXmquuv5oXOKpGeT6aGrr3o3Gc9AlVa6JBfUSOCnbxGGZF+/0ooI7KrPuUSztUdU5A==
  dependencies:
    ansi-regex "^5.0.1"
```

Thanks to @dappelt for submitting phylum-dev/vuln-reach#67 and bringing this to my attention.

cd-work:
This fixes multiple issues in the v1 yarn parser:
 - Add support for path dependencies
 - Add support for git:// dependencies
 - Fix aliased dependencies
 - Ignore http(s) dependencies

Closes #1299.


lockfile/src/javascript.rs
```diff
@@ -499,23 +499,59 @@ mod tests {
...
+        for p in [with_trailing_newlines, &without_trailing_newlines] {
+            let pkgs = YarnLock.parse(p).unwrap();
 
-            assert_eq!(pkgs[0].name, "@yarnpkg/lockfile");
-            assert_eq!(pkgs[0].version, PackageVersion::FirstParty("1.1.0".into()));
+            assert_eq!(pkgs.len(), 20);
 
-            assert_eq!(pkgs[3].name, "cliui");
-            assert_eq!(pkgs[3].version, PackageVersion::FirstParty("7.0.4".into()));
+            let expected_pkgs = [
```
mathew-horner, line 510: Hmm, I see that this appears to be a subset of the packages defined in the fixture. What's the reason for having more packages over there than we are asserting on here?

lockfile/src/parsers/yarn.rs
```diff
@@ -29,28 +35,80 @@ fn entry(input: &str) -> IResult<&str, Package> {
...
-            package_type: PackageType::Npm,
-        })
-    })
+fn parse_entry(input: &str) -> IResult<&str, Option<Package>> {
+    let (input, (name, version)) = context("entry", tuple((entry_name, entry_version)))(input)?;
+
+    let version = match version {
+        Some(version) => version,
+        None => return Ok((input, None)),
+    };
```
mathew-horner, line 44: nit: seems like a good place for let/else.
```suggestion
    let Some(version) = version else {
        return Ok((input, None));
    };
```

lockfile/src/parsers/yarn.rs
```diff
@@ -29,28 +35,80 @@ fn entry(input: &str) -> IResult<&str, Package> {
...
+
+    let package = Package { version, name: name.to_string(), package_type: PackageType::Npm };
+
+    Ok((input, Some(package)))
 }
 
 fn entry_name(input: &str) -> IResult<&str, &str> {
+    // Strip optional quotes.
     let (i, _) = opt(tag(r#"""#))(input)?;
```
mathew-horner, line 53: nit again (I know it wasn't changed in this PR), but any reason to use `i` and not shadow `input` like we do in most of the other parsing functions we have in this file?

lockfile/src/parsers/yarn.rs
```diff
@@ -29,28 +35,80 @@ fn entry(input: &str) -> IResult<&str, Package> {
...
+        return git_dep(input);
+    }
+
+    // Ignore HTTP(S) dependencies.
+    //
+    // These could be either git or tar dependencies, so to avoid miscategorization
+    // we just ignore them.
+    if input.starts_with("@http://") || input.starts_with("@https://") {
+        return Ok((input, None));
+    }
```
mathew-horner, line 83: Could you fill me in on the reasoning for ignoring these dependencies or link to any discussion threads about it? I know it says to avoid miscategorization between the two options, but I'm trying to understand the bigger picture of why this problem means we have to completely leave them out of analysis.

maxrake (changes requested):
The code was minimally reviewed. The updated tests are great...and passing in CI to help reassure of the correctness of the changes.

A CHANGELOG update is needed.

lockfile/src/parsers/yarn.rs
cd-work: I do not think `let else` is ever a better option.

lockfile/src/parsers/yarn.rs
cd-work: Happy to change this. I don't like this either.

lockfile/src/parsers/yarn.rs
cd-work: I mean the comment says it all. We don't know if it's a link to a tar package or a git dependency. So if we don't know, we should omit it rather than reporting it incorrectly.

lockfile/src/javascript.rs
mathew-horner: Sounds good. That's kind of what I figured: make sure they don't fail to parse but don't really need explicit assertions for them.

lockfile/src/parsers/yarn.rs
mathew-horner: Interesting, why's that?

cd-work: It's pointless extra syntax that doesn't make code more readable. There's no reason to add new syntax for readers just to save a single line. It's addition to Rust is completely misguided imo. Makes a language that already has way too much special syntax even harder to learn and read.

lockfile/src/parsers/yarn.rs
mathew-horner: I think I answered my own question. Looks like we submit packages as triples (not purls) and we [don't submit HTTP packages at all anyways](https://github.com/phylum-dev/cli/blob/main/cli/src/commands/parse.rs#L363). I figured we'd have some ability to process packages at arbitrary URLs / git repos, but I guess we don't.

lockfile/src/parsers/yarn.rs
mathew-horner: Fair enough 👍

maxrake (approved):

PR #1306, opened 2023-11-28
Add subprocess for sandboxed lockfile generation

### LINKED ISSUE #1296: Enable sandboxed lockfile generation in extensions
#1292 disabled the sandbox for lockfile generation in extensions, with this explanation:

> Currently the lockfile generation sandbox does not spawn a separate process, thus enabling the sandbox for the calling process directly. Since extensions might be doing other things after generating the lockfile for a manifest, this could prevent them from operating correctly.

> This patch removes the sandboxing for lockfile generation when calling `parseLockfile` from an extension. In the future it should be possible to enable this again by spawning a separate process for lockfile generation.

This request is that future placeholder, to enable the sandbox again. Spawning a separate process for lockfile generation and using IPC is one possibility. There may be others.

cd-work:
Before this patch lockfile generation would always happen in the CLI's process, which inevitably applied the sandbox to the process itself making the execution environment after generation severely limited.

To allow for the CLI and extensions to use sandboxed lockfile generation without having to spawn a separate process, the generation itself is now always executed in a separate process if sandboxing is requested.

Closes #1296.

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -564,6 +564,22 @@ pub fn add_subcommands(command: Command) -> Command {
                 .about("Find all lockfile and manifest paths")
                 .hide(true),
         )
+        .subcommand(
+            Command::new("generate-lockfile")
+                .args(&[
+                    Arg::new("lockfile-type")
+                        .value_name("TYPE")
+                        .required(true)
+                        .help("Lockfile type whose generator will be used")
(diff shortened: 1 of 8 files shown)


maxrake (approved):
The code was reviewed and the changes were tested locally, with the same setup that produced errors before. What failed previously is working now. LGTM.

Issue #1271, opened 2023-10-27
Rename uses of "lockfile" term


maxrake:
With the introduction of manifest file support, the CLI is now capable of handling both lockfiles and manifests. Agreement was reached on using the term "dependency file" to represent either a lockfile or a manifest (e.g., generically or before the distinction is clear). Further, `depfile` may be used in situations were the abbreviation is warranted.

The language in this repository needs to change to reflect this...in arguments, interfaces, documentation, code, comments, help output, etc.

This will likely cause breaking changes so it should be done all at one time and in coordination with consumers of the CLI and it's crates. Here is a sampling of the expected changes that are likely to be breaking (but not necessarily a complete list):

* `find-lockable-files` command
* `--lockfile-type` arguments
* `--lockfile` optional arguments (e.g., `init`)
* `lockfiles` key in `.phylum_project`
* `lockfiles` key in `phylum status --json` output
* Extension API
  * `parseLockfile`
  * types and members of types

Care should be taken to ensure a global search and replace is not attempted since there are still instances where it makes sense to talk specifically about _lockfiles_ or specifically about _manifests_.

Creating a release of the CLI with these changes should be done in coordination with documentation updates and releases of the tools that depend on the CLI.

PR #1310, opened 2023-12-04
Switch to dependency file terminology

cd-work:
This patch makes sweeping breaking changes to change the internal and
public terminology to refer to a manifest or lockfile as "dependency
file".

This change should also clarify some CLI interfaces where the previous
terminology was asking for a lockfile, but a manifest file was accepted
without clearly stating it.

Closes https://github.com/phylum-dev/cli/issues/1271.

---

Currently there's two major unsolved things I'm aware of:
 - Types in `phylum_types` are not changed, this might be a good opportunity to get rid of it where possible since deprecation is planned?
 - `LockfileFormat` enum is unchanged, this makes some sense to me since the format is still about one specific lockfile type, but I did change the public CLI interface and some internal code to refer to it as "dependency file type"

### DIFF
--- a/.phylum_project
+++ b/.phylum_project
@@ -1,7 +1,7 @@
-id: d84b0bb3-fb64-43cb-a473-dc96b54db662
+id: f0ea0603-6916-47bc-ad73-8156d7dc40fc
 name: cli
-created_at: 2023-01-26T20:02:11.276987834+01:00
-group_name: integrations
+created_at: 2023-12-04T20:51:24.124463210+01:00
+group_name: null
 lockfiles:
-- path: Cargo.lock
+- path: ./Cargo.lock
   type: cargo
--- a/.phylum_project
+++ b/.phylum_project
@@ -1,7 +1,7 @@
-id: f0ea0603-6916-47bc-ad73-8156d7dc40fc
+id: d84b0bb3-fb64-43cb-a473-dc96b54db662
 name: cli
-created_at: 2023-12-04T20:51:24.124463210+01:00
-group_name: null
+created_at: 2023-01-26T20:02:11.276987834+01:00
+group_name: integrations
 lockfiles:
-- path: ./Cargo.lock
+- path: Cargo.lock
   type: cargo
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Renamed multiple CLI arguments to avoid the term `lockfile` in places where
     manifests are also accepted
+- Renamed `lockfiles` key in `phylum status --json` output to `dependency_files`
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `no-generation` option for `parse`/`analyze` to disable lockfile generation
 - Optional `--project` and `--group` arguments for `phylum project status`
 
+### Changed
+
+- Renamed multiple CLI arguments to avoid the term `lockfile` in places where
+    manifests are also accepted
+
 ### Fixed
 
 - Aliased dependency names in `package-lock.json`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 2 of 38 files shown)


kylewillmon:
> Types in phylum_types are not changed, this might be a good opportunity to get rid of it where possible since deprecation is planned?

I agree. Any types needed for CLI can (and should) be moved to this repo. If we find that they still need to be shared (which I doubt in most cases), they can be exported from here

kylewillmon:
> `LockfileFormat` enum is unchanged, this makes some sense to me since the format is still about one specific lockfile type, but I did change the public CLI interface and some internal code to refer to it as \"dependency file type\"

This seems reasonable and matches with our system of matching all manifest formats to a single associated lockfile format.

.phylum_project
```diff
@@ -1,7 +1,7 @@
-id: d84b0bb3-fb64-43cb-a473-dc96b54db662
+id: f0ea0603-6916-47bc-ad73-8156d7dc40fc
 name: cli
-created_at: 2023-01-26T20:02:11.276987834+01:00
-group_name: integrations
+created_at: 2023-12-04T20:51:24.124463210+01:00
+group_name: null
```
maxrake, line 4: This looks like a `phylum init` was done locally, without the `--group integrations` option, and then committed. Changing the project to not have a group will remove the visibility previously enjoyed by everyone in the `integrations` group.

Now...changing the group to something else is another story...and may be warranted now.

maxrake (changes requested):
Full review coming...but wanted to at least point out the missing CHANGELOG entry for this breaking change...which will likely result in a v6.0.0 major version release.

maxrake (changes requested):
Do we want to rename the `phylum_lockfile` package now?

extensions/phylum.ts
```diff
@@ -105,21 +112,21 @@ export class PhylumApi {
...
   /**
-   * Analyze dependencies in a lockfile.
+   * Run Phylum analysis on a list of packages.
    *
    * Packages are expected in the following format:
    *
    * ```
    * [
-   *   { name: "accepts", version: "1.3.8", type: "npm", lockfile: "/path/to/lockfile" },
+   *   { name: "accepts", version: "1.3.8", type: "npm", origin: "/path/to/lockfile" },
```
maxrake, line 121:
```suggestion
   *   { name: "accepts", version: "1.3.8", type: "npm", origin: "/path/to/dependency.file" },
```

docs/commands/phylum_analyze.md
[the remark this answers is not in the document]
maxrake: After the changes, there are still two files that contain instances of `--lockfile`:

* `doc_templates/phylum_analyze.md`
* `doc_templates/phylum_package.md`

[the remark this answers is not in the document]
cd-work: The `phylum_package` docs were actually *completely* broken. I think there might have been a copy/paste issue.

maxrake (changes requested):
A few issues were created to follow up with the primary consumers of the CLI and it's extension API:

* https://github.com/phylum-dev/community-extensions/issues/29
* https://github.com/phylum-dev/phylum-ci/issues/363

There may still be changes needed in the `phylum-dev/documentation` repository.

CHANGELOG.md
maxrake, line 1: The new entries in this file should be in the `Unreleased` section.

extensions/CHANGELOG.md
maxrake, line 1: The new entries in this file should be in the `Unreleased` section.

extensions/CHANGELOG.md
```diff
@@ -16,6 +16,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `sandboxGeneration` parameter for `parseLockfile` to disable the lockfile
     generation sandbox
 
+### Changed
+
+- Renamed `parseLockfile` to `PhylumApi::parseDependencyFile`
+- Removed `lockfile` field from `PhylumApi::Package` type
+- Removed `PhylumApi::Lockfile` type in favor of `PhylumApi::DependencyFile`
+- Changed `PhylumApi::analyze` packages type `PhylumApi::PackageWithOrigin`
```
maxrake, line 24:
```suggestion
- Changed `PhylumApi::analyze` packages type to `PhylumApi::PackageWithOrigin`
```

maxrake (changes requested):
Almost there...

maxrake (changes requested):
The `extensions/CHANGELOG.md` still has the new entries in the v5.9.0 section instead of the "Unreleased" section.

maxrake (approved):
LGTM

Issue #1313, opened 2023-12-05
`phylum package` should return CVE data.


mathew-horner:
# Overview
CVE details are now returned from the API for issues. We should be displaying those to the user when they call `phylum package`.

You can see an example the requisite additions in this PR: https://github.com/phylum-dev/phylum-types/pull/86 (closed in favor of moving the types to CLI).

# Acceptance Criteria
- [ ] Vulnerability issues should display CVE details from the API when `phylum package` is called.

PR #1314, opened 2023-12-07
Move submit package API types to CLI and add vuln data.

mathew-horner:
In favor of https://github.com/phylum-dev/phylum-types/pull/86, I've moved the `PackageSubmitResponse` and all its child types from `phylum-types` to CLI. I've updated any other code that was referencing these types to the local instances.

Unfortunately a couple `From` impls were necessary to maintain compatibility with the code that still is using `PackageStatusExtended`... I guess I could move that too, but I didn't because it's out of scope for the main purpose of this PR.

closes https://github.com/phylum-dev/cli/issues/1313


cli/src/types.rs
```diff
@@ -126,6 +129,289 @@ pub struct RevokeTokenRequest<'a> {
     pub name: &'a str,
 }
 
+/// Response body for `/data/packages/submit`.
```
mathew-horner, line 132: I touched up some of these doc comments and removed ones that seemed redundant... lmk what you think.

cli/src/types.rs
```diff
@@ -126,18 +129,285 @@ pub struct RevokeTokenRequest<'a> {
...
 
-    #[test]
-    fn test_risk_level_ordering() {
-        assert!(
-            RiskLevel::Info < RiskLevel::Low
-                && RiskLevel::Low < RiskLevel::Medium
-                && RiskLevel::Medium < RiskLevel::High
-                && RiskLevel::High < RiskLevel::Critical,
-            "Ordering of risk levels is invalid"
-        );
```
mathew-horner, line 141: This test seemed unnecessary. The way `Ord` works is very well defined.

mathew-horner:
In order to test these changes, you'll need to point your CLI at staging. The API changes have not been cut to prod yet (deserialization will still succeed since it's an optional field, you just won't see it).

kylewillmon (changes requested):

kylewillmon (approved):

PR #1320, opened 2023-12-15
Move generic depfile parsing to phylum_lockfile

cd-work:
This patch moves a large chunk of the dependency file parsing from the CLI into `phylum_lockfile` to allow others to make use of it.

The interface of the functions has been reworked based on the assumption that filesystem access might not be present, allowing pure in-memory for API usage.


kylewillmon (changes requested):
I still have more testing to do, but I won't get to that until tomorrow and I've already found some items that need addressing

lockfile/src/parse_depfile.rs
```diff
@@ -0,0 +1,250 @@
...
+    // Return the original lockfile parsing error.
+    match lockfile_error {
+        // Report parsing errors only for lockfiles.
+        Some(err) if !maybe_manifest => Err(err),
+        _ => Err(ParseError::ManifestWithoutGeneration(id.into())),
+    }
+}
+
+/// Attempt to get packages from an unknown lockfile type
+fn try_get_packages(id: &str, contents: &str) -> Result<ParsedLockfile, anyhow::Error> {
```
kylewillmon, line 120: I think this function should stay in CLI. Or maybe even be removed entirely. It has always been a bit of a hack to try every parser and I'd rather not extend that behavior anywhere else if we can avoid it.

lockfile/src/parse_depfile.rs
```diff
@@ -0,0 +1,250 @@
...
+    let packages = parse_lockfile_content(&generated_lockfile, parser)?;
+
+    Ok(ParsedLockfile::new(id, format, packages))
+}
+
+/// Attempt to parse a lockfile.
+fn parse_lockfile_content(
+    content: &str,
+    parser: &dyn Parse,
+) -> Result<Vec<PackageDescriptor>, ParseError> {
```
kylewillmon, line 168: If we remove `try_get_packages()`, I think this function would be a better entry point than `parse_depfile()` for applications that don't support lockfile generation.

The `parser` argument could be changed to `format: LockfileFormat` for consistency. If the caller doesn't know the lockfile format, they can use `get_path_format()`, which is already public.

The cases where `get_path_format()` doesn't work (either because no format matches or because the caller doesn't have a file path) might be worth further discussion... But I think `try_get_packages()` is too brittle to be a good answer there.

lockfile/src/parse_depfile.rs
```diff
@@ -0,0 +1,250 @@
...
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
+}
+
+/// Lockfile with all its dependencies.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ParsedLockfile {
+    /// User-facing lockfile identifier (i.e. a file path or name).
+    #[serde(alias = "path")]
+    pub id: String,
```
kylewillmon, line 28: I would rather keep the name `path` for this variable.

Also, since this is only user-facing, we might be able to avoid it entirely in this crate. Currently, it's only used for error messages and `.api_packages()` (which isn't used directly by this crate).

Perhaps we could have functions in this crate return `Vec<PackageDescriptor>` instead of using this struct. Alternately, we could just remove the `id` field from this struct. CLI would then need its own struct to add the `path` where needed.

kylewillmon (changes requested):

lockfile/src/parse_depfile.rs
cd-work: I don't agree with this. **Especially** in a scenario where filesystem access is not available, this becomes useful for determining lockfile contents. The behavior might be a bit hacky, but I'd argue that it works extremely well with zero known real-world conflicts and since parsing is solely done in memory it's also not like this is difficult for us to do.

lockfile/src/parse_depfile.rs
cd-work: I'd rather have this struct in a central location where it is easy to share between other crates. That way we won't have different decisions for this problem in every single crate.

lockfile/src/parse_depfile.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    // Parse with the identified parser.
+    let parser = format.parser();
+
+    // Check if file could be a lockfile based on file name.
+    let maybe_lockfile = pseudopath.as_ref().map_or(false, |path| parser.is_path_lockfile(path));
+
+    // Attempt to parse the identified lockfile.
+    let mut lockfile_error = None;
+    if maybe_lockfile {
```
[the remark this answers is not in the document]
cd-work: Yeah the suggestion is what we did before, so just an oversight.

lockfile/src/parse_depfile.rs
```diff
@@ -0,0 +1,250 @@
...
+    let pseudopath = file_name.map(PathBuf::from);
+
+    // Try to determine the dependency file format.
+    let format = format.or_else(|| pseudopath.as_ref().and_then(crate::get_path_format));
+
+    // Attempt to parse with all known parsers as fallback.
+    let format = match format {
+        Some(format) => format,
+        None => return Ok(try_get_packages(id, contents)?),
+    };
```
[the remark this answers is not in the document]
cd-work: Right I just didn't consider manifests at all since that is handled by CLI.

lockfile/src/parse_depfile.rs
cd-work: Really not a fan of what this did to our `generate_lockfile` method.

lockfile/src/parse_depfile.rs
```diff
@@ -0,0 +1,249 @@
...
+///
+/// # Features
+///
+/// Parsing manifests requires the `generator` feature.
+pub fn parse_depfile(
+    contents: &str,
+    file_name: Option<&OsStr>,
+    format: Option<LockfileFormat>,
+    path: &str,
+    _generation_path: Option<PathBuf>,
```
kylewillmon, line 69: Can't this just be a `bool`?

```suggestion
    _generate_lockfiles: bool,
```

If set to `true`, it can simply use the path from the `file_name` parameter.

For example, the current caller [here][1]:

[1]: https://github.com/phylum-dev/cli/blob/37a219e5218b049a3f1be3db5a242427a6783de1/cli/src/commands/parse.rs#L183-L186

```rs
let generation_path = generate_lockfiles.then(|| path.clone());
let file_name = path.file_name();

phylum_lockfile::parse_depfile(&contents, file_name, format, &id, generation_path)
```

Could simply be this:

```rs
phylum_lockfile::parse_depfile(&contents, Some(path), format, &id, generate_lockfiles)
```

kylewillmon (changes requested):

lockfile/src/parse_depfile.rs
cd-work: No, because `file_name` is a file name. Not a path.

kylewillmon: That's an arbitrary restriction though, right? The code would still function if a full path was passed for that argument.

cd-work: We need two different things here:
 - A file name to determine lockfile/manifest format and maybe_manifest/lockfile without filesystem access
 - A path that is used for lockfile generation

The real path is only needed for lockfile generation, so using it as an option makes sense to me.

This change is mainly due to how an API is interacted with. Say you upload a file, then the full path even if accessible isn't really useful. But the file name can still be passed to the API to help with lockfile identification.

kylewillmon (dismissed):

lockfile/src/parse_depfile.rs
kylewillmon: I agree. But I think `fs_path` and `display_path` are well-named given the situation.

cli/src/commands/parse.rs
```diff
@@ -396,48 +316,120 @@ fn relative_path(from: &Path, to: &Path) -> Result<PathBuf> {
...
+    }
+
+    Ok(command)
+}
+
+/// Create sandbox for dependency file parsing.
+///
+/// This sandbox will automatically add all exceptions necessary to generate
+/// lockfiles for any ecosystem.
+fn depfile_parsing_sandbox(canonical_manifest_path: &Path) -> Result<Birdcage> {
```
kylewillmon, line 353: The parsing sandbox doesn't include access to Phylum's config file which results in a few nasty log messages

```
❗ Config access denied: Operation not permitted (os error 1)
[2023-12-21T16:05:16Z ERROR phylum] Failed to save config: Operation not permitted (os error 1)
```

These aren't usually printed since we capture stderr and they don't cause the process to fail (it uses a default config instead), but I noticed it when I caused a separate error in the subprocess...

This seems a bit complicated to resolve because the config file path can be overridden on the command line. So I suppose we should give access to the effective config file and pass the config file argument to the subprocess...

Or, since the config doesn't actually affect the behavior of `phylum parse-sandboxed`, we could disable config loading for the subprocess (but that would require adding a `--no-config` top-level option).

And finally, because the default config doesn't have a `last_update` timestamp... The subprocess does an update check...

```
       ---------------- Update Available ----------------

       A new version of the Phylum CLI is available. Run

       	phylum update

       to update to the latest version!

       --------------------------------------------------
```

It be nice to avoid that too...

cli/src/commands/parse.rs
```diff
@@ -113,108 +146,49 @@ pub fn parse_depfile(
...
+        let path = path.canonicalize().map_err(anyhow::Error::from)?;
+        let lockfile_type = format.map(|format| format.to_string());
+        let mut command = parse_sandboxed_command(
+            &path,
+            &display_path,
+            lockfile_type.as_ref(),
+            generate_lockfiles,
+            false,
+        )?;
+        let output = command.output().map_err(anyhow::Error::from)?;
```
kylewillmon, line 168: Since stderr is now captured (and discarded on success), we've lost log messages. In particular, I think [this `eprintln!()`][1] was useful and now it is hidden.

[1]: https://github.com/phylum-dev/cli/blob/ce3a5691db1a4b0621d922c77d49834ee57b871f/lockfile/src/parse_depfile.rs#L148

Removing the log message may be good for extensions (which should control their own output). But it's a negative for the CLI itself.

### Before

```
> phylum parse Cargo.toml >/dev/null
Generating lockfile for manifest "Cargo.toml" using Cargo…
```

### After

```
> ./target/debug/phylum parse Cargo.toml >/dev/null
```

lockfile/src/parse_depfile.rs
```diff
@@ -0,0 +1,249 @@
...
+            Ok(packages) => return Ok(ParsedLockfile::new(path, format, packages)),
+            // Store error on failure.
+            Err(err) => lockfile_error = Some(err),
+        }
+    }
+
+    // Generate lockfile if path might be a manifest and feature and option are
+    // enabled.
+    #[cfg(feature = "generator")]
+    if let Some(generation_path) = _generation_path {
```
kylewillmon, line 102: We still need to check `maybe_manifest`... Otherwise, we'll attempt lockfile generation on all sorts of random files.

```
> echo "BAD" > Cargo.lock; ./target/debug/phylum parse Cargo.lock
❗ Error: could not parse dependency file "Cargo.lock"

Caused by:
    0: Dependency file parsing failed
    1: subprocess failed:

[ ... SNIP (config errors, etc. addressed in separate comments) ...]

       Generating lockfile for manifest "Cargo.lock" using Cargo…
       ❗ Error: Lockfile generation failed! For details, see: https://docs.phylum.io/docs/lockfile_generation

       Caused by:
           package manager quit unexpectedly (code: Some(101)):

           error: the manifest-path must be a path to a Cargo.toml file
```

```suggestion
    if let Some(generation_path) = _generation_path.filter(|_| maybe_manifest) {
```

cli/src/commands/parse.rs
```diff
@@ -113,108 +146,49 @@ pub fn parse_depfile(
...
-
-    // Attempt to parse with all known parsers as fallback.
-    let (format, lockfile) = match format {
-        Some(format) => format,
-        None => return Ok(try_get_packages(path, project_root)?),
+    let (format, path) = match find_depfile_format(&path, depfile_type) {
+        Some((format, Some(path))) => (Some(format), path),
+        Some((format, None)) => (Some(format), path),
+        None => (None, path),
     };
```
kylewillmon, line 153: The previous code tracked `path` and `lockfile` separately, but now we combine them. This technically has one minor drawback.

Previously, if the CLI was given a manifest file and it found a matching lockfile, it would switch to parsing the lockfile. But if that parsing failed, it would **fallback to lockfile generation** like so:

```
> echo "BAD" > Cargo.lock; phylum parse Cargo.toml >/dev/null
⚠️  "Cargo.toml" is not a lockfile, using "/Users/kyle/code/cli/Cargo.lock" instead
Generating lockfile for manifest "Cargo.toml" using Cargo…
```

Now the code completely switches to `Cargo.lock` and fails if the file is invalid with no way to fallback to the originally provided file.

I don't know if this fallback was actually useful or not... But it's noteworthy that we have lost it now.

kylewillmon (changes requested):

cli/src/commands/parse.rs
cd-work: I could also remove the error, how likely are people running into this anyway?

cli/src/commands/parse.rs
cd-work: This is how master behaves too. So it's at least not a regression from the status quo.

kylewillmon: This is not at all how `main` behaves... Take this sample Rust project:

```toml
[package]
name = "tmp"
version = "0.1.0"

[dependencies]
log = "0.4.20"
```

Overwrite `Cargo.lock` and then running `phylum parse` on `Cargo.toml` looks like this:

### `main`

```
> echo "BAD" > Cargo.lock; ~/code/cli/target/debug/phylum parse Cargo.toml
⚠️  "Cargo.toml" is not a lockfile, using "/Users/kyle/tmp/Cargo.lock" instead
Generating lockfile for manifest "Cargo.toml" using Cargo…
[
  {
    "name": "log",
    "version": "0.4.20",
    "type": "cargo",
    "lockfile": "Cargo.toml"
  }
]                                                                                                                                                                                                                                                                                                                               

```

### `move_parse`

```
> echo "BAD" > Cargo.lock; ~/code/cli/target/debug/phylum parse Cargo.toml
⚠️  "Cargo.toml" is not a lockfile, using "/Users/kyle/tmp/Cargo.lock" instead
❗ Error: could not parse dependency file "Cargo.toml"

Caused by:
    0: Dependency file parsing failed
    1: subprocess failed:
       ❗ Config access denied: Operation not permitted (os error 1)
       [2023-12-21T19:18:02Z ERROR phylum] Failed to save config: Operation not permitted (os error 1)
       ---------------- Update Available ----------------

       A new version of the Phylum CLI is available. Run

       	phylum update

       to update to the latest version!

       --------------------------------------------------


       Generating lockfile for manifest "Cargo.lock" using Cargo…
       ❗ Error: Lockfile generation failed! For details, see: https://docs.phylum.io/docs/lockfile_generation

       Caused by:
           package manager quit unexpectedly (code: Some(101)):

           error: the manifest-path must be a path to a Cargo.toml file
```

cd-work: Yeah my bad, I tested with the wrong version.

kylewillmon (approved):

PR #1322, opened 2023-12-15
Allow /opt/gradle for Lockfile generation

kylewillmon:
Gradle [install instructions][1] list extracting to `/opt/gradle` as one method of installing the tool. This patch should make those installs function correctly.

[1]: https://gradle.org/install/

### DIFF
--- a/cli/src/commands/generate_lockfile.rs
+++ b/cli/src/commands/generate_lockfile.rs
@@ -128,6 +128,7 @@ fn lockfile_generation_sandbox(canonical_manifest_path: &Path) -> Result<Birdcag
     )?;
     // Gradle.
     permissions::add_exception(&mut birdcage, Exception::WriteAndRead(home.join(".gradle")))?;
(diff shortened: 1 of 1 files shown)


cd-work (approved):

maxrake:
`CHANGELOG.md` entry? Or, add it in when a bugfix release/RC goes out?

cli/src/commands/generate_lockfile.rs
```diff
@@ -128,6 +128,7 @@ fn lockfile_generation_sandbox(canonical_manifest_path: &Path) -> Result<Birdcag
     )?;
     // Gradle.
     permissions::add_exception(&mut birdcage, Exception::WriteAndRead(home.join(".gradle")))?;
+    permissions::add_exception(&mut birdcage, Exception::Read("/opt/gradle".into()))?;
```
maxrake, line 131: Why `Read` instead of `ExecuteAndRead`? Was this tested?

PR #1327, opened 2023-12-21
Switch documentation hosting from Readme.com to Docusaurus

maxrake:
This is part of a larger effort to move <https://docs.phylum.io/> from
Readme.com to Docusaurus, as a statically generated site. The new site
will host all the documentation, regardless of where the docs originate.
The CLI docs from this repository are included in the
`phylum-dev/documentation` repository as a git submodule. The format of
the markdown docs has been changed to account for operating in the new
structure and with Docusaurus as the static site generator.

This PR also removes the doc update job from the release workflow.
The release workflow will be updated in a different PR to include a
documentation update trigger step in the `Release` job. This will look a
lot like the existing trigger steps there. It can't happen until the
`documentation` repo has a `repository_dispatch` workflow that allows
for such triggering. In the mean time, any CLI releases that occur before
this change will need to be manually synced with the `documentation` repo
and deployed from there. Ideally, this will not be needed at all since
the trigger workflow is expected before the next CLI release.

Actions taken include:

* Update doc templates to remove unused Readme.com front matter
* Remove all doc template files that only existed to specify a category
  * The sub-categories are removed in favor of simpler command updates
* Update `gendocs` xtask
  * Remove Readme.com front matter in the default header
  * Write an H1 header instead of frontmatter as the HEADER
* Update links
  * Update absolute links where the the doc is not included in the site
  * Update links to be relative where the doc is included in the site
  * Add a section to the `doc_templates/README.md` about adding links
* Format markdown docs
  * Update `clap_markdown` crate
    * Use H2 headings instead of H3 so as not to "skip" a level
    * Ensure relative links include the `.md` extension
  * Remove frontmatter from all docs and replace it with an H1 header
  * Update `doc_templates`
    * Use H2 headings instead of H3 so as not to "skip" a level
* Regenerate all docs with to include changes made here
  * Use `cargo run -p xtask gendocs`

## Testing

The changes from this PR can be viewed now at <https://docs-stg.phylum.io/>.


.github/workflows/release.yml
maxrake, line 1: Using this comment to keep the PR open until coordination between the `documentation` repo and the corresponding changes there (https://github.com/phylum-dev/documentation/pull/75) are complete.

xtask/src/gendocs.rs
```diff
@@ -15,11 +15,10 @@ const OUTPUT_DIR: &str = "./docs/commands";
...
 
 /// File header inserted at the top of each page.
-const HEADER: &str = "---
-title: {PH-TITLE}
-category: 6255e67693d5200013b1fa3e
-hidden: false
----";
+// NOTE: This is just a H1 header right now, but additional Docusaurus metadata
+//       will likely be added in the future, as frontmatter, and it will be
+//       easier to do so with this structure left in place.
```
cd-work, line 20: Don't think this is necessary.

extensions/duplicates/README.md
```diff
@@ -6,9 +6,9 @@ dependency graph.
 [phylum-cli]: https://github.com/phylum-dev/cli
 
 This extension serves as a simple example and has its source code fully
-explained in our [extension example documentation].
+explained in our [extension example documentation][example].
 
-[extension example documentation]: https://docs.phylum.io/docs/extension_example
+[example]: https://docs.phylum.io/cli/extensions/extension_example
```
cd-work, line 11: Did this not work or why did you change the link ID here? That usually shouldn't cause a problem.

cd-work (approved):

xtask/src/gendocs.rs
maxrake: Roger that. It has been removed.

extensions/duplicates/README.md
maxrake: It worked, but spilled over past the 80 character mark. The change was an attempt to keep all lines within that limit. It has been changed back.

cd-work (approved):

.github/workflows/release.yml
maxrake: Discussions with @kylewillmon and @furi0us333 today resulted in a green light to proceed with this cutover. The changes here are good to be merged now.

PR #1334, opened 2024-01-10
Remove unrelated output from `parse-sandboxed`

### LINKED ISSUE #1328: Avoid unrelated output in `parse-sandboxed`
The `parse-sandboxed` command is designed to be run from within CLI. As a result any output it writes which isn't specific to its operation (like config errors or update notifications) will be duplicated.

The command should be reworked to ensure no output is written that isn't directly related to its purpose.

Currently the following messages are written to stderr:
 - Config access error (we can maybe pass `--config /dev/null` or add `--no-config` option)
 - New version available (we already have selfmanage compile-time feature, it should also be a runtime option)

cd-work:
This patch removes all output from `parse-sandboxed` that isn't directly related to its purpose of parsing dependency files. This includes omitting both config and update messages.

Closes #1328.

### DIFF
--- a/cli/src/app.rs
+++ b/cli/src/app.rs
@@ -39,6 +39,11 @@ pub fn app() -> Command {
                 .value_name("FILE")
                 .help("Sets a custom config file")
                 .value_hint(ValueHint::FilePath),
+            Arg::new("no-config")
+                .long("no-config")
+                .help("Ignore all configuration files")
+                .conflicts_with("config")
+                .action(ArgAction::SetTrue),
             Arg::new("timeout")
                 .short('t')
                 .long("timeout")
--- a/cli/src/bin/phylum.rs
+++ b/cli/src/bin/phylum.rs
@@ -1,8 +1,7 @@
 use std::fmt::Display;
-use std::path::{Path, PathBuf};
 use std::time::UNIX_EPOCH;
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
 #[cfg(feature = "selfmanage")]
 use clap::ArgMatches;
 use env_logger::Env;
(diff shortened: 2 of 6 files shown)


cli/src/app.rs
```diff
@@ -39,6 +39,11 @@ pub fn app() -> Command {
                 .value_name("FILE")
                 .help("Sets a custom config file")
                 .value_hint(ValueHint::FilePath),
+            Arg::new("no-config")
```
cd-work, line 42: I decided to go for a separate option here since `--config /dev/null` would not work on Windows and I see no reason to make something so simple platform-dependent.

cli/src/bin/phylum.rs
```diff
@@ -104,41 +99,33 @@ async fn handle_commands() -> CommandResult {
...
     let timeout = matches.get_one::<String>("timeout").and_then(|t| t.parse::<u64>().ok());
 
-    // Check for updates if enabled and if we haven't explicitly invoked `update`.
-    if cfg!(feature = "selfmanage") && matches.subcommand_matches("update").is_none() {
-        check_for_updates(&mut config, &config_path).await?;
+    let (subcommand, sub_matches) = matches.subcommand().unwrap();
+
+    // Check for updates if enabled and if we haven't explicitly invoked `update` or
+    // `parse-sandboxed`.
+    if cfg!(feature = "selfmanage") && subcommand != "update" && subcommand != "parse-sandboxed" {
```
cd-work, line 116: Decided *against* a separate flag for update checks since we're already matching for subcommands to avoid printing messages for the update command itself. If there is a real reason for a dedicated flag, we can always change it in the future.

cli/src/config.rs
```diff
@@ -86,6 +100,30 @@ where
...
+    let config_path = matches
+        .get_one::<String>("config")
+        .and_then(|config_path| shellexpand::env(config_path).ok())
+        .map(|config_path| PathBuf::from(config_path.to_string()))
+        .unwrap_or(settings_path);
+
+    log::debug!("Reading config from {}", config_path.to_string_lossy());
+    let mut config: Config = read_configuration(&config_path).map_err(|err| {
+        anyhow!("Failed to read configuration at `{}`: {}", config_path.to_string_lossy(), err)
+    })?;
```
kylewillmon, line 120: I know this code was only moved. But I can't help but notice the `.map_err()` here should probably be `.with_context()` instead.

```suggestion
    let mut config: Config = read_configuration(&config_path).with_context(|| {
        format!("Failed to read configuration at `{}`", config_path.to_string_lossy())
    })?;
```

kylewillmon (changes requested):

cli/src/config.rs
```diff
@@ -77,6 +81,16 @@ impl Config {
...
     }
+
+    /// Write updates to the configuration file.
+    pub fn save(&self) -> Result<()> {
+        let path = match &self.path {
+            Some(path) => path,
+            None => return Ok(()),
+        };
+
+        save_config(path, self)
```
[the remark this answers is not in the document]
cd-work: Yeah that's quite interesting actually: We use this to write `.phylum_project`. It makes sense since it basically is just an atomic serialization utility, but maybe the name isn't ideal.

cli/src/bin/phylum.rs
```diff
@@ -104,41 +99,33 @@ async fn handle_commands() -> CommandResult {
...
     let timeout = matches.get_one::<String>("timeout").and_then(|t| t.parse::<u64>().ok());
 
-    // Check for updates if enabled and if we haven't explicitly invoked `update`.
-    if cfg!(feature = "selfmanage") && matches.subcommand_matches("update").is_none() {
-        check_for_updates(&mut config, &config_path).await?;
+    let (subcommand, sub_matches) = matches.subcommand().unwrap();
+
+    // Check for updates if enabled and if we haven't explicitly invoked `update` or
+    // `parse-sandboxed`.
+    if cfg!(feature = "selfmanage") && config.path.is_some() && subcommand != "update" {
```
kylewillmon, line 116: The comment is incorrect now. I should have mentioned that in my suggestion, sorry...

kylewillmon (changes requested):

cli/src/app.rs
maxrake, line 42: This addition requires regenerating the docs with `cargo run -p xtask gendocs`

maxrake (changes requested):

kylewillmon (approved):

PR #1335, opened 2024-01-10
Forward parse-sandboxed STDERR on success

### LINKED ISSUE #1329: Logging from `parse-sandboxed` should be forwarded
Currently anything written to STDERR from `parse-sandboxed` is swallowed unless the command fails. Once unrelated messages are removed from STDERR (https://github.com/phylum-dev/cli/issues/1328) we should allow both user-facing and hidden log messages to always be forwarded (`Stdio::inherit()`?) to the parent process.

cd-work:
Previously the STDERR of the sandboxed `parse-sandboxed` subcommand was only forwarded if the command failed. Now the STDERR is always inherited regardless of exit status.

To ensure optimal troubleshooting in case of a failure, the STDOUT is now also printed to STDOUT in case of a failure.

Closes #1329.

### DIFF
--- a/cli/src/commands/parse.rs
+++ b/cli/src/commands/parse.rs
@@ -2,7 +2,7 @@
 
 use std::borrow::Cow;
 use std::path::{Path, PathBuf};
-use std::process::Command;
+use std::process::{Command, Stdio};
 use std::result::Result as StdResult;
 use std::str::FromStr;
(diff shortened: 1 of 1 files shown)


kylewillmon (approved):

maxrake (approved):
This appears to work as intended. LGTM.

PR #1337, opened 2024-01-11
Parse manifest files with non-standard names

maxrake:
This change allows for parsing manifest files with names that don't
match the expected standards. For example, `pip` manifest files can be
named *anything* and they are therefore hard to identify. Pip lockfiles
can also be named anything but they will at least get picked up by the
fallback approach of attempting to parse with all known parsers
(assuming a "true" lockfile with no loose requirement entries).

A new custom error (`UnknownManifestFormat`) and exit code (21) were
created to be used when a dependency file is provided without a file
type and does not parse as a known lockfile, making it a likely manifest
file.

A change was made to report errors with the `pip` version check that
happens in the `pip` lockfile generator.

Additionally, a change was made to ensure `ManifestWithoutGeneration`
errors use the display path instead of the full path.

Prerequisite checks were also added for lockfile generation to account
for known manifest file names. Previously, some lockfile generators
completely ignored the filename provided and just called the relevant
tool in the directory, leading to some unexpected successes.

## Testing

The behavior before this change:

```
❯ phylum --version
phylum v6.0.1

❯ cat dev-requirements.txt
pyyaml

❯ phylum parse dev-requirements.txt
❗ Error: could not parse dependency file "dev-requirements.txt"

Caused by:
    Failed to identify type for lockfile "dev-requirements.txt"

❌ 1 ❯ phylum parse -t pip dev-requirements.txt
❗ Error: could not parse dependency file "dev-requirements.txt"

Caused by:
    0: Failed to parse lockfile
    1: Failed to parse requirements file
    2: 0: at line 1, in Tag:
       pyyaml
             ^
```

The behavior after this change:

```
phylum v6.0.1-27-gacb4869

❯ cat dev-requirements.txt
pyyaml

❯ target/debug/phylum parse dev-requirements.txt

❗ Could not parse manifest: Parsing "dev-requirements.txt" requires a type to be specified

❌ 21 ❯ target/debug/phylum parse -t pip dev-requirements.txt
Generating lockfile for manifest "dev-requirements.txt" using Pip…
[
  {
    "name": "PyYAML",
    "version": "6.0.1",
    "type": "pypi",
    "lockfile": "dev-requirements.txt"
  }
]

❯ target/debug/phylum parse -t msbuild dev-requirements.txt
❗ Error: Failed to parse lockfile

Caused by:
    0: Syntax: 1:1 Unexpected characters outside the root element: p
    1: 1:1 Unexpected characters outside the root element: p

❗ Error: Could not parse dependency file "dev-requirements.txt" as "msbuild" type

Caused by:
    Dependency file parsing failed

❌ 1 ❯ target/debug/phylum parse -t npm dev-requirements.txt
Generating lockfile for manifest "dev-requirements.txt" using Npm…
❗ Error: Failed to parse lockfile

Caused by:
    expected value at line 1 column 1

❗ Error: Could not parse dependency file "dev-requirements.txt" as "npm" type

Caused by:
    Dependency file parsing failed
```


lockfile/src/parse_depfile.rs
```diff
@@ -94,10 +99,9 @@ pub fn parse_depfile(
...
     }
 
-    // Generate lockfile if path might be a manifest and feature and option are
-    // enabled.
+    // Generate lockfile for likely manifests when feature and option are enabled.
     #[cfg(feature = "generator")]
-    if let Some(generation_path) = _generation_path.filter(|_| maybe_manifest) {
+    if let Some(generation_path) = _generation_path.filter(|_| manifest_likely) {
         return Ok(generate_lockfile(&generation_path, &path, format, parser)?);
     }
```
kylewillmon, line 106: This can now call `generate_lockfile()` with a `format` that doesn't support lockfile generation... For example:

```
> ./target/debug/phylum parse -t msbuild Cargo.toml
❗ Error: unsupported manifest file "Cargo.toml"

❗ Error: could not parse dependency file "Cargo.toml"

Caused by:
    Dependency file parsing failed
```

Of course, that's a pretty unusual case... But I think it would be slightly better to return the error that we've stored from attempting to parse it as a lockfile (i.e., `lockfile_error`).

lockfile/src/parse_depfile.rs
```diff
@@ -80,10 +83,12 @@ pub fn parse_depfile(
     let pseudopath = Path::new(&path);
     let maybe_lockfile = parser.is_path_lockfile(pseudopath);
     let maybe_manifest = parser.is_path_manifest(pseudopath);
+    let lockfile_likely = maybe_lockfile || !maybe_manifest;
+    let manifest_likely = !maybe_lockfile || maybe_manifest;
```
kylewillmon, line 87: Some of our lockfile generators completely ignore the filename provided and just call the relevant tool in the directory. This can lead to weird successes... For example, in the CLI repository:

```
> ./target/debug/phylum parse -t npm Cargo.toml
Generating lockfile for manifest "Cargo.toml" using Npm…
[]                                                                                                                                                                                                                                                                                                                              
```

kylewillmon (changes requested):
Definitely a nice improvement. But there are a few edge cases worth discussing (even if we decide not to fix them).

lockfile/src/parse_depfile.rs
cd-work: I preferred when these variables were not extracted, since this does not properly capture what they represent. `maybe_lockfile` signifies that it's likely a lockfile. The `!maybe_manifest` does not make it more likely that it is a lockfile.

These variables are implying a certainty where there is none. We in fact know **less** about a file that is neither lockfile nor manifest, we're just trying everything as a hail mary.

maxrake: The variables will be put back so `lockfile_likely` and `manifest_likely` are no longer used.

@kylewillmon, are you suggesting a change is needed here? Like, maybe the lockfile generators need to be updated to _not_ ignore the filename provided?

cd-work: The cargo generator should never be run when the type is explicitly specified as npm. It should first try parsing as npm lockfile and then try npm generation.

kylewillmon: > It should first try parsing as npm lockfile and then try npm generation.

That is exactly what happens. The unexpected behavior here is that npm generation **succeeds**

cd-work: Oh I didn't even realize that.

lockfile/src/parse_depfile.rs
maxrake: I _think_ I was able to make the requested change. I'm not confident it was done in the most Rusty way so please feel free to provide better implementations.

lockfile/src/parse_depfile.rs
maxrake: A more detailed comment was added here but no code related to the behavior was changed. If code changes are desired, I will need help to do it.

lockfile/src/parse_depfile.rs
```diff
@@ -83,11 +86,16 @@ pub fn parse_depfile(
...
+    // point to not be a valid/parseable lockfile but may parse as a manifest with
+    // a non-standard name.
     #[cfg(feature = "generator")]
-    if let Some(generation_path) = _generation_path.filter(|_| maybe_manifest) {
-        return Ok(generate_lockfile(&generation_path, &path, format, parser)?);
+    if let Some(generation_path) = _generation_path.filter(|_| !maybe_lockfile || maybe_manifest) {
+        let result = generate_lockfile(&generation_path, &path, format, parser);
+        if result.is_ok() {
+            return Ok(result?);
+        }
```
kylewillmon, line 98: 1. This would be better written as a `match` (since it avoids double-checking `result` with both `.is_ok()` and `?`).
2. Sometimes we want to return the error (specifically when `maybe_manifest` is true)
3. As mentioned [here](https://github.com/phylum-dev/cli/pull/1337#discussion_r1450951131), some parsers don't support generation, so we might as well skip it.

This patch fixes all 3 issues.

```suggestion
        if parser.generator().is_some() {
            match generate_lockfile(&generation_path, &path, format, parser) {
                Ok(depfile) => return Ok(depfile),
                // Discard errors for unknown files. The error from the lockfile parser can be used
                // instead.
                Err(_) if !maybe_manifest => {},
                Err(err) => return Err(err.into()),
            }
        }
```

lockfile/src/parse_depfile.rs
kylewillmon: > Like, maybe the lockfile generators need to be updated to not ignore the filename provided?

I do think this code change is warranted. Updating the generators would fix this issue.

kylewillmon (changes requested):

lockfile/src/parse_depfile.rs
maxrake: This suggestion was accepted.

lockfile/src/parse_depfile.rs
maxrake: The lockfile generators have been updated.

lockfile_generator/src/gradle.rs
```diff
@@ -24,4 +25,14 @@ impl Generator for Gradle {
     fn tool(&self) -> &'static str {
         "Gradle"
     }
+
+    fn check_prerequisites(&self, manifest_path: &Path) -> Result<()> {
+        if manifest_path.file_name() != Some(OsStr::new("build.gradle"))
+            && manifest_path.file_name() != Some(OsStr::new("build.gradle.kts"))
```
kylewillmon, line 31: It took my brain a minute to confirm that this condition is correctly written with `&&` instead of `||`. Negatives are hard...

This might be easier to read if the condition was reversed (i.e., `if <correct file name> { Ok(()) } else { Err(...) }`)

Or it could be written with a `match` statement like so:

```patch
diff --git a/lockfile_generator/src/gradle.rs b/lockfile_generator/src/gradle.rs
index ea3f7d7..3b0b7f7 100644
--- a/lockfile_generator/src/gradle.rs
+++ b/lockfile_generator/src/gradle.rs
@@ -1,6 +1,5 @@
 //! Java gradle ecosystem.

-use std::ffi::OsStr;
 use std::process::Command;

@@ -27,12 +26,9 @@ impl Generator for Gradle {
     }

     fn check_prerequisites(&self, manifest_path: &Path) -> Result<()> {
-        if manifest_path.file_name() != Some(OsStr::new("build.gradle"))
-            && manifest_path.file_name() != Some(OsStr::new("build.gradle.kts"))
-        {
-            Err(Error::InvalidManifest(manifest_path.to_path_buf()))
-        } else {
-            Ok(())
+        match manifest_path.file_name().and_then(|f| f.to_str()) {
+            Some("build.gradle" | "build.gradle.kts") => Ok(()),
+            _ => Err(Error::InvalidManifest(manifest_path.to_path_buf())),
         }
     }
 }
```

But it's not terrible as is, so I won't hold up this review.

kylewillmon (approved):

lockfile_generator/src/gradle.rs
maxrake: Funny...I thought the same thing and had it the inverted way initially but then thought maybe _that_ would be a problem later if someone wants to add new prerequisites and the pattern of "failure cases first" was not in place. It has been changed.

kylewillmon (approved):

PR #1351, opened 2024-01-31
Fix sandbox exceptions for local pip installs

### LINKED ISSUE #1349: Allow more valid `pip` sources
# Overview

It has been observed that users with their primary/active `python3`/`pip` binary located in `~/.local` are not able to perform lockfile generation on pip manifest files. This is a valid location for `pip` to exist and should be supported. Updates to sandbox exceptions are likely needed.

## Additional Guidance

A workaround exists to bypass the sandbox with the `--skip-sandbox` option. However, this is not recommended unless the fully resolved set of dependencies are already known and trusted since arbitrary code execution is possible without the sandbox.

## Acceptance Criteria

- [ ] Lockfile generation for `pip` manifests succeeds when the primary `pip` binary is located in `~/.local` and no other instances of `pip` exist
- [ ] The `pip` extension sandbox permissions are updated to match
- [ ] [Documentation](https://docs.phylum.io/cli/lockfile_generation) is updated so that users experiencing this error will get more background and/or actionable next steps when visiting the suggested link

cd-work:
This fixes lockfile generation with pip when it is installed in `~/.local/lib/python*/site-packages/pip`.

It also fixes an issue where `pyenv` installed through the package manager wouldn't allow for lockfile generation.

Closes #1349.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Fixed
+
+- Python lockfile generation with pip in ~/.local
+- Python lockfile generation with pyenv
+
 ## 6.1.0 - 2024-01-29
 
 ### Added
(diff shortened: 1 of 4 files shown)


maxrake (dismissed):
Verified with local testing. LGTM.

louislang (approved):
👍

maxrake (approved):
LGTM

PR #1356, opened 2024-02-07
Bump birdcage to 0.7.1

### LINKED ISSUE #1350: Provide more information on system minimum requirements
# Overview

User feedback has shown that there are times when system minimum requirements are not met and actions like lockfile generation failed, but with less than helpful output such that the source of the problem and possible corrective actions were not provided. Here is one example of such output:

```
---TRIMMED---
Caused by:
    0: Lockfile generation failed! For details, see: https://docs.phylum.io/docs/lockfile_generation
    1: subprocess failed:
       ❗ Error: input/output error: Function not implemented (os error 38)
```

In this case, the `Error` message was confusing and the provided link did not yield any more detail to resolve the error. It turned out to be the case that the system kernel version was too old to support one of the required/requested sandbox permissions, but that was only uncovered by developers of the tool after reaching out for their support. The user had no clue what was wrong or what they could do themselves to figure that out.

This issue is meant to help users in as many ways as possible. Actions which may help include, but are not limited to:

* Put clear prerequisite minimums (kernel, glibc, etc) in the appropriate doc locations (CLI? birdcage?)
* Update installers to check for the prerequisites and fail with helpful messages when they aren't met
* Update error messages to provide more hints as to the source of the failure and/or links to resources that can provide more detail

## Acceptance Criteria

- [ ] Documentation is updated to minimally specify the sources of system pre-requisites and why/when certain actions may not work when those aren't met
- [x] Error output is improved to provide more detail and/or actionable next steps

cd-work:
Closes #1350.

---

I wouldn't make **any** documentation changes for this, however @maxrake has expressed interest in adding extra documentation. Could you go into additional details on what exactly you'd want to see documented?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -387,9 +387,9 @@ dependencies = [
 
 [[package]]
 name = "birdcage"
-version = "0.7.0"
+version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4270b12541e12bb8d8a323c015a1c317625b9ba511a027f05c4047456943f835"
+checksum = "3ed46d13ad2637d8609d5614963ebccde3dcb554e995a86bd4ff9d800b4bf14d"
 dependencies = [
  "bitflags 2.4.2",
  "libc",


kylewillmon (approved):

maxrake:
The idea behind the requested documentation updates is to [provide users with context and actionable next steps to take when they experience errors](https://clig.dev/#errors). Improving error messages is part of the solution (as was done in the Birdcage update here). The error messages only go so far, given their brevity, and the common text in a lot of these messages is to get more details at the https://docs.phylum.io/cli/lockfile_generation site.

However, that page does not have a troubleshooting section or information on common errors, leaving users scratching their head as to what went wrong and what they should do next. To improve the user experience for these error cases, it is expected that at least the following would exist on the `docs/lockfile_generation.md` page: 

* Minimum system requirements
  * Specify the sources of system pre-requisites (e.g., kernel, glibc, etc.)
  * Explain broadly which actions may not work when these aren't met
  * Explain why exact versions may not always be possible (like was done [in this comment](https://github.com/phylum-dev/birdcage/pull/69#discussion_r1481961961))
    * This could help to explain why an install-time check is not possible and that encountering these errors is expected/common and not the user's fault
* When lockfile generation is expected to fail
  * Primary example is when used in a Docker container
    * This causes a panic and the message `Sandbox::spawn must be called from a single-threaded process`
  * Options for what to do instead (e.g., `--skip-sandbox`?)
* Lockfile generation sandbox exceptions/permissions
  * What they are and when new ones might be needed
  * What an error related to a missing exception looks like
  * How to determine which exceptions/permissions are missing
  * How to reach out for support in adding new exceptions
  * This section would resolve the uncompleted acceptance criteria for #1349
* How to reach out for "level 2" support
  * Could be email, Discord, submitting a bug ticket, or anything else
  * Be explicit about the options and don't assume anything about the user's background
  * Make it clear which information is needed for the investigation to be successful
  * Could be reassuring to user's to let them know that these requests are expected and necessary to improve the product
    * We don't want any users to give up in frustration because they think the product is broken and don't know how to help make it better

This list is not exhaustive and will likely change over time as more users discover more errors.

cd-work:
> Minimum system requirements 

We don't know these and they might change at any time without our knowledge.

> Explain why exact versions may not always be possible (like was done https://github.com/phylum-dev/birdcage/pull/69#discussion_r1481961961)

What does the user gain from us going on about why we can't provide information? That doesn't help them with troubleshooting at all.

> Primary example is when used in a Docker container
>> This causes a panic and the message Sandbox::spawn must be called from a single-threaded process

Docker containers aren't all created equal. It is possible to run birdcage sandboxing inside Docker as far as I'm aware.

Did you test it and run into this error message? Because that is **not** the error message I would expect when running inside of Docker.

> Lockfile generation sandbox exceptions/permissions 

All of them? Because there's **a lot**. Most of them nobody needs to care about.

> How to reach out for "level 2" support 

Shouldn't this be covered by our general documentation? It makes no difference what kind of issue you run into when it comes to this. We don't need to repeat the same thing over and over for this.

maxrake:
FYI...the discussion was continued out of band. Some of the changes are represented in #1357 and the conversation will continue there.

PR #1357, opened 2024-02-09
Add docs for lockfile generation sandbox

cd-work:
This documents the behavior of the sandbox in connection to automatic lockfile generation when running `analyze`/`parse` on manifest files.

See #1350.

### DIFF
--- a/docs/lockfile_generation.md
+++ b/docs/lockfile_generation.md
@@ -1,11 +1,14 @@
 # Lockfile Generation
 
-Lockfiles can be directly parsed and analyzed by Phylum's CLI. However since
+Lockfiles can be directly parsed and analyzed by Phylum's CLI. However, since
 manifest files only specify loose version requirements, it is necessary for the
-CLI to internaly generate the corresponding lockfile.
+CLI to internally generate the corresponding lockfile.
 
-No lockfile generation will take place if the `--no-generation` flag is passed
-to `parse` or `analyze`.
+No lockfile generation will take place if the `--no-generation` CLI flag is
+passed to [`phylum parse`] or [`phylum analyze`].
+
+[`phylum parse`]: ../cli/commands/phylum_parse.md
+[`phylum analyze`]: ../cli/commands/phylum_analyze.md
 
 ## Lockfile generators
 
--- a/docs/lockfile_generation.md
+++ b/docs/lockfile_generation.md
@@ -1,8 +1,16 @@
 # Lockfile Generation
 
-The Phylum CLI can generate a lockfile when it is given a manifest file.
+Lockfiles can be directly parsed and analyzed by Phylum's CLI. However since
+manifest files only specify loose version requirements, it is necessary for the
+CLI to internaly generate the corresponding lockfile.
+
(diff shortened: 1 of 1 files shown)


docs/lockfile_generation.md
```diff
@@ -1,9 +1,17 @@
 # Lockfile Generation
 
-The Phylum CLI can generate a lockfile when it is given a manifest file.
+Lockfiles can be directly parsed and analyzed by Phylum's CLI. However since
+manifest files only specify loose version requirements, it is necessary for the
+CLI to internaly generate the corresponding lockfile.
```
cd-work, line 5: Expanded this based on what @kylewillmon said about a little introduction for this page. Still brief, but giving a bit more detail on when generation actually takes place.

docs/lockfile_generation.md
```diff
@@ -1,9 +1,17 @@
...
+
+No lockfile generation will take place if the `--no-generation` flag is passed
+to `parse` or `analyze`.
```
cd-work, line 8: Previously this was under the "lockfile detection" heading, but it doesn't really have anything to do with lockfile detection.

docs/lockfile_generation.md
```diff
@@ -72,6 +78,23 @@ a lockfile (e.g., `pip freeze > requirements.txt`).
 Phylum handles these files by first attempting to analyze them as a lockfile. If anything in the file is not fully
 specified, this will fail, and Phylum will silence the error and proceed to lockfile generation.
 
+## Sandboxing
```
maxrake, line 81: This new section is a welcome addition. It could be made better still by including information about what it looks like when the sandbox fails due to a missing exception/permission (the list of "what ecosystem tools commonly need") and what to do about it.

A quick review of the existing documentation did not turn up any specific directions about the order of preference for seeking support and if/where/how bug tickets are filed. The closest page appears to be the [Contact Us](https://docs.phylum.io/support/contact_us) one...but that is meant to cover the entire Phylum solution and is very broad (does not mention bug tickets for any repository, let alone the CLI).

maxrake (changes requested):

docs/lockfile_generation.md
cd-work: > It could be made better still by including information about what it looks like when the sandbox fails due to a missing exception/permission

This could be literally anything. We're executing arbitrary code after all.

> A quick review of the existing documentation did not turn up any specific directions about the order of preference for seeking support and if/where/how bug tickets are filed.

If we want to add a dedicated paragraph somewhere about contacting us for support I think opening a separate issue makes sense. I don't think it makes sense to go into detail about how to do that right here. Once that is done we could sprinkle short "contact us" sentences with links to that section throughout other pages.

maxrake: Sure...a dedicated contact section/page can be done elsewhere.

I still think that users experiencing a sandbox failure due to a missing permission are going to be left wondering what to do next...but maybe that will be alleviated with the (soon to be added) contact preferences text. This can always be updated if/when users express confusion about the errors and what to do about them.

maxrake (approved):

PR #1368, opened 2024-02-28
Fix panic in SPDX parser

### LINKED ISSUE #1364: SPDX panics while parsing
Input:

```
:
PackageName:
```

Tested with `cargo run --bin phylum -- parse -t spdx <file>`.

cd-work:
Closes #1364.


cd-work:
Converted to draft for now because I fixed one more crash and found at least another one.

cd-work:
Ran libfuzz for a while and only found the SPDX issue.

Ran afl (which usually performs better) and almost immediately found both the SPDX and the C# issue. Ran it for a couple hours (over 10 million execs) and found no more crashes.

maxrake (changes requested):
Requesting CHANGELOG entry for bug fix(es).

Code was reviewed and functionality was confirmed with local testing. The panic is no longer present, but it appears the test input file from #1364 is still not valid SPDX (which is expected?):

<details><summary>Details (click to expand)</summary>
<p>

```
❯ phylum --version
phylum v6.1.2

❯ cat test_file.spdx
:
PackageName:

❯ phylum parse -t spdx test_file.spdx
thread 'main' panicked at /Users/phylum/.cargo/registry/src/index.crates.io-6f17d22bba15001f/nom-7.1.3/src/internal.rs:41:9:
Cannot call `finish()` on `Err(Err::Incomplete(_))`: this result means that the parser does not have enough data to decide, you should gather more data and try to reapply  the parser instead

❗ Error: Could not parse dependency file "test_file.spdx" as "spdx" type

Caused by:
    Dependency file parsing failed

❌ 2 ❯ target/debug/phylum --version
phylum v6.1.2-3-g8efd1d7

❯ target/debug/phylum parse -t spdx test_file.spdx
❗ Error: Failed to parse lockfile

Caused by:
    0: at line 3, in CrLf:

    ^

    1: at line 2, in package:
    PackageName:
    ^

    2: at line 1, in Many1:
    :
    ^



❗ Error: Could not parse dependency file "test_file.spdx" as "spdx" type

Caused by:
    Dependency file parsing failed
```

</p>
</details>

cd-work:
> but it appears the test input file from https://github.com/phylum-dev/cli/issues/1364 is still not valid SPDX (which is expected?):

Yes.

lockfile/src/csharp.rs
```diff
@@ -28,14 +29,15 @@ impl Parse for PackagesLock {
...
+                    _ => return Err(anyhow!("invalid dependency {name:?}: {dependency:?}")),
                 };
 
-                Package { version, name, package_type: PackageType::Nuget }
+                Ok(Package { version, name, package_type: PackageType::Nuget })
             })
-            .collect();
+            .collect::<Result<_, _>>()?;
 
         Ok(packages)
```
kylewillmon, line 42: The `?` is unnecessary since you can just return the `Result`

maxrake (approved):
LGTM

maxrake (approved):

PR #1374, opened 2024-03-12
Remove `phylum group transfer` subcommand

mathew-horner:
This endpoint will be going away in the near future because we are getting rid of owners, thus the CLI command should also be removed.

Is it okay for older versions to just 404 on group ownership transfer? It's a feature that is hardly used (5x in the last year, and I assume most, if not all, of those were likely done in the UI).

ref https://github.com/phylum-dev/api/issues/1254


cd-work (approved):
> Is it okay for older versions to just 404 on group ownership transfer? It's a feature that is hardly used (5x in the last year, and I assume most, if not all, of those were likely done in the UI).

No way around this considering we'll remove it from API sooner or later. It will inevitably happen at *some* point unless we keep an old version of API hosted until the end of time, which is not feasible.

PR #1375, opened 2024-03-12
Remove deserialization of `owner_email` and `is_owner fields` in group listing structures

mathew-horner:
These fields are going away soon: https://github.com/phylum-dev/api/issues/1266

We probably will need some sort of transitionary period here, so we don't break `phylum group list` on older versions. I was thinking that we could return `is_owner` always set to false and maybe something like `"<deprecated>"` for the owner email, and that will get displayed in the groups table in older versions of the CLI. What do you guys think?


extensions/phylum.ts
```diff
@@ -310,10 +310,8 @@ export class PhylumApi {
    *     {
    *       created_at: "2022-05-02T14:25:24.184866508Z",
    *       last_modified: "2022-05-02T14:25:24.599950299Z",
-   *       owner_email: "null@phylum.io",
    *       group_name: "phlock",
    *       is_admin: false,
-   *       is_owner: true
    *     }
    *   ]
    * }
```
mathew-horner, line 317: What is the impact of breaking the extension API here? Acceptable or no?

cd-work: Unlikely to cause any issues I think.

cd-work (dismissed):
This should be fine, but my main concern here is that we're removing something without integrating the replacement for it.

Have you put in any thought on if or how the admin user role should be represented in the CLI?

mathew-horner:
> This should be fine, but my main concern here is that we're removing something without integrating the replacement for it.
> 
> Have you put in any thought on if or how the admin user role should be represented in the CLI?

The structure still includes `is_admin`. I think we probably want to replace that with a `role` field eventually, like we did for the group member list endpoint, for the same forward compatability reasons.

cd-work:
> The structure still includes is_admin. I think we probably want to replace that with a role field eventually, like we did for the group member list endpoint, for the same forward compatability reasons.

Do you have any plan to show this information in the `group list` table?

mathew-horner:
> > The structure still includes is_admin. I think we probably want to replace that with a role field eventually, like we did for the group member list endpoint, for the same forward compatability reasons.
> 
> Do you have any plan to show this information in the `group list` table?

I don't, but we could probably add it fairly easily. Maybe after we add a `role` to this response we can display their role in each group in the table?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Removed
 
 - Removed `phylum group transfer` subcommand
+- Removed owner email from `phylum group list` results
```
cd-work, line 18:
```suggestion
- Owner email from `phylum group list` results
```

Forgot to mention that for the previous changelog entry, but generally we don't add `Removed` in front of every removed entry, since it's redundant.

Feel free to also change the other entry.

cd-work (changes requested):

cd-work (approved):

Issue #1384, opened 2024-04-04
Use `SPDX` download location to determine ecosystem


ejortega:
# Overview
The download location can be parsed and matched with known package registries.
This is only done when package locators or when a PURL is present, but can also be used when if we have enough information from the package fields.

# How To Reproduce
Steps to reproduce this behavior:
1. Attempt to parse a SPDX sbom missing an external ref
2. See error for missing external ref or package location

# Expected Behavior
Should be able to handle sbom packages with just name, version and download location.

PR #1385, opened 2024-04-04
Use `SPDX` download location to determine package registry

ejortega:
This uses the download location to match with known package registries for packages without an external ref.

Closes https://github.com/phylum-dev/cli/issues/1384

## Checklist

### DIFF
--- a/lockfile/src/spdx.rs
+++ b/lockfile/src/spdx.rs
@@ -83,6 +83,7 @@ fn type_from_url(url: &str) -> anyhow::Result<PackageType> {
         Ok(PackageType::PyPi)
     } else if url.starts_with("https://repo1.maven.org")
         || url.starts_with("https://search.maven.org")
+        || url.starts_with("central.sonatype.com")
     {
         Ok(PackageType::Maven)
     } else if url.starts_with("https://api.nuget.org") {
@@ -145,17 +146,17 @@ impl TryFrom<&PackageInformation> for Package {
 
     fn try_from(pkg_info: &PackageInformation) -> anyhow::Result<Self> {
         // Attempt to determine package data from external refs.
-        let pkg_result =
-            pkg_info.external_refs.iter().find_map(|external| match external.reference_category {
-                ReferenceCategory::PackageManager => {
-                    if external.reference_type == "purl" {
-                        Some(from_purl(&external.reference_locator, pkg_info))
-                    } else {
-                        Some(from_locator(&external.reference_type, &external.reference_locator))
-                    }
+        let pkg_result = pkg_info.external_refs.iter().find_map(|external| {
+            match (&external.reference_category, external.reference_type.as_str()) {
+                (ReferenceCategory::PackageManager, "purl") => {
+                    Some(from_purl(&external.reference_locator, pkg_info))
+                },
+                (ReferenceCategory::PackageManager, _) => {
+                    Some(from_locator(&external.reference_type, &external.reference_locator))
                 },
                 _ => None,
-            });
+            }
+        });
 
         pkg_result.unwrap_or_else(|| {
             let version = pkg_info
--- a/lockfile/src/spdx.rs
+++ b/lockfile/src/spdx.rs
@@ -28,6 +28,7 @@ pub(crate) struct PackageInformation {
     pub(crate) name: String,
     pub(crate) version_info: Option<String>,
     pub(crate) download_location: String,
+    #[serde(default)]
     pub(crate) external_refs: Vec<ExternalRefs>,
 }
 
@@ -124,7 +125,8 @@ impl TryFrom<&PackageInformation> for Package {
     type Error = anyhow::Error;
 
     fn try_from(pkg_info: &PackageInformation) -> anyhow::Result<Self> {
-        pkg_info
+        // Attempt to determine package data from external refs.
+        let pkg_result = pkg_info
             .external_refs
             .iter()
             .find_map(|external| match external.reference_category {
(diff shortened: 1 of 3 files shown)


lockfile/src/spdx.rs
```diff
@@ -272,42 +291,6 @@ mod tests {
         }
     }
 
-    #[test]
```
ejortega, line 275: Test is removed as this is now an `UnknownEcosystem`.

lockfile/src/spdx.rs
```diff
@@ -137,7 +139,24 @@ impl TryFrom<&PackageInformation> for Package {
...
+                        name: pkg_info.name.to_owned(),
+                        version: PackageVersion::FirstParty(version.to_owned()),
+                        package_type,
+                    }),
+                    (Some(_), Err(_)) => Err(UnknownEcosystem.into()),
+                    (None, Ok(_)) => Err(anyhow!("Missing package version {}", pkg_info.name)),
+                    _ => Err(e),
+                }
+            },
+        }
```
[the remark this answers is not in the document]
cd-work: >  I think it reads pretty cleanly here

It never does. The `ok_or_else`/`map_err` solution is far better.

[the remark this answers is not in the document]
mathew-horner: less lines = better, everyone knows this 😉

lockfile/src/spdx.rs
ejortega: https://github.com/phylum-dev/cli/blob/2aee6488a96046716886c39d02fe7d03e82c3a09/lockfile/src/spdx.rs#L161

We just let those pass. I could make a test that has has a list of know and unknown and show they aren't captured?

mathew-horner: Yeah if you have another test case that is supposed to test for the succesful case we could probably add a package that covers this behavior to the fixture. I just think we want that behavior covered somewhere.

lockfile/src/spdx.rs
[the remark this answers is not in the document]
cd-work: :golf:

lockfile/src/spdx.rs
mathew-horner: I'm confused; we were talking about missing test coverage, not a behavior change?

ejortega: The test was originally removed due to behavior change. The behavior being that this used to fail parsing if we didn't have an external reference with package locator or a PURL to determine the triplet (package name, version, ecosystem). If we came across an unknown ecosystem with the external reference, we gave a warning and continued to parse the remaining packages.

Now that this uses the download location to determine the ecosystem, this test should no long fail regarding missing a PURL or locator since we can fallback on using the downloadlocation. If it's a url from a known registry then we can determine enough information to send off for processing, otherwise it is now an unknown ecosystem.

lockfile/src/spdx.rs
```diff
@@ -74,14 +74,16 @@ pub(crate) struct Relationship {
...
+        || url.starts_with("https://registry.yarnpkg.com")
     {
         Ok(PackageType::Npm)
     } else if url.starts_with("https://rubygems.org") {
         Ok(PackageType::RubyGems)
     } else if url.starts_with("https://files.pythonhosted.org") {
         Ok(PackageType::PyPi)
-    } else if url.starts_with("https://repo1.maven.org") {
+    } else if url.starts_with("https://repo1.maven.org")
+        || url.starts_with("https://search.maven.org")
```
cd-work, line 85: This seems to go to `central.sonatype.com`. Is there a chance the name will change in the future and it should also be added immediately?

lockfile/src/spdx.rs
```diff
@@ -142,10 +144,9 @@ impl TryFrom<&PackageInformation> for Package {
     type Error = anyhow::Error;
 
     fn try_from(pkg_info: &PackageInformation) -> anyhow::Result<Self> {
-        pkg_info
-            .external_refs
-            .iter()
-            .find_map(|external| match external.reference_category {
+        // Attempt to determine package data from external refs.
+        let pkg_result =
+            pkg_info.external_refs.iter().find_map(|external| match external.reference_category {
```
cd-work, line 149: This is mostly a guess, but I think if you match on both `reference_category` and `reference_type` then you'll avoid the line break on the assignment due to causing `find_map` to create a dedicated block (instead of direct match) and at the same time remove the extra conditionals inside the match.

cd-work (changes requested):

lockfile/src/spdx.rs
ejortega: I can add it, I think that change was fairly recent. The [API guide](https://central.sonatype.org/search/rest-api-guide/) still shows `search.maven.org`.

lockfile/src/spdx.rs
mathew-horner: Yes, I was just saying that we should have _some_ test or even just a singular package in another one of the tests that shows that if you get an `UnknownEcosystem`, then the whole process doesn't fail.

cd-work (approved):

lockfile/src/spdx.rs
ejortega: Ah gotcha, it's covered by the current [test](https://github.com/phylum-dev/cli/blob/d8e80a11ec38cc690f6251b75b801c6b27c8b53b/lockfile/src/spdx.rs#L370). The parsing comes back empty without failing if only one package and it's an `UnknownEcosystem`

mathew-horner (approved):

Issue #1386, opened 2024-04-04
Issue parsing `effective-pom.xml` files with `ch1252` encoding


ejortega:
# Overview
The parser fails when deserializing https://github.com/eugenp/tutorials/blob/master/maven-modules/maven-pom-types/effective-pom.xml

# How To Reproduce
Steps to reproduce this behavior:
1. Analyze an effective-pom xml with ch1252 encoding
2. See `Unsupported encoding: Cp1252` error

PR #1387, opened 2024-04-04
Use quick-xml to parse pom files

ejortega:
Uses `quick-xml` crate for parsing pom files. It attempts to automatically detect the encoding and will fallback to `utf-8` which does successfully parse the file.

Closes https://github.com/phylum-dev/cli/issues/1386

## Checklist


lockfile/Cargo.toml
```diff
@@ -18,6 +18,7 @@ log = "0.4.6"
 nom = "7.1.1"
 phylum_types = { git = "https://github.com/phylum-dev/phylum-types", branch = "development" }
 purl = "0.1.1"
+quick-xml = { version = "0.31.0", features = ["encoding", "serialize"] }
```
cd-work, line 21: Is there a good reason to use both serde-xml-rs and quick-xml? I'd prefer just using one of them.

cd-work (changes requested):

lockfile/Cargo.toml
ejortega: The `csharp` parser uses a method [non_contiguous_seq_elements](https://docs.rs/serde-xml-rs/0.6.0/serde_xml_rs/de/struct.Deserializer.html#method.non_contiguous_seq_elements) [here](https://github.com/phylum-dev/cli/blob/920aee4696d08b9819729ac88b1fe70e2131507a/lockfile/src/csharp.rs#L147) that's not available in `quick-xml`.

The tests pass if I remove it but it was explicitly added so I'm guessing it was needed at some point. I could look at writing a custom deserializer if we don't want to keep both crates.

ejortega: https://docs.rs/quick-xml/0.31.0/quick_xml/index.html#overlapped-lists

I think this handles the case I described. Will swap out the crate entirely.

lockfile/src/csharp.rs
```diff
@@ -95,16 +94,19 @@ pub struct CSProj;
 
 #[derive(Debug, Deserialize, PartialEq, Eq)]
 pub struct PackageReference {
-    #[serde(alias = "Include", default)]
+    #[serde(rename = "@Include", default)]
```
cd-work, line 97: Curious why these were changed, considering alias should be a superset here?

lockfile/src/csharp.rs
```diff
@@ -143,9 +147,11 @@ impl Parse for CSProj {
     /// Parses `.csproj` files into a vec of packages
     fn parse(&self, data: &str) -> anyhow::Result<Vec<Package>> {
         let data = data.trim_start_matches(UTF8_BOM);
-        let mut de =
-            Deserializer::new_from_reader(data.as_bytes()).non_contiguous_seq_elements(true);
-        let parsed = Project::deserialize(&mut de)?;
+        // let mut de =
+        // Deserializer::new_from_reader(data.as_bytes()).
+        // non_contiguous_seq_elements(true); let parsed =
+        // Project::deserialize(&mut de)?;
```
cd-work, line 153: I'd assume this patch is a WIP, please feel free to re-request a review once it's done.

cd-work (changes requested):

lockfile/src/csharp.rs
ejortega: Reverted, I was following the readme, but it didn't need to be changed except for the adding @ for the attribute.

lockfile/src/csharp.rs
ejortega: Forgot to cleanup. Fixed now

lockfile/src/csharp.rs
```diff
@@ -95,11 +94,14 @@ pub struct CSProj;
...
-    #[serde(alias = "Include", default)]
+    #[serde(alias = "@Include", default)]
     pub name: String,
 
-    #[serde(alias = "Version", default)]
+    #[serde(alias = "@Version", default)]
     pub version: String,
+
+    #[serde(alias = "@version", default)]
+    pub version_lower: String,
```
cd-work, line 104: What is this? It sounds unrelated to the patch itself and sounds suspiciously like a manifest (opposed to lockfile) thing.

lockfile/src/csharp.rs
```diff
@@ -266,4 +268,24 @@ mod tests {
...
+            <bar>1</bar>
+            <bar>2</bar>
+            <baz>Hello, world</baz>
+            <bar>3</bar>
+            <bar>4</bar>
+        </foo>
+        "##;
+        let foo: Foo = quick_xml::de::from_str(s).unwrap();
+        assert_eq!(foo, Foo { bar: vec![1, 2, 3, 4], baz: "Hello, world".to_string() });
+    }
```
cd-work, line 290: Why are we testing `quick_xml` in our tests? This seems unrelated to our actual objectives.

cd-work (changes requested):

lockfile/src/csharp.rs
ejortega: Ah, this is due to https://github.com/phylum-dev/cli/blob/9fa177b58862c68fa7db793182e0af5b87570a10/lockfile/src/csharp.rs#L147. We didn't have a test case that covered the use of [non_contiguous_seq_elements](https://docs.rs/serde-xml-rs/0.6.0/serde_xml_rs/de/struct.Deserializer.html#method.non_contiguous_seq_elements).  It's disabled by default but was specifically enabled so it was probably encountered at some point. `quick-xml` has [overlapped-lists](https://docs.rs/quick-xml/0.31.0/quick_xml/index.html#overlapped-lists) feature and I wanted to verify it worked accordingly.

cd-work: Is it possible to have a test more focused on our actual usage of this feature? I would have imagined it's part of the new test data you added.

ejortega: I haven't added any new test data for csharp other than this test. The other work is `SPDX` related, but I'll look at generating something closer to our usage or find something that has it.

ejortega: Updated the overlapped test case.

lockfile/src/csharp.rs
ejortega: Removed, didn't realize you could use multiple alias', but the test file has two version attributes (Version and version).

cd-work (approved):

PR #1390, opened 2024-04-08
Only set content-type header with content

matt-phylum:
This PR changes the `PhylumApi.fetch` method to only set the `Content-Type` header when it is sending content. Most API endpoints either expect a JSON request body with `Content-Type: application/json`, or expect no content (and don't care about `Content-Type` at all). There are a couple API endpoints that optionally accept a request body which may or may not be JSON, and for those endpoints the `Content-Type` header must be set to JSON if and only if the request body is JSON. A null body is not JSON, so sending a request to one of those endpoints using the usual `PhylumApi.fetch("v0", "/endpoint", { method: "POST" })` results in a 400 error when the server tries to decode a 0-byte body as JSON.

This PR doesn't attempt to do any sort of autodetection to determine if a request body is actually JSON. If there is a body and the type is not explicitly specified, it gets sent as JSON. This means if you want to send a text body you still need to set `Content-Type: text/plain`.

See this extension PR comment thread https://github.com/phylum-dev/community-extensions/pull/43#discussion_r1554221398

## Checklist

- [ ] Have you updated CHANGELOG.md (or extensions/CHANGELOG.md), if applicable

### DIFF
--- a/extensions/phylum.ts
+++ b/extensions/phylum.ts
@@ -64,8 +64,8 @@ export type RejectionSource = {
   reason: string | null;
 };
 
-async function requestHeaders(headersInit?: HeadersInit): Promise<Headers> {
-  const headers = new Headers(headersInit);
+async function ensureRequestHeaders(init: RequestInit) {
+  const headers = init.headers = new Headers(init.headers);
 
   // Set Authorization header if it is missing.
   if (!headers.has("Authorization")) {
@@ -74,11 +74,9 @@ async function requestHeaders(headersInit?: HeadersInit): Promise<Headers> {
   }
 
   // Set Content-Type header if it is missing.
-  if (!headers.has("Content-Type")) {
+  if (init.body && !headers.has("Content-Type")) {
     headers.set("Content-Type", "application/json");
   }
-
-  return headers;
 }
 
 export class PhylumApi {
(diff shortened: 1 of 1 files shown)


extensions/phylum.ts
```diff
@@ -74,11 +74,9 @@ async function requestHeaders(headersInit?: HeadersInit): Promise<Headers> {
   }
 
   // Set Content-Type header if it is missing.
-  if (!headers.has("Content-Type")) {
+  if (init.body && !headers.has("Content-Type")) {
```
kylewillmon, line 77: This has me wondering about the `{body: ""}` edge case... I have no idea whether that counts as a request with a body or not... But it's definitely not valid JSON, so I'm going to avoid that rabbit hole.

kylewillmon (approved):

extensions/phylum.ts
matt-phylum: `{body: ""}` won't get the header because `""` is not truthy. I think as far as HTTP requests are concerned, there isn't a difference between an empty body and no body. I considered checking if it was `null` or `undefined`, but the interface allows either and it's much easier to check if it's truthy than to check that it's not null and not undefined.

kylewillmon: > it's much easier to check if it's truthy than to check that it's not null and not undefined.

I wouldn't say "much easier". You can check for both null and undefined using `init.body == null` (thanks to loose equality)

But, as noted, `""` isn't valid JSON and may not even be a valid body. So setting `Content-Type: application/json` makes no sense for that request. The current code is correct.

Issue #1396, opened 2024-04-16
Improve `Go` dependency parsing


ejortega:
# Overview
The current lockfile generator creates a `go.sum` file and parses the output for dependencies. However the `go.sum` contents is not really a lockfile and analysis results are skewed due to these additional versions even if they aren't used to build the application.

Update the go.sum parser to ignore hashes for /go.mod. These only indicate that a package was considered and not that the package is actually used. There will still be too many dependencies but should have improved results.

Newer go projects supporting go directive 1.17 and higher can use the go.mod parser.

# Acceptance Criteria
- [ ] Update the go.sum parser to ignore hashes for /go.mod.
- [ ] Replace the Go lockfile generator with a go.mod parser which only works for files with a go directive of 1.17 or higher.

cd-work:
How do you envision this to be done? Our lockfile analysis cannot run arbitrary commands. While we could use this for lockfile analysis, I don't see us being able to use this for lockfile analysis.

Generally speaking it doesn't sound like this output is ever intended to land in a file, which is a prerequisite for lockfile parsing.

kylewillmon:
> While we could use this for lockfile generation, I don't see us being able to use this for lockfile analysis.

That's exactly the plan. The Go lockfile generator currently uses `go get -d`. The plan is to update that generator to use `go list` instead.

> Generally speaking it doesn't sound like this output is ever intended to land in a file

That is true, but the output contains enough information to generate a virtual `go.sum` file, just like our [pip generator][1] does with `pip install --dry-run` to `requirements.txt`

[1]: https://github.com/phylum-dev/cli/blob/920aee4696d08b9819729ac88b1fe70e2131507a/lockfile_generator/src/pip.rs#L53-L110

cd-work:
Yeah I apologize I must have been completely blind when reading this yesterday. It states clearly multiple times that this refers to lockfile generation, so not sure why I thought this had anything to do with our normal lockfile analysis.

kylewillmon:
I've been doing some research on this... Go is complicated and the behavior has changed a lot... Even `go list` has issues because it does not respect module pruning. 

## Proposal

To improve our handling, I think we can do a few things:

1. Update the `go.sum` parser to ignore hashes for `/go.mod`. These only indicate that a package was considered and not that the package is actually used.
2. Replace the Go lockfile generator with a `go.mod` parser which only works for files with a `go` directive of 1.17 or higher.

[...]

PR #1397, opened 2024-04-18
Update go lockfile generator to use output from `go list`

ejortega:
Changes the `go` lockfile generator to use the output of `go list -m -json all` from the previous `go.sum` generation using `go get -d`.

Closes https://github.com/phylum-dev/cli/issues/1396

## Checklist

### DIFF
--- a/lockfile/src/golang.rs
+++ b/lockfile/src/golang.rs
@@ -66,4 +66,69 @@ mod tests {
             assert!(pkgs.contains(&expected_pkg));
         }
     }
+
+    #[cfg(feature = "generator")]
+    #[test]
+    fn lock_generate_goproj() {
+        let path = Path::new("../tests/fixtures/lock_generate_go/go.mod");
+        let lockfile = GoGenerator.generate_lockfile(path).unwrap();
+
+        let pkgs = GoSum.parse(&lockfile).unwrap();
+        assert_eq!(pkgs.len(), 9);
+
+        let expected_pkgs = [
+            // Go.mod direct dependencies.
+            Package {
+                name: "github.com/go-chi/chi/v5".into(),
+                version: PackageVersion::FirstParty("v5.0.12".into()),
+                package_type: PackageType::Golang,
+            },
+            Package {
+                name: "github.com/rs/zerolog".into(),
+                version: PackageVersion::FirstParty("v1.32.0".into()),
+                package_type: PackageType::Golang,
+            },
+            // Go.mod indirect dependencies.
+            Package {
+                name: "github.com/mattn/go-colorable".into(),
+                version: PackageVersion::FirstParty("v0.1.13".into()),
+                package_type: PackageType::Golang,
+            },
+            Package {
+                name: "github.com/mattn/go-isatty".into(),
+                version: PackageVersion::FirstParty("v0.0.19".into()),
+                package_type: PackageType::Golang,
+            },
+            Package {
+                name: "golang.org/x/sys".into(),
+                version: PackageVersion::FirstParty("v0.12.0".into()),
+                package_type: PackageType::Golang,
+            },
+            // Transitive dependencies.
+            Package {
+                name: "github.com/coreos/go-systemd/v22".into(),
+                version: PackageVersion::FirstParty("v22.5.0".into()),
+                package_type: PackageType::Golang,
+            },
+            Package {
+                name: "github.com/godbus/dbus/v5".into(),
+                version: PackageVersion::FirstParty("v5.0.4".into()),
+                package_type: PackageType::Golang,
+            },
+            Package {
+                name: "github.com/pkg/errors".into(),
+                version: PackageVersion::FirstParty("v0.9.1".into()),
+                package_type: PackageType::Golang,
+            },
+            Package {
+                name: "github.com/rs/xid".into(),
+                version: PackageVersion::FirstParty("v1.5.0".into()),
+                package_type: PackageType::Golang,
+            },
+        ];
+
+        for expected_pkg in expected_pkgs {
+            assert!(pkgs.contains(&expected_pkg), "missing package {expected_pkg:?}");
+        }
+    }
 }
--- a/lockfile/src/golang.rs
+++ b/lockfile/src/golang.rs
@@ -73,10 +73,9 @@ mod tests {
         let path = Path::new("../tests/fixtures/lock_generate_go/go.mod");
         let lockfile = GoGenerator.generate_lockfile(path).unwrap();
 
-        let pkgs = GoSum.parse(&lockfile).unwrap();
-        assert_eq!(pkgs.len(), 9);
+        let mut actual_pkgs = GoSum.parse(&lockfile).unwrap();
 
-        let expected_pkgs = [
+        let mut expected_pkgs = [
             // Go.mod direct dependencies.
             Package {
                 name: "github.com/go-chi/chi/v5".into(),
@@ -127,8 +126,9 @@ mod tests {
             },
         ];
 
-        for expected_pkg in expected_pkgs {
-            assert!(pkgs.contains(&expected_pkg), "missing package {expected_pkg:?}");
-        }
+        expected_pkgs.sort();
+        actual_pkgs.sort();
+
+        assert_eq!(expected_pkgs, *actual_pkgs);
     }
 }
--- a/lockfile_generator/src/go.rs
+++ b/lockfile_generator/src/go.rs
@@ -18,23 +18,6 @@ struct Module {
     version: Option<String>,
 }
 
-#[derive(Serialize, Deserialize, Debug)]
-struct GoDep {
-    name: String,
-    version: String,
-}
-
-fn create_go_dep(module: Module) -> Option<GoDep> {
-    // Skip main module.
-    if module.main.unwrap_or(false) {
-        return None;
-    }
-
-    let version = module.version?;
-
-    Some(GoDep { name: module.path, version })
-}
-
 pub struct Go;
 
 impl Generator for Go {
@@ -83,14 +66,19 @@ impl Generator for Go {
             let stderr = String::from_utf8_lossy(&output.stderr);
             return Err(Error::NonZeroExit(code, stderr.into()));
         }
-        // Parse go list STDOUT.
-        let stdout = String::from_utf8(output.stdout)?;
-        let stream = serde_json::Deserializer::from_str(&stdout).into_iter::<Module>();
-        let packages = stream.filter_map(|res| res.ok()).filter_map(create_go_dep);
 
+        // Parse go list STDOUT.
+        let stream = serde_json::Deserializer::from_slice(&output.stdout).into_iter::<Module>();
         let mut lockfile = String::new();
-        for pkg in packages {
-            let _ = writeln!(lockfile, "{} {} h1:h1", pkg.name, pkg.version);
+        for res in stream {
+            let module = res?;
+            if module.main == Some(true) {
+                // Skip main module.
+                continue;
+            }
+            if let Some(version) = module.version {
+                let _ = writeln!(lockfile, "{} {} h1:FAKEHASH", module.path, version);
+            }
         }
 
         Ok(lockfile)
--- a/lockfile_generator/src/go.rs
+++ b/lockfile_generator/src/go.rs
@@ -1,29 +1,99 @@
 //! Go ecosystem.
 
 use std::ffi::OsStr;
+use std::fmt::Write;
+use std::fs;
 use std::path::{Path, PathBuf};
-use std::process::Command;
+use std::process::{Command, Stdio};
+
+use serde::{Deserialize, Serialize};
 
 use crate::{Error, Generator, Result};
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "PascalCase")]
+struct Module {
+    path: String,
+    main: Option<bool>,
+    version: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+struct GoDep {
+    name: String,
+    version: String,
+}
+
+fn create_go_dep(module: Module) -> Option<GoDep> {
+    // Skip main module.
+    if module.main.unwrap_or(false) {
+        return None;
+    }
+
+    let version = module.version?;
+
+    Some(GoDep { name: module.path, version })
+}
 
 pub struct Go;
 
 impl Generator for Go {
-    fn lockfile_path(&self, manifest_path: &Path) -> Result<PathBuf> {
-        let project_path = manifest_path
-            .parent()
-            .ok_or_else(|| Error::InvalidManifest(manifest_path.to_path_buf()))?;
-        Ok(project_path.join("go.sum"))
+    fn lockfile_path(&self, _manifest_path: &Path) -> Result<PathBuf> {
+        // NOTE: Go's `generate_lockfile` will never write to disk.
+        unreachable!()
     }
 
     fn command(&self, _manifest_path: &Path) -> Command {
         let mut command = Command::new("go");
-        command.args(["get", "-d"]);
+        command.args(["list", "-m", "-json", "all"]);
         command
     }
 
     fn tool(&self) -> &'static str {
         "Go"
+    }
+
+    /// Generate dependencies from `go list` output.
+    ///
+    /// Since the `go list` never writes any actual lockfile to the disk, we
+    /// provide a custom method here which parses this output and transforms it
+    /// into a `go.sum` format our lockfile parser expects.
+    fn generate_lockfile(&self, manifest_path: &Path) -> Result<String> {
+        let canonicalized = fs::canonicalize(manifest_path)?;
+        let project_path = canonicalized
+            .parent()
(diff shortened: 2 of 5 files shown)


lockfile_generator/src/go.rs
```diff
@@ -1,31 +1,101 @@
...
+            return Err(Error::NonZeroExit(code, stderr.into()));
+        }
+        // Parse go list STDOUT.
+        let stdout = String::from_utf8(output.stdout)?;
+        let stream = serde_json::Deserializer::from_str(&stdout).into_iter::<Module>();
+        let packages = stream.filter_map(|res| res.ok()).filter_map(create_go_dep);
+
+        let mut lockfile = String::new();
+        for pkg in packages {
+            let _ = writeln!(lockfile, "{} {} h1:h1", pkg.name, pkg.version);
```
kylewillmon, line 93: It's probably a good idea to make it obvious that the second `h1` here is a fake hash value.

```suggestion
            let _ = writeln!(lockfile, "{} {} h1:FAKEHASH", pkg.name, pkg.version);
```

Also, [the hash is base64-encoded][1]. So as an added bonus, `FAKEHASH` is valid base64.

[1]: https://go.dev/ref/mod#go-sum-files

tests/fixtures/lock_generate_go/go.sum
```diff
@@ -0,0 +1,17 @@
+github.com/coreos/go-systemd/v22 v22.5.0/go.mod h1:Y58oyj3AT4RCenI/lSvhwexgC+NSVTIJ3seZv2GcEnc=
```
kylewillmon, line 1: This is a bit beyond the scope of this PR, but I'm curious about this...

This dependency (and a few others) only have hashes for `/go.mod` and not for the entire package. But this appears to be a legitimate dependency... How does `go` validate the rest of the package?

kylewillmon (changes requested):

tests/fixtures/lock_generate_go/main.go
cd-work, line 1: Why are we adding an entire project here? Is this necessary to make the command work? If so, is it possible that will cause issues with our lockfile generator considering previously we only needed a single file, while now auxiliary files need to be present in the same directory?

The actual code part certainly isn't necessary, so it should probably be removed since it just distracts from the bits we care about.

tests/fixtures/lock_generate_go/go.sum
ejortega: Actually while answering @cd-work question, it appears the command `go list -m -json all` is parsing the `go.sum`. We might be able to just adjust the parser to produce the same results.

tests/fixtures/lock_generate_go/main.go
cd-work: > I was following the example for [csharp](https://github.com/phylum-dev/cli/tree/main/tests/fixtures/lock_generate_csproj), which also includes a project and even a .gitignore, which is probably unnecessary.

The csharp is a minimal generated project, your code doesn't look minimal to me?

ejortega: I created it solely to get a few dependencies in. I can make it more minimal

tests/fixtures/lock_generate_go/go.sum
matt-phylum: I thought this work started because in newer versions of go, `go list` was parsing the go.sum file and then recursing into go.sum files downloaded from the internet.

ejortega: We use the `go.sum` as a lockfile, which is incorrect. It can lead to having false positives as includes packages not used to in a project.

cd-work: > It can lead to having false positives as includes packages not used to in a project.

This is true for multiple ecosystems, is it not? Cargo will add things to the lockfile as long as they're in the manifest, it doesn't matter if it's used or not. I don't really see an issue with that.

ejortega: I wasn't aware of that from cargo, but according to a customer, the output from `go list` was a 95% reduction in the amount of packages listed in the go.sum.

lockfile/src/golang.rs
```diff
@@ -66,4 +66,69 @@ mod tests {
...
     }
+
+    #[cfg(feature = "generator")]
+    #[test]
+    fn lock_generate_goproj() {
+        let path = Path::new("../tests/fixtures/lock_generate_go/go.mod");
+        let lockfile = GoGenerator.generate_lockfile(path).unwrap();
+
+        let pkgs = GoSum.parse(&lockfile).unwrap();
+        assert_eq!(pkgs.len(), 9);
```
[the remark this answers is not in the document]
cd-work: Has not really be an issue in the past. Figuring it out is usually very straight-forward and logging all packages is easy.

tests/fixtures/lock_generate_go/go.sum
matt-phylum: It does not work offline with these files:

```
root@bf8b095c5ee5:/app# go list -m -json all
go: github.com/go-chi/chi/v5@v5.0.12: Get "https://proxy.golang.org/github.com/go-chi/chi/v5/@v/v5.0.12.info": dial tcp: lookup proxy.golang.org on 192.168.65.7:53: dial udp 192.168.65.7:53: connect: network is unreachable
go: github.com/mattn/go-colorable@v0.1.13: Get "https://proxy.golang.org/github.com/mattn/go-colorable/@v/v0.1.13.info": dial tcp: lookup proxy.golang.org on 192.168.65.7:53: dial udp 192.168.65.7:53: connect: network is unreachable
go: github.com/mattn/go-isatty@v0.0.19: Get "https://proxy.golang.org/github.com/mattn/go-isatty/@v/v0.0.19.info": dial tcp: lookup proxy.golang.org on 192.168.65.7:53: dial udp 192.168.65.7:53: connect: network is unreachable
go: github.com/rs/zerolog@v1.32.0: Get "https://proxy.golang.org/github.com/rs/zerolog/@v/v1.32.0.info": dial tcp: lookup proxy.golang.org on 192.168.65.7:53: dial udp 192.168.65.7:53: connect: network is unreachable
go: golang.org/x/sys@v0.12.0: Get "https://proxy.golang.org/golang.org/x/sys/@v/v0.12.0.info": dial tcp: lookup proxy.golang.org on 192.168.65.7:53: dial udp 192.168.65.7:53: connect: network is unreachable
go: github.com/go-chi/chi/v5@v5.0.12: Get "https://proxy.golang.org/github.com/go-chi/chi/v5/@v/v5.0.12.mod": dial tcp: lookup proxy.golang.org on 192.168.65.7:53: dial udp 192.168.65.7:53: connect: network is unreachable
```

It tries to resolve the packages from go.mod, even if go.sum is available.

ejortega: Ah I tried the same thing but it worked however had network up. With `go get -d` and `go tidy`, there's some messages to stdout that shows the files being downloaded, but not with `go list -m -json all`.

lockfile/src/golang.rs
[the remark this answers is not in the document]
cd-work: Your sorting doesn't actually fix this here.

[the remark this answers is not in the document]
ejortega: Then I'm not sure what the issue is.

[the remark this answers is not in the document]
matt-phylum: When asserting that the collections are equal, don't compare the lengths. Alternatively, if you're doing tests for whether the collection contains the expected elements, comparing the length last would give more useful output in the most likely failure case where something is missing. When the length assertion fails it just says the length is wrong and you have to step through with a debugger or modify the code in order to know why the lengths are different.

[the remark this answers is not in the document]
ejortega: Gotcha, I was just being consistent with [other tests](https://github.com/phylum-dev/cli/blob/c6e1f78deda3c02bef73d6a70252a8599430d01d/lockfile/src/csharp.rs#L166). No one seemed to have an issue there. There's a test further below that actually tests if they are equal, but I'll remove this length check.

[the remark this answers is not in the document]
cd-work: > No one seemed to have an issue there.

As I've said, this has never been an issue. I don't mind it at all personally.

That said, most other tests do this because comparing the entire vec is just unnecessary noise, so they compare the length to check for approximate correctness and then take a few edgecase samples to confirm specific values.

tests/fixtures/lock_generate_go/go.sum
kylewillmon: This still didn't sit right with me... So I decided to test this by adding some source code to the `tests/fixtures/lock_generate_go`. Specifically, I added this file (as `main.go`):

```go
package example

import (
  "io"

  "github.com/rs/zerolog/journald"
)

func Foo() io.Writer {
  return journald.NewJournalDWriter()
}
```

And then running `go build` fails with the following error:

```sh-session
> go build
/Users/kyle/go/pkg/mod/github.com/rs/zerolog@v1.32.0/journald/journald.go:27:2: missing go.sum entry for module providing package github.com/coreos/go-systemd/v22/journal (imported by github.com/rs/zerolog/journald); to add:
	go get github.com/rs/zerolog/journald@v1.32.0
```

Running the requested command yields the following:

```sh-session
> go get github.com/rs/zerolog/journald@v1.32.0

> git diff
diff --git a/tests/fixtures/lock_generate_go/go.mod b/tests/fixtures/lock_generate_go/go.mod
index d73b228..5750659 100644
--- a/tests/fixtures/lock_generate_go/go.mod
+++ b/tests/fixtures/lock_generate_go/go.mod
@@ -8,6 +8,7 @@ require (
 )

 require (
+       github.com/coreos/go-systemd/v22 v22.5.0 // indirect
        github.com/mattn/go-colorable v0.1.13 // indirect
        github.com/mattn/go-isatty v0.0.19 // indirect
        golang.org/x/sys v0.12.0 // indirect
diff --git a/tests/fixtures/lock_generate_go/go.sum b/tests/fixtures/lock_generate_go/go.sum
index b0fa8e1..91a4d60 100644
--- a/tests/fixtures/lock_generate_go/go.sum
+++ b/tests/fixtures/lock_generate_go/go.sum
@@ -1,3 +1,4 @@
+github.com/coreos/go-systemd/v22 v22.5.0 h1:RrqgGjYQKalulkV8NGVIfkXQf6YYmOyiJKk8iXXhfZs=
 github.com/coreos/go-systemd/v22 v22.5.0/go.mod h1:Y58oyj3AT4RCenI/lSvhwexgC+NSVTIJ3seZv2GcEnc=
 github.com/go-chi/chi/v5 v5.0.12 h1:9euLV5sTrTNTRUU9POmDUvfxyj6LAABLUcEWO+JJb4s=
 github.com/go-chi/chi/v5 v5.0.12/go.mod h1:DslCQbL2OYiznFReuXYUmQ2hGd1aDpCnlMNITLSKoi8=
```

So once code is actually required, it adds the hash to `go.sum`. But more importantly, **it adds the dependency to `go.mod`**!!

### What do the docs say?

The [docs][1] for `go.mod` state:

[1]: https://go.dev/doc/modules/gomod-ref#go-notes

> - At go 1.17 or higher:
>   - The go.mod file includes an explicit require directive for each module that provides any package transitively imported by a package or test in the main module. (At go 1.16 and lower, an indirect dependency is included only if minimal version selection would otherwise select a different version.) This extra information enables module graph pruning and lazy module loading.

In other words, for **all supported versions of go** (currently oldest supported is 1.21), `go.mod` includes all direct and transitive dependencies with the precise version that will be used.

### Conclusion (TLDR)

We have been going about this all wrong... `go.mod` is a lockfile and we do not need `go.sum` at all.

kylewillmon (changes requested):
It seems we have completely misunderstood the Go ecosystem... `go.mod` is a lockfile and we should parse it rather than going through all of this trouble to generate something equivalent.

tests/fixtures/lock_generate_go/go.sum
kylewillmon: > In other words, for **all supported versions of go** (currently oldest supported is 1.21), `go.mod` includes all direct and transitive dependencies with the precise version that will be used.

This is unfortunately not true... The Go version that is relevant is the `go` directive in `go.mod`, which is the minimum supported go version and is allowed to be lower than the oldest support go version... That complicates things slightly and puts us into a world similar to Python where `go.mod` **might be** a lockfile, but we don't know until we attempt to parse it.

ejortega: For that same example, if you run `go list -m -json all`. You'll get the following:
```json
root@8d18781e51ee:/app# go list -m -json all
{
        "Path": "cli/example",
        "Main": true,
        "Dir": "/app",
        "GoMod": "/app/go.mod",
        "GoVersion": "1.22.2"
}
{
        "Path": "github.com/coreos/go-systemd/v22",
        "Version": "v22.5.0",
        "Time": "2022-11-07T13:52:27Z",
        "Indirect": true,
        "Dir": "/go/pkg/mod/github.com/coreos/go-systemd/v22@v22.5.0",
        "GoMod": "/go/pkg/mod/cache/download/github.com/coreos/go-systemd/v22/@v/v22.5.0.mod",
        "GoVersion": "1.12"
}
{
        "Path": "github.com/godbus/dbus/v5",
        "Version": "v5.0.4",
        "Time": "2021-03-23T03:10:11Z",
        "Indirect": true,
        "GoMod": "/go/pkg/mod/cache/download/github.com/godbus/dbus/v5/@v/v5.0.4.mod",
        "GoVersion": "1.12"
}
{
        "Path": "github.com/mattn/go-colorable",
        "Version": "v0.1.13",
        "Time": "2022-08-15T05:53:26Z",
        "Indirect": true,
        "Dir": "/go/pkg/mod/github.com/mattn/go-colorable@v0.1.13",
        "GoMod": "/go/pkg/mod/cache/download/github.com/mattn/go-colorable/@v/v0.1.13.mod",
        "GoVersion": "1.15"
}
{
        "Path": "github.com/mattn/go-isatty",
        "Version": "v0.0.19",
        "Time": "2023-03-22T15:51:17Z",
        "Indirect": true,
        "Dir": "/go/pkg/mod/github.com/mattn/go-isatty@v0.0.19",
        "GoMod": "/go/pkg/mod/cache/download/github.com/mattn/go-isatty/@v/v0.0.19.mod",
        "GoVersion": "1.15"
}
{
        "Path": "github.com/pkg/errors",
        "Version": "v0.9.1",
        "Time": "2020-01-14T19:47:44Z",
        "Indirect": true,
        "GoMod": "/go/pkg/mod/cache/download/github.com/pkg/errors/@v/v0.9.1.mod"
}
{
        "Path": "github.com/rs/xid",
        "Version": "v1.5.0",
        "Time": "2023-04-12T03:56:21Z",
        "Indirect": true,
        "GoMod": "/go/pkg/mod/cache/download/github.com/rs/xid/@v/v1.5.0.mod",
        "GoVersion": "1.12"
}
{
        "Path": "github.com/rs/zerolog",
        "Version": "v1.32.0",
        "Time": "2024-02-01T00:56:59Z",
        "Dir": "/go/pkg/mod/github.com/rs/zerolog@v1.32.0",
        "GoMod": "/go/pkg/mod/cache/download/github.com/rs/zerolog/@v/v1.32.0.mod",
        "GoVersion": "1.15"
}
{
        "Path": "golang.org/x/sys",
        "Version": "v0.12.0",
        "Time": "2023-08-25T20:53:44Z",
        "Indirect": true,
        "Dir": "/go/pkg/mod/golang.org/x/sys@v0.12.0",
        "GoMod": "/go/pkg/mod/cache/download/golang.org/x/sys/@v/v0.12.0.mod",
        "GoVersion": "1.17"
}
```

There's a few more dependencies that are not listed in the current `go.mod` file. I'm guessing as those functions/modules are not used in the current example or maybe because the mod file itself is `go 1.15`?

`github.com/pkg/errors` for example is listed as a dependency in the `go list` results because it's a direct dependency of `github.com/rs/zerolog`,  but not showing as an indirect in the current go.mod.

If you look at the mod file for `github.com/rs/zerolog`, you'll see `github.com/pkg/errors` is a direct dependency.
```bash
root@8d18781e51ee:/app# cat /go/pkg/mod/github.com/rs/zerolog\@v1.32.0/go.mod
module github.com/rs/zerolog

go 1.15

require (
        github.com/coreos/go-systemd/v22 v22.5.0
        github.com/mattn/go-colorable v0.1.13
        github.com/mattn/go-isatty v0.0.19 // indirect
        github.com/pkg/errors v0.9.1
        github.com/rs/xid v1.5.0
        golang.org/x/sys v0.12.0 // indirect
)
```

kylewillmon: > `github.com/pkg/errors` for example is listed as a dependency in the `go list` results because it's a direct dependency of `github.com/rs/zerolog`, but not showing as an indirect in the current go.mod.

This is due to module pruning. `github.com/pkg/errors` is unused, so it is not included in the build and cannot effect the runtime behavior. For that same reason, we do not need to include it in Phylum analysis.

ejortega:
Closing this PR as we'll be parsing the `go.mod` and `go.sum` directly.

PR #1399, opened 2024-04-18
Bump to latest deno

kylewillmon:


cli/src/commands/extensions/api.rs
```diff
@@ -251,6 +251,10 @@ async fn get_access_token(
 #[op2(async)]
 #[serde]
 async fn get_refresh_token(op_state: Rc<RefCell<OpState>>) -> Result<RefreshToken> {
+    get_refresh_token_impl(op_state).await
+}
+
+async fn get_refresh_token_impl(op_state: Rc<RefCell<OpState>>) -> Result<RefreshToken> {
```
matt-phylum, line 257: There should probably be a comment explaining why this needs to be duplicated. (because `#[op2(async)]` replaces the function with a function that describes the function instead of doing what it's written to do?)

matt-phylum (approved):

kylewillmon:
Seems like this is going to need more time and testing... I'll come back to this when I've got time.

extensions/phylum.ts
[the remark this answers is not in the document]
maxrake: Indeed...this file is referenced in `docs/extensions/extension_api.md`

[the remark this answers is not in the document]
kylewillmon: Alright. I've replaced this file with a [global definition file][1], `extensions/phylum.d.ts`

[1]: https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-d-ts.html

As an added bonus, this restores type-checking and has allowed me to remove 892c072

It's a bit annoying to have the type definition separate from the implementation... But this is how many libraries operate (including Deno itself), so hopefully it will work out for us.

cli/src/commands/extensions/api.rs
kylewillmon: I decided to refactor this a bit instead and just implement a `refresh_token()` helper function. See b534da3

It's a bit more efficient this way.

cli/src/deno.rs
```diff
@@ -25,8 +32,19 @@ use crate::commands::extensions::state::ExtensionState;
...
+    dir "js",
+    "api.js",
+    "api_version.js",
+    "main.js");
+
+/// Importable "phylum" module for backwards compatibility.
+const PHYLUM_MODULE: &str = "
+export const PhylumApi = Phylum;
+export const ApiVersion = Phylum.ApiVersion;
+";
```
maxrake, line 47: Even with this backwards compatibility, should the documentation be updated to match the _current_ method of using the global `Phylum` object directly? If so, it looks like `PhylumApi` shows up in these doc files:

* `docs/extensions/extension_example.md`
* `docs/extensions/extension_manifest.md`
* `docs/extensions/extension_rest_api.md`
* `docs/extensions/extension_sandboxing.md`

maxrake (commented):
This is a non-exhaustive review for now.

This level of change seems to warrant an `extensions/CHANGELOG.md` entry.

cli/src/commands/extensions/mod.rs
```diff
@@ -26,8 +26,6 @@ pub mod permissions;
 pub mod state;
 
 const EXTENSION_SKELETON: &[u8] = b"\
-import { PhylumApi } from 'phylum';
-
 console.log('Hello, World!');
 ";
```
kylewillmon, line 30: I'm not sure what the skeleton should be now... Suggestions are welcome.

maxrake (approved):
All code was reviewed. Not all of it was understood. The documentation looks good now.

PR #1413, opened 2024-05-07
Add `go.mod` parser

ejortega:
## Checklist

- [ ] Have you updated CHANGELOG.md (or extensions/CHANGELOG.md), if applicable


lockfile/src/golang.rs
```diff
@@ -37,6 +37,50 @@ impl Parse for GoSum {
...
 }
 
+pub struct GoDeps {
+    pub go: String,
+    pub modules: Vec<Package>,
+}
+
+fn check_go_directive(version: &str) -> anyhow::Result<()> {
+    let numeric_part = version.split(|c: char| !c.is_numeric()).collect::<Vec<&str>>();
+    if let (Some(major), Some(minor)) = (numeric_part.first(), numeric_part.get(1)) {
```
cd-work, line 47: No need to collect if you just `next()` `next()`?

lockfile/src/golang.rs
```diff
@@ -37,6 +37,50 @@ impl Parse for GoSum {
...
+        let major: u32 = major.parse().unwrap_or(0);
+        let minor: u32 = minor.parse().unwrap_or(0);
+
+        // Check if version meets the criteria.
+        if major < 1 || minor < 17 {
```
cd-work, line 52: What about `2.1`?

lockfile/src/golang.rs
```diff
@@ -37,6 +37,50 @@ impl Parse for GoSum {
...
+            bail!("Minimum supported go directive is 1.17")
+        }
+    } else {
+        bail!("Error parsing go directive")
```
cd-work, line 56: There's currently only 4 instances of `bail!` in CLI, so it is not frequently used. None of these were written by me either.

What is the advantage of using a non-obvious macro here over standard Rust functionality? How is a new contributor supposed to distinguish between `bail!` and `panic!` if all the relevant bits of what `bail!` does are hidden inside the macro?

lockfile/src/golang.rs
```diff
@@ -66,4 +110,258 @@ mod tests {
...
+            Package {
+                name: "golang.org/x/sys".into(),
+                version: PackageVersion::FirstParty("v0.12.0".into()),
+                package_type: PackageType::Golang,
+            },
+        ];
+
+        for expected_pkg in expected_pkgs {
+            assert!(pkgs.contains(&expected_pkg));
+        }
```
cd-work, line 149: The whole point of the `expected_pkg` stuff in our existing tests is to confirm a subset of packages are correct, while the original file itself contains a larger set of packages which possibly might cause issues.

If the test file is a minimal example specifically for testing against anyway, there's no need to check like this. You can just compare the entire vec instead. If you are concerned about order, then sorting should be pretty easy.

lockfile/src/golang.rs
```diff
@@ -66,4 +110,258 @@ mod tests {
...
+            go 1.17
+
+            require (
+                example.com/othermodule v1.2.3
+                example.com/thismodule v1.2.3
+                example.com/thatmodule v1.2.3
+            )
+
+            replace example.com/thatmodule => ../thatmodule
+            replace example.com/thismodule v1.2.3 => example.com/newmodule v3.2.1
```
cd-work, line 189: Couldn't this be part of our sample fixture file? I like having a file that contains all kinds of edgecases since that makes it easier to test things.

Same goes for most of the other tests.

lockfile/src/lib.rs
```diff
@@ -161,9 +164,10 @@ impl Iterator for LockfileFormatIter {
             9 => LockfileFormat::NugetLock,
             10 => LockfileFormat::Msbuild,
             11 => LockfileFormat::Go,
-            12 => LockfileFormat::Cargo,
-            13 => LockfileFormat::Spdx,
-            14 => LockfileFormat::CycloneDX,
+            12 => LockfileFormat::GoMod,
```
cd-work, line 167: We prefer the old go format over the new one. Did you order it this way because of the failure chance?

lockfile/src/parsers/go_mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        // Check if the replacement module version is available.
+        match &replacement.version {
+            Some(version) => {
+                // Remove the specific module version.
+                let dep = Module {
+                    path: module_path.to_owned(),
+                    version: version.to_owned(),
+                    indirect: false,
+                };
+                // Remove all modules wtih the same name and version.
```
cd-work, line 110:
```suggestion
                // Remove all modules with the same name and version.
```

lockfile/src/parsers/go_mod.rs
```diff
@@ -0,0 +1,240 @@
...
+
+        // Check if the replacement module version is available.
+        match &replacement.version {
+            Some(version) => {
+                // Remove the specific module version.
+                let dep = Module {
+                    path: module_path.to_owned(),
+                    version: version.to_owned(),
+                    indirect: false,
+                };
```
cd-work, line 109: These comments don't make sense to me. How does this remove the specific module version and how does the following statement remove all modules with the same name and version?

cd-work (changes requested):

lockfile/src/golang.rs
cd-work: Yeah it's just harder to read imo.

lockfile/src/golang.rs
ejortega: Combined the tests and sample fixture except for the unsupported go directive test case.

cd-work (dismissed):
I did not really review the nom code, but as long as it completes the tests that is fine by me.

ejortega:
@cd-work rebased with main

cd-work (approved):

PR #1415, opened 2024-05-08
Move CHANGELOG entry for process cleanup

kylewillmon:
This entry was mistakenly put in the wrong section. 760cefd was not included in the last release.


maxrake (approved):
LGTM

Issue #1421, opened 2024-05-13
Double lockfiles for Go projects


kylewillmon:
Since the addition of the `gomod` parser in 7eb96a01, Phylum now detects both `go.mod` and `go.sum` as lockfiles. For example:

```sh-session
> phylum status
Project: null
Group: null
Project Root: null
Dependency Files:
 - path: ./go.mod
   type: gomod
 - path: ./go.sum
   type: go
```

Users can override this with a `.phylum_project` file, but this default behavior is not good for anyone. This will also impact the GitHub App, which is likely where most users will run into this issue.

kylewillmon:
As far as I can tell, there are three options:

1. Modify `phylum_lockfile::find_lockfiles()` and friends so that the `gomod` parser is ignored. This would mean that users would only get that parser if it is specifically requested (by command line options or `.phylum_project` file).
2. Find some way to make `go.mod` the "primary" lockfile and only use `go.sum` if parsing `go.mod` fails.
3. Remove the `gomod` parser entirely.

cd-work:
I like 1), but I'd make it a little smarter than suggested. In our depfile logic we already have precedence for conditionally filtering out certain files, we could do the same with the `go.mod` file. That said, I'm not sure if this extra complication would be useful considering that both files are pretty much **always** present in any Go project from what I understand?

kylewillmon:
> In our depfile logic we already have precedence for conditionally filtering out certain files, we could do the same with the `go.mod` file.

This is referring to [our handling of Python files][1], right?

[1]: https://github.com/phylum-dev/cli/blob/d6d2eee4fff6742f8db89fde33197a5a059a5cd6/lockfile/src/lib.rs#L395-L401

I had forgotten about that code... It might be a bit more annoying to do something like that for Go because `go.mod` and `go.sum` are both considered lockfiles. So in the current code, it would require special code in both `find_lockfiles_at()` and `find_depfiles_at()`.

> That said, I'm not sure if this extra complication would be useful considering that both files are pretty much **always** present in any Go project from what I understand?

This is also my understanding. I have not seen any Go project where only one of the two files is present.. Although we may need to double-check Go workspaces since I believe that have multiple `go.mod` files but only a single `go.sum` file...

Even so, preferring `go.sum` in all situations would resolve this issue.

ejortega:
According to the docs. [No, go.sum is not a lock file. The go.mod files in a build provide enough information for 100% reproducible builds.
[...]

PR #1422, opened 2024-05-15
Use the `go.sum` parser as the default lockfile format for golang

ejortega:
Updates the lockfile detection to ignore go.mod files and default to the go.sum parser.

If the go.sum is missing, the go.mod file is recognized and can be parsed if the go directive is >= 1.17.

If the go directive criteria is not satisfied, the user will receive a message that the lockfile generation is required.

```bash
❯ phylum analyze -p test

❗ Could not parse manifest: Parsing "go.mod" requires lockfile generation, but it was disabled
```

Closes https://github.com/phylum-dev/cli/issues/1421

## Checklist

### DIFF
--- a/lockfile/src/lib.rs
+++ b/lockfile/src/lib.rs
@@ -237,8 +237,7 @@ pub struct ThirdPartyVersion {
 ///
 /// The file does not need to exist.
 pub fn get_path_format<P: AsRef<Path>>(path: P) -> Option<LockfileFormat> {
-    LockfileFormat::iter()
-        .find(|f| f != &LockfileFormat::GoMod && f.parser().is_path_lockfile(path.as_ref()))
+    LockfileFormat::iter().find(|f| f.parser().is_path_lockfile(path.as_ref()))
 }
 
 /// Identify a dependency file's format based on its path.
@@ -250,8 +249,7 @@ pub fn get_depfile_path_format<P: AsRef<Path>>(path: P) -> Option<LockfileFormat
     let path = path.as_ref();
     LockfileFormat::iter().find(|format| {
         let parser = format.parser();
-        format != &LockfileFormat::GoMod
-            && (parser.is_path_lockfile(path) || parser.is_path_manifest(path))
+        parser.is_path_lockfile(path) || parser.is_path_manifest(path)
     })
 }
 
@@ -270,7 +268,7 @@ pub fn find_manifest_lockfile<P: AsRef<Path>>(path: P) -> Option<(PathBuf, Lockf
     // Find matching format and lockfile in the manifest's directory.
     LockfileFormat::iter()
         // Check if file is a valid manifest.
-        .filter(|format| format != &LockfileFormat::GoMod && format.parser().is_path_manifest(path))
+        .filter(|format| format.parser().is_path_manifest(path))
         // Try to find the associated lockfile.
         .find_map(|format| {
             let lockfile_path = WalkDir::new(manifest_dir)
--- a/lockfile/src/lib.rs
+++ b/lockfile/src/lib.rs
@@ -331,8 +331,8 @@ impl DepFiles {
                 let parser = format.parser();
 
                 let mut format_found = false;
-                // GoMod can reprsent a manifest and lockfile which causes duplicate
-                // lockfiles being submitted when a go.sum is present. This removees
+                // GoMod can represent a manifest and lockfile which causes duplicate
+                // lockfiles being submitted when a go.sum is present. This removes
                 // go.mod files from being automatically recognized as a lockfile.
                 if format != LockfileFormat::GoMod && parser.is_path_lockfile(path) {
                     depfiles.lockfiles.push((path.to_path_buf(), format));
--- a/lockfile/src/lib.rs
+++ b/lockfile/src/lib.rs
@@ -237,7 +237,8 @@ pub struct ThirdPartyVersion {
 ///
 /// The file does not need to exist.
 pub fn get_path_format<P: AsRef<Path>>(path: P) -> Option<LockfileFormat> {
-    LockfileFormat::iter().find(|f| f.parser().is_path_lockfile(path.as_ref()))
+    LockfileFormat::iter()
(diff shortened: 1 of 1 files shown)


lockfile/src/lib.rs
```diff
@@ -237,7 +237,8 @@ pub struct ThirdPartyVersion {
 ///
 /// The file does not need to exist.
 pub fn get_path_format<P: AsRef<Path>>(path: P) -> Option<LockfileFormat> {
-    LockfileFormat::iter().find(|f| f.parser().is_path_lockfile(path.as_ref()))
+    LockfileFormat::iter()
+        .find(|f| f != &LockfileFormat::GoMod && f.parser().is_path_lockfile(path.as_ref()))
 }
```
cd-work, line 242: Is it necessary to change this? It should still be possible to detect go files on their own.

I'd assume this was changed because of the GitHub app?

lockfile/src/lib.rs
```diff
@@ -249,7 +250,8 @@ pub fn get_depfile_path_format<P: AsRef<Path>>(path: P) -> Option<LockfileFormat
     let path = path.as_ref();
     LockfileFormat::iter().find(|format| {
         let parser = format.parser();
-        parser.is_path_lockfile(path) || parser.is_path_manifest(path)
+        format != &LockfileFormat::GoMod
+            && (parser.is_path_lockfile(path) || parser.is_path_manifest(path))
```
cd-work, line 254: Same comment as the path format function. It's still a valid format, it just shouldn't be suggested together with other go lockfiles.

lockfile/src/lib.rs
```diff
@@ -268,7 +270,7 @@ pub fn find_manifest_lockfile<P: AsRef<Path>>(path: P) -> Option<(PathBuf, Lockf
     // Find matching format and lockfile in the manifest's directory.
     LockfileFormat::iter()
         // Check if file is a valid manifest.
-        .filter(|format| format.parser().is_path_manifest(path))
+        .filter(|format| format != &LockfileFormat::GoMod && format.parser().is_path_manifest(path))
```
cd-work, line 273: A note here on why go.mod is ignored seems prudent.

lockfile/src/lib.rs
```diff
@@ -331,7 +333,7 @@ impl DepFiles {
                 let parser = format.parser();
 
                 let mut format_found = false;
-                if parser.is_path_lockfile(path) {
+                if format != LockfileFormat::GoMod && parser.is_path_lockfile(path) {
```
cd-work, line 336: A comment here would be nice too.

cd-work (changes requested):

lockfile/src/lib.rs
ejortega: Yes this is for the GitHub app. However maybe the change to prefer go.sum first should be done in the app?

lockfile/src/lib.rs
ejortega: Now that I think on this a little more. I'm not sure this function needs to change. It's not in use from what I can tell and its purpose is to identify a lockfile given a manifest.

lockfile/src/lib.rs
ejortega: @kylewillmon @matt-phylum thoughts?

matt-phylum: We need some way to prefer the correct file type or iteratively work through them. It's okay if we detect that both files may be lockfiles we're interested in, grab both, and then decide that we only needed one of them, but right now we detect which file we _are_ interested in, grab that file, and error if we can't use it, and this code I think is supposed to make the current code work.

IIRC it is possible, although undocumented, for users to specify lockfiles instead of us autodetecting, which may be another option if we have a small number of users for which they need to use the older scheme.

matt-phylum: If the autodetection had some way to specify that these two files represent the same dependencies (as opposed to go.mod in one folder and go.lock in a different folder, or go.mod and package-lock.json in the same folder) as options for a single dependency source, we could support trying and failing individual files.

matt-phylum: Wouldn't CLI have the same problem? Without this code, if it detects both go.mod and go.sum, does it have a way to pick the correct file or does it use both unless told otherwise?

ejortega: > Wouldn't CLI have the same problem? Without this code, if it detects both go.mod and go.sum, does it have a way to pick the correct file or does it use both unless told otherwise?

Yes, I removed it from being detected in the `find_at` function. You can still target it manually via `phylum analyze -t gomod go.mod`.

ejortega: I removed this change, however we'll need a new issue for the GitHub app.

lockfile/src/lib.rs
ejortega: Removed

lockfile/src/lib.rs
ejortega: Removed this change

lockfile/src/lib.rs
```diff
@@ -331,7 +331,10 @@ impl DepFiles {
                 let parser = format.parser();
 
                 let mut format_found = false;
-                if parser.is_path_lockfile(path) {
+                // GoMod can reprsent a manifest and lockfile which causes duplicate
```
cd-work, line 334:
```suggestion
                // GoMod can represent a manifest and lockfile which causes duplicate
```

lockfile/src/lib.rs
```diff
@@ -331,7 +331,10 @@ impl DepFiles {
...
+                // lockfiles being submitted when a go.sum is present. This removees
```
cd-work, line 335:
```suggestion
                // lockfiles being submitted when a go.sum is present. This removes
```

cd-work (changes requested):

cd-work (approved):
Analyzing `go.sum` by passing the path works.
Analyzing `go.mod` by passing the path works and doesn't fall back to "lockfile".
Analyzing without path only uses `go.mod`.
Init suggests `go.sum` only if `go.mod` is not available.

LGTM

PR #1428, opened 2024-05-24
Add STDOUT to lockfile generation error message

### LINKED ISSUE #1425: Hidden error messages for dotnet lockfile generator
The lockfile generators capture and return stderr from the ecosystem tool when an error occurs. Unfortunately, `dotnet` prints error messages to stdout...

Because of this, we need to return (or redirect) stdout for `dotnet` (at a minimum) so that users can see the appropriate message when an error occurs.

cd-work:
Previously only the STDERR was printed whenever lockfile generation encountered a non-zero exit code from the package manager, however some package managers like `dotnet` print their errors to STDOUT.

This patch reworks the output in these error cases to print both STDERR and STDOUT and formats things a little nicer to make it easy to keep the two apart.

Closes #1425.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Sandboxed processes sticking around after CLI is killed with a signal
 - Lockfiles with local versions breaking the pip parser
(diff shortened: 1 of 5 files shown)


maxrake (changes requested):
Tested locally and it works as intended. Request adding CHANGELOG entry.

maxrake (approved):

PR #1435, opened 2024-06-03
Prioritize msbuild parser

kylewillmon:
The nugetlock parser is theoretically better, but many clients have found that the nugetlock lockfile generation does not work without a very specific environment because of the complexity of the build system...

This patch is an effort to work in more scenarios by switching the default parser to msbuild


maxrake (approved):
This was (somewhat) tested locally. It was hard to replicate the same results as would be found in CI with a `linux/amd64` platform Docker image running on a macOS system. The changes in this PR are minimal and review of the code that determines which parser and/or generator will be used led me to believe the changes here are correct. There is a small question about whether the documentation will need to be updated based on this change...particularly the [lockfile generation page](https://docs.phylum.io/cli/lockfile_generation). Approving now to get a solution out there and the docs can be updated later, if needed.

PR #1440, opened 2024-06-07
Add PNPM lockfile version 5 support

### LINKED ISSUE #1438: Support pnpm lockfile version 5
Errors occur when parsing a pnpm lockfile version 5 because package IDs use a forward slash to separate the name and version (`/foo/1.0.0`) instead of an `@` (`/foo@1.0.0`). See [the lockfile spec][1].

[1]: https://github.com/pnpm/spec/blob/master/lockfile/5.md#packages

cd-work:
Closes #1438.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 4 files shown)


maxrake (changes requested):
The code was reviewed and confirmed with local testing. The inclusion of a new test fixture is appreciated! The only requested change is to add a CHANGELOG entry for this new feature.

PR #1443, opened 2024-06-11
Allow execute for maven for lockfile generation

kylewillmon:
Maven has a `libexec` folder that contains scripts that get executed when generating lockfiles. This patch allows execute so that those scripts are not blocked.


maxrake (changes requested):
Looks like it would work. CHANGELOG entry requested.

maxrake (approved):
Tested locally and it works as intended with the `homebrew` install of Maven on macOS. LGTM.

PR #1450, opened 2024-06-17
Improve robustness of find-permissions extensions

cd-work:
This patch adds two minor changes that should make the `find-permissions` extensions easier to use for people unfamiliar with it.

To ensure we don't incorrectly log that `/` is required when the command always fails, the extension now does an explicit check on `/` and outputs a detailed error on the issue.

While I wasn't able to reproduce this issue myself, there have been some reports on failures due to permission errors. To avoid the possibility of that we now explicitly handle `Deno.readDir` failures and just assume the corresponding directory is wholly required.

To improve consistency, all `console.error` calls have been changed to use red color.

### DIFF
--- a/extensions/find-permissions/main.ts
+++ b/extensions/find-permissions/main.ts
@@ -52,14 +52,14 @@ const check_read = Deno.args.includes("--read");
 
 // Ensure at least one type of permission is specified.
 if (!check_write && !check_read) {
-  console.error("Expected at least one of `--read`, `--write`");
+  console.error(`${red("Expected at least one of `--read`, `--write`")}`);
   Deno.exit(111);
 }
 
 // Ensure test executable was passed.
 const test_bin_path = getArgOption("--bin");
 if (!test_bin_path) {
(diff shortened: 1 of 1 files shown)


extensions/find-permissions/main.ts
```diff
@@ -107,6 +107,14 @@ for (let i = 0; true; i++) {
 // Required sandboxing exceptions.
 const requiredPaths: string[] = [];
 
+// Abort if it doesn't work with all paths allowed.
+if (!test(["/"])) {
+  let msg =
```
maxrake, line 112:
```suggestion
  const msg =
```
This is the suggestion from the deno checks failure.

maxrake (approved):
I like these changes and think they will help. The only suggestion is to fix a deno checks failure.

extensions/find-permissions/main.ts
cd-work: Interesting, I didn't find this when running it locally. But I guess as long as CI finds...

maxrake (approved):
LGTM

PR #1451, opened 2024-06-17
Remove extraneous output from maven generator

cd-work:
This patch silences the maven lockfile generator to avoid part of this output ending up in the target effective-pom.xml.


maxrake (approved):
LGTM

PR #1452, opened 2024-06-18
Fix OpenJDK path detection

cd-work:
This fixes several severe issues with c198a44, which was incorrectly matching whole path segments.


maxrake (approved):
I've read [the docs for `ends_with`](https://doc.rust-lang.org/std/path/struct.Path.html#method.ends_with) and it looks to be correct now.

PR #1457, opened 2024-06-25
Add init script for gradle lockfile generation

cd-work:
This patch adds a temporary init script using Gradle's `--init-script` CLI parameter to force enable lockfile writing during lockfile generation.


maxrake (approved):
Code was reviewed and tested as part of a pair programming session. LGTM.

PR #1460, opened 2024-06-26
Add support for older Gradle lockfiles

### LINKED ISSUE #1459: Support legacy Gradle lockfiles
## Overview

Gradle versions prior to v7 produce lockfiles differently than the newer v7 and v8 releases. These legacy versions produce multiple lockfiles and store them in a `gradle/dependency-locks` directory with `*.lockfile` names. Each file is named by the configuration it locks and has the `lockfile` extension. The CLI, parser, and lockfile generator should be updated to account for these legacy Gradle lockfiles.

## Additional Info

Possible CLI changes required:

* Lockfile identification, [here](https://github.com/phylum-dev/cli/blob/main/lockfile/src/java.rs#L33-L35)
  * `lockfile/src/java.rs`
* Lockfile parsing, if the format for older versions is different
  * `lockfile/src/parsers/gradle_dep.rs`
* Lockfile generation, for the `lockfile_path` plus maybe some other differences
  * `lockfile_generator/src/gradle.rs`
* [Documentation](https://docs.phylum.io/cli/supported_lockfiles)
  * `docs/supported_lockfiles.md`
* New test fixture(s) for older version of this lockfile

## Resources

* [Gradle v5.5.1 dependency locking documentation](https://docs.gradle.org/5.5.1/userguide/dependency_locking.html)
* [Gradle v6 dependency locking documentation](https://docs.gradle.org/6.0/userguide/dependency_locking.html#lock_state_location_and_format)
* [PR where the format/options changed](https://github.com/gradle/gradle/pull/12561)

## Questions

* Are versions of Gradle prior to v6 targeted for inclusion in this issue?
  * Answer: Yes. Specifically v5.5.1 and newer.
* If an older version of `gradle` is in use and a lockfile does not exist for the CLI to parse, should the lockfile generator be updated (if needed) to ensure it creates a lockfile in the older/matching format?
  * It appears the command for generating lockfiles has remained constant: `gradle dependencies --write-locks`
  * Still need to confirm if the `--init-script` option is also consistent
* Can we get real-world examples of lockfiles generated with Gradle v6?

## Acceptance Criteria

- [x] Lockfiles generated with Gradle v5 are recognized
- [x] Lockfiles generated with Gradle v5 can be parsed
- [ ] Lockfiles generated with Gradle v6 are recognized
- [ ] Lockfiles generated with Gradle v6 can be parsed
- [x] Lockfiles generated with Gradle v7 and v8 continue to be recognized and parsed
- [x] ~Test fixture(s) are added to account for the older format~
  - The format did not change, only the path/name
- [x] Documentation is updated to reflect the additional supported lockfile(s)

cd-work:
This patch accepts Gradle lockfiles inside the `gradle/dependency-locks` directory, which all share Gradle's `.lockfile` suffix but have various different file names.

Closes #1459.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Fixed
+
+- Support for gradle lockfiles in `gradle/dependency-locks/`
+
 ## 6.6.2 - 2024-06-25
 
 ### Fixed
--- a/docs/lockfile_generation.md
+++ b/docs/lockfile_generation.md
@@ -25,7 +25,7 @@ and which tools must be installed to facilitate automatic lockfile generation:
 | `poetry`      | `pyproject.toml` | [`poetry`][poetry]          |
 | `gem`         | `Gemfile`        | `bundle` (from [Bundler][]) |
 | `mvn`         | `pom.xml`        | `mvn` (from [Maven][])      |
-| `gradle`      | `build.gradle` <br/> `build.gradle.kts`   | [`gradle`][gradle]          |
+| `gradle`      | `build.gradle` <br/> `build.gradle.kts`   | [`gradle`][gradle] version 7.0.0+ |
 | `go`          | `go.mod`         | [`go`][go]                  |
 | `cargo`       | `Cargo.toml`     | [`cargo`][cargo]            |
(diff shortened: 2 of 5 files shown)


maxrake (approved):
Code was reviewed and tested as part of a pair programming session. 
The one thing we didn't confirm at this time was if the lockfile generation would work with the init script.
LGTM.

PR #1467, opened 2024-07-08
Fix package-lock.json lockfile parsing failures

cd-work:
This fixes an issue where the lockfile parser for JavaScript's `package-lock.json` was too restrictive and would fail parsing valid lockfiles.

As a solution all packages without explicit `resolved` field are now ignored, since these are the only packages we're capable of analyzing anyway.

### DIFF
--- a/lockfile/src/javascript.rs
+++ b/lockfile/src/javascript.rs
@@ -12,6 +12,7 @@ use lockfile_generator::pnpm::Pnpm as PnpmGenerator;
 use lockfile_generator::yarn::Yarn as YarnGenerator;
 #[cfg(feature = "generator")]
 use lockfile_generator::Generator;
+use log::debug;
 use nom::error::convert_error;
 use nom::Finish;
 use phylum_types::types::package::PackageType;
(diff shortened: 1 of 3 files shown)


lockfile/src/javascript.rs
```diff
@@ -67,19 +68,15 @@ impl Parse for PackageLock {
...
-                }
-
                 // Get dependency type.
-                let resolved = get_field(keys, "resolved")
-                    .ok_or_else(|| anyhow!("Dependency '{name}' is missing \"resolved\" key"))?;
+                let resolved = match get_field(keys, "resolved") {
+                    Some(resolved) => resolved,
+                    // Ignore packages without clear resolution details.
+                    None => {
+                        debug!("ignoring package without `resolved` field: {name}");
```
maxrake, line 76: While testing this branch against the lockfile that generated this PR, I could not get this debug line to show in the output. Is there a specific configuration/environment/etc. required to see the debug statement?

```
phylum v6.6.4-3-gcac3dbf

❯ target/debug/phylum -vvv parse package-lock.json > parse_output.json
[2024-07-08T15:13:22Z DEBUG phylum_cli::config] Reading config from /Users/maxrake/.config/phylum/settings.yaml
[2024-07-08T15:13:22Z DEBUG phylum] Skipping update check...
[2024-07-08T15:13:22Z INFO  phylum_project] Found project configuration file at "/Users/maxrake/dev/phylum/localdev/cli/.phylum_project"
```

lockfile/src/javascript.rs
maxrake, line 1: This change/fix requires a CHANGELOG entry.

maxrake (changes requested):
The changes were tested locally and appear to work as intended. There is a question about debug output and a request to add a CHANGELOG entry.

lockfile/src/javascript.rs
cd-work: `--skip-sandbox`

PR #1473, opened 2024-07-12
Fix msbuild parsing with lowercase version field

cd-work:
This fixes an issue where the `.csproj` parser would ignore uppercase
version fields on `PackageReference`.

### DIFF
--- a/lockfile/src/csharp.rs
+++ b/lockfile/src/csharp.rs
@@ -248,14 +248,24 @@ mod tests {
     #[test]
     fn lock_parse_csproj() {
         let pkgs = CSProj.parse(include_str!("../../tests/fixtures/sample.csproj")).unwrap();
+        assert_eq!(pkgs.len(), 6);
 
-        assert_eq!(pkgs.len(), 5);
-        assert_eq!(pkgs[0].name, "Microsoft.NETFramework.ReferenceAssemblies");
-        assert_eq!(pkgs[0].version, PackageVersion::FirstParty("1.0.0".into()));
+        let expected_pkgs = [
+            Package {
+                name: "Microsoft.NETFramework.ReferenceAssemblies".into(),
+                version: PackageVersion::FirstParty("4.5.0".into()),
+                package_type: PackageType::Nuget,
+            },
+            Package {
+                name: "Microsoft.NETCore.UniversalWindowsPlatform".into(),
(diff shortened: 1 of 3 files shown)


lockfile/src/csharp.rs
```diff
@@ -248,14 +248,24 @@ mod tests {
...
         let pkgs = CSProj.parse(include_str!("../../tests/fixtures/sample.csproj")).unwrap();
+        assert_eq!(pkgs.len(), 6);
 
-        assert_eq!(pkgs.len(), 5);
-        assert_eq!(pkgs[0].name, "Microsoft.NETFramework.ReferenceAssemblies");
-        assert_eq!(pkgs[0].version, PackageVersion::FirstParty("1.0.0".into()));
+        let expected_pkgs = [
+            Package {
+                name: "Microsoft.NETFramework.ReferenceAssemblies".into(),
+                version: PackageVersion::FirstParty("4.5.0".into()),
```
kylewillmon, line 256: I'm not sure why the version was changed here... It's `1.0.0` in `sample.csproj`

```suggestion
                version: PackageVersion::FirstParty("1.0.0".into()),
```

kylewillmon: Oh. This version was from the old check of `pkgs.last().unwrap()`. But the package name is from `pkgs[0].name`...

Anyways, I'll leave you to take care of that however you see fit.

lockfile/src/csharp.rs
```diff
@@ -97,7 +97,7 @@ pub struct PackageReference {
     #[serde(alias = "@Include", default)]
     pub name: String,
 
-    #[serde(alias = "@Version", alias = "@version", default)]
+    #[serde(alias = "@Version", alias = "@version", alias = "Version", default)]
```
ejortega, line 100: Do you want to include lowercase as well?

kylewillmon (approved):

lockfile/src/csharp.rs
kylewillmon: He includes it in the test. It works with lowercase...

ejortega: Ah missed thank, thanks

cd-work: The default is lowercase, since all of these are just aliases it goes back to the field name.

PR #1479, opened 2024-08-06
Update `UserGroup` API type

cd-work:
This updates the response of the `phylum group list --json` command to include all the latest fields of the corresponding API endpoint.

### DIFF
--- a/cli/src/types.rs
+++ b/cli/src/types.rs
@@ -416,9 +416,16 @@ pub struct PurlWithOrigin {
 pub struct UserGroup {
     pub created_at: DateTime<Utc>,
     pub last_modified: DateTime<Utc>,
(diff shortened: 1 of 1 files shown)


cli/src/types.rs
```diff
@@ -416,9 +416,16 @@ pub struct PurlWithOrigin {
 pub struct UserGroup {
     pub created_at: DateTime<Utc>,
     pub last_modified: DateTime<Utc>,
+    pub group_id: Option<String>,
```
cd-work, line 419: https://github.com/phylum-dev/api/pull/1503 technically doesn't break things, it just removes the admin information. So technically we could wait until that PR is in prod and then remove the `Option` here.

kylewillmon (approved):

PR #1481, opened 2024-08-27
Add organization management subcommand

cd-work:
This patch adds the new `phylum org` subcommand, which allows for listing organizations (`phylum org list`) and managing organization members (`phylum org member`).

Since the organization will likely be relevant for most of the subcommands in the future, a global `-o`/`--org` option has been added to allow for easily overriding the target operation.

Most users will likely always operate under the same organization, so the `phylum org link` and `phylum org unlink` subcommands were added to allow for managing a default organization for all subcommands.

### DIFF
--- a/cli/src/commands/org.rs
+++ b/cli/src/commands/org.rs
@@ -146,11 +146,23 @@ pub async fn handle_member_remove(
 
 /// Prompt user for org selection.
 async fn prompt_org(api: &PhylumApi) -> Result<Option<String>> {
-    // Check if user is part of any organizations.
+    // Get orgs the user is a member of.
     let orgs_response = api.orgs().await?;
-    if orgs_response.organizations.is_empty() {
-        print_user_failure!("User is not part of any organizations");
-        return Ok(None);
+    match orgs_response.organizations.as_slice() {
+        // Print error if user is not part of any orgs.
+        [] => {
+            print_user_failure!("User is not part of any organizations");
+            return Ok(None);
+        },
+        // Short-circuit if user is only part of one org.
+        [org] => {
+            print_user_success!(
+                "Automatically selected the only organization you are a member of: {:?}",
+                org.name
+            );
+            return Ok(Some(org.name.clone()));
+        },
+        _ => (),
     }
 
     // Ask user to select one of their orgs.
--- a/cli/src/commands/org.rs
+++ b/cli/src/commands/org.rs
@@ -1,0 +1,133 @@
+//! Subcommand `phylum org`.
+
+use clap::ArgMatches;
+use reqwest::StatusCode;
+
+use crate::api::PhylumApi;
+use crate::commands::{CommandResult, ExitCode};
+use crate::config::Config;
+use crate::format::Format;
+use crate::{print_user_failure, print_user_success, print_user_warning};
+
+/// Handle `phylum org` subcommand.
+pub async fn handle_org(api: &PhylumApi, matches: &ArgMatches, config: Config) -> CommandResult {
+    match matches.subcommand() {
+        Some(("list", matches)) => handle_org_list(api, matches).await,
+        Some(("link", matches)) => handle_org_link(api, matches, config).await,
+        Some(("unlink", _)) => handle_org_unlink(config).await,
+        Some(("member", sub_matches)) => {
+            let org = match config.org() {
+                Some(org) => org,
+                None => {
+                    print_user_failure!("Arguments missing target organization");
+                    print_user_failure!("Either add `--org` flag, or update your configuration:");
+                    print_user_failure!("    phylum org link <ORG>");
+                    return Ok(ExitCode::MissingOrg);
+                },
+            };
+
+            match sub_matches.subcommand() {
+                Some(("list", sub_matches)) => handle_member_list(api, sub_matches, org).await,
+                Some(("add", sub_matches)) => handle_member_add(api, sub_matches, org).await,
+                Some(("remove", sub_matches)) => handle_member_remove(api, sub_matches, org).await,
+                _ => unreachable!("invalid clap configuration"),
+            }
+        },
+        _ => unreachable!("invalid clap configuration"),
+    }
+}
+
+/// Handle `phylum org list` subcommand.
+pub async fn handle_org_list(api: &PhylumApi, matches: &ArgMatches) -> CommandResult {
+    let response = api.orgs().await?;
+
+    let pretty = !matches.get_flag("json");
+    response.write_stdout(pretty);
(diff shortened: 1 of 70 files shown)


cd-work:
Converted to draft for now since @kylewillmon has made the great suggestion to allow for interactive org selection if no org is specified for `phylum org link`.

Might even be worth considering if we should add that interactive org selection when no org is configured as default or provided through `-o`.

cli/src/format.rs
```diff
@@ -19,10 +19,14 @@ use vulnreach_types::Vulnerability;
...
 use crate::types::{
-    HistoryJob, Issue, ListUserGroupsResponse, Package, PolicyEvaluationResponse,
-    PolicyEvaluationResponseRaw, RiskLevel, UserGroup, UserToken,
+    HistoryJob, Issue, ListUserGroupsResponse, OrgMember, OrgMembersResponse, OrgsResponse,
+    Package, PolicyEvaluationResponse, PolicyEvaluationResponseRaw, RiskLevel, UserGroup,
+    UserToken,
 };
 
+// Maximum length of email column.
+const MAX_EMAIL_WIDTH: usize = 25;
```
maxrake, line 28: With the understanding that this is a refactor of an existing constant (pulling it up), this width seems low. AFAICT, bumping this to 35 (or even as high as 46) will still limit the total MAX width of the largest use cases to 80.

cli/src/commands/org.rs
```diff
@@ -0,0 +1,133 @@
...
+    match matches.subcommand() {
+        Some(("list", matches)) => handle_org_list(api, matches).await,
+        Some(("link", matches)) => handle_org_link(api, matches, config).await,
+        Some(("unlink", _)) => handle_org_unlink(config).await,
+        Some(("member", sub_matches)) => {
+            let org = match config.org() {
+                Some(org) => org,
+                None => {
+                    print_user_failure!("Arguments missing target organization");
+                    print_user_failure!("Either add `--org` flag, or update your configuration:");
```
maxrake, line 23:
```suggestion
                    print_user_failure!("Missing target organization");
                    print_user_failure!("Either add `--org` flag, or update configuration:");
```
The first time I encountered this error message, I didn't quite parse it correctly. Namely, that there is an option to add a missing argument and a different option to update the user configuration to set the org globally. This is just a suggestion for making it a bit more clear and shorter.

cli/src/commands/org.rs
```diff
@@ -0,0 +1,133 @@
...
+    let users = matches.get_many::<String>("user").unwrap();
+
+    for user in users {
+        match api.org_member_add(org, user).await {
+            Ok(()) => print_user_success!("Successfully added {user:?} to organization {org:?}"),
+            Err(err) if err.status() == Some(StatusCode::CONFLICT) => {
+                print_user_warning!("User {user:?} is already a member of organization {org:?}");
+                return Ok(ExitCode::AlreadyExists);
+            },
+            Err(err) => return Err(err.into()),
```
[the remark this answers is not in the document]
mathew-horner: Honestly, this is probably an insufficiency of the API. Our error responses are not the greatest.

cli/src/format.rs
cd-work: The appropriate email width has been determined to be 25 in all currently relevant cases. There's no reason to change it when this is the best value for what it is used for.

doc_templates/phylum_org_link.md
```diff
@@ -0,0 +1,10 @@
+{PH-HEADER}
+
+{PH-MARKDOWN}
+
+## Examples
+
+```sh
+# Set `sample` as the default organization for all future operations
+$ phylum org link sample
```
maxrake, line 9:
```suggestion
# Interactively select an organization for all future operations
$ phylum org link

# Set `sample` as the default organization for all future operations
$ phylum org link sample
```
This is just a suggestion for making this example look more like the one in `doc_templates/phylum_init.md`. The same change _could_ be made for each of the other `org` commands that allows for interactively selecting the organization, but that is likely too much. `phylum org link` feels different, though, in this respect.

maxrake (dismissed):
The code was reviewed and the changes were tested locally. The new commands all appear to work as intended. The change to allow for selecting an org when one is not provided is really nice. One possible UX improvement to that might be to automatically use the org when only one org is returned. If so, it might also be worth adding the org name used to the output so there is nothing hidden from the user.

The remaining comments provided are all suggestions and this could go out like it is. LGTM.

cli/src/config.rs
```diff
@@ -56,17 +56,23 @@ pub struct Config {
     ignore_certs_cli: bool,
     #[serde(deserialize_with = "default_option_bool")]
     ignore_certs: bool,
+    #[serde(rename = "organization")]
+    org: Option<String>,
+    #[serde(skip)]
+    org_cli: Option<String>,
```
mathew-horner, line 62: nit: Seems a bit strange that this struct mixes values that are persisted between runs and values that are provided as arguments to the CLI, but not that big of a deal.

mathew-horner (dismissed):
Few questions / optional nits. Otherwise LGTM

cli/src/commands/org.rs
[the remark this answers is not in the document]
cd-work: Yes.

cli/src/config.rs
```diff
@@ -56,17 +56,23 @@ pub struct Config {
     ignore_certs_cli: bool,
     #[serde(deserialize_with = "default_option_bool")]
     ignore_certs: bool,
+    #[serde(rename = "organization")]
+    org: Option<String>,
```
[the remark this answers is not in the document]
cd-work: This code consistently refers to organizations with the `org` shorthand. I'd like to keep that consistent if possible so I decided to go for a rename here.

The alternative would be using `organization` everywhere, but I think that due to the length and the ubiquity of the `org` shorthand, it's better to just have `org` as the canonical form for anything that isn't directly user-facing.

cli/src/config.rs
cd-work: This is already true for the certs stuff. I'd argue it's not that unusual to keep layered non-persistent configuration options stored together with the persistent information. This way users cannot accidentally ignore the config override which makes dealing with this sort of layered configuration significantly easier.

maxrake (dismissed):
LGTM

maxrake (approved):
The most recent set of changes were reviewed and tested locally. The UX is nice. LGTM.

mathew-horner (approved):
Looks good, but just a heads up: there are some merge conflicts to resolve.

maxrake (approved):
LGTM

PR #1485, opened 2024-08-29
Build without sandboxing on Windows

kylewillmon:
With this patch, all sandboxing code is guarded by `#[cfg(unix)]`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -161,7 +161,7 @@ jobs:
   # This job reports the results of the matrixes above
   test:
     if: always()
-    needs: [clippy, test-matrix, all-features, oldstable]
+    needs: [windows, clippy, test-matrix, all-features, oldstable]
     runs-on: ubuntu-latest
     steps:
       - if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,18 @@ on:
     - cron: '30 5 * * 1'
 
 jobs:
+  windows:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Install Rust toolchain
+        run: rustup toolchain install --no-self-update stable --profile minimal
+
+      - name: Test
+        run: cargo +stable test --workspace --exclude xtask --locked --no-default-features
+
   rustfmt:
     if: github.event_name != 'schedule'
     runs-on: ubuntu-20.04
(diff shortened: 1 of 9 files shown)


lockfile_generator/src/lib.rs
```diff
@@ -157,12 +158,12 @@ impl Display for Error {
...
-                    f,
-                    "package manager quit unexpectedly (code: {:?}, signal: {:?})",
-                    output.status.code(),
-                    output.status.signal()
-                )?;
+                write!(f, "package manager quit unexpectedly (code: {:?}", output.status.code())?;
+                #[cfg(unix)]
+                {
+                    write!(f, ", signal: {:?}", output.status.signal())?;
+                }
```
cd-work, line 165: I'd assume the dedicated block here is necessary because of the way the macro is built? Usually a single statement that isn't an assignment can have cfg directly attached I thought.

Also this is likely cleaner by just printing `None` on Windows. I doubt anyone would mind.

cd-work (changes requested):
I like the choice of making this a platform-specific thing rather than a feature, makes it more about not being possible rather than making it sound like sandboxing is a feature you might want to disable in certain scenarios.

I've got one minor nit but I'm fine with it going either way.

maxrake:
I'm still reviewing...but there are [three warnings on the Windows job](https://github.com/phylum-dev/cli/actions/runs/10623062657/job/29448683382?pr=1485#step:4:568). Do those need to be accounted for here? If so, then enabling warnings as errors should also be pursued.

kylewillmon:
> Do those need to be accounted for here?

They're pretty simple to handle, so sure.

> If so, then enabling warnings as errors should also be pursued.

This might just anger people... So I'd rather not do that for now.

lockfile_generator/src/lib.rs
kylewillmon: I've removed the block. I'm not sure how to print `None` on windows with making the code more complicated... So I've left that bit out.

maxrake:
> > If so, then enabling warnings as errors should also be pursued.
> 
> This might just anger people... So I'd rather not do that for now.

Roger that. Maybe case in point...there are [two new warnings now](https://github.com/phylum-dev/cli/actions/runs/10623206777/job/29449162414?pr=1485#step:4:587).

maxrake (approved):
LGTM

.github/workflows/test.yml
```diff
@@ -13,6 +13,18 @@ on:
     - cron: '30 5 * * 1'
 
 jobs:
+  windows:
```
maxrake, line 16: This new job should be added to the test roll-up job at the end of this workflow:

```
  test:
    if: always()
    needs: [clippy, test-matrix, all-features, oldstable, windows]
```

maxrake (changes requested):
One change requested.

cd-work (approved):

maxrake (approved):
LGTM

PR #1486, opened 2024-08-30
Add windows binary to release

kylewillmon:
This patch adds a windows executable to the release artifacts. This build does not include extensions or sandboxing, so I have left it unsigned and will not be recommending it as a first choice for Windows.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,6 +53,10 @@ jobs:
             os: macos-14
           - target: aarch64-apple-darwin
             os: macos-14
+          - target: x86_64-pc-windows-msvc
+            os: windows-latest
+            extension: .exe
+            flags: --no-default-features
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout the repo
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -68,20 +72,25 @@ jobs:
...
+      - if: ${{ runner.os == 'Windows' }}
+        name: Install Rust toolchain
+        run: rustup toolchain install stable --profile minimal --target ${{ matrix.target }}
+
+      - if: ${{ runner.os != 'Windows' }}
+        name: Install Rust toolchain
         run: |
           curl https://sh.rustup.rs -sSf \
             | sh -s -- -y --default-toolchain stable --profile minimal --target ${{ matrix.target }}
           echo "$HOME/.cargo/bin" >> "$GITHUB_PATH"
```
mathew-horner, line 84: It looks like the non-Windows command downloads Rustup before invoking it, while the Windows command just invokes it. How does that work?

kylewillmon: The only reason we need to download rustup is because it's not available on our self-hosted runners. All the GitHub runners already have it installed

maxrake:
> This build does not include extensions or sandboxing, so I have left it unsigned and will not be recommending it as a first choice for Windows.

Are there other choices coming for Windows? Or, does this just mean the current suggestion for Windows installation will remain the first choice:

> On Windows, we recommend using the Linux binaries under [WSL](https://learn.microsoft.com/en-us/windows/wsl/)

.github/workflows/release.yml
```diff
@@ -116,12 +125,19 @@ jobs:
...
+          done
 
       - name: Upload release artifacts
         uses: actions/upload-artifact@834a144ee995460fba8ed112a2fc961b36a5ec5a # v4.3.6
         with:
           name: release-archives
-          path: phylum-*.zip
+          path: |
+            phylum-*.zip
+            phylum-*.exe
```
maxrake, line 140: Because of this new name/extension, the Windows artifact won't be included in the release assets. The `Upload release assets` step of the `Release` job will need to be updated if that is the goal.

maxrake (approved):
Approving for now in case time is of the essence, but there is a suquestion that should still be addressed.

maxrake (approved):
LGTM

Issue #1482, opened 2024-08-28
Org support for non-org subcommands


cd-work:
In https://github.com/phylum-dev/cli/pull/1481 initial support for org management itself was added to the CLI, however none of the other subcommands were changed to support organizations.

The following subcommands will likely need to have org support added explicitly:
 - analyze (#1490)
 - group (https://github.com/phylum-dev/cli/pull/1488)
 - history (#1490)
 - init (#1490)
 - project status/create/delete/update/list/link (#1490)

PR #1488, opened 2024-09-04
Add org support to `phylum group`

cd-work:
This patch adds org support to the `list`/`create`/`delete` subcommands of `phylum group`. The `member` subcommand is unchanged, since user management is performed on the org level instead.

The `create` and `delete` subcommands will always use the currently linked organization, with no way to temporarily override it. While this could be annoying to some, it avoids adding extra flags that would get deprecated in the future. Most users likely always or never work within organizations and for people working in-between it's easiest to avoid linking orgs and just using `--org` instead as required.

See #1482.

### DIFF
--- a/cli/src/commands/group.rs
+++ b/cli/src/commands/group.rs
@@ -40,7 +40,8 @@ pub async fn handle_group_create(
 ) -> CommandResult {
     let group_name = matches.get_one::<String>("group_name").unwrap();
 
-    let response = if let Some(org) = config.org() {
+    let org = config.org();
+    let response = if let Some(org) = org {
         api.org_create_group(org, group_name).await
     } else {
         api.create_group(group_name).await.map(|_| ())
@@ -55,6 +56,12 @@ pub async fn handle_group_create(
             print_user_failure!("Group '{}' already exists", group_name);
             Ok(ExitCode::AlreadyExists)
         },
+        Err(PhylumApiError::Response(ResponseError { code: StatusCode::FORBIDDEN, .. }))
+            if org.is_some() =>
+        {
+            print_user_failure!("Authorization failed, only organization admins can create groups");
+            Ok(ExitCode::NotAuthenticated)
+        },
         Err(err) => Err(err.into()),
     }
 }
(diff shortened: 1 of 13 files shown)


cli/src/commands/group.rs
```diff
@@ -29,11 +33,22 @@ pub async fn handle_group(api: &PhylumApi, matches: &ArgMatches) -> CommandResul
...
+    } else {
+        api.create_group(group_name).await.map(|_| ())
+    };
+
+    match response {
+        Ok(_) => {
+            print_user_success!("Successfully created group {}", group_name);
             Ok(ExitCode::Ok)
         },
         Err(PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. })) => {
```
maxrake, line 54: I still think it would be a nicer UX to make it clear that org groups can't be created if you are not the org admin. The current output in that case is not great for users:

```
❯ target/debug/phylum -c ~/.config/phylum/staging.settings.yaml group create --org phylum delme_org_group_create_test
❗ Error: HTTP request error (403 Forbidden):

access forbidden, insufficient privileges, re-authentication will not help
```

Is a change required in API or could this be improved here?

maxrake (changes requested):
The code was reviewed and tested locally. It appears to function as intended. There is an open question and a CHANGELOG entry is requested.

Additionally, it is requested that documentation be updated to reflect some of these new changes and idiosyncrasies between subcommands. Adding examples or details to the relevant `doc_templates` is likely the easiest path.

cli/src/commands/group.rs
cd-work: Org groups are currently always accessible to all members as far as I know. So I think in the scenario where this returns a 403 we can be certain that the issue is due to not being an admin.

kylewillmon: @maxrake Can you open an API ticket for this?

Even if CLI produces a better message, the API response is still lacking.

maxrake (approved):
Reviewed the code changes and tested locally. LGTM.

PR #1490, opened 2024-09-06
Add org support for init/project/analyze/history

cd-work:
This patch uses the `Config::org` function to add support for org
support in all subcommands that were still missing org support.

Generally when an org is linked, the org mode is always preferred. Only
in scenarios where no group is present at all, like `phylum project
delete <NAME>`, the personal projects will be used instead.

Since this could potentially lead to a lot of confusion with the project
management subcommands, all project subcommands have been changed to
print projects in a fixed format that lists the corresponding org and
group:

```
✅ Successfully deleted project cli (org: phylum, group: org-group)
```

```
✅ Successfully deleted project demo (org: -, group: -)
```

Closes https://github.com/phylum-dev/cli/issues/1482.

### DIFF
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -261,14 +261,21 @@ impl PhylumApi {
     pub async fn create_project(
         &self,
         name: impl Into<String>,
+        org: Option<&str>,
         group: Option<String>,
         repository_url: Option<String>,
     ) -> Result<ProjectId> {
+        let group_name = match (org, group) {
+            (Some(org), Some(group)) => Some(format!("{org}/{group}")),
+            (None, Some(group)) => Some(group),
+            (Some(_), None) | (None, None) => None,
+        };
+
         let url = endpoints::create_project(&self.config.connection.uri)?;
         let body = CreateProjectRequest {
             repository_url,
+            group_name,
             default_label: None,
-            group_name: group,
             name: name.into(),
         };
         let response: CreateProjectResponse = self.post(url, body).await?;
@@ -279,18 +286,21 @@ impl PhylumApi {
     pub async fn update_project(
         &self,
         project_id: &str,
+        org: Option<String>,
         group: Option<String>,
         name: impl Into<String>,
         repository_url: Option<String>,
         default_label: Option<String>,
     ) -> Result<ProjectId> {
-        let url = endpoints::project(&self.config.connection.uri, project_id)?;
-        let body = UpdateProjectRequest {
-            repository_url,
-            default_label,
-            name: name.into(),
-            group_name: group,
+        let group_name = match (org, group) {
+            (Some(org), Some(group)) => Some(format!("{org}/{group}")),
+            (None, Some(group)) => Some(group),
+            (Some(_), None) | (None, None) => None,
         };
+
+        let url = endpoints::project(&self.config.connection.uri, project_id)?;
+        let body =
+            UpdateProjectRequest { repository_url, default_label, name: name.into(), group_name };
         let response: CreateProjectResponse = self.put(url, body).await?;
         Ok(response.id)
     }
@@ -312,14 +322,24 @@ impl PhylumApi {
     /// equivalent to filtering with [`str::contains`].
     pub async fn get_projects(
         &self,
+        org: Option<&str>,
         group: Option<&str>,
         name_filter: Option<&str>,
     ) -> Result<Vec<ProjectListEntry>> {
         let mut uri = endpoints::projects(&self.config.connection.uri)?;
 
         // Add filter query parameters.
-        if let Some(group) = group {
-            uri.query_pairs_mut().append_pair("filter.group", group);
+        match (org, group) {
+            (Some(org), Some(group)) => {
+                uri.query_pairs_mut().append_pair("filter.group", &format!("{org}/{group}"));
+            },
+            (Some(org), None) => {
+                uri.query_pairs_mut().append_pair("filter.organization", org);
+            },
+            (None, Some(group)) => {
+                uri.query_pairs_mut().append_pair("filter.group", group);
+            },
+            (None, None) => (),
         }
         if let Some(name_filter) = name_filter {
             uri.query_pairs_mut().append_pair("filter.name", name_filter);
@@ -415,19 +435,11 @@ impl PhylumApi {
     pub async fn get_project_history(
         &self,
         project_name: &str,
+        org_name: Option<&str>,
         group_name: Option<&str>,
     ) -> Result<Vec<HistoryJob>> {
-        let project_id = self.get_project_id(project_name, group_name).await?.to_string();
-
-        let url = match group_name {
-            Some(group_name) => endpoints::get_group_project_history(
-                &self.config.connection.uri,
-                &project_id,
-                group_name,
-            )?,
-            None => endpoints::get_project_history(&self.config.connection.uri, &project_id)?,
-        };
-
+        let project_id = self.get_project_id(project_name, org_name, group_name).await?.to_string();
+        let url = endpoints::get_project_history(&self.config.connection.uri, &project_id)?;
         self.get(url).await
     }
 
@@ -435,9 +447,10 @@ impl PhylumApi {
     pub async fn get_project_id(
         &self,
         project_name: &str,
+        org_name: Option<&str>,
         group_name: Option<&str>,
     ) -> Result<ProjectId> {
-        let projects = self.get_projects(group_name, Some(project_name)).await?;
+        let projects = self.get_projects(org_name, group_name, Some(project_name)).await?;
 
         projects
             .iter()
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::HashSet;
 use std::time::Duration;
 
@@ -26,8 +27,8 @@ use crate::auth::{
 };
 use crate::config::{AuthInfo, Config};
 use crate::types::{
-    AddOrgUserRequest, AnalysisPackageDescriptor, CreateProjectRequest, GetProjectResponse,
-    HistoryJob, ListUserGroupsResponse, OrgGroup, OrgGroupsResponse, OrgMembersResponse,
+    AddOrgUserRequest, AnalysisPackageDescriptor, ApiOrgGroup, CreateProjectRequest,
+    GetProjectResponse, HistoryJob, ListUserGroupsResponse, OrgGroupsResponse, OrgMembersResponse,
     OrgsResponse, PackageSpecifier, PackageSubmitResponse, Paginated, PingResponse,
     PolicyEvaluationRequest, PolicyEvaluationResponse, PolicyEvaluationResponseRaw,
     ProjectListEntry, RevokeTokenRequest, SubmitPackageRequest, UpdateProjectRequest, UserToken,
@@ -435,10 +436,9 @@ impl PhylumApi {
     pub async fn get_project_history(
         &self,
         project_name: &str,
-        org_name: Option<&str>,
-        group_name: Option<&str>,
+        group: Option<Group>,
     ) -> Result<Vec<HistoryJob>> {
-        let project_id = self.get_project_id(project_name, org_name, group_name).await?.to_string();
+        let project_id = self.get_project_id(project_name, group).await?.to_string();
         let url = endpoints::get_project_history(&self.config.connection.uri, &project_id)?;
         self.get(url).await
     }
@@ -447,14 +447,17 @@ impl PhylumApi {
     pub async fn get_project_id(
         &self,
         project_name: &str,
-        org_name: Option<&str>,
-        group_name: Option<&str>,
+        group: Option<Group>,
     ) -> Result<ProjectId> {
-        let projects = self.get_projects(org_name, group_name, Some(project_name)).await?;
+        let (org, group, combined_format) = group.as_ref().map_or((None, None, None), |group| {
+            (group.org(), Some(group.name()), Some(group.combined_format().into_owned()))
+        });
+
+        let projects = self.get_projects(org, group, Some(project_name)).await?;
 
         projects
             .iter()
-            .find(|project| project.name == project_name)
+            .find(|project| project.name == project_name && project.group_name == combined_format)
             .ok_or_else(|| anyhow!("No project found with name {:?}", project_name).into())
             .map(|project| project.id)
     }
@@ -517,7 +520,7 @@ impl PhylumApi {
     /// Create a new organization group.
     pub async fn org_create_group(&self, org_name: &str, group_name: &str) -> Result<()> {
         let url = endpoints::org_groups(&self.config.connection.uri, org_name)?;
-        let body = OrgGroup { name: group_name.into() };
+        let body = ApiOrgGroup { name: group_name.into() };
         self.send_request_raw(Method::POST, url, Some(body)).await?;
         Ok(())
     }
@@ -584,6 +587,66 @@ impl PhylumApi {
     pub fn config(&self) -> &Config {
         &self.config
     }
+}
+
+/// Phylum group types.
+pub enum Group {
+    Legacy(String),
+    Org(OrgGroup),
+}
+
+impl Group {
+    /// Create a new group.
+    fn new(org: String, group: String) -> Self {
+        Self::Org(OrgGroup { org, name: group })
+    }
+
+    /// Create a group without organization.
+    fn new_legacy(group: String) -> Self {
+        Self::Legacy(group)
+    }
+
+    /// Create a group from an optional group name.
+    pub fn try_new<O, G>(org: Option<O>, group: Option<G>) -> Option<Self>
+    where
+        O: Into<String>,
+        G: Into<String>,
+    {
+        group.map(|group| match org {
+            Some(org) => Self::new(org.into(), group.into()),
+            None => Self::new_legacy(group.into()),
+        })
+    }
+
+    /// Get the group's organization.
+    pub fn org(&self) -> Option<&str> {
+        match self {
+            Self::Legacy(_) => None,
+            Self::Org(OrgGroup { org, .. }) => Some(org.as_str()),
+        }
+    }
+
+    /// Get the group's name.
+    pub fn name(&self) -> &str {
+        match self {
+            Self::Legacy(name) => name.as_str(),
+            Self::Org(OrgGroup { name, .. }) => name.as_str(),
+        }
+    }
+
+    /// Format the group as a single string.
+    pub fn combined_format(&self) -> Cow<'_, str> {
+        match self {
+            Self::Legacy(name) => Cow::Borrowed(name),
+            Self::Org(OrgGroup { org, name }) => Cow::Owned(format!("{org}/{name}")),
+        }
+    }
+}
+
+/// Group under an organization.
+pub struct OrgGroup {
+    org: String,
+    name: String,
 }
 
 /// Tests
--- a/cli/src/api/mod.rs
+++ b/cli/src/api/mod.rs
@@ -596,16 +596,6 @@ pub enum Group {
 }
 
 impl Group {
-    /// Create a new group.
-    fn new(org: String, group: String) -> Self {
-        Self::Org(OrgGroup { org, name: group })
-    }
-
-    /// Create a group without organization.
-    fn new_legacy(group: String) -> Self {
-        Self::Legacy(group)
-    }
-
     /// Create a group from an optional group name.
     pub fn try_new<O, G>(org: Option<O>, group: Option<G>) -> Option<Self>
     where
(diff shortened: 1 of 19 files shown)


cli/src/api/mod.rs
```diff
@@ -415,33 +436,28 @@ impl PhylumApi {
...
     pub async fn get_project_id(
         &self,
         project_name: &str,
-        group_name: Option<&str>,
+        group: Option<Group>,
     ) -> Result<ProjectId> {
-        let projects = self.get_projects(group_name, Some(project_name)).await?;
+        let (org, group, combined_format) = group.as_ref().map_or((None, None, None), |group| {
+            (group.org(), Some(group.name()), Some(group.combined_format().into_owned()))
+        });
```
matt-phylum, line 454: It seems like it would be easier to understand if we could write

```rust
let org = group.org();
let group = group.group(); // this naming isn't right
let group_path = group.combined_format();
```

The group parameter contains the organization-qualified group name (the group path?), but it's just called "group" and then another variable is defined named "group" which contains just the unqualified name of the group. It seems like the second "group" should be "group_name", but in the other functions "group_name" contains the qualified group name, which is here called "combined_format" instead.

cli/src/api/mod.rs
```diff
@@ -573,6 +589,66 @@ impl PhylumApi {
...
 }
 
+/// Phylum group types.
+pub enum Group {
+    Legacy(String),
+    Org(OrgGroup),
+}
+
+impl Group {
+    /// Create a new group.
```
matt-phylum, line 599: This function doesn't create a new group. It creates a `Group` value referencing an organization group.

cli/src/api/mod.rs
```diff
@@ -573,6 +589,66 @@ impl PhylumApi {
...
+    pub fn try_new<O, G>(org: Option<O>, group: Option<G>) -> Option<Self>
+    where
+        O: Into<String>,
+        G: Into<String>,
+    {
+        group.map(|group| match org {
+            Some(org) => Self::new(org.into(), group.into()),
+            None => Self::new_legacy(group.into()),
+        })
+    }
```
matt-phylum, line 619: Should these be "from" functions instead of "new" functions? They're just moving the input parameters into enum variants. Maybe we don't need `new` or `new_legacy` at all. I also wonder, if we're not having a `None` variant, would it be better to just have `struct Group { organization: Option<String>, name: String }` instead of an enum.

cli/src/api/mod.rs
```diff
@@ -573,6 +589,66 @@ impl PhylumApi {
...
+        }
+    }
+
+    /// Format the group as a single string.
+    pub fn combined_format(&self) -> Cow<'_, str> {
+        match self {
+            Self::Legacy(name) => Cow::Borrowed(name),
+            Self::Org(OrgGroup { org, name }) => Cow::Owned(format!("{org}/{name}")),
+        }
+    }
```
matt-phylum, line 643: Here the group name is the group name and the group path is the combined format, but in functions like `create_project` the "combined format" value is used as the group name.

cli/src/api/mod.rs
```diff
@@ -573,6 +589,66 @@ impl PhylumApi {
...
+}
+
+/// Group under an organization.
+pub struct OrgGroup {
+    org: String,
+    name: String,
+}
```
matt-phylum, line 650: Does this struct have any value? It seems like it's intended to be used as a private implementation detail of `Group`. Its members are always accessed through `Group`. It'd probably be better to just put the values into the variant unless we're planning to have a function that returns an `Option<OrgGroup>`.

cli/src/commands/extensions/api.rs
```diff
@@ -168,12 +168,16 @@ async fn analyze(
...
     #[string] label: Option<String>,
+    #[string] organization: Option<String>,
 ) -> Result<JobId> {
     let state = ExtensionState::from(op_state);
     let api = state.api().await?;
 
     let (project, group) = match (project, group) {
-        (Some(project), group) => (api.get_project_id(&project, group.as_deref()).await?, None),
+        (Some(project), group) => {
+            let org_group = Group::try_new(organization, group);
```
matt-phylum, line 178: This value is called `org_group`, but it may be a legacy group.

cli/src/commands/extensions/api.rs
```diff
@@ -356,18 +361,21 @@ async fn create_project(
...
-    match api.create_project(&name, group.clone(), repository_url).await {
-        Err(PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. })) => api
-            .get_project_id(&name, group.as_deref())
-            .await
-            .map(|id| CreatedProject { id, status: CreatedProjectStatus::Exists })
-            .map_err(|e| e.into()),
+    match api.create_project(&name, organization.as_deref(), group.clone(), repository_url).await {
+        Err(PhylumApiError::Response(ResponseError { code: StatusCode::CONFLICT, .. })) => {
+            let org_group = Group::try_new(organization, group);
+            api.get_project_id(&name, org_group)
```
matt-phylum, line 374: It seems weird here that `create_project` takes organization and group strings, but `get_project_id` takes a `Group`. They should probably all take `&Group`?

cli/src/commands/project.rs
```diff
@@ -143,42 +76,108 @@ pub async fn status(api: &PhylumApi, matches: &ArgMatches) -> StdResult<(), Phyl
...
     project.write_stdout(pretty_print);
 
-    Ok(())
+    Ok(ExitCode::Ok)
+}
+
+/// List the projects in this account.
+async fn handle_list_projects(api: &PhylumApi, matches: &ArgMatches) -> CommandResult {
+    let group = matches.get_one::<String>("group").map(|g| g.as_str());
+    let org = matches.get_one::<String>("org").map(|o| o.as_str());
```
matt-phylum, line 85: What happens if the user receives a "org/group" format group name from an API call or something and then passes it as a group name to a command? It seems like we're not expecting it but it will _sometimes_ work anyway because _some_ of the commands only need the "org/group" form and we never check whether "group" is actually "org/group". Probably if the group parameter contains an organization, that should be used instead of the organization from the configuration.

cli/src/format.rs
```diff
@@ -176,8 +176,19 @@ impl Format for Vec<ProjectListEntry> {
...
+                    match project.group_name.as_deref().unwrap_or("").split_once('/') {
+                        Some((org_name, _)) => {
+                            print::truncate(org_name, MAX_NAME_WIDTH).into_owned()
+                        },
+                        None => String::new(),
+                    }
+                }),
                 ("Group Name", |project| {
-                    let group_name = project.group_name.as_deref().unwrap_or("");
+                    let org_and_group = project.group_name.as_deref().unwrap_or("");
```
matt-phylum, line 188: I'm surprised clippy isn't complaining about not using `unwrap_or_default()`.

extensions/phylum.d.ts
```diff
@@ -112,6 +112,7 @@ declare namespace Phylum {
     project?: string,
     group?: string,
     label?: string,
+    organization?: string,
```
matt-phylum, line 115: This is pretty ugly. `analyze(packages, project, group)` becomes `analyze(packages, project, group, null, organization)`. It seems common in ECMAScript that we'd have a `analyze(packages: (PackageWithOrigin | PurlWithOrigin)[], options: { project?: string, group?: string, label?: string, organization?: string })` form to write `analyze(packages, { project: project, group: group, organization: organization })` without having to count positional null parameters.

Organization isn't added to the documentation. I assume both group and organization are defaulted from `.phylum_project` if not specified, and then the organization is defaulted from the CLI configuration if it's not in `.phylum_project` either? Or is it that `group` and `organization` are required if `project` is specified and ignored when it's not specified (because we're using a project ID instead of a project name?)?

cli/src/api/mod.rs
cd-work: Your suggestion doesn't work, it's just an entirely different thing and does not behave the same way as mine.

Having `group_name` over `group` doesn't really improve things. It's just redundant.

cli/src/api/mod.rs
cd-work: I don't think this is particularly misleading on a constructor. It does create a new `Group`, that's exactly what a constructor does.

cli/src/api/mod.rs
cd-work: Yes, these should be `new` functions. 'From' impls on random tuples or strings don't make sense. I don't think changing anything here would improve things.

cli/src/api/mod.rs
cd-work: Yes, because those are API functions and `group_name` is what the API calls it for now. That behavior is not exposed outside of those functions at all so I don't see why it would be an issue?

cli/src/api/mod.rs
cd-work: Moving structs outside of enums is **always** the better option. Anonymous structs in enums are strictly less powerful and should be avoided.

cli/src/commands/extensions/api.rs
cd-work: It's a combination of org and group, where org may be `None`. The naming distinguishes a group from a group stored together with its org and as such I don't see any issue here. I've **intentionally** made this change after having it as `group` initially and I don't think going back to just group would improve things.

cli/src/api/mod.rs
matt-phylum: It's less powerful, but also less verbose. We aren't using any of the additional capabilities.

cli/src/commands/extensions/api.rs
cd-work: There's no necessity for `Group` in the `create_project` endpoint, so for the sake of simplicity it was avoided. While it would be fine to use `Group` here, I don't think there's any benefit to changing it other than introducing additional code on the callsite. So that would effectively just be a net negative?

cli/src/commands/project.rs
cd-work: > What happens if the user receives a "org/group" format group name from an API

They don't.

> It seems like we're not expecting it but it will sometimes work anyway because some of the commands only need the "org/group" form and we never check whether "group" is actually "org/group". Probably if the group parameter contains an organization, that should be used instead of the organization from the configuration.

If it sometimes works, there's not really an issue with it? I see no point in supporting a format that we explicitly do **not** want to show to the user.

cli/src/format.rs
cd-work: It's just much shorter like this, `unwrap_or_default` was kind of a mistake anyway.

extensions/phylum.d.ts
cd-work: I've added it to the docs. I also don't like this format, but it's the best choice for avoiding issues with existing implementations.

cli/src/commands/project.rs
matt-phylum: Isn't that the format we return from the API, in which case it will also be the format returned when using the JSON output?

cd-work: It is, for now. It's never going to be returned in production and a lot of this code is going to go away in the future.

cli/src/api/mod.rs
cd-work: Not *yet*. Verbosity rarely matters much, but the additional capabilities are important and quite difficult to adopt at a later point in time.

cli/src/commands/project.rs
matt-phylum: Will it really go away? We're passing this value as a parameter when filtering the project listing.

cli/src/api/mod.rs
matt-phylum: It creates a new `Group`, but not a new group, which is implemented by a different function.

cli/src/api/mod.rs
matt-phylum: I'd rather `Group::from_legacy` or `Group::legacy` or just `Group::Legacy` over `Group::new_legacy`.

cli/src/api/mod.rs
matt-phylum: The code is difficult to follow because we use the same few variable names, "group" "group name" "org group," in different places to refer to the name of the group, the organization/group path of the group (which may be just the group name for legacy groups), and a struct containing the name of the organization and group names. They're all overloaded names and they're all roughly the same concept so context doesn't help.

cli/src/commands/extensions/api.rs
matt-phylum: In the code for an extension to create a group, it takes the organization and group as separate strings and passes them to one function as separate strings and another function as a `Group`.

In general, I think it's better for two related values like this to be the same parameter so there are less parameters.

There are several places that duplicate the (fairly trivial) `combined_format` code when working with the organization and group as separate parameters.

extensions/phylum.d.ts
matt-phylum: The documentation doesn't specify what happens if you don't provide an organization name, and the documentation hasn't been updated for any of the other functions that got an organization parameter.

It's possible in Typescript to overload the function signature, but since it's just Javascript it doesn't actually overload the function. The callee needs to expect to receive either four optional strings or a single object, which seems like a pain with Deno op2. I guess since we're already using `#[serde]` for `packages` we could use an untagged enum to handle `project_or_options` being a string or object and then throw if `project_or_options` is the options object and any of `group` `label` `organization` are set (if we even allow passing a positional organization this way). That allows named options without breaking existing callers.

cli/src/api/mod.rs
cd-work: I don't think that would surprise anyone? This function is not taking an API argument? I can remove the doc comment but I don't really see the issue.

cli/src/api/mod.rs
cd-work: `Group::Legacy` or `Group::legacy` are both not really an option, since they stray from Rust standards.

`Group::from_legacy` implies what you pass is a legacy group, which could be argued but really at that point you'd expect it to be `Group::Legacy` as parameter.

Because of that `Group::new_legacy` is the only option that is both accurate and follows common rust patterns.

cli/src/api/mod.rs
cd-work: That is correct, but that is also largely due to the way it was initially implemented and it will change in the future.

Generally I think calling this combined format `org_group` while `group` is just the group and `org` is just the org makes sense and that's what I've tried to use in the places this new struct has been used so far.

cli/src/commands/extensions/api.rs
cd-work: The combined format part is a valid argument, but this is something that will go away in the future so I don't want to design API boundaries around it.

Less parameters is good when it makes calling something easier. But in this scenario this just means having to create a new group in every place where we'd otherwise just pass the variables in directly otherwise. I'm not sure what we would gain here? At that point I'd rather remove the struct entirely, but that comes with the drawbacks of reduced API guarantees, since it was introduced **after** discovering a bug with the implementation when using separate parameters.

cli/src/commands/project.rs
cd-work: Yes, the combined path format should go away from the API entirely. That is the plan.

extensions/phylum.d.ts
cd-work: > The documentation doesn't specify what happens if you don't provide an organization name, and the documentation hasn't been updated for any of the other functions that got an organization parameter.

It hasn't been updated for any of the other functions because none of them document parameters. I've considered just removing the parameter docs from this function too, but didn't want to do so in this PR.

Not providing an organization name is the status quo, so I don't think it needs to be explicitly documented. That's specifically the goal of these API changes.

> It's possible in Typescript to overload the function signature, but since it's just Javascript it doesn't actually overload the function.

We have basically only string arguments here. Overloading is not an option. The only viable choice that doesn't confuse **everyone** involved is adding a new parameter to the end. It might make sense to rework our API once legacy groups go away to streamline everything a bit, but I don't think it makes sense to perform drastic changes to the extension API in this patch, when there's absolutely no need to.

cli/src/api/mod.rs
matt-phylum: `Group::Legacy(name)` is the Rust standard for enums, and we already have it. I don't see why we would also need `Group::new_legacy(name)` which does exactly the same thing, except maybe to avoid writing `Group::Org(OrgGroup { organization, name })`, which I still believe to be needless complication.

However, neither `Group::new_legacy` nor `Group::new` are ever called. We probably don't need them.

cli/src/api/mod.rs
matt-phylum: It would have been redundant before but I don't think it's redundant now. We're moving from a model where a group is identified by its name to a model where a group is identified by the name of the organization and the name of the group within the organization, which is leading to this problem where "group" can refer to the group or "group" can refer to the unqualified name of the group, which is just an attribute of the group and not enough to identify the group.

I think this is further complicated by the design of the CLI where the organization is being implied in contrast to the backend that the CLI uses where the organization is explicit. On the user-facing side, "group" means the unqualified name of the group within the configured organization, but then in the CLI command handler the unqualified name is upgraded into a qualified name, and from then on "group" _sometimes_ means the qualified group name. In this code we're decomposing the qualified "group" into its parts, and we're replacing `group: QualifiedGroup` with `org: GroupQualifier, group: UnqualifiedGroup`. That's the name overloading problem I have with other parts of this PR.

In this case though, I was more worried about `let (a, b, c) = o.as_ref().map_or((None, None, None), |v| (v.a(), Some(v.b()), Some(v.c().into_owned())));` We could write `let (a, b, c) = o.as_ref().map(|v| (v.a(), Some(v.b()), Some(v.c().into_owned()))).unwrap_or_default();`, but I don't think this is really much better. We could write `let (a, b, c) = match &o { Some(v) => (v.a(), Some(v.b()), Some(v.c().into_owned())), None => (None, None, None) }`. Originally I was thinking it would be nice if the parameter wasn't `Option<Group>` or if `Option<Group>` implemented a trait such that we could obtain information about the group or lack of group without mapping (eg `let org = group.org()`), but I think really what would be best here is if we didn't decompose the qualified group name at all: `let combined_format = group.map(|g| g.combined_format().into_owned());` here and then `let projects = self.get_projects(group, Some(project_name)).await?` later. It simplifies the code here in the caller and also slightly simplifies the code in the callee as well.

I think the `into_owned()` can be avoided by using `project.name == project_name && project.group_name.as_deref() == combined_format.as_deref())` for the filter.

cli/src/commands/extensions/api.rs
cd-work: The APIs being called are different and as such the implication of calling them incorrectly are also different. The places that don't use the `Group` type simply don't have the same issues as the places that do use it.

extensions/phylum.d.ts
cd-work: > However, if we're not documenting the organization name parameter because the caller isn't supposed to provide an organization name, do we even need the organization name parameter?

We don't document any of the parameters in most other functions. Have you checked out the code of this file? We really don't need to overcomplicate things here, I'd much rather keep the changes to the extension API minimal so people can actually keep using the extensions they have.

cli/src/commands/extensions/api.rs
matt-phylum: `Group::try_new(org, group)` is a trivial amount of code that packages two pointers into a single value. The change to the caller is insignificant, all callees have one less positional parameter, callees can use functions implemented for `Group`, callees making API calls can match on an enum instead of a pair of options, and we can't accidentally transpose parameters except for the call to `try_new` which should always be consistent.

I didn't count, but it seems like there are more places where the current code is converting back and forth between `Group` and org+group than there are places that need to create a `Group` from external input.

This commit shows how the code is simplified by using `Group` consistently for functions that deal with specific groups: https://github.com/phylum-dev/cli/commit/b0699ed2378f179fe7f91acf1e4040b067fdba6a

I fixed a bug where the current `handle_update_project` is passing a `None` `org` to `api.update_project`, but a potentially `Some` `org` to all other functions it calls. It was obvious once the code was changed to use `Group`.

cli/src/commands/project.rs
```diff
@@ -258,6 +257,7 @@ pub async fn update_project(
 
     api.update_project(
         &project_id,
+        None,
```
matt-phylum, line 255:
```suggestion
        org,
```

matt-phylum (approved):

matt-phylum (approved):

cli/src/commands/extensions/api.rs
cd-work: > This commit shows how the code is simplified by using Group consistently for functions that deal with specific groups: https://github.com/phylum-dev/cli/commit/b0699ed2378f179fe7f91acf1e4040b067fdba6a

It really doesn't. Inlining code that never should be inlined, just goes to show that it clearly causes additional friction:

```
let group = Group::try_new(config.org(), matches.get_one::<String>("group"));
```

The main "benefit" you're gaining here is in reusing the functions for the combined path, which are irrelevant since they're not going to be there long-term anyway.

PR #1493, opened 2024-09-10
Remove unneeded install in release workflow

kylewillmon:
Installing `cmake` is unnecessary and just causes [warnings][1] on our release workflow. The patch removes it from the workflow.

[1]: https://github.com/phylum-dev/cli/actions/runs/10798611911


kylewillmon:
I've manually kicked off [this workflow](https://github.com/phylum-dev/cli/actions/runs/10799343560) to prove that this still works.

PR #1495, opened 2024-09-11
Update `yarn` extension sandbox exception

maxrake:
This change allows for local project directory write access
for the building stage of the `yarn` extension. This brings
it into parity with the rest of the package manager extensions.

The need for this change was discovered by a community edition
user who reported in Discord how the extension was failing due
to a sandbox access violation:

```
Error: EROFS: read-only file system, rmdir '/home/sum/DEV/ETH/YD/yd-marketplace/node_modules/.store'
```

The same user also noted their confusion over the language
provided by the extension after this failure:

> Please submit your lockfile to Phylum should this error persist.

They did not have a lockfile present (just a `package.json` manifest)
and noted that the extension did not create a `yarn.lock` lockfile
for them to submit.

Therefore, the language of this message has been updated to be less
specific about lockfiles and more about "dependency files."

Finally, a troubleshooting section was added to the `yarn` extension's
documentation. This is an attempt to better prepare users for what to
do when the extension fails. After the language is ironed out here for
this extension, the plan is to add it to each of the other package
manager extensions by updating this review.

### DIFF
--- a/extensions/yarn/README.md
+++ b/extensions/yarn/README.md
@@ -3,6 +3,10 @@
 A [Phylum CLI][phylum-cli] extension that checks your [yarn][yarn] dependencies
 through [Phylum][phylum] before installing them.
 
+[phylum-cli]: https://github.com/phylum-dev/cli
+[yarn]: https://yarnpkg.com/
+[phylum]: https://phylum.io
+
 ## Installation
 
 This is a pre-installed extension and may be available without any additional
@@ -42,6 +46,39 @@ In cases where Phylum still needs to process some of the packages, the command
 will exit with a warning **without** installing the packages. Once the analysis
 is complete, another attempt can be made.
 
-[phylum]: https://phylum.io
-[phylum-cli]: https://github.com/phylum-dev/cli
-[yarn]: https://yarnpkg.com/
+## Troubleshooting
+
+There are some common reasons this extension may fail.
+
+### Sandbox violations
+
+The underlying package manager may fail due to operating within the Phylum
+sandbox. These errors are usually displayed in the output and say something
+about not being able to access a particular file or network resource.
+
+One possibility is that this is a valid sandbox violation that has not been
+accounted for with an exception. If you believe this is the case, please
+[submit an issue][issue] or [contact us][contact] to discuss adding exceptions.
+
+Another possibility is that a malicious action was taken by one of the packages.
+Blocking such malicious actions is a primary goal for using this extension but
+it may be hard to know what is malicious and what is not. If you need help
+making that determination, please [contact us][contact] and provide the log
+output along with any other relevant details.
+
+[issue]: https://github.com/phylum-dev/cli/issues/new/choose
+[contact]: https://docs.phylum.io/support/contact_us
+
+### Package violations
+
+Packages analyzed by this extension may not pass the
+[Phylum default policy][policy]. When this happens the extension will prevent
+the requested action and instead print a list of policy violations.
+
+There is not currently a method for allowing policy violations discovered by
+this extension. Instead, inspect the findings and determine if the offending
+entries are truly necessary. If not, update your project dependencies and try
+again. If they are, then the only recourse is to try the action again without
+the Phylum package manager extension.
+
+[policy]: https://docs.phylum.io/knowledge_base/policy
--- a/extensions/yarn/README.md
+++ b/extensions/yarn/README.md
@@ -53,18 +53,19 @@ There are some common reasons this extension may fail.
 ### Sandbox violations
 
 The underlying package manager may fail due to operating within the Phylum
-sandbox. These errors are usually displayed in the output and say something
-about not being able to access a particular file or network resource.
+sandbox. These errors may be displayed in the output and say something about not
+being able to access a particular resource.
 
 One possibility is that this is a valid sandbox violation that has not been
-accounted for with an exception. If you believe this is the case, please
-[submit an issue][issue] or [contact us][contact] to discuss adding exceptions.
+accounted for with an exception. Another possibility is that a malicious action
+was taken by one of the packages. Blocking such malicious actions is a primary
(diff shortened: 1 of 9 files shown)


extensions/yarn/README.md
maxrake, line 1: This language may change as part of this PR. Once it is ironed out, the plan is to update this PR with the same language added to each of the other package manager extensions.

extensions/yarn/README.md
```diff
@@ -42,6 +46,39 @@ In cases where Phylum still needs to process some of the packages, the command
...
+[Phylum default policy][policy]. When this happens the extension will prevent
+the requested action and instead print a list of policy violations.
+
+There is not currently a method for allowing policy violations discovered by
+this extension. Instead, inspect the findings and determine if the offending
+entries are truly necessary. If not, update your project dependencies and try
+again. If they are, then the only recourse is to try the action again without
+the Phylum package manager extension.
+
+[policy]: https://docs.phylum.io/knowledge_base/policy
```
cd-work, line 84: Why would this only be added to the yarn extension? Also this seems **extremely** verbose, I wouldn't expect more than one paragraph of this size for each of these headings. I doubt users care to this degree.

cd-work (changes requested):

extensions/yarn/README.md
maxrake: As noted in the PR summary and annotated with a file-level reminder comment, the plan is to iron out the language once and update this PR with the same language added to each of the other package manager extensions.

Alternative wording is welcome. It has already been seen that documentation of this sort is desired by real users. The one that kicked off these changes observed that they spent considerable time attempting to get the `yarn` extension working and did not understand what to do with the failed output provided by it.

extensions/yarn/README.md
```diff
@@ -42,6 +46,39 @@ In cases where Phylum still needs to process some of the packages, the command
...
-[yarn]: https://yarnpkg.com/
+## Troubleshooting
+
+There are some common reasons this extension may fail.
+
+### Sandbox violations
+
+The underlying package manager may fail due to operating within the Phylum
+sandbox. These errors are usually displayed in the output and say something
+about not being able to access a particular file or network resource.
```
cd-work, line 57: This is inaccurate, at least for all extensions. Often times there are no error messages and while network resource errors might be possible, I think we have networking enabled pretty much everywhere?

extensions/yarn/README.md
```diff
@@ -42,6 +46,39 @@ In cases where Phylum still needs to process some of the packages, the command
...
+
+One possibility is that this is a valid sandbox violation that has not been
+accounted for with an exception. If you believe this is the case, please
+[submit an issue][issue] or [contact us][contact] to discuss adding exceptions.
+
+Another possibility is that a malicious action was taken by one of the packages.
```
cd-work, line 63: Malicious activity would only be a failure if it fails loudly, which is probably the minority of cases.

extensions/yarn/README.md
```diff
@@ -42,6 +46,39 @@ In cases where Phylum still needs to process some of the packages, the command
...
+Blocking such malicious actions is a primary goal for using this extension but
+it may be hard to know what is malicious and what is not. If you need help
+making that determination, please [contact us][contact] and provide the log
+output along with any other relevant details.
```
cd-work, line 67: This is effectively just the same as the last paragraph. Why do we need to reiterate it again here?

extensions/yarn/README.md
```diff
@@ -42,6 +46,39 @@ In cases where Phylum still needs to process some of the packages, the command
...
+
+Packages analyzed by this extension may not pass the
+[Phylum default policy][policy]. When this happens the extension will prevent
+the requested action and instead print a list of policy violations.
+
+There is not currently a method for allowing policy violations discovered by
+this extension. Instead, inspect the findings and determine if the offending
+entries are truly necessary. If not, update your project dependencies and try
+again. If they are, then the only recourse is to try the action again without
+the Phylum package manager extension.
```
cd-work, line 82: The patch to change this would be shorter than the paragraph explaining the reason why it is this way if we actually want to change this behavior.

cd-work (changes requested):

extensions/yarn/README.md
maxrake: I will attempt to reword this to be more universal. The `yarn` and `npm` extensions *disable* networking during the sandbox building step. The remaining package manager extensions *enable* networking for all `runSandboxed` calls.

extensions/yarn/README.md
maxrake: Blocking malicious activity is why users want to use these extensions (based on feedback from the user on Discord that initiated this PR). Noting it is worthwhile, regardless of the frequency of occurrence. Alternate wording suggestions are welcome.

extensions/yarn/README.md
maxrake: The "contact us" part? I will work to consolidate that.

extensions/yarn/README.md
maxrake: It would be great to change this behavior. Until that happens, this documentation is helpful. Even if/once that behavior changes, documentation will be needed to explain it.

extensions/yarn/README.md
cd-work: I'm not saying malicious activity isn't the only cause of failure. I'm saying malicious activity might be blocked without causing failure.

extensions/yarn/README.md
cd-work: Yeah, I think once is fine for that.

cd-work (commented):
I still don't think the verbosity present in this patch will be beneficial to the user. But if you're happy with it then we can just add it to all extensions and get this merged, I don't believe the readmes for the individual extensions get much action anyway.

maxrake:
I'm happy with the changes. I am in the camp that verbose user documentation is a good thing.

cd-work (approved):

PR #1496, opened 2024-09-13
Fix maven lockfile generation on Windows

kylewillmon:
With some testing, we discovered that running Maven on Windows was failing for a few reasons:

1. Maven uses a batch script called `mvn.cmd` on Windows. And `Command` does not automatically find batch scripts (because it emulates the behavior of `CreateProcess` and not `cmd.exe`)
2. We were using paths from `fs::canonicalize()` as the working directory for spawned processes. However, these paths use the `\\?\` prefix which is not supported by many Windows binaries (including `cmd.exe` itself)

This patch fixes both issues and adds a bit more information to the error output.


lockfile_generator/Cargo.toml
```diff
@@ -13,3 +13,4 @@ anyhow = "1.0.75"
 glob = "0.3.1"
 thiserror = "1.0.49"
 tempfile = "3.3.0"
+dunce = "1.0.5"
```
cd-work, line 16: Do we need a dependency for this? Its entire code is under 300 lines, including tests.

cd-work (changes requested):

lockfile_generator/Cargo.toml
kylewillmon: Strictly speaking, no... I could copy the code and include a license notice... I even considered just blindly dropping the prefix which would work 99.9% of the time.

But `dunce` seems to be a pretty safe dependency to add. It's fairly popular. It's used directly by rust-lang as a part of `miri`. And it is maintained by a developer that seems pretty active in the Rust community. In fact, [the same maintainer](https://crates.io/users/kornelski?sort=downloads) also maintains `security-framework`, `urlencoding`, and `xml-rs` (among many others) and these crates are already being used directly or indirectly by CLI.

cd-work: I personally would **not** have added a dependency for this. I don't consider it good practice for dependencies with essentially no code.

However since it's not some random tiny crate nobody else uses, I will leave the final judgement up to you. I just wanted to make sure it was carefully considered on your part.

PR #1497, opened 2024-09-17
Bump to v7.0.0

kylewillmon:
### Added

- Organization management under the `phylum org` subcommand
- Organization support for existing subcommands
- `phylum project update --default-label` option to set a project's default label
- `phylum project list --no-group` flag to only show personal projects
- Full sandbox write access to project directory for building with `yarn` extension

### Fixed

- Lockfile parsing of bun-generated yarn lockfiles
- Maven lockfile generation on Windows

### Removed

- `phylum batch` subcommand

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+## 7.0.0 - 2024-09-17
+
 ### Added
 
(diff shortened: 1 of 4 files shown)


maxrake (approved):
Finally moving past the spooky v6.6.6 ...LGTM

I may wait a beat on approving the `phylum-ci` Docker image creation workflow...just to ensure I am around to make changes or roll it back in case it breaks the integrations for some reason. It shouldn't...but just in case...

PR #1498, opened 2024-09-19
Enable extensions for Windows release artifacts

maxrake:
This change enables the extensions feature when building Windows release artifacts. Doing so will enable the `phylum-ci` tool to continue to operate with its extension-based analysis for the coming Windows standalone binary. Additional changes include:

* Consolidate `windows` job in `Test` workflow with `test-matrix` job
* Add Windows runner to `clippy` and `oldstable` `Test` workflow jobs
  * Fix Windows clippy findings
* Remove extraneous `cmake` installs on macOS runners
* Fix `shellcheck` and `actionlint` findings
* Format throughout

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,7 +56,7 @@ jobs:
           - target: x86_64-pc-windows-msvc
             os: windows-latest
             extension: .exe
-            flags: --no-default-features
+            flags: --no-default-features -F extensions
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout the repo
@@ -71,6 +71,10 @@ jobs:
       - if: ${{ runner.os == 'Linux' }}
         name: Install protoc
         run: sudo apt update && sudo apt install -y protobuf-compiler
+
+      - if: ${{ runner.os == 'Windows' }}
+        name: Install protoc
+        run: choco install protoc
 
       - if: ${{ runner.os == 'Windows' }}
         name: Install Rust toolchain
@@ -191,7 +195,7 @@ jobs:
         env:
           OPENSSL_KEY: ${{ secrets.OPENSSL_KEY }}
         run: |
-          for archive in phylum-*.zip;
+          for archive in phylum-*.zip phylum-*.exe;
           do
             openssl dgst -sha256 -sign <(printf "%s" "$OPENSSL_KEY") -out "${archive}.signature" "${archive}"
           done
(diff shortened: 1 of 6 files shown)


maxrake:
Do we want to add `windows-latest` to the matrix for the `clippy` and `oldstable` jobs?

I'm guessing adding it to the `all-features` job doesn't make as much sense given that we don't want (can't have) all the features enabled for Windows targets.

.github/workflows/release.yml
```diff
@@ -191,7 +195,7 @@ jobs:
         env:
           OPENSSL_KEY: ${{ secrets.OPENSSL_KEY }}
         run: |
-          for archive in phylum-*.zip;
+          for archive in phylum-*.zip phylum-*.exe;
```
kylewillmon, line 198: I intentionally left the windows binary unsigned since it is not fully supported and does not contain all of the features of our other builds.

.github/workflows/test.yml
```diff
@@ -142,26 +137,34 @@ jobs:
...
+        run: find . -iname "*.sh" -print0 | xargs -0 shellcheck -o all -S style -s sh
 
-  # This job reports the results of the matrixes above
+  # This job reports the results of the test jobs above and is used
+  # to enforce status checks in the repo settings without needing
+  # to update those settings every time the test jobs are updated.
   test:
     if: always()
-    needs: [windows, clippy, test-matrix, all-features, oldstable]
+    needs: [rustfmt, clippy, test-matrix, oldstable, all-features, deno-checks, shellcheck]
```
kylewillmon, line 167: `rustfmt`, `oldstable`, and `deno-checks` are already required by repo settings... There is no reason to include them here (especially in a PR that leaves those jobs unchanged)

kylewillmon (changes requested):

.github/workflows/release.yml
maxrake: It has been reverted.

.github/workflows/test.yml
maxrake: It has been reverted. Only the jobs with matrices are included in this test rollup job now.

.github/workflows/test.yml
```diff
@@ -31,25 +31,29 @@ jobs:
...
 
-      - if: ${{ runner.os == 'macOS' }}
-        name: Install protoc
-        run: brew install protobuf
-
-      - if: ${{ runner.os == 'Linux' }}
-        name: Install protoc
-        run: sudo apt install -y protobuf-compiler
+      - name: Install protoc
+        run: ${{ matrix.protoc }}
```
kylewillmon: I'm not sure I can explain why... But I preferred this better before. Putting the install command in the variables for the matrix just feels weird...

More importantly, the `oldstable` matrix still does it the other way. These should probably match.

kylewillmon (changes requested):

kylewillmon (approved):

PR #1500, opened 2024-09-20
Fix Windows depfile paths in Phylum project files

maxrake:
The changes made here are emulating those made in #1496, but for the `phylum_project` crate. This allows for more natural paths on Windows systems. Other changes made include:

* Update `.phylum_project` file for this repo
  * Include the output/format from the current `phylum init` command
* Ensure `all-features` job in `Test` workflow has matching strategy

## Testing

Output from a *nix system (macOS):

```
## Latest released version
❯ phylum --version
phylum v7.0.0

❯ phylum status
Project: cli
Group: integrations
Project Root: /Users/maxrake/dev/phylum/localdev/cli
Dependency Files:
 - path: /Users/maxrake/dev/phylum/localdev/cli/./Cargo.lock
   type: cargo

## Version made from the changes in this PR
phylum v7.0.0-3-g35f1873

❯ target/debug/phylum status
Project: cli
Group: integrations
Project Root: /Users/maxrake/dev/phylum/localdev/cli
Dependency Files:
 - path: /Users/maxrake/dev/phylum/localdev/cli/./Cargo.lock
   type: cargo
```

Output from a Windows system:

```
## Latest released version
C:\Users\maxrake\Downloads\phylum\phylum-ci>phylum --version
phylum v7.0.0

C:\Users\maxrake\Downloads\phylum\phylum-ci>phylum status
Project: phylum-ci
Group: phylum_bot
Project Root: \\?\C:\Users\maxrake\Downloads\phylum\phylum-ci
Dependency Files:
 - path: \\?\C:\Users\maxrake\Downloads\phylum\phylum-ci\poetry.lock
   type: poetry
   
## Version made from the changes in this PR
C:\Users\maxrake\Downloads\phylum\phylum-ci>phylum --version
phylum v7.0.0-3-g35f1873

C:\Users\maxrake\Downloads\phylum\phylum-ci>phylum status
Project: phylum-ci
Group: phylum_bot
Project Root: C:\Users\maxrake\Downloads\phylum\phylum-ci
Dependency Files:
 - path: C:\Users\maxrake\Downloads\phylum\phylum-ci\./poetry.lock
   type: poetry
```

While the output for both OS types is less than ideal, with the inclusion of the final relative path component, it doesn't affect the operation of `phylum-ci`, which makes use of the output of this command to populate dependency file paths.

Ideally,
`/Users/maxrake/dev/phylum/localdev/cli/./Cargo.lock`
would show as
`/Users/maxrake/dev/phylum/localdev/cli/Cargo.lock`
and
`C:\Users\maxrake\Downloads\phylum\phylum-ci\./poetry.lock`
would show as
`C:\Users\maxrake\Downloads\phylum\phylum-ci\poetry.lock`.

It is simply a visual nuisance and the paths still resolve correctly so pursuing a "fix" to this will be left to another time.

### DIFF
--- a/phylum_project/src/lib.rs
+++ b/phylum_project/src/lib.rs
@@ -51,8 +51,7 @@ impl ProjectConfig {
                 .iter()
                 .map(|depfile| {
                     let path = self.root.join(&depfile.path);
-                    let dunce_path = dunce::simplified(&path);
-                    DepfileConfig::new(dunce_path, depfile.depfile_type.clone())
+                    DepfileConfig::new(dunce::simplified(&path), depfile.depfile_type.clone())
                 })
                 .collect();
         }
@@ -109,8 +108,7 @@ pub fn find_project_conf(
     // If `path` is like `.`, `path.parent()` is `None`.
     // Convert to a canonicalized path so removing components navigates up the
     // directory hierarchy.
-    let mut path = starting_directory.as_ref().canonicalize().ok()?;
-    path = dunce::simplified(&path).to_path_buf();
+    let mut path = dunce::canonicalize(starting_directory.as_ref()).ok()?;
 
     for _ in 0..max_depth {
         let conf_path = path.join(PROJ_CONF_FILE);
--- a/phylum_project/src/lib.rs
+++ b/phylum_project/src/lib.rs
@@ -51,7 +51,8 @@ impl ProjectConfig {
                 .iter()
                 .map(|depfile| {
                     let path = self.root.join(&depfile.path);
-                    DepfileConfig::new(path, depfile.depfile_type.clone())
+                    let dunce_path = dunce::simplified(&path);
+                    DepfileConfig::new(dunce_path, depfile.depfile_type.clone())
                 })
                 .collect();
         }
@@ -109,6 +110,7 @@ pub fn find_project_conf(
     // Convert to a canonicalized path so removing components navigates up the
     // directory hierarchy.
     let mut path = starting_directory.as_ref().canonicalize().ok()?;
+    path = dunce::simplified(&path).to_path_buf();
 
     for _ in 0..max_depth {
         let conf_path = path.join(PROJ_CONF_FILE);
@@ -157,7 +159,7 @@ mod tests {
         let mut config =
             ProjectConfig::new(PROJECT_ID, PROJECT_NAME.to_owned(), Some(GROUP_NAME.to_owned()));
         config.set_depfiles(vec![DepfileConfig {
-            path: PathBuf::from("Cargo.lock"),
+            path: PathBuf::from("./Cargo.lock"),
             depfile_type: "cargo".to_owned(),
         }]);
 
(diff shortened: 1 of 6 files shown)


phylum_project/src/lib.rs
```diff
@@ -157,7 +159,7 @@ mod tests {
         let mut config =
             ProjectConfig::new(PROJECT_ID, PROJECT_NAME.to_owned(), Some(GROUP_NAME.to_owned()));
         config.set_depfiles(vec![DepfileConfig {
-            path: PathBuf::from("Cargo.lock"),
+            path: PathBuf::from("./Cargo.lock"),
             depfile_type: "cargo".to_owned(),
         }]);
```
kylewillmon, line 164: Wait... Does this not work without `./` anymore? That seems like a problem...

kylewillmon (changes requested):

phylum_project/src/lib.rs
maxrake: It works with both `Cargo.lock` and `./Cargo.lock`. It was changed in the test cases to `./Cargo.lock` to be more representative of how the config will look now (see the updated `.phylum_project` file in this PR for how `phylum init` populates it now).

phylum_project/src/lib.rs
```diff
@@ -109,6 +110,7 @@ pub fn find_project_conf(
     // Convert to a canonicalized path so removing components navigates up the
     // directory hierarchy.
     let mut path = starting_directory.as_ref().canonicalize().ok()?;
+    path = dunce::simplified(&path).to_path_buf();
```
kylewillmon, line 113: One of the benefits of `dunce::simplified()` is that it's a no-op on non-Windows platforms... But when you add `.to_path_buf()`, it's a copy everywhere... That might be optimized away, but who knows...

You can just replace the call to `canonicalize` here instead of using `simplified`

```suggestion
    let mut path = dunce::canonicalize(starting_directory.as_ref()).ok()?;
```

phylum_project/src/lib.rs
```diff
@@ -51,7 +51,8 @@ impl ProjectConfig {
                 .iter()
                 .map(|depfile| {
                     let path = self.root.join(&depfile.path);
-                    DepfileConfig::new(path, depfile.depfile_type.clone())
+                    let dunce_path = dunce::simplified(&path);
+                    DepfileConfig::new(dunce_path, depfile.depfile_type.clone())
```
kylewillmon, line 55: The variable name `dunce_path` doesn't really mean anything (unless you're calling the path stupid...). You can just use variable shadowing instead:

```suggestion
                    let path = dunce::simplified(&path);
                    DepfileConfig::new(path, depfile.depfile_type.clone())
```

Or avoid the variable altogether (reformatting is probably needed for this one):

```suggestion
                    DepfileConfig::new(dunce::simplified(path), depfile.depfile_type.clone())
```

phylum_project/src/lib.rs
kylewillmon: Hmm... I'm not sure when that change to `phylum init` happened.

kylewillmon (changes requested):

phylum_project/src/lib.rs
maxrake: Good stuff. I knew I wouldn't get all this Rust stuff right...thanks!

phylum_project/src/lib.rs
maxrake: Thank you for the suggestions. I learned something with the first one but went with the second one.

PR #1502, opened 2024-09-24
Add spellcheck to CI

cd-work:


cd-work:
Discarding this patch because there's plenty of intentionally misspelled text in CLI and having to annotate that is more of a hassle than having a typo here or there which doesn't affect anyone unless they end up in the CLI's output.

PR #1506, opened 2024-09-26
Use pre-installed rustup

kylewillmon:
All of our runners (including self-hosted runners) now have rustup available. So this patch removes the use of curl to install it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -76,19 +76,11 @@ jobs:
         name: Install protoc
         run: choco install protoc
 
-      - if: ${{ runner.os == 'Windows' }}
-        name: Install Rust toolchain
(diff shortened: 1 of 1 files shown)


kylewillmon:
I've manually triggered the release workflow on this branch [here](https://github.com/phylum-dev/cli/actions/runs/11060320840).

maxrake:
Still not working on the manually triggered release workflow. I'll hold off a bit on this review.

kylewillmon:
> Still not working on the manually triggered release workflow.

It looks like one of the old runners was still around to pick up the job... I assumed enough time had passed that they would all be using the new image. A re-run should resolve the issue (as soon as GitHub lets me re-run it)

maxrake (approved):
LGTM

PR #1517, opened 2024-10-10
Fix unprocessed packages showing as complete

### LINKED ISSUE #1516: `phylum package` does not indicate unprocessed packages
## Overview

The `phylum package` command does not indicate when a package has been submitted but not processed. Instead, it seems to indicate the package is processed and free of findings.

## How To Reproduce

1. Find a package that was recently published
2. Use `phylum package` subcommand to get info on it
3. See that the status appears to be complete, with "Num Vulns" of `0` and "Risk Vectors" of 100

Here is an example, with an `awscli` package uploaded ~1 hour before this attempt:

```
❯ phylum --version
phylum v7.1.1

## What it looks like for a package that has been "submitted" but not "processed"
❯ phylum package pypi awscli 1.35.3

 Package Name:                awscli    Package Version:                           1.35.3
 License:         Apache License 2.0    Last updated:       2024-10-09T19:22:46.332+00:00
 Num Deps:                         6    Num Vulns:                                      0
 Ecosystem:                     pypi


 Risk Vectors:

    Total Risk:              100
    Author Risk:             100
    Engineering Risk:        100
    License Risk:            100
    Malicious Code Risk:     100
    Vulnerability Risk:      100

## What it looks like for a package that has not been "submitted" (doesn't exist in this case)
❯ phylum package pypi awscli 1.35.4
⚠️  Thank you for submitting this package. Please check back later for results.

## What it looks like when analyzing the still unprocessed package
❯ echo "awscli==1.35.3" > requirements-delme.txt

## It shows as "INCOMPLETE" and "unprocessed" ...which is correct/expected
❯ phylum analyze requirements-delme.txt
✅ Successfully parsed dependency file "requirements-delme.txt" as type "pip"
✅ Job ID: 707e53a5-39d3-4044-bc13-fed35561d23e

Phylum Supply Chain Risk Analysis — INCOMPLETE

The analysis contains 1 unprocessed package, preventing a complete risk
analysis. Phylum is currently processing these packages and should complete
soon. Please wait for up to 30 minutes, then re-run the analysis.

You can find the interactive report here:
  https://app.phylum.io/projects/d84b0bb3-fb64-43cb-a473-dc96b54db662?group=integrations&label=uncategorized
```

## Expected Behavior

Some indication is given when the package is still unprocessed by Phylum, like how the UI does it:

image

## Additional Context

None at this time

cd-work:
Closes #1516.

### DIFF
--- a/cli/src/commands/packages.rs
+++ b/cli/src/commands/packages.rs
@@ -35,7 +35,7 @@ pub async fn handle_get_package(api: &PhylumApi, matches: &clap::ArgMatches) -> 
     };
 
     match resp {
-        PackageSubmitResponse::AlreadyProcessed(mut resp) => {
+        PackageSubmitResponse::AlreadyProcessed(mut resp) if resp.complete => {
             let filter = matches.get_one::<String>("filter").and_then(|v| Filter::from_str(v).ok());
             if let Some(filter) = filter {
                 resp.filter(&filter);
@@ -43,14 +43,14 @@ pub async fn handle_get_package(api: &PhylumApi, matches: &clap::ArgMatches) -> 
 
             resp.write_stdout(pretty_print);
         },
-        PackageSubmitResponse::AlreadySubmitted => {
-            print_user_warning!(
-                "Package is still processing. Please check back later for results."
-            );
-        },
         PackageSubmitResponse::New => {
             print_user_warning!(
                 "Thank you for submitting this package. Please check back later for results."
+            );
+        },
+        _ => {
+            print_user_warning!(
+                "Package is still processing. Please check back later for results."
             );
         },
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Fixed
+
+- `phylum package` subcommand showing unprocessed packages as complete
+
 ## 7.1.1 - 2024-10-09
 
 ### Changed


cli/src/commands/packages.rs
```diff
@@ -35,22 +35,22 @@ pub async fn handle_get_package(api: &PhylumApi, matches: &clap::ArgMatches) ->
     };
 
     match resp {
-        PackageSubmitResponse::AlreadyProcessed(mut resp) => {
+        PackageSubmitResponse::AlreadyProcessed(mut resp) if resp.complete => {
```
matt-phylum, line 38: `resp.complete` should always be `true`. If it's false that's a bug.

cd-work: Well then there is a bug.

cd-work: Well then the API is broken.

maxrake (approved):
Confirmed with local testing, by picking a package that had just been published to PyPI ~4 minutes prior (selected from this query: `https://pypi.org/search/?q=&o=-created&c=License+%3A%3A+OSI+Approved`):

```
❯ phylum --version
phylum v7.1.1

❯ phylum package pypi aw-api 2.0.4

 Package Name:    aw-api    Package Version:                            2.0.4
 License:            MIT    Last updated:       2024-10-10T21:09:31.781+00:00
 Num Deps:             4    Num Vulns:                                      0
 Ecosystem:         pypi


 Risk Vectors:

    Total Risk:              100
    Author Risk:             100
    Engineering Risk:        100
    License Risk:            100
    Malicious Code Risk:     100
    Vulnerability Risk:      100


phylum v7.1.1-2-gb6b238a

❯ target/debug/phylum package pypi aw-api 2.0.4
⚠️  Package is still processing. Please check back later for results.

## Submit an older version that had never been seen by Phylum
❯ target/debug/phylum package pypi aw-api 1.1.0
⚠️  Thank you for submitting this package. Please check back later for results.
```

cli/src/commands/packages.rs
kylewillmon: I can confirm that something is broken for some of these packages... We're still investigating the cause...

The `complete` field should probably be deprecated and removed. I don't think this is a CLI bug at all.

cd-work: Easy enough to remove in the future. For now this works around the API bug with just a single line changed.

PR #1530, opened 2024-11-01
Allow whitespace in `pip` requirement specifiers

maxrake:
A bug was discovered recently when attempting to parse a `pip` lockfile. That is, a `requirements*.txt` file containing exact versions (i.e., `==`) in all of it's requirement specifiers. Such a file should parse as a lockfile and not fall back to lockfile generation. However, if whitespace is included around the `==`, the parsing fails.

Whitespace is allowed in requirement specifiers. This change accounts for that. It also updates the test fixture to prevent regressions.

Ref: https://pip.pypa.io/en/stable/reference/requirement-specifiers/

### DIFF
--- a/tests/fixtures/requirements-locked.txt
+++ b/tests/fixtures/requirements-locked.txt
@@ -11,7 +11,8 @@ amqp==5.0.9 --hash=sha256:77fd4e1249d8c9923de34907236b747ced06e5467ecac1a7bb7115
     --hash=sha256:8c2f9abd47a9e8df7f0c3f091ce9497d011dc3b31effcf4c85a6e2b50f4114ef
     # via kombu
 
-attrs==20.2.0   # This is an inline comment
+# Whitespace is allowed in requirement specifiers
+attrs == 20.2.0   # This is an inline comment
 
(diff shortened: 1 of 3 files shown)


tests/fixtures/requirements-locked.txt
```diff
@@ -11,7 +11,8 @@ amqp==5.0.9 --hash=sha256:77fd4e1249d8c9923de34907236b747ced06e5467ecac1a7bb7115
     --hash=sha256:8c2f9abd47a9e8df7f0c3f091ce9497d011dc3b31effcf4c85a6e2b50f4114ef
     # via kombu
 
-attrs==20.2.0   # This is an inline comment
+# Whitespace is allowed in requirement specifiers
+attrs == 20.2.0   # This is an inline comment
```
mathew-horner, line 15: Is it possible to have both variants in the fixture?

maxrake: Yes. Both variants are already in this fixture. For example `alembic==1.10.3` is on line 8.

mathew-horner (approved):

PR #1539, opened 2024-11-22
Fix PNPM v5 version parsing

### LINKED ISSUE #1538: Parsing `pnpm-lock.yaml` can result in incorrect versions
# Overview
The parsing of pnpm lockfiles can sometimes result in incorrect versions.

# How To Reproduce
I will provide a sample file in slack.

# Expected Behavior
Correctly parse pnpm lockfiles.

# Additional Context
Entries in the lockfile like this seem to be the cause:
image

The parser returns this:
image

But the correct version is just `1.2.2`
image

cd-work:
This fixes an issue with the lockfile parser of PNPM v5 lockfiles which would incorrectly include metadata in the package version.

Closes #1538.

### DIFF
--- a/lockfile/src/javascript.rs
+++ b/lockfile/src/javascript.rs
@@ -344,9 +344,9 @@ impl PnpmLock {
     ///
     /// This parses the combined name and version used as an index for the
     /// `packages` map.
-    fn parse_key(mut key: &str, version: PnpmVersion) -> anyhow::Result<(String, String)> {
+    fn parse_key(mut key: &str, pnpm_version: PnpmVersion) -> anyhow::Result<(String, String)> {
         // Strip prefix from `version < 9` lockfiles.
-        if version < PnpmVersion::V9 {
+        if pnpm_version < PnpmVersion::V9 {
             key = key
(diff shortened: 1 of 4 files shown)


lockfile/src/javascript.rs
```diff
@@ -356,16 +356,23 @@ impl PnpmLock {
         let name = key.split_once('(').map(|(name, _)| name).unwrap_or(key);
 
         // Get version separator based on PNPM version.
-        let version_separator = match version {
+        let version_separator = match pnpm_version {
```
maxrake, line 359: Not exactly related to this PR, but should the supported PNPM lockfile formats be listed in `docs/supported_lockfiles.md`? We do that for a few other lockfiles that have multiple recognized formats.

maxrake (approved):
One question, but approving now since the changes here address the linked issue.

lockfile/src/javascript.rs
cd-work: Probably? Though we do support a wide enough version range to where I've yet to find a lockfile that we cannot parse (other than bugs like this).

PR #1543, opened 2024-11-25
Ignore platform-specific gems

### LINKED ISSUE #1540: `gem` lockfile parser does not account for platform-specific gems
## Overview

Parsing `Gemfile.lock` lockfiles fails when multiple platforms are specified and platform-specific gems are included.

## How To Reproduce

I will provide a sample file in Slack...but can likely be recreated by:

1. Taking an existing `Gemfile.lock`
2. Add an additional platform
3. Add a dependency that provides platform-specific gems (e.g., `ffi`)
4. Update the lockfile
5. Try to parse it with Phylum CLI
6. See failure like this:

```
❯ phylum parse --no-generation Gemfile.lock
❗ Error: Could not parse dependency file "Gemfile.lock" as "auto" type

Caused by:
    0: Failed to parse lockfile
    1: Failed to parse gem lockfile
    2: 0: at line 29, in Tag:
           ffi (1.17.0-x86_64-linux-gnu)
```

## Expected Behavior

Parse `gem` lockfiles with platform-specific dependencies.

## Additional Context

It is not clear if the Phylum processing pipeline acquires and analyzes each platform-specific gem separately. If so, the platform should be taken into account. If not, perhaps the platform portion of the version can be ignored during parsing.

References:
* [`bundler` documentation for supporting other platforms](https://bundler.io/v2.5/man/bundle-lock.1.html#SUPPORTING-OTHER-PLATFORMS)
* [related SO post](https://stackoverflow.com/q/48761493)
* [blog post about understanding the `Platforms` section](https://www.moncefbelyamani.com/understanding-the-gemfile-lock-file/#platforms)

cd-work:
This fixes an issue in the Ruby `Gemfile.lock` parser where version-specific dependencies like `ffi (1.17.0-x86_64-linux-gnu)` would cause the parser to crash.

Since new Ruby Gems are not allowed to contain dashes, this patch simply discards Gems that contain a `-` in their version. There are technically two versions of `asciidoctor-reducer` (`1.0.0-rc.1` and `1.0.0-beta.1`) which contain dashes, but those are neither commonly used, nor a security risk.

Closes #1540.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - `pnpm` version 5 parser including metadata in package versions
+- Platform-specific dependencies ignored by the `Gemfile.lock` parser
 
 ## 7.1.4 - 2024-11-07
 
--- a/lockfile/src/parsers/gem.rs
+++ b/lockfile/src/parsers/gem.rs
@@ -1,7 +1,7 @@
 use nom::branch::alt;
 use nom::bytes::complete::{tag, take_until};
 use nom::character::complete::{line_ending, not_line_ending, satisfy, space0};
-use nom::combinator::{eof, opt, recognize};
+use nom::combinator::{opt, recognize};
 use nom::error::{VerboseError, VerboseErrorKind};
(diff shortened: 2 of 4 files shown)


maxrake (approved):
The code was reviewed and the changes confirmed with local testing. LGTM.

PR #1545, opened 2024-11-27
Add C# *.packages.config support

cd-work:
This adds support for `packages.config` files for the C# ecosystem under the new `NugetConfig` lockfile name.

The file format itself seems to be relatively simple, with just a list of package names and optional versions. The current test fixture is based on examples of this lockfile found on <https://github.com>.

Both `packages.config` on its own and `*.packages.config` are supported as valid names. Since this seems to be a lockfile there is no need for lockfile generation.

### DIFF
--- a/lockfile/src/lib.rs
+++ b/lockfile/src/lib.rs
@@ -17,7 +17,7 @@ use thiserror::Error;
 use walkdir::WalkDir;
 
 pub use crate::cargo::Cargo;
-pub use crate::csharp::{CSProj, PackagesLock};
+pub use crate::csharp::{CSProj, PackagesConfig, PackagesLock};
 pub use crate::cyclonedx::CycloneDX;
 pub use crate::golang::{GoMod, GoSum};
 pub use crate::java::{GradleLock, Pom};
@@ -61,6 +61,7 @@ pub enum LockfileFormat {
     #[serde(alias = "nuget")]
     Msbuild,
     NugetLock,
+    NugetConfig,
     GoMod,
     Go,
     Cargo,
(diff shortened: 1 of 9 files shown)


maxrake:
Can the files be named as `*packages.config` as well? Or, will there always be a `.` first (`*.packages.config`)? I plan to look at this PR in more detail next week, but keeping someone else on the reviewer list who knows C# better is recommended.

cd-work:
> Can the files be named as *packages.config as well? Or, will there always be a . first (*.packages.config)?

https://github.com/search?q=path%3A**%2F*packages.config+language%3AXML&type=code suggests to me that only `packages.config` and `*.packages.config` are valid. This makes sense to me because it matches the new lockfile version and it's just generally sensible.

lockfile/src/lib.rs
```diff
@@ -140,7 +143,7 @@ impl LockfileFormat {
 }
 
 /// An iterator of all supported lockfile formats.
-pub struct LockfileFormatIter(u8);
+pub struct LockfileFormatIter(usize);
```
matt-phylum, line 146: Instead of `usize` here, it could be `std::slice::Iter<'static, LockfileFormat>`. The size of the struct increases by 16x instead of 8x like this code, but `next` would become `Some(*self.0.next()?)`.

furi0us333:
I don't know anything about the nuget ecosystem, but I see that the customer sent over a file named `farm.packages.config`. Unfortunately, I have no context about the rest of the files in their project. Let me know if you would like me to ask for any specific additional information.

cd-work:
@furi0us333 I've already asked Louis to double-check in the meeting with ICE today. All the evidence points to the file being renamed by ICE, but I'd like to double-check before we move forward.

matt-phylum (dismissed):
The changelog and some comments still say `*.packages.config`.

matt-phylum (dismissed):

matt-phylum (approved):

PR #1546, opened 2024-11-29
Fix certificate check option naming

### LINKED ISSUE #480: `ignore_certs` is plural but `--no-check-certificate` is singular
The `ignore_certs` config file option is plural, but the `--no-check-certificate` command line option is singular.

cd-work:
This fixes an inconsistency between the `ignore_certs` configuration option and the `--no-check-certificate` CLI flag by changing the CLI flag to be `--ignore-certs` with an alias for the old version.

Closes #480.


matt-phylum (approved):
`--no-check-certificate` is the name of the parameter used by wget, but if it's just wget that doesn't seem like a standard that we'd want to stick with. I like when the name contains something like "insecure", but it's probably okay.

PR #1549, opened 2024-12-04
Fix nightly clippy warnings

cd-work:
This fixes nightly clippy warnings, mainly the new suggestion to prefer `Option::is_some_and`/`Result::is_ok_and` over `.map_or(false, …)`. Despite these warnings not being stable yet, they are good suggestions to improve the code quality.

### DIFF
--- a/lockfile/src/spdx.rs
+++ b/lockfile/src/spdx.rs
@@ -192,8 +192,8 @@ impl Parse for Spdx {
             .relationships
             .into_iter()
             .filter_map(|r| {
-                if r.relationship_type.as_ref().map_or(false, |t| t == "DESCRIBES")
-                    && r.spdx_element_id.as_ref().map_or(false, |t| t == &spdx_info.spdx_id)
+                if r.relationship_type.as_deref() == Some("DESCRIBES")
+                    && r.spdx_element_id.as_ref() == Some(&spdx_info.spdx_id)
                 {
                     r.related_spdx_element
                 } else {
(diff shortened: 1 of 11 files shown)


lockfile/src/spdx.rs
```diff
@@ -192,8 +192,8 @@ impl Parse for Spdx {
             .relationships
             .into_iter()
             .filter_map(|r| {
-                if r.relationship_type.as_ref().map_or(false, |t| t == "DESCRIBES")
-                    && r.spdx_element_id.as_ref().map_or(false, |t| t == &spdx_info.spdx_id)
+                if r.relationship_type.as_deref() == Some("DESCRIBES")
+                    && r.spdx_element_id.as_ref() == Some(&spdx_info.spdx_id)
```
maxrake, line 196: Why was the first instance of `.as_ref().map_or(false,` transformed into `.as_deref()` and the second instance wasn't? I'm not saying it's wrong...just trying to learn something.

maxrake (approved):
One question, but LGTM.

lockfile/src/spdx.rs
cd-work: The first instance compares to a literal, which is a `&str`. So we need to deref to turn the `String` into a `&str`.

The second instance compares a `String` to another `String`. We don't want to move any of those strings since we need them at a later stage, so we turn both of them into string references `&String`.

With pure strings this usually works automatically since all the comparisons are implemented across string variants. However once you start wrapping it in other types (like `Option`) you lose those benefits.

PR #1553, opened 2024-12-10
Fix Windows Gradle manifests incorrectly retained

maxrake:
This change fixes a bug that is specific to Windows systems. When legacy Gradle (prior to v7.0.0) lockfiles exist (e.g.,
`gradle/dependency-locks/*.lockfile`) along with a manifest at a higher level, the manifest is meant to be ignored when finding dependency files to analyze. However, on Windows systems this was not the case before this change. Instead, both the manifest and the lockfiles were retained. After this change, only the lockfiles are retained when both are present.

A failing test was added first to confirm the incorrect behavior. Then, a change was made to handle paths as `Path` objects instead of strings. That way, the generic `/` path separator character will be used correctly regardless of the runtime system.

Additionally, an error message was updated to reflect the correct dependency file language for Gradle parsing errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
-- Gradle manifests used on Windows when legacy lockfiles exist
+- Gradle manifests incorrectly retained on Windows
 
 ## 7.1.5 - 2024-11-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support for C#'s `packages.*.config` lockfile type
 - `phylum firewall log` command to browse firewall activity log
 
+### Fixed
+
+- Gradle manifests used on Windows when legacy lockfiles exist
+
 ## 7.1.5 - 2024-11-26
 
 ### Fixed
--- a/lockfile/src/lib.rs
+++ b/lockfile/src/lib.rs
@@ -372,9 +372,8 @@ pub fn find_depfiles_at(root: impl AsRef<Path>) -> Vec<(PathBuf, LockfileFormat)
             // Legacy Gradle (before v7) lockfiles are in a subdirectory,
             // so we truncate these directories to get the effective
             // directory these lockfiles were created for.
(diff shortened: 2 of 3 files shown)


maxrake:
There may be other instances of incorrectly manipulating paths as strings but finding and fixing those will be handled separately, in an effort to get a quick fix out for a specific customer affected by this Gradle issue.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support for C#'s `packages.*.config` lockfile type
 - `phylum firewall log` command to browse firewall activity log
 
+### Fixed
+
+- Gradle manifests used on Windows when legacy lockfiles exist
```
cd-work, line 18: Not sure I like the `legacy` phrasing here. Sounds a bit like we're using outdated lockfiles over updated manifests when realistically the entire project is "legacy".

lockfile/src/lib.rs
```diff
@@ -372,9 +372,8 @@ pub fn find_depfiles_at(root: impl AsRef<Path>) -> Vec<(PathBuf, LockfileFormat)
             // Legacy Gradle (before v7) lockfiles are in a subdirectory,
             // so we truncate these directories to get the effective
             // directory these lockfiles were created for.
-            let dir_str = lockfile_dir.to_string_lossy();
             if lockfile_format == &LockfileFormat::Gradle
-                && dir_str.ends_with("/gradle/dependency-locks")
+                && lockfile_dir.ends_with("gradle/dependency-locks")
```
cd-work, line 376: I wasn't aware that this would actually work considering the following statement in Rust's docs:

> Only considers whole path components to match.

But I've tested it and this does indeed work. While it doesn't behave as I would expect, it does behave like we want here. The path **must** include both components in their entirety.

The only notable thing is that `/path/gradle/dependency-locks/` also matches, so both with and without the trailing slash matches. But I don't think this should be an issue in our case, since we're matching on a directory anyway.

cd-work (dismissed):

CHANGELOG.md
maxrake: I can change it if desired. How about `Gradle manifests incorrectly retained on Windows`?

maxrake: It has been updated.

cd-work (approved):

PR #1554, opened 2024-12-10
Bump to v7.2.0

maxrake:
Release-Version: v7.2.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Certificate check option name doesn't match configuration option
 - Gradle manifests incorrectly retained on Windows
 
 ## 7.1.5 - 2024-11-26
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,13 +8,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+## 7.2.0 - 2024-12-10
+
 ### Added
 
 - Support for C#'s `packages.*.config` lockfile type
 - `phylum firewall log` command to browse firewall activity log
 
 ### Fixed
 
+- Certificate check option name doesn't match configuration option
```
cd-work, line 20: I feel like this should be moved to changed, since it's not just a fix but technically a breaking change if we consider the command line interface a sort of API.

One might also consider the same reasoning for a 8.0.0, but I don't think anyone actually uses this option so it should be fine?

cd-work (changes requested):

CHANGELOG.md
maxrake: I'll make the change to `CHANGED` category, but I don't think it is breaking since an alias to the old name was retained. Removing that alias, if and when we ever do, would be a breaking change IMO.

cd-work: Right I forgot about the alias, should be fine as non-breaking!

cd-work (approved):

Issue #725, opened 2022-10-07
0.x dependencies are never updated


kylewillmon:
We have a weekly `cargo update` for minor version updates and weekly Dependabot run for major version updates. However, it turns out that neither of these handles bumping dependencies from `0.x` to `0.x+1`

_Originally posted by @kylewillmon in https://github.com/phylum-dev/cli/issues/707#issuecomment-1265607742_

kylewillmon:
Thanks to #889, we now know that Dependabot **will** update these minor versions if there is a security alert on the package.

This issue remains relevant, but that is at least a little bit of comfort.

kylewillmon:
A bit of a workaround here is to periodically use cargo's unstable [update-breaking][1] feature to make sure we're using the latest version possible:

[1]: https://doc.rust-lang.org/cargo/reference/unstable.html#update-breaking

```
cargo +nightly -Zunstable-options update --breaking
```

PR #1563, opened 2024-12-26
Bump dependencies

kylewillmon:
Bump a few dependencies that won't get updated in our weekly updates.

Since these projects are all using 0.x versions, dependeabot doesn't detect them as breaking changes, even though they are (See #725)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -311,7 +311,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "swc_macros_common",
- "syn 2.0.91",
+ "syn 2.0.92",
 ]
 
 [[package]]
@@ -347,7 +347,7 @@ checksum = "c7c24de15d275a1ecfd47a380fb4d5ec9bfe0933f309ed5e705b775596a3574d"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.91",
+ "syn 2.0.92",
 ]
 
 [[package]]
@@ -358,7 +358,7 @@ checksum = "721cae7de5c34fbb2acd27e21e6d2cf7b886dce0c27388d46c4e6c47ea4318dd"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.91",
+ "syn 2.0.92",
 ]
 
 [[package]]
@@ -529,7 +529,7 @@ dependencies = [
  "regex",
  "rustc-hash 1.1.0",
  "shlex",
- "syn 2.0.91",
+ "syn 2.0.92",
  "which 4.4.2",
 ]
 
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -2951,12 +2944,9 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
-version = "0.15.2"
+version = "0.15.0"
```
kylewillmon, line 2947: How weird... I'll have to look into why this was downgraded...

cli/Cargo.toml
```diff
@@ -77,7 +77,7 @@ birdcage = { version = "0.8.1" }
 
 [target.'cfg(windows)'.dependencies]
 # HACK: The Win32_Security feature is required by deno_io, but their crate didn't enable it
-windows-sys = { version = "0.52.0", features = ["Win32_Security"] }
+windows-sys = { version = "0.59.0", features = ["Win32_Security"] }
```
maxrake, line 80: Looks like this bump introduced some breaking changes.

cli/Cargo.toml
```diff
@@ -24,15 +24,15 @@ axum = "0.7.4"
...
 bytes = "1.1.0"
 chrono = { version = "^0.4", default-features = false, features = ["serde", "clock"] }
-cidr = "0.2.0"
+cidr = "0.3.0"
 clap = { version = "4.0.9", features = ["string", "wrap_help"] }
 console = "0.15.2"
 dashmap = "6.0.1"
 deno_ast = { version = "0.41.2", features = ["transpiling"], optional = true }
 deno_core = { version = "0.307.0", optional = true }
 deno_runtime = { version = "0.177.0", optional = true }
```
maxrake, line 33: Is it possible to bump these as well? It *might* [make DBot happy](https://github.com/phylum-dev/cli/security/dependabot/33).

maxrake (changes requested):

cli/Cargo.toml
kylewillmon: That's my bad for not reading the comment right above the line... I shouldn't have bumped this.

cli/Cargo.toml
kylewillmon: It's possible with some code changes that I didn't really want to get into at the moment.

But it wouldn't fix the `idna` issue because we have more paths to that crate. For example:

```
phylum-cli
└── git2 v0.19.0
    └── url v2.4.1
        └── idna v0.4.0
```

maxrake: Roger that...a separate PR then.

maxrake (approved):
LGTM

PR #1565, opened 2024-12-30
Replace use of STAGING API with PROD

maxrake:
This change removes the use of the Phylum STAGING environment.


maxrake:
If this PR is approved and merged, a follow-on action is to remove the `PHYLUM_TOKEN_STAGING` secret from the repo settings.

maxrake:
Closing now that STAGING is back up...

PR #1575, opened 2025-02-17
Fix test failure to to phylum.io redirect

cd-work:
This fixes an issue caused by phylum.io redirecting to veracode.com, which caused our sandbox to complain that network access permissions to veracode.com were missing.

Some other tests also used `phylum.io` as dummy URIs. While these were never hit, they've been replaced for consistency and future-proofing.

There are still several instances referencing `api.phylum.io` or `docs.phylum.io`, which have been left unchanged since no Veracode alternative exists yet.

### DIFF
--- a/cli/src/permissions.rs
+++ b/cli/src/permissions.rs
@@ -264,7 +264,7 @@ impl From<&Permissions> for PermissionsOptions {
 
         let mut allow_net = value.net.clone();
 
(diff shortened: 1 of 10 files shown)


louislang (approved):

maxrake (approved):
LGTM

PR #1577, opened 2025-02-18
Improve error for init with invalid org

### LINKED ISSUE #1576: Cryptic error in `phylum init` when org is set wrong
If you set your organization to a nonexistent value, `phylum init` fails and the error message is very unhelpful

```
> phylum init
❗Error: HTTP request error (404 Not Found):

could not find any matching resource
```

In fairness to the CLI, if you set that value using `phylum org link`, a helpful warning is printed...

```
> phylum org link nonexistent
⚠️ Could not access organization "nonexistent", future Phylum commands may fail unexpectedly
✅ Successfully set default organization to "nonexistent"
```

But in practice, this warning is too far removed from the error to be useful. Other commands like `phylum project list`, `phylum status`, and `phylum group list` still work just fine.

We should add some additional context to the error here at least (e.g. `Error: Could not list groups in organization "blah"`)

cd-work:
This adds some error details to `phylum init` when executed with an organization that either doesn't exist, or the user has no access to.

Closes #1576.


mhorner-vera (approved):

cli/src/commands/init.rs
maxrake, line 1: CHANGELOG entry needed.

maxrake (changes requested):
Confirmed the changes work as expected with local testing.

The request is to add a CHANGELOG entry.

maxrake (approved):
LGTM

PR #1581, opened 2025-03-03
add default exception for nix

matt-phylum:
With Nix (including NixOS), software is installed into /nix/store and resolved using environment variables or configuration files or symbolic links. For example, if you install the jdk package your java installation will be somewhere like `/nix/store/3dhyjzr2j852wxgsaij64xgm74h6wgfp-openjdk-21.0.5+11/bin/java` and resolved using symbolic links or `PATH` depending on the installation method. Therefore, if `/nix/store` isn't readable and executable you won't be able to run anything installed using Nix (or practically anything at all on NixOS).

There is a chance that this allows a malicious package to read sensitive files under `/nix/store`. This should be uncommon. Users aren't supposed to put secrets directly into the Nix store because the files all have 0444 or 0555 permissions. However, it doesn't seem unlikely that a user might be using Nix to build private source code, which would leave a copy of that source code in the store where it would be made accessible by this change. I doubt it's a big enough deal that Nix support would require querying for and whitelisting specific packages.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Added
+
+- Nix store access to the default sandbox
+
 ### Fixed
 
 - Unclear error when running `phylum init` with an invalid organization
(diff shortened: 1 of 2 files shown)


matt-phylum:
This also applies to ld.so on NixOS, which will be somewhere like `/nix/store/pzdwqz1l5w8xyk1aqyb53r87kdalxc7f-glibc-2.40-36-bin/bin/ld.so`.

cd-work (approved):

maxrake (approved):
LGTM

PR #1595, opened 2025-03-27
`version` and `purl` fields in `components` list are now optional in CycloneDX SBOMs.

mhorner-vera:
Since CycloneDX v1.4, `version` is no longer a required field: https://cyclonedx.org/docs/1.4/json/#components_items_version

We also made the determination in Slack to just skip components that have a missing PURL. This patch has made an exception only for the missing PURL case, and no others. If we want to relax our input requirements further, we can discuss that here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- `version` and `purl` fields in `components` list are now optional in CycloneDX SBOMs.
+- `version` and `purl` fields in `components` list are now optional in CycloneDX SBOMs
 
 ## 7.4.0 - 2025-03-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Changed
+
+- `version` and `purl` fields in `components` list are now optional in CycloneDX SBOMs.
+
 ## 7.4.0 - 2025-03-20
 
 ### Added
--- a/lockfile/src/cyclonedx.rs
+++ b/lockfile/src/cyclonedx.rs
@@ -12,8 +12,7 @@ use crate::{determine_package_version, formatted_package_name, Package, Parse, U
 /// Define the generic trait for components.
 trait Component {
     fn component_type(&self) -> &str;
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Changed
+
+- `version` and `purl` fields in `components` list are now optional in CycloneDX SBOMs.
```
cd-work, line 13:
```suggestion
- `version` and `purl` fields in `components` list are now optional in CycloneDX SBOMs
```

I was going to also suggest changing it up a bit to be below 80 columns, but iirc the cli changelog is more of a ~100 column limit and generally 80 can be a bit tight. So I think it should be fine as-is without the full stop.

lockfile/src/cyclonedx.rs
```diff
@@ -40,8 +39,9 @@ struct Components<T> {
 struct XmlComponent {
     #[serde(rename = "@type")]
     component_type: String,
-    name: String,
-    version: String,
+    #[serde(rename = "name")]
+    _name: String,
```
cd-work, line 43: Why do we read this at all if it's not used?

lockfile/src/cyclonedx.rs
```diff
@@ -78,8 +74,9 @@ impl Component for XmlComponent {
 struct JsonComponent {
     #[serde(rename = "type")]
     component_type: String,
-    name: String,
-    version: String,
+    #[serde(rename = "name")]
+    _name: String,
```
cd-work, line 78: See XML _name question.

cd-work (changes requested):
Just to be clear: The PURL is also **optional** according to the spec?

lockfile/src/cyclonedx.rs
mhorner-vera: I left it only because it is required by the spec. If we don't care about that, we can remove it.

mhorner-vera:
> Just to be clear: The PURL is also optional according to the spec?

Yes

lockfile/src/cyclonedx.rs
cd-work: I'd argue that it's not our job to verify full spec compliance as long as it's not a violation that affects our ability to parse the file directly. Spec failures are a convenient excuse not to add hacks to work around some issues, but I don't think we need to be pedantic about it.

cd-work (approved):

